Issue #30, opened 2020-11-11
Run parallel builds and integration tests on all supported platforms


UnseenWizzard:
Currently we build and test on Linux only, but release Linux, Windows and Mac versions of the tool. 

To be more confident in those, we should run at least the integration tests on all platforms to continuously ensure that these releases function. 

In trial we've already seen that there are the following issues with this: 

- linting check is using a shell script that is not compatible with windows (it does this to get around gofmt not returning error codes on unformated files)
- integration tests running in parallel influence each other, resulting in some tests timing out or failing

UnseenWizzard:
Closing as OS differences were primarily due to file path/reference handling, which v2 abstracts away. 
Running E2E tests that run as a Go test execution on cmd library level, not compiled CLI level, would currently not have a notable benefit.

PR #32, opened 2020-11-12
Temporarily run integration tests just on linux

didiladi:
This commit changes the workflow which test executed when building
integration tests of PRs coming from forks (Run Integration Tests
for Forks). It temporarily disables builds to run on MacOS and Windows.

The reason for that is #30, builds are currently failing on windows.
Furthermore, there seems to be a problem with concurrent integration
test execution.

We should revert this change again anlongside the fix for #30.

### DIFF
--- a/.github/workflows/fork-integration-test.yml
+++ b/.github/workflows/fork-integration-test.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


UnseenWizzard:
nitpick: Could you add the reason "why" running on different OSs in parallel is disabled to your commit message? 
relates to issue #30

didiladi:
> nitpick: Could you add the reason "why" running on different OSs in parallel is disabled to your commit message?
> relates to issue #30

Changed the commit message and PR description.

UnseenWizzard (approved):

Issue #22, opened 2020-10-22
Config loads properties from other configs


patrickpichler:
There is currently an issue when defining multiple configs in a single yaml with similar names.

E.g. have two configs with one called `dashboard` and the other `dashboard-availability`.

`dashboard` will now add all properties from `dashboard-avilability` to its properties map.

This causes checks such as `HasDependencyOn(Config)` to als search for dependencies
on this wrong properties.

The cause for this is in `conifg.go` line 84:

```go
func filterProperties(id string, properties map[string]map[string]string) map[string]map[string]string {

	result := make(map[string]map[string]string)
	for key, value := range properties {
		if strings.HasPrefix(key, id) {
			result[key] = value
		}
	}
	return result
}
```

Simply testing for `HasPrefix` isn't enough here.

PR #35, opened 2020-11-13
Fix issue that loads properties from configs that start with the same

baumgarb:
name like the config that is being processed at the moment

Resolves: #22

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -66,6 +66,17 @@ func TestFilterProperties(t *testing.T) {
 	assert.Check(t, properties["Captains"] != nil)
 }
 
+func TestFilterPropertiesToReturnExactMatchOnlyForConfigName(t *testing.T) {
+	m := make(map[string]map[string]string)
+
+	m["dashboard"] = make(map[string]string)
+	m["dashboard-availability"] = make(map[string]string)
+
+	properties := filterProperties("dashboard", m)
+
+	assert.Check(t, len(properties) == 1)
+}
+
 func TestGetConfigStringWithEnvironmentOverride(t *testing.T) {
 
 	m := getTestProperties()
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -19,6 +19,7 @@
 package config
 
 import (
+	"fmt"
 	"os"
 	"reflect"
 	"testing"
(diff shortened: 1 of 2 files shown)


pkg/config/config_test.go
```diff
@@ -66,6 +66,17 @@ func TestFilterProperties(t *testing.T) {
 	assert.Check(t, properties["Captains"] != nil)
 }
 
+func TestFilterPropertiesToReturnExactMatchOnlyForConfigName(t *testing.T) {
```
didiladi, line 69: Can you also add a test for two properties with groups in their names?

E.g.:
```
m["dashboard"] = make(map[string]string)
m["dashboard"]["myProperty"] = "test"
m["dashboard.devEnv"]  = make(map[string]string)
m["dashboard.devEnv"]["myProperty"] = "test for dev"

m["dashboard-availability.devEnv"] = make(map[string]string)
```

didiladi (changes requested):
Cool, thanks for your contribution. The change looks good. Would be cool to have an additional test here, thanks!

pkg/config/config_test.go
baumgarb: Thanks, yeah that makes a ton of sense. I've created the test and updated the PR already.

pkg/config/config_test.go
```diff
@@ -66,6 +66,43 @@ func TestFilterProperties(t *testing.T) {
...
+	m["dashboard.dev"]["prop2"] = "C"
+	m["dashboard-availability.dev"]["prop1"] = "E"
+	m["dashboard-availability.dev"]["prop2"] = "F"
+
+	properties := filterProperties("dashboard.dev", m)
+
+	assert.Check(t, len(properties) == 1)
+	assert.Check(t, properties["dashboard.dev"] != nil)
+	assert.Check(t, len(properties["dashboard.dev"]) == 2)
+}
```
didiladi, line 104: Sorry about nitpicking, but these two assertions would make sense here:

```
assert.Equal(t, "B", properties["dashboard.dev"]["prop1"])
assert.Equal(t, "C", properties["dashboard.dev"]["prop2"])
```

baumgarb: I thought about that but I deliberately chose against it. I can add them if really want, but I would rather create a separate test for this because if I add these assertions we're testing different things with one tiny unit test. The test name says it's only testing for exact match of config name, but the assertions you'd like to add here are testing a different thing. They're testing if more specific properties, if any, are overwriting more general ones. 

Or what do you think? I'll wait for your answer before adding / modifying the tests here.

didiladi: I see. Yes, you are completely right. By doing that, your test would be not tesing just a single thing anymore. Would be cool to have such a test with the assertions, since we are missing it currently.

baumgarb: All right, I've implemented two more tests. Can you have a look? Thanks!

didiladi: Cool, thank you!

didiladi:
> ###  Integration tests failed on ubuntu-latest

The failing integration tests have nothing to do with the changes in this PR.

didiladi (approved):

baumgarb:
Do I need to do anything to merge this PR now? I think it says "Only those with write access to this repository can merge pull requests."

didiladi:
> Do I need to do anything to merge this PR now? I think it says "Only those with write access to this repository can merge pull requests."

Hey, I'm currently in the progress of fixing the integration tests (#63). Once this is done, I'll re-run the integration tests for your change and merge the PR :)

PR #39, opened 2020-11-13
Fix issue when parallel integrations tests are failing on kubernetes credentials

tatomir146:
As `endpointUrl` needs to be unique, used name (which is generated with timestamp). Using hardcoded `endpointUrl` results in error 
```JSON 
{"error":{"code":400,"message":"Constraints violated.","constraintViolations":[{"path":"endpointUrl","message":"URL is already in use.","parameterLocation":"PAYLOAD_BODY","location":null}]}}
```
Using `name` instead of hardcoded url, because it also contains timestamp when test is started, meaning `endpointUrl` will be different every time tests are executed.

### DIFF
--- a/cmd/monaco/test-resources/integration-all-configs/project/kubernetes-credentials/kubernetes.json
+++ b/cmd/monaco/test-resources/integration-all-configs/project/kubernetes-credentials/kubernetes.json
@@ -1,9 +1,9 @@
 {
     "active": "{{ .active }}",
     "label": "{{ .name }}",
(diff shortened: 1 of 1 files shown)


didiladi:
Relates to #30

didiladi (approved):

UnseenWizzard:
Only thing I keep criticizing is commit messages ;) 

Would be cool if your explanation on why this fixes an integration test issue from your PR commit is actually in the commit. 
Otherwise agree, looks good

UnseenWizzard (approved):

Issue #68, opened 2020-11-27
Support for service request naming rules


163a:
Enhancement request to support for service request naming rules:
/api/config/v1/service/requestNaming

https://www.dynatrace.com/support/help/shortlink/api-request-naming

PR #69, opened 2020-11-27
Global Service Request Naming Support

Dynatrace-Adam-Gardner:
Solves #68

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -57,7 +57,7 @@ var apiMap = map[string]string{
 	"conditional-naming-host":         "/api/config/v1/conditionalNaming/host",
 	"conditional-naming-service":      "/api/config/v1/conditionalNaming/service",
 	"maintenance-window":              "/api/config/v1/maintenanceWindows",
-    "request-naming":                  "/api/config/v1/service/requestNaming",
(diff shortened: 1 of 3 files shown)


UnseenWizzard (approved):

UnseenWizzard:
Great stuff getting this in so quickly @Dynatrace-Adam-Gardner ! 

For releasing this as 1.1.0 I'd wait for a few more PRs that should happen today

pkg/api/api.go
```diff
@@ -57,6 +57,7 @@ var apiMap = map[string]string{
 	"conditional-naming-host":         "/api/config/v1/conditionalNaming/host",
 	"conditional-naming-service":      "/api/config/v1/conditionalNaming/service",
 	"maintenance-window":              "/api/config/v1/maintenanceWindows",
+	"request-naming":                  "/api/config/v1/service/requestNaming",
```
didiladi, line 60: Should we use a folder name which has `service` in its name? E.g. `request-naming-service`?

This way we would indicate that its a service api. Furthermore, if there will be a request naming api for a different type of entity in the future, it would be less confusing. Its then also aligned with the other folder names like `conditional-naming-service`.

What do you think?

PR #72, opened 2020-11-28
Feature/windows builds

UnseenWizzard:
Makes it possible to run all Make targets on windows (gnuwin make). 

Also updates CONTRIBUTING.md to only reference the make targets rather than documenting possibly outdated go commands

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,22 +1,31 @@
 BINARY=monaco
 
-.PHONY: build clean test integration-test
+.PHONY: lint format build install clean test integration-test test-package
 
 lint:
+ifeq ($(OS),Windows_NT)
+	@.\tools\check-format.cmd
+else
 	@sh ./tools/check-format.sh
+endif
 
 format:
 	@gofmt -w .
 
-build: clean
+build: clean lint
 	@echo Build ${BINARY}
 	@go build ./...
 	@go build -o ./bin/${BINARY} ./cmd/monaco
 
 clean:
 	@echo Remove ${BINARY} and bin/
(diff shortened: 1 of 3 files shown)


tatomir146 (approved):

didiladi (approved):

UnseenWizzard:
@patrickpichler I'd appreciate a look from you as well! after all you're the biggest Makefile proponent and so far sole contributor to it ;)

Makefile
```diff
@@ -1,24 +1,41 @@
...
+	@.\tools\check-format.cmd
+else
 	@sh ./tools/check-format.sh
+endif
 
 format:
 	@gofmt -w .
 
-build: clean
+mocks:
```
patrickpichler, line 15: You need to add `mock` also to the `.PHONY` tasks

patrickpichler (commented):
As I do not have a windows machine, I cannot really test it. Would be nice if somebody with a windows machine could test it.

Makefile
UnseenWizzard: thx, missed that one

UnseenWizzard:
> As I do not have a windows machine, I cannot really test it. Would be nice if somebody with a windows machine could test it.

I developed this on my windows PC while I was waiting for something to download ;) 

Tested on Win 10 & WSL Ubuntu 20

Makefile
UnseenWizzard: Fixed

patrickpichler (approved):
Would be nice if you could squash the commits ^^

PR #74, opened 2020-11-29
Encapsulate REST code in DynatraceClient

didiladi:
This commit introduces a new client (`DynatraceClient`) for performing
basic CRUD operations on any Dynatrace API supported by monaco.

It encapsulates the configuration-specific inconsistencies of certain
APIs in one place to provide a common interface to work with. After
all: A user of DynatraceClient shouldn't care about the implementation
details of each individual Dynatrace API.

Its design is intentionally not dependant on the Config and Environment
interfaces included in monaco. This makes sure, that DynatraceClient
can be used as a base for future tooling, which relies on a
standardized way to access Dynatrace APIs.

The new client implementation opens the door for future improvements,
like concurrent web request execution and the implementation of
various caching strategies.


pkg/api/api.go
```diff
@@ -87,6 +89,10 @@ func newApi(id string, apiPath string) Api {
 }
 
 func NewApi(id string, apiPath string) Api {
+
+	// TODO log warning if the user tries to create an API with a id not present in map above
```
UnseenWizzard, line 93: That is currently only possible if changing the code though, isn't it?

cmd/monaco/main.go
```diff
@@ -351,10 +353,22 @@ func deleteConfigs(apis map[string]api.Api, environments map[string]environment.
 		for name, environment := range environments {
 			util.Log.Info("Deleting %d configs for environment %s...", len(configs), name)
 
+			apiToken, err := environment.GetToken()
+			if err != nil {
+				return err
+			}
+			client := rest.NewDynatraceClient(environment.GetEnvironmentUrl(), apiToken)
```
UnseenWizzard, line 360: Is there a reason to create one client per environment for execute and another for delete? 

Maybe we could create the client per env once somewhere early in the execution and reuse them - e.g. keep a map of env->client that can be accessed in both execute and delete?

UnseenWizzard:
Cool fast way to move to fewer client objects and a simpler rest call wrapper to be reused!

pkg/api/api.go
didiladi: I'm afraid, no. Any user can import a package (pkg/` folder) in an external application and use it. I guess we should really start with validating the arguments passed to our public API. That's not just here the case, but everywhere across `monaco`.

cmd/monaco/main.go
didiladi: Good idea, I already thought of that. However, I wanted to keep it as simple as possible for now. Later, we can do a lot of improvements like parallel execution, caching of the initial GET request results, ... :smile:

pkg/api/api.go
UnseenWizzard: ah, good point :+1:

UnseenWizzard:
In how far is this still work in progress @didiladi ?

didiladi:
@UnseenWizzard this is ready to be reviewed. Its currently in a state I'm comftable with. There are some follow-up actions to this - I don't want to make the review even more difficult:

* Properly handle the error in [request.go](https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/pull/74/files#diff-76bc1ca3c519d2fed86bc08c4d9d9722951fc57cde145898dc447d8dc442e975R79)
* Paralled request execution ()
* Caching of the initial GET request - we currently perform the GET for every config, no matter if there was already one applied for the same API.
* To create just one client per environment (as @UnseenWizzard wrote [here](https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/pull/74#discussion_r533957137))

If you find the time, I'd be very happy with a review @UnseenWizzard, @patrickpichler, @CruzanCaramele, @tatomir146 :small_airplane:

didiladi:
Note that this PR breaks the public API of the `request` package. However, after merging this PR, we will have a stable interface which can be used by various other tools.

UnseenWizzard (dismissed):

patrickpichler (approved):

patrickpichler (commented):
I just noticed the integration tests are failing. 

Other than that, the PR looks good to me :+1:

didiladi:
> Integration tests are failing

Yes, I'm on it. Doesn't seem to come from my changes though.

didiladi:
It was not related to my changes:

```
Response was: {"error":{"code":429,"message":"Rate limit of 200 requests/minute exceeded! Reset time: 2020-12-17 16:37:47 GMT (1608223067402)"}}, responsible config: test-resources/integration-multi-environment/cinema-infrastructure/auto-tag/application-tagging.json
```

I just re-triggered the workflow - it worked now. I'll work on finishing the rate limit strategy (#89) next.

patrickpichler (approved):

patrickpichler:
Wouldn't a rebase be better than a merge? 
That way we would have a clean linear history

didiladi:
> Wouldn't a rebase be better than a merge?
> That way we would have a clean linear history

That's why we activated the "Squash Rebase" merge strategy :1st_place_medal:

patrickpichler:
That is true ^^

PR #79, opened 2020-12-01
Delete configs only if apply finished without errors (#20)

### LINKED ISSUE #20: Don't delete configs if apply didn't work
Currently, the tool deletes the configs in the `delete.yaml` file independently of the success of the prior applying of configs.

If the apply fails (for any config) we should skip the delete.

neonalps:
This commit deletes configs only if there were no errors during previous deployment/validation.

fixes #20


didiladi (approved):
:party

Issue #21, opened 2020-10-20
Change handling of backslash/slash in configs


patrickpichler:
As of right now, all `/` in the config will be replaced with `\` on windows. This
could lead to some strange behaviors, such as replacing `/` in names.

This is done solely to handle paths on windows, but if you think about it, there
is no need to replace every `/` in the config, as we read files from disk only once
at the beginning. Afterwards `/` could always be used to reference other configs.

PR #83, opened 2020-12-02
Fix #60: Allow path separators in variables

### LINKED ISSUE #60: Slashes can not be used in variables - Caused by current OS path-seperator handling
**Describe the bug**
Monaco fails for synthetic tests when a variable uses special characters. However passing the same string directly via JSON works.

**How to reproduce**

```
my_config:
  - ua_string: "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.198 Safari/537.36"
```
and
```
{
...
  "script": {
    "configuration": {
      "userAgent": "{{ .ua_string }}"
    }
  }
...
}
```

Fails with (line 11 is the userAgent line):
```
Response was: {"error":{"code":400,"message":"Could not map JSON at 'script' near line 11 column 20"}}, responsible config: projects\baseconfig\synthetic-monitor\myfile.json
```

Whereas removing that YAML variable and pushing this, works just fine.

```
{
...
  "script": {
    "configuration": {
      "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.198 Safari/537.36"
    }
  }
...
}
```

baumgarb:
Fix #60: Allow path separators in variables

Normalize path separators (slash, backslash) only if a variable /
property value starts with one. Otherwise, assume that it is not
a path and therefore normalizing is not needed (e.g. in case of a
user agent string).

### DIFF
--- a/cmd/monaco/special_character_in_config_integration_test.go
+++ b/cmd/monaco/special_character_in_config_integration_test.go
@@ -1,0 +1,43 @@
+// +build integration
+
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"testing"
+
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util"
+	"gotest.tools/assert"
+)
+
+func TestIntegrationDoNotNormalizePathSeparatorsInUserAgentString(t *testing.T) {
+
+	const specialCharConfigFolder = "test-resources/special-character-in-config/"
+	const specialCharEnvironmentsFile = specialCharConfigFolder + "environments.yaml"
+
+	RunIntegrationWithCleanup(t, specialCharConfigFolder, specialCharEnvironmentsFile, "SpecialCharacterInConfig", func(fileReader util.FileReader) {
+
+		statusCode := RunImpl([]string{
+			"monaco",
+			"--environments", specialCharEnvironmentsFile,
+			specialCharConfigFolder,
+		}, fileReader)
+
+		assert.Equal(t, statusCode, 0)
+	})
+}
--- a/pkg/util/parse.go
+++ b/pkg/util/parse.go
@@ -92,7 +92,7 @@ func convert(original map[string]interface{}) (err error, typed map[string]map[s
 						case string:
 							switch v3 := v3.(type) {
 							case string:
-								if referencesJSON(v3) || appearsToReferenceVariableInAnotherYaml(v3) {
+								if referencesConfigJSON(k1, v3) || appearsToReferenceVariableInAnotherYaml(v3) {
 									m2Inner[k3] = ReplacePathSeparators(v3)
 								} else {
 									m2Inner[k3] = v3
@@ -127,7 +127,10 @@ func appearsToReferenceVariableInAnotherYaml(s string) bool {
 	return true
 }
 
-func referencesJSON(s string) bool {
+func referencesConfigJSON(yamlSection, s string) bool {
+	if yamlSection != "config" {
+		return false
+	}
 	return strings.HasSuffix(s, ".json")
 }
 
--- a/pkg/util/parse.go
+++ b/pkg/util/parse.go
@@ -91,7 +91,7 @@ func convert(original map[string]interface{}) (err error, typed map[string]map[s
 						case string:
 							switch v3 := v3.(type) {
 							case string:
-								if !strings.HasPrefix(v3, "http") {
+								if startsWithPathSeparator(v3) {
 									m2Inner[k3] = ReplacePathSeparators(v3)
 								} else {
 									m2Inner[k3] = v3
@@ -112,4 +112,9 @@ func convert(original map[string]interface{}) (err error, typed map[string]map[s
 		}
 	}
 	return nil, m2
+}
+
+func startsWithPathSeparator(s string) bool {
+	firstChar := string(s[0])
+	return firstChar == "/" || firstChar == "\\"
 }
--- a/pkg/util/parse.go
+++ b/pkg/util/parse.go
@@ -19,6 +19,7 @@ package util
 import (
 	"errors"
 	"os"
+	"regexp"
 	"strings"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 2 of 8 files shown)


baumgarb:
@UnseenWizzard @patrickpichler : fyi guys, I tried to solve it never ever replacing any path separators in variables, but that makes a couple of tests fail on Windows (e.g. TestCreateProjectsFromFolderList) where e.g. `transitional-dependency-test/anjie/garkbit/dashboard/dashboard.yaml` references the dashboard from another project. Therefore, I changed the implementation so that we still consider it a path where we should normalize path separators when the variable starts with a Slash. 

Doing that made another test fail :-) This time, the `TestLoadProjectsThrowsErrorOnCircularConfigDependecy` test failed because it loads loads configurations in a circular manner. In this test the paths did not start with a slash or backslash, therefore the newly introduced logic described above didn't work. To fix this I had to change the test resources for the circular dependency test so that the paths start with a slash now. 

So if you guys think this is okay and this PR gets approved it means that we're actually no longer downwards compatible. Any configuration out there which might use relative paths would no longer work. 

If you guys think this is not okay and we should come up with another solution then pls let me know what to do, I'm not entirely sure what paths between configurations and projects should be supported.

UnseenWizzard:
Thanks @baumgarb, will take a look tomorrow! From just the description the drawbacks sound a bit discouraging tbh . 

@didiladi please review this as well.

baumgarb:
Yes you're absolutely right. Just wanted to point out the drawbacks straight away and get the discussion going on how to solve this in a downwards-compatible way. I think we're probably not too far away from a satisfying solution :)

patrickpichler:
I guess we could do some probing to determine if a value is a path or not.

We could leverage the fact that as of right now, we only support `.name` and
`.id` cross references.

patrickpichler:
Also as pointed out in #21 we maybe should agree on only having `/` as 
path separator in the config and loading paths via `filepath.ToSlash`.

patrickpichler:
Maybe something like [this](https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/pull/86/files#diff-69b879735174386e2ae963f274ef5cb30f6b79652eb5db75b58f6ba63445f6a5R79)

UnseenWizzard:
I like @patrickpichler's idea of moving to explicitly handling path references! 

There's two places we expect paths in a YAML:

In the `config:` section, that maps each configuration to a JSON

And when referencing a dependency - though we will need to discuss if we want to be able to reference other variables, and if we would block that case.

@didiladi, @CruzanCaramele what's your take?

baumgarb:
> Maybe something like [this](https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/pull/86/files#diff-69b879735174386e2ae963f274ef5cb30f6b79652eb5db75b58f6ba63445f6a5R79)

I like that, sounds reasonable. This way we'll keep it downwards-compatible and we can easily identify a valid relative or absolute cross reference. 

If we ever need to reference other variables we just need to keep in mind to add them to the glob pattern.

didiladi:
I also like @patrickpichler's idea and the fact that we can stay downwards-compatible.

At a later point in time, we do need to support referencing other yaml variables (besides `id` and `name`). We had the case already twice in our configuration repository, where someone tried to reference variables cross-yaml - then they were astonished that this is not yet working. :sweat_smile:

cmd/monaco/special_character_in_config_integration_test.go
```diff
@@ -0,0 +1,43 @@
...
+package main
+
+import (
+	"testing"
+
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util"
+	"gotest.tools/assert"
+)
+
+func TestIntegrationDoNotNormalizePathSeparatorsInUserAgentString(t *testing.T) {
```
didiladi, line 28: :1st_place_medal:

baumgarb:
@patrickpichler @UnseenWizzard @didiladi : Hey guys! After a while I found some time again. I've pushed the changes, we're now backwards-compatible and replace the path separator only if a variable's value ends with ".json" or if the variable's value appears to be a reference to a variable in another YAML file. 

Hope that we're getting closer with that approach now? Happy to hear your feedback on this, thanks a lot!

didiladi (dismissed):
Cool! I like that this solution explicitly states what's a valid path. I like the naming of the functions!

patrickpichler:
Looks good :+1: 
I will close the other PR :ok_hand:

baumgarb:
@didiladi since the integration tests ran through fine I guess we could merge this one now?

pkg/util/parse.go
```diff
@@ -113,3 +114,29 @@ func convert(original map[string]interface{}) (err error, typed map[string]map[s
...
+	}
+	if doesNotReferenceKnownVariable(s) {
+		// As of right now there's only a limited number of variables that can be referenced. If s points to something else let's bail out here.
+		return false
+	}
+	return true
+}
+
+func referencesJSON(s string) bool {
+	return strings.HasSuffix(s, ".json")
```
UnseenWizzard, line 131: I'm wondering if there is any usecase/type of Dynatrace config in which I'd pass a .json filename as actual content? 
For our JSON references, I think they will always be in the initial `config:` object. 

Not sure if we should be stricter in this check already, or just wait and see if it becomes an issue. 
If we wait and see, maybe placing this comment in the method might be helpful? 
@didiladi ?

UnseenWizzard (commented):
Cool stuff, thanks @baumgarb :+1: 

Let's decide on which way to go with my .json comment/worry, then let's merge it :)

pkg/util/parse.go
didiladi: Hmm, that's a good point @UnseenWizzard :+1: I like the idea to check for the `config` object.

didiladi: But that could also be done in a follow-up PR in my opinon.

UnseenWizzard: I guess we should sanity check that the current handling at least doesn't cause in issue in any of our config repos?

If not I agree, we may move to base it on the config as a follow up only if needed

didiladi: I just found an example in our config repo: the extension name usually ends with `.json`. Ok, so we need to add the check in this PR. Thanks for being so persistent @UnseenWizzard :smile:

didiladi: And we don't run into this problem in the integration tests, because the name of our example doesn't end with `.json`: https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/blob/main/cmd/monaco/test-resources/integration-all-configs/project/extension/plugins.yaml#L6

patrickpichler: We could solve this issue by requiring some sort of prefix for variable references.
Something like `$/other/file.id` / `$other/file.name`. 

And it would maybe make sense to only allow for `/` in config references. We
could then after detection of a reference call `filepath.FromSlash` to get the 
os specific file path.

UnseenWizzard: persisting a quick discussion with @patrickpichler : 

We think for now we should limit to replacing paths in config:, not on all .jsons

Long term we should discuss the suggestions on how we can better handle general variables/references to paths (e.g. marking them)

patrickpichler: I added a new discussion ticket for this #98

baumgarb: @didiladi @UnseenWizzard @patrickpichler : I've changed it so that references to JSONs are only replaced if they appear in the YAML config section. Hope I understood it correctly?

UnseenWizzard (approved):
with that last change to where we expect json this looks safe to me :) 
Awesome stuff @baumgarb ! 

We should probably sanity check on some big internal configurations before merging, just to be safe?

didiladi:
I just ran it internally against our configs -> Worked like a charm. Thanks @baumgarb. :rocket:

PR #88, opened 2020-12-07
Enhance log messages for syntax errors detected in JSONs (fix #77)

### LINKED ISSUE #77: Add config line number to detected error
Enhancement request to add the line number of the configuration when validator detects error in the configuration file(s).

`2020-12-01 14:36:10 ERROR Validation of sandbox failed with error projects/t24/request-naming-service/t24_webrequest_enquiry_operation.json is not a valid json! Error: invalid character '"' after object key:value pair`

didiladi:
This commit enhances the log messages displayed to the user if a json
contained a syntax error. Previously, we just displayed the message of
the original cause to the user. This resulted in a error message like
this:

```
2020-12-01 14:36:10 ERROR Validation of sandbox failed with error \
  operation.json is not a valid json! Error: invalid character '"' \
  after object key:value pair
```

This commit introduces a second log which indicates where in the
file the syntax error is. Besides the line and character numbers,
we also display now line (+previous line), which contain the syntax
error.

E.g. if the user forgets to put a comma after a key:value pair:

```
2020-12-07 19:52:12 ERROR File did not contain valid json:
 --> operation.json:7:4
   |    "no": "comma"
 7 |    "boolean": true
   |    ^^^
   - Cause: invalid character '"' after object key:value pair
```

E.g. if the user forgets the closing curly brace for an object:

```
2020-12-07 19:52:12 ERROR File did not contain valid json:
 --> test.json:6:2
   |    "foo": "bar"
 6 |  ]
   |  ^^^
   - Cause: invalid character ']' after object key:value pair
```

This log line format was heavily inspired by the Rust compiler.


pkg/util/json.go
```diff
@@ -0,0 +1,148 @@
...
+func ValidateJson(jsonString string, filename string) error {
+
+	var j string
+	err := json.Unmarshal([]byte(jsonString), &j)
+
+	if jsonError, ok := err.(*json.SyntaxError); ok {
+		return mapError(jsonString, filename, int(jsonError.Offset), err)
+	}
+	if _, ok := err.(*json.UnmarshalTypeError); ok {
+		// TODO actually check against the model
```
UnseenWizzard, line 100: This `TODO` is the only issue I have with this PR ;) 
Is this to do as part of you still working on it, or a future enhancement? 
If the first, I'll wait to accept the PR, if the latter could you create and issue and reference it in this TODO please ?

didiladi: :smile: No, the TODO is for a future enhancement. We don't have any model information currently.

UnseenWizzard: Ok, could you create an issue for it and reference the id in the TODO comment, then please?

Todo's in code are way to easy to forget about :D

pkg/util/json.go
```diff
@@ -0,0 +1,148 @@
...
+			whiteSpace, e.Cause.Error())
+	}
+}
+
+// ValidateJson validates whether the json file is correct, by using the internal validation done
+//when unmarshalling to a an object. As none of our jsons can actually be unmarshalled
+//to a string, we catch that error, but report any other error as fatal.
+func ValidateJson(jsonString string, filename string) error {
+
+	var j string
```
patrickpichler, line 93: Why do we unmarshal the json as a string? 

This seems a bit odd to me. Wouldn't something like`interface{}` be better?

didiladi: I guess we didn't have any use case for `interface{}` yet. That's why `config.GetConfigForEnvironment(..)` returns the json as a `string`. But right, that would be something to tackle in the future. I just created #101 for it.

pkg/util/json.go
didiladi: Just created #104.

didiladi: ... and added it to the comment.

didiladi: ... and I lost the approval because of that. :sob:

PR #89, opened 2020-12-08
Fix #87: implement simple strategy to prevent running into rate limit

### LINKED ISSUE #87: Introduce strategy to prevent reaching API rate limit
The APIs of Dynatrace have some environment-wide API rate limits defined.

Those are currently:
* environment APIs: 50 requests / minute
* config APIs: 200 requests / minute

More about the API rate limits can be seen here: https://www.dynatrace.com/support/help/extend-dynatrace/dynatrace-api/basics/access-limit/

Implement a strategy to avoid hitting the API rate limit. This can be:
* Adjusting the speed of requests against the API
* Introducing a wait period after the limit was hit

didiladi:
This commit introduces a simple strategy to prevent monaco from running
into rate-limiting errors. The strategy reads the rate limit headers
from the response and suspends the current goroutine (sleeps) until the
current rate limit is being reset again.

More details:

If the HTTP status response code of any web request done by monaco is
429 (Too Many Requests), the newly implemented rate-limiting strategy
reads the `X-RateLimit-Reset` header from the response and tries to wait
until the point in time, the rate is being reset again.

Since we can't rely on the client-side timestamp returned by
`time.Now()`, we sleep:
  * a minimal duration of 5 Seconds, or
  * a maximal duration of up to 1 minute
before performing the same web request again which returned the 429
status code.

The whole process is repeated at max 5 times until an error is returned
to the user.

For more information about the rate limiting in Dynatrace, please consult
[the official documentation](https://www.dynatrace.com/support/help/dynatrace-api/basics/access-limit/).


pkg/rest/request.go
```diff
@@ -71,14 +73,26 @@ func requestWithBody(method string, url string, body io.Reader, apiToken string)
...
-		println("HTTP Request failed with Error: " + err.Error())
-		return Response{}
-	}
-	defer func() {
-		err = resp.Body.Close()
-	}()
-	body, err := ioutil.ReadAll(resp.Body)
-	return Response{resp.StatusCode, body}
+
+	rateLimitStrategy := createRateLimitStrategy()
```
patrickpichler, line 77: Would it maybe make sense to move this to a global var?

pkg/rest/rate_limit.go
```diff
@@ -0,0 +1,131 @@
...
+		util.Log.Debug("simpleSleepRateLimitStrategy: Reset sleep duration to %f seconds...", sleepDuration.Seconds())
+	}
+	if sleepDuration.Nanoseconds() > maxWaitTimeInNanoseconds.Nanoseconds() {
+		sleepDuration = maxWaitTimeInNanoseconds
+		util.Log.Debug("simpleSleepRateLimitStrategy: Reset sleep duration to %f seconds...", sleepDuration.Seconds())
+	}
+	return sleepDuration
+}
+
+func (s *simpleSleepRateLimitStrategy) convertMicrosecondsToUnixTime(timeInMicroseconds int64) time.Time {
```
tatomir146, line 125: should be as part of utils?

tatomir146 (dismissed):

pkg/rest/rate_limit.go
```diff
@@ -0,0 +1,131 @@
...
+type simpleSleepRateLimitStrategy struct{}
+
+func (s *simpleSleepRateLimitStrategy) executeRequest(timelineProvider util.TimelineProvider, callback func() Response) (Response, error) {
+
+	response := callback()
+
+	maxIterationCount := 5
+	currentIteration := 0
+
+	for response.StatusCode == http.StatusTooManyRequests && currentIteration <= maxIterationCount {
```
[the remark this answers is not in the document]
didiladi: you are right, classic off-by-one error :+1:

pkg/rest/rate_limit.go
```diff
@@ -0,0 +1,131 @@
...
+
+		limit, humanReadableTimestamp, timeInMicroseconds, err := s.extractRateLimitHeaders(response)
```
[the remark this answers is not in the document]
didiladi: No, since Dynatrace returns all timestamps in UTC. I'm guessing the rate limiting headers are no exception here. But  now that I think of it... you are right - the `timelineProvider` needs to return UTC as well, I'll check the code again. Great input!

pkg/rest/rate_limit.go
didiladi: You are right. I'll move this function to `time.go`. :+1:

pkg/rest/request.go
didiladi: You are right, makes more sense as a global var. "Global" in the package-global sense - it makes no sense to expose this outside of the `rest` package. Thanks!

pkg/rest/request.go
```diff
@@ -71,14 +73,26 @@ func requestWithBody(method string, url string, body io.Reader, apiToken string)
...
+
+	// TODO properly handle error
```
UnseenWizzard, line 79: If I read it correctly, errors inside `rateLimitStrategy` only happen if we fail to parse the expected rate limit headers? 

I find it interesting that the request callback here swallows any errors but the client.Do error (so did the method it was before) 

Are there any tests that check how the RateLimitStrategy handles a blank `Response` object from the callback?

pkg/rest/request.go
```diff
@@ -71,14 +73,26 @@ func requestWithBody(method string, url string, body io.Reader, apiToken string)
...
+	response, _ := rateLimitStrategy.executeRequest(util.NewTimelineProvider(), func() Response {
+		resp, err := client.Do(request)
+		if err != nil {
+			println("HTTP Request failed with Error: " + err.Error())
```
UnseenWizzard, line 83: just noticed this uses a println and not the Logger, could you please change that while you're changing things? :)

didiladi: Sure thing, good catch. Eagle-eye @UnseenWizzard.

pkg/rest/request.go
didiladi: On second thought - we need to add a strategy for a simple exponential backoff (+jitter) soon. So, I'd rather leave it here as a preparation for later.

pkg/rest/request.go
didiladi: You are right, I'm going to return the empty response if the `rateLimitStrategy` returns an `error`.

Good catch! There are no tests for handling the empty response in `simpleSleepRateLimitStrategy`. I will add one.

pkg/rest/rate_limit.go
didiladi: done in aff4dc7

pkg/rest/rate_limit.go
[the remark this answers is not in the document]
didiladi: done in 3924199

pkg/rest/rate_limit.go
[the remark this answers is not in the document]
didiladi: Done in 9cda9ba

pkg/rest/request.go
didiladi: Done in 3fa1d52.

pkg/rest/request.go
didiladi: Done in 808c37c

UnseenWizzard (approved):

Issue #6, opened 2020-10-06
Problem with using flags


tatomir146:
If using flags without equals "=" there is a bug if specifying project parameter as latest:
`command... -p my_project`

2020/04/16 11:07:02 Project my_project does not exist!: stat my_project/my_project: no such file or directory

Following works:
`command... -p=my_project`

The last parameter gets interpreted as configuration root parameter

patrickpichler:
This should be fixed by #100

PR #100, opened 2020-12-18
Introduce new cli design

patrickpichler:
This is a first draft for the new cli design. 

It would be nice to get some feedback if the structure is something we want 
to go forward with in the future.


patrickpichler:
@dynatrace-oss/mac-maintainers

patrickpichler:
@didiladi I added the possibility to switch between new and old cli.

Would be nice if you could give it a try :+1:

didiladi:
> @didiladi I added the possibility to switch between new and old cli.
> 
> Would be nice if you could give it a try

Awesome, I'll take a look at it on Monday!

patrickpichler:
Actually this also fixes #6

cmd/monaco/deploy.go
```diff
@@ -0,0 +1,262 @@
+package main
```
didiladi, line 1: I like that you moved the functionality to its own file! :+1:

cmd/monaco/main.go
```diff
@@ -17,22 +17,14 @@
...
-	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/config"
-	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/delete"
-	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/environment"
-	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/project"
-	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/rest"
+
 	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util"
 	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/version"
+
+	"github.com/urfave/cli/v2"
```
didiladi, line 27: I really like this CLI. Compared to cobra it produces more readable code, awesome! :rocket:

cmd/monaco/main.go
```diff
@@ -44,380 +36,211 @@ func Run(args []string) int {
...
-	for i, err := range errorList {
-		configIssue := fmt.Sprintf("environmentfile-issue-%d", i)
-		deploymentErrors[configIssue] = err
-	}
-
-	err := util.SetupLogging(verbose)
-
-	if err != nil {
-		util.Log.Error("Error writing log file: %s", err.Error())
+	if newCli, ok := os.LookupEnv("NEW_CLI"); ok && newCli != "0" {
```
didiladi, line 42: Everyone ok with calling the env var `NEW_CLI` @dynatrace-oss/mac-maintainers?

cmd/monaco/main.go
```diff
@@ -44,380 +36,211 @@ func Run(args []string) int {
...
+		app = buildExperimentalCli(fileReader)
```
didiladi, line 43: Can you add a log line which states something like:

```
You are using the new CLI structure which is currently in Beta. Please provide feedback here: https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/issues/45.
We plan to make this CLI GA in version 2.0.0
```

cmd/monaco/main.go
```diff
@@ -44,380 +36,211 @@ func Run(args []string) int {
...
+	} else {
+		app = buildCli(fileReader)
```
didiladi, line 45: And here something like:

```
You are currently using the old CLI structure which will be used by default until monaco version 2.0.0.
Check out the beta of the new CLI by adding the environment variable `NEW_CLI`. We can't wait for your feedback.
```

cmd/monaco/main.go
```diff
@@ -44,380 +36,211 @@ func Run(args []string) int {
...
-			util.Log.Info("Validation finished without errors")
-		} else {
-			util.Log.Info("Deployment finished without errors")
-		}
-		err = deleteConfigs(apis, environments, path, dryRun, fileReader)
-		if err != nil {
-			util.Log.Error("Deletion of configs failed with error %s\n", err)
-			statusCode = -1
-		}
+		log.Fatal(err)
```
didiladi, line 51: Isn't `log.Fatal` already exiting? Can you change it to `log.Error`?

cmd/monaco/main.go
```diff
@@ -44,380 +36,211 @@ func Run(args []string) int {
...
 }
 
-func parseInputCommand(args []string, fileReader util.FileReader) (dryRun bool, verbose bool, environments map[string]environment.Environment, project string, path string, errorList []error, flagError error) {
-
-	// define flags
-	var environmentsFile string
-	var specificEnvironment string
-	var versionFlag bool
+func buildCli(fileReader util.FileReader) *cli.App {
+	app := cli.NewApp()
```
didiladi, line 59: Please add `app.EnableBashCompletion = true` to enable the bash completion.

https://github.com/urfave/cli/blob/master/docs/v2/manual.md#bash-completion

cmd/monaco/main.go
```diff
@@ -44,380 +36,211 @@ func Run(args []string) int {
...
 
-	// parse flags
-	shorthand := " (shorthand)"
-	dryRunUsage := "Set dry-run flag to just validate configurations instead of deploying."
+	app.Version = version.MonitoringAsCode
```
didiladi, line 61: Don't you need to set `app.Name` here as well?

didiladi (commented):
Cool, awesome PR, thank you for the initiative to make the CLI better understandable to users! :+1: 

@dynatrace-oss/mac-maintainers I guess we need to agree on a release schedule for this. Deprecation of the old CLI in in 1.5 and replacement (breaking change) in 2.0 would be feasible. What do you think?

cmd/monaco/main.go
patrickpichler: Yeah I had a look at cobra and to be honest it seems way
to complicated for what it is trying to achieve.

cmd/monaco/main.go
patrickpichler: I didn't include it by default, as this needs some more ritual to get working:
https://github.com/urfave/cli/blob/master/docs/v2/manual.md#enabling

And I guess this will break by us switching the cli based on an env variable, 
as most people will use it like `NEW_CLI=1 ./monaco` (at least I assume this).

patrickpichler: But I can activate it if needed :+1:

cmd/monaco/main.go
patrickpichler: Should we always display this log message?

The issue I see with this is that this kinda breaks `--version` as it 
now prints more than just the version.

patrickpichler: But whoever depends on just the version printing can fix their
scripts via `./monaco --version | tail -1`. I guess this is not 
that big of a deal then.

didiladi: Hmm, right! I did not think of the version. Yeah, I guess other tools print more than the version as well. So you would always have to do some extracting/parsing afterwards. E.g. kubectl looks like that:

```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.2", GitCommit:"66049e3b21efe110454d67df4fa62b08ea79a19b", GitTreeState:"clean", BuildDate:"2019-05-16T16:23:09Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
The connection to the server 69C15CD5DF601CD3C6660C5439F1C18B.gr7.us-east-1.eks.amazonaws.com:5000 was refused - did you specify the right host o
```

patrickpichler: Also I used `fmt.Print` instead of `util.Log` to print this info, as I personally think
it adds no value to the log file and is quite annoying to have the logger prefixing
the date and log level before the message.

cmd/monaco/main.go
didiladi: Hmm, I should have read more than the first paragraph :smile:. It, would make sense to activate/document the completion in its own ticket. Maybe there is a way to implement it with some graceful handling.

patrickpichler: I guess a new ticket is the way to go, as there are too many unanswered questions. 

We could provide the completions for the old cli and once it is deprecated switch over
to the new. 

But I guess this would required two user interactions.

Another way would be to serve the completions via a command like `monaco completions bash`.
The user the would put something like `source $(monaco compeltions bash)` into their .bashrc`
and it would automatically switch over to the new completions once we update.

cmd/monaco/main.go
UnseenWizzard: If we don't set the name that seems to result in this: 

```
NAME:
   monaco - A new cli application
```

Is there a way we can get rid of the "A new cli application" without hardcoding the name?

UnseenWizzard: otherwise looks no

patrickpichler: Yes we simply have to set the `Usage`. What shall I set it to?

didiladi: Maybe the following:

> Automates the deployment of Dynatrace Monitoring Configuration to one or multiple Dynatrace environments.

like we have in the repo description?

patrickpichler: Usage text should be updated now :+1:

didiladi (dismissed):

UnseenWizzard (approved):

Issue #105, opened 2020-12-22
Handle breaking change in Dashboard API in version 1.208+


didiladi:
**Description**
There is a breaking change in version `1.208+` coming for the Dashboard API:

```
1) POST - /dashboards - Create a new dashboard

BEFORE:
• the owner property in the dashboardMetadata object within the request body must be null or undefined, otherwise the validation fails
• the sharingDetails property is present in dashboardMetadata

1.208+:
• It becomes mandatory to specify the owner property as a non-null value, otherwise the validation fails.
• dropped sharingDetails from dashboardMetadata
```

```
2) PUT - /dashboards/{id} - Update a dashboard

BEFORE:
• the owner property is not properly validated to always be not-null
• the sharingDetails property is present in dashboardMetadata

1.208+:
• the owner property will be always validated to be not-null
• dropped sharingDetails from dashboardMetadata
```

```
3) GET - /dashboards/{id} - Get a dashboard
BEFORE:
• the sharingDetails property is present in dashboardMetadata

1.208+:
• dropped sharingDetails from dashboardMetadata
```

Since this API is still in early-adoper state, its version will not be increased.

This is especially painful if the same configuration needs to be applied to environments in different versions. `monaco` should handle the apply of the config for the user in a graceful fashion. 

**Describe the solution you'd like**
If you try to apply a dasboard configuration with the old (pre `1.208`) format to an enviromnent which is on `1.208+`:
* `monaco` should insert a `owners` property into the json before throwing it to the API.
* the user should get a warning message, indicating that the configuration needs to be updated.

If you try to apply a dasboard configuration with the new (`1.208+`) format to an enviromnent which is still on an older version `pre 1.208`:
* `monaco` should remove the `owners` property from the json before throwing it to the API.

**Describe alternatives you've considered**
If you don't need to deploy configuration to multiple environments in different versions, you can fix your configurations like so:

Configuration yaml (note the double quotes `'` and `"` for the owner):

```
myDashboard:
  - name: "My Dashboard"
  - owner: '"my.email@example.com"'
```

Dashboard JSON (there are no quotes for `owners`):

```
{
  "dashboardMetadata": {
    "name": "{{ .name }}",
    "shared": true,
    "owner": {{ .owner }},
   ...
```

tatomir146:
@didiladi this would be the first time `monaco` modifies params on upload depending on cluster version?

didiladi:
[...]

PR #112, opened 2021-01-12
Implemented Monaco download functionality

diegorqc:
This will add the download functionality to monaco. It allows to specify a configuration to download an specific list of apis or download everything by default.
List of changes:
There is a README file to explain the basic usage.
Includes unit test for most of the code.

Once it's merge, we will need to check some missing pieces like replacing the id's of dependent configs and other small improvements.


cmd/monaco/main.go
```diff
@@ -61,7 +63,7 @@ func RunImpl(args []string, fileReader util.FileReader) (statusCode int) {
 
 	statusCode = 0
 
-	dryRun, verbose, environments, projectNameToDeploy, path, errorList, flagError := parseInputCommand(args, fileReader)
+	dryRun, verbose, environments, projectNameToDeploy, downloadFlag, downloadSpecificAPI, path, errorList, flagError := parseInputCommand(args, fileReader)
```
didiladi, line 66: Might make sense to create a `struct` which holds all the results. Retuning 9 arguments here becomes very hard to read.

cmd/monaco/main.go
```diff
@@ -270,7 +287,7 @@ func execute(environment environment.Environment, projects []project.Project, dr
 			if err != nil {
 				return err
 			}
-			name = config.GetApi().GetId() + "/" + name
+			name = filepath.Join(config.GetApi().GetId(), name)
```
didiladi, line 290: :1st_place_medal:

pkg/download/README.md
```diff
@@ -0,0 +1,22 @@
+# Download configuration
+
+This feature allows you to download the configuration from a Dynatrace tenant as Monaco files. You can use this feature to avoid starting from scratch when using Monaco.
```
didiladi, line 3: Please use the term `environment` instead of `tenant`. Environment is the term, which is used everywhere else.

pkg/download/download.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/environment"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/rest"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util/files"
+)
+
+var baseProjectName = "" //for folder naming
+var cont = 0
+
+//GetConfigs Entry point that retrieves the specified configurations from a Dynatrace tenant
```
didiladi, line 21: `environment`

pkg/download/download.go
```diff
@@ -0,0 +1,168 @@
...
+func GetConfigs(environments map[string]environment.Environment, path string, downloadSpecificAPI string) (statusCode int) {
+
+	list, err := getAPIList(downloadSpecificAPI)
+	if err != nil {
+		util.Log.Fatal("The API list contains invalid values. Run monaco command to see the available options\n" + err.Error())
```
didiladi, line 26: Isn't `util.Log.Fatal` exiting the program? Please use `util.Log.Error`.

pkg/download/download.go
```diff
@@ -0,0 +1,168 @@
...
+		statusCode = -1
+	}
+	if statusCode == -1 {
+		return statusCode
+	}
+	return 0
+}
+
+//returns the list of API filter if the download specific flag is used, otherwise returns all the API's
+func getAPIList(downloadSpecificAPI string) (map[string]api.Api, error) {
```
didiladi, line 56: It would make sense to return an array of errors here:

```
func getAPIList(downloadSpecificAPI string) (map[string]api.Api, []error) {
```

The concatenation of error messages (line 76) is probably not the most user-friendly approach for logging the errors raised in this method. It would lead to error messages like the following:

```
2020-12-03 11:12:09 ERROR Value foo is not a valid API name.
 Value bar is not a valid API name.
```

More readable would be:

```
2020-12-03 11:12:09 ERROR Value foo is not a valid API name.
2020-12-03 11:12:09 ERROR Value bar is not a valid API name.
```

pkg/download/download.go
```diff
@@ -0,0 +1,168 @@
...
+		if isApi == false {
+			result = false
+			errString += errString + fmt.Sprintf(" \t - Value %s is not a valid API name.\n", cleanAPI)
+		} else {
+			filterAPI := availableApis[cleanAPI]
+			path := transFormSpecialCasesAPIPath(filterAPI.GetId(), filterAPI.GetApiPath())
+			filterAPIList[cleanAPI] = api.NewApi(filterAPI.GetId(), path)
+		}
+	}
+	if result == false {
```
didiladi, line 83: You could check for the error array length `> 0` here. Then you could get rid of the `result` variable.

pkg/download/download.go
```diff
@@ -0,0 +1,168 @@
...
+		util.Log.Error("error retrieving token: %s %v", err)
+		return err
+	}
+	client, err := rest.NewDynatraceClient(environment.GetEnvironmentUrl(), token)
+	if err != nil {
+		util.Log.Error("error creating dynatrace client: %s %v", err)
+		return err
+	}
+	for _, api := range listApis {
+		util.Log.Info(" --- GETTING CONFIGS for %s", api.GetId())
```
didiladi, line 123: Is that debug output? I'm asking because of the uppercase spelling.

pkg/download/download.go
```diff
@@ -0,0 +1,168 @@
...
+		jcreator := jsoncreator.NewJSONCreator()
+		ycreator := yamlcreator.NewYamlConfig()
+		err = createConfigsFromAPI(api, token, creator, fullpath, client, jcreator, ycreator)
+		if err != nil {
+			util.Log.Error("error configs for api: %s %v", api.GetId(), err)
+		}
+	}
+	util.Log.Info("END downloading info %s", projectName)
```
didiladi, line 131: Debug output?

pkg/download/download.go
```diff
@@ -0,0 +1,168 @@
...
+	jcreator jsoncreator.JSONCreator, ycreator yamlcreator.YamlCreator) (err error) {
+	//retrieves all objects for the specific api
+	values, err := client.List(api)
+
+	if len(values) == 0 {
+		util.Log.Info("No elements for API %s", api.GetId())
+		return nil
+	}
+	subPath := filepath.Join(fullpath, api.GetId())
+	creator.CreateFolder(subPath)
```
didiladi, line 145: Please handle the error which can be returned by calling this method.

pkg/download/download.go
```diff
@@ -0,0 +1,168 @@
...
+
+	for _, val := range values {
+		util.Log.Debug("getting detail %s", val)
+		cont++
+		util.Log.Debug("REQUEST counter %v", cont)
+		name, filter, err := jcreator.CreateJSONConfig(client, api, val, creator, subPath)
+		if err != nil {
```
didiladi, line 152: Does it make sense to collect these errors and return them?

pkg/download/jsoncreator/json_creator.go
```diff
@@ -0,0 +1,115 @@
...
+	switch apiID {
+	case "dashboard":
+		if dat["dashboardMetadata"] != nil {
+			metadata := dat["dashboardMetadata"].(map[string]interface{})
+			if metadata["preset"] != nil && metadata["preset"] == true {
+				return true
+			}
+		}
+		return false
+	case "synthetic-location":
```
didiladi, line 104: Wouldn't be a `default: return false` enough? Why do you need the cases?

pkg/download/download.go
```diff
@@ -0,0 +1,168 @@
...
+		return err
+	}
+	client, err := rest.NewDynatraceClient(environment.GetEnvironmentUrl(), token)
+	if err != nil {
+		util.Log.Error("error creating dynatrace client: %s %v", err)
+		return err
+	}
+	for _, api := range listApis {
+		util.Log.Info(" --- GETTING CONFIGS for %s", api.GetId())
+		jcreator := jsoncreator.NewJSONCreator()
```
didiladi, line 124: That's a very elegant solution to encapsulate all json and yaml operations. :100:

didiladi (commented):
Cool, awesome work! I really like this changest! I especially like that you added a ton of unit tests for the download functionality. During my review, I just found some minor things, mostly cosmetic stuff. :1st_place_medal:

didiladi:
One more thing: In my opinion it makes sese to add an integration test, which runs against 2 "real" environments:
* Let's configure an environment with some test configs and download the configs from that and
* apply the configs to a separate, empty environment and assert for the config presence there. 
* After that, we can clean the configs from the second environment again

What do you think?

didiladi:
> The download functionality is somehow urgent for some of our clients, so maybe the integration test can be added in a second pull request. What do you think?

Sure, fine by me. :+1:

didiladi:
The integration tests failed because of the breaking API change in the dashboard API (#105). It has nothing to do with the changes for this PR.

```
2021-01-15 14:14:09 ERROR Deployment to environment2 failed with error Failed to upsert DT object Star Wars_1610720048410552328MultiEnvironment (HTTP 400)!
    Response was: {"error":{"code":400,"message":"Constraints violated.","constraintViolations":[{"path":"dashboardMetadata.owner","message":"may not be null","parameterLocation":"PAYLOAD_BODY","location":null}]}}, responsible config: test-resources/integration-multi-environment/star-wars/dashboard/overview-dashboard.json
2021-01-15 14:14:09 ERROR Deployment to environment1 failed with error Failed to upsert DT object Star Wars_1610720048410552328MultiEnvironment (HTTP 400)!
    Response was: {"error":{"code":400,"message":"Constraints violated.","constraintViolations":[{"path":"dashboardMetadata.owner","message":"may not be null","parameterLocation":"PAYLOAD_BODY","location":null}]}}, responsible config: test-resources/integration-multi-environment/star-wars/dashboard/overview-dashboard.json
```

pkg/download/README.md
diegorqc: fix wording for all environment mentions

pkg/download/download.go
diegorqc: thanks

pkg/download/jsoncreator/json_creator.go
diegorqc: yes, that's correct for now only default cases and dashboard

pkg/download/download.go
diegorqc: yes I don't think that the application should stop the download process if only one of the API is failing

pkg/download/download.go
diegorqc: done

pkg/download/download.go
diegorqc: not really, I have fix some of the log wording now

pkg/download/download.go
diegorqc: sure, looks a lot better, implemented []error array instead of the string concat

pkg/download/download.go
diegorqc: done

pkg/download/download.go
diegorqc: fixed

cmd/monaco/main.go
diegorqc: this was trickier, I have created the struct and it works fine but I couldn't implement a unit test for this one without refactoring a lot of the method, the flag.Parse() should be call outside this method for it to be able to be unit tested https://stackoverflow.com/questions/38711653/unit-test-different-flag-values or https://eli.thegreenplace.net/2020/testing-flag-parsing-in-go-programs/. I think that the code in main for the upload functionality should be moved to a new class and main should be refactor. If nobody is working on it I can create a new pull request in a couple of weeks to do this. Also saw the new experimental CLI but this would break the merge of this funcionality and it's still leaving upload code on main.go file

cmd/monaco/main.go
```diff
@@ -61,20 +63,20 @@ func RunImpl(args []string, fileReader util.FileReader) (statusCode int) {
 
 	statusCode = 0
 
-	dryRun, verbose, environments, projectNameToDeploy, path, errorList, flagError := parseInputCommand(args, fileReader)
+	input, err := parseInputCommand(args, fileReader)
```
UnseenWizzard, line 66: I think you're a rebase away from @patrickpichler 's input handling/CLI changes. 
I assume there will be merge conflicts in this file...? @patrickpichler can maybe help out with that if anything is too entangled

cmd/monaco/main.go
```diff
@@ -83,7 +85,15 @@ func RunImpl(args []string, fileReader util.FileReader) (statusCode int) {
...
 	apis := createApis()
-
+	if input.downloadFlag {
+		statusCode = download.GetConfigs(input.environments, input.path, input.downloadSpecificAPI)
+	} else {
+		statusCode = syncConfigs(input.dryRun, input.environments, input.projectNameToDeploy, apis, input.path, fileReader, deploymentErrors)
+	}
+	return statusCode
+}
+func syncConfigs(dryRun bool, environments map[string]environment.Environment, projectNameToDeploy string,
```
UnseenWizzard, line 95: That step should still be 'deployConfigs' (or similar) - we don't technically synchronize, we push the local config over anything on the cluster

pkg/download/README.md
```diff
@@ -0,0 +1,22 @@
+# Download configuration
+
+This feature allows you to download the configuration from a Dynatrace environment as Monaco files. You can use this feature to avoid starting from scratch when using Monaco.
+
+### steps
+1. Create an environment file. 
+2. Run the monaco command using the download flag ``-download``
```
UnseenWizzard, line 7: Generally we aim to document parameters as `--long-name` the way most CLI tools handle them, even if Go flags allows both/single `-`

pkg/download/download.go
UnseenWizzard: I agree with @diegorqc , as long as we ensure that all errors are logged in a usable manner at the end this is similar to how we handle upload.
Just cause one environment fails to download a single object, it's more user friendly (and realistic) to continue on for the X other configs and environments

UnseenWizzard: With that I would consider collecting and adding these errors to the one's you log with `publishErrors` though rather than logging them here

pkg/download/jsoncreator/json_creator.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, false, err
+	}
+	err = json.Unmarshal(resp, &dat)
+	if err != nil {
+		util.Log.Error("error transforming %s from json to object", name)
+		return nil, false, err
+	}
+	filter = isDefaultEntity(api.GetId(), dat)
+	if filter == true {
+		util.Log.Debug("Object has been filter", name)
```
UnseenWizzard, line 64: It might be helpful for debugging if we logged why it's been filtered.

Maybe something like 

```suggestion
		util.Log.Debug("Non-user-created default Object has been filtered out", name)
```

pkg/project/project_loader.go
```diff
@@ -32,7 +33,9 @@ import (
 // it also resolves all project dependencies
 // if no -p parameter specified, then it creates a list of all projects
 func LoadProjectsToDeploy(specificProjectToDeploy string, apis map[string]api.Api, path string, fileReader util.FileReader) (projectsToDeploy []Project, err error) {
-	projectsFolder := filepath.Join(".", path)
+	//projectsFolder := filepath.Join(".", path)
+	projectsFolder := path
+	//absolute path solution, test if the folder exist leave it as it is, otherwise add the dot
```
UnseenWizzard, line 38: This sounds like this is left over from some local test? 
Should this change go in or be dropped? 

If it should be merged, please make sure it's a dedicated commit, as it seems unrelated to the download functionality(?)

UnseenWizzard (commented):
Cool stuff @diegorqc! 
Let's just clear up the few open comments & make sure this is rebased to the current `main`

pkg/project/project_loader.go
diegorqc: Hi Nico Just implemented your suggestions alongside the integration with the new api 2.0 in a new brach with a new pull request since there were quite a few changes in main.go

PR #122, opened 2021-01-20
download feature implemented

diegorqc:
Download functionality implemented using CLI 2.0 with commands.
Main changes are:
Move deploy to it's own package to follow golang guidelines and avoid vscode errors.
Implemented download functionality using download command.
Add base documentation for download process


pkg/download/jsoncreator/json_creator.go
```diff
@@ -0,0 +1,115 @@
...
+	if err != nil {
+		util.Log.Error("error writing detail %s", api.GetId())
+		return "", false, err
+	}
+	return name, false, nil
+}
+
+func getDetailFromAPI(client rest.DynatraceClient, api api.Api, name string) (dat map[string]interface{}, filter bool, err error) {
+
+	name = strings.Replace(name, "%", "%25", -1) // Replace the real % character in the name with %25 (encoded version)
```
patrickpichler, line 51: Do we only want to encode `%` or url encode the whole string? 

If we want the whole string url encoded, we could use `url.QueryEscape`.

pkg/download/download.go
```diff
@@ -0,0 +1,183 @@
...
+		util.Log.Error("error creating dynatrace client: %s %v", err)
+		return err
+	}
+	for _, api := range listApis {
+		util.Log.Info(" --- GETTING CONFIGS for %s", api.GetId())
+		jcreator := jsoncreator.NewJSONCreator()
+		ycreator := yamlcreator.NewYamlConfig()
+		errs := createConfigsFromAPI(api, token, creator, fullpath, client, jcreator, ycreator)
+		if len(errs) > 0 {
+			publishErrors(errs)
```
UnseenWizzard, line 144: Hm, if you publish them here they will also be logged per environment, while any individual errors (like creating client) will be returned and then logged in bulk on line 55

What I meant was to return a list of errors from this method, to include in the bulk log

cmd/monaco/main.go
```diff
@@ -213,56 +215,105 @@ Examples:
...
+				ctx.Bool("dry-run"),
+				ctx.Bool("verbose"),
+			)
 		},
 	}
-
-	return app
+	return command
+}
+func getDownloadCommand(fileReader util.FileReader) cli.Command {
```
patrickpichler, line 275: Hurray finally some more commands to see how cool the new cli is :laughing:

pkg/download/jsoncreator/json_creator_util.go
```diff
@@ -0,0 +1,15 @@
+package jsoncreator
```
patrickpichler, line 1: I guess this file should also be tagged with `unit` and `integration`

pkg/project/project_loader.go
```diff
@@ -32,6 +33,12 @@ import (
 // it also resolves all project dependencies
 // if no -p parameter specified, then it creates a list of all projects
 func LoadProjectsToDeploy(specificProjectToDeploy string, apis map[string]api.Api, path string, fileReader util.FileReader) (projectsToDeploy []Project, err error) {
+	// projectsFolder := ""
```
UnseenWizzard, line 36: Comment from the other PR still stands. 
Please drop the commented out code, or actually put it in (as it's own commit) if it's a necessary bugfix for something :)

patrickpichler (changes requested):

pkg/util/files/creator.go
```diff
@@ -0,0 +1,75 @@
...
+func NewDiskFileCreator() FileCreator {
+	el := &fileCreator{}
+	el.fileManager = afero.NewOsFs()
+	return el
+}
+
+//CreateFolder creates a folder in the specified path
+func (a *fileCreator) CreateFolder(path string) (fullpath string, err error) {
+	//path should be sanitized
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
UnseenWizzard: Having taken a quick look at the library it does indeed seem like it might be a good way to replace the self-made abstraction layer in files.go

At a glance it seems to have some nice features in terms of testing. 

@didiladi what's your take?

From my side: If we intend to use it more thoroughly it would seem reasonable to add it.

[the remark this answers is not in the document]
didiladi: It makes sense to simplify/replace the abstraction layer in `files.go` if all use-cases like, e.g. in-memory filesystem are also supported in `afero`. But that would need some more investigation. As we don't want to block @diegorqc's PR here, for now, I'd say, its good as it is - let's revisit that in a followup ticket. :) Everyone ok with that?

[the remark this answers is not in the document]
patrickpichler: Sound good to me :+1:

pkg/download/download.go
```diff
@@ -0,0 +1,183 @@
...
+	requestedApis := strings.Split(downloadSpecificAPI, ",")
+	result := true
+	errString := ""
+
+	for _, id := range requestedApis {
+		cleanAPI := strings.TrimSpace(id)
+		isAPI := api.IsApi(cleanAPI)
+		if isAPI == false {
+			result = false
+			errString += errString + fmt.Sprintf(" \t - Value %s is not a valid API name.\n", cleanAPI)
```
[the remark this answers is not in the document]
diegorqc: yes you are correct, fixing that in a new commit

pkg/api/api_util.go
```diff
@@ -0,0 +1,14 @@
+package api
```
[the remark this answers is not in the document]
diegorqc: added

pkg/download/download.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/api"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/download/jsoncreator"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/download/yamlcreator"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/environment"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/rest"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util/files"
+)
+
+var baseProjectName = "" //for folder naming
```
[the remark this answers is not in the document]
diegorqc: I have removed that, initially it was suppose to provide a base project folder but it's no longer required

pkg/download/download.go
diegorqc: after thinking about it, since the other components are logging errors when being called, I have change the code to avoid using any bulk log otherwise the logs for the other components would appear in one place and the main logs in another

pkg/download/jsoncreator/json_creator_util.go
diegorqc: I've tried adding tags before the package definition but then vscode automatic test detection fails to detect anything that doesn't start with package in the first line. Not sure how are you running the test locally, but I find really useful the visual studio code intelisense for golang to speed development. I might have to google more to find a workaround for vscode or github tags

pkg/project/project_loader.go
diegorqc: done, I will add another pr just for this since this would have it's own unit test added. It's going to be a fix to support absolute paths, right now I found that that line blocks abs path usage

pkg/util/files/creator.go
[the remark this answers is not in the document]
diegorqc: yes ``afero`` is a good library to mock the filesystem, it provides some good examples about how to mock everything and make it unit testable. I have modified line 34 to use ``afero.Exist``

pkg/download/jsoncreator/json_creator_util.go
patrickpichler: I guess this is caused by golang-lsp ignoring all tagged files by default.

Here is the issue for it. It also contains info how to specify tags it should use. 
https://github.com/golang/go/issues/29202

patrickpichler (approved):

didiladi (approved):

UnseenWizzard (dismissed):

didiladi:
@diegorqc Looks like the integration tests failed with a compile error:

```
pkg/util/files/creator_integration_test.go:12:2: undefined: fileCreateTest
....
FAIL	github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util/files [build failed]
```

https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/runs/1748210174?check_suite_focus=true

didiladi (approved):

PR #140, opened 2021-01-27
Always transform references to absolute paths

### LINKED ISSUE #126: Reference evaluation changes when using -p= in command
Hi,

When I run Monaco, I get issues with references. (I also don't understand how they are supposed to be done properly, documentation lacks a good example/explanation):

If I add -p=template, I get issues, if I omit it and adapt the path, it works without issues:

```
# monaco -e=environments.yaml -p=template perform/monaco/exercise-5/projects
2021-01-20 15:39:57 INFO  Dynatrace Monitoring as Code v1.0.1
2021-01-20 15:39:57 INFO  Executing projects in this order: 
2021-01-20 15:39:57 INFO  	1: perform/monaco/exercise-5/projects/template (2 configs)
2021-01-20 15:39:57 INFO  Processing environment perform...
2021-01-20 15:39:57 INFO  	Processing project perform/monaco/exercise-5/projects/template...
2021-01-20 15:39:57 ERROR Deployment to perform failed with error Id 'application/selling-app-one' was not available. Please make sure the reference exists.
```

```
# monaco -e=environments.yaml perform/monaco/exercise-5/projects/template
2021-01-20 15:40:07 INFO  Dynatrace Monitoring as Code v1.0.1
2021-01-20 15:40:07 INFO  Executing projects in this order: 
2021-01-20 15:40:07 INFO  	1: perform/monaco/exercise-5/projects/template (2 configs)
2021-01-20 15:40:07 INFO  Processing environment perform...
2021-01-20 15:40:07 INFO  	Processing project perform/monaco/exercise-5/projects/template...
2021-01-20 15:40:07 INFO  Deployment finished without errors

```
File paths are:

```
perform/monaco/exercise-5/projects
perform/monaco/exercise-5/projects/template
perform/monaco/exercise-5/projects/template/app-detection-rule
perform/monaco/exercise-5/projects/template/app-detection-rule/rule.json
perform/monaco/exercise-5/projects/template/app-detection-rule/rule.yaml
perform/monaco/exercise-5/projects/template/application
perform/monaco/exercise-5/projects/template/application/application.yaml
perform/monaco/exercise-5/projects/template/application/application.json
```

Involved yaml files are:

```
$ cat perform/monaco/exercise-5/projects/template/app-detection-rule/rule.yaml
config:
  - selling-app-one-rule: "rule.json"
  
selling-app-one-rule:
  - name: "application/selling-app-one.name"
  - applicationId: "application/selling-app-one.id"
  - pattern: "https://selling-app-one.corp"

$ cat perform/monaco/exercise-5/projects/template/application/application.yaml
config:
    - selling-app-one: "application.json"
  
selling-app-one:
    - name: "selling-app-one"
```

patrickpichler:
In order to ease the handling of dependencies, references are now always
transformed to absolute paths. That way one can be sure that if he
encounters a cross reference, it is absolute.

Fixes #126

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -237,13 +237,13 @@ func (c *configImpl) replaceDependencies(data map[string]map[string]string, dict
 }
 
 func (c *configImpl) parseDependency(dependency string, dict map[string]api.DynatraceEntity) (string, error) {
-	absolutePath := false
+	absoluteFromProjectRoot := false
 
 	// in case of an absolute path within the dependency:
 	if strings.HasPrefix(dependency, string(os.PathSeparator)) {
 		// remove prefix "/"
 		dependency = dependency[1:]
-		absolutePath = true
+		absoluteFromProjectRoot = true
 	}
 
 	id, access, err := splitDependency(dependency)
@@ -251,7 +251,7 @@ func (c *configImpl) parseDependency(dependency string, dict map[string]api.Dyna
 		return "", err
 	}
 
-	if !absolutePath {
+	if !absoluteFromProjectRoot {
 		id = filepath.Join(c.project, id)
 	}
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 
@@ -236,17 +237,24 @@ func (c *configImpl) replaceDependencies(data map[string]map[string]string, dict
 }
 
 func (c *configImpl) parseDependency(dependency string, dict map[string]api.DynatraceEntity) (string, error) {
+	absolutePath := false
 
 	// in case of an absolute path within the dependency:
 	if strings.HasPrefix(dependency, string(os.PathSeparator)) {
 		// remove prefix "/"
 		dependency = dependency[1:]
+		absolutePath = true
 	}
 
 	id, access, err := splitDependency(dependency)
 	if err != nil {
 		return "", err
 	}
+
+	if !absolutePath {
+		id = filepath.Join(c.project, id)
+	}
+
 	dtObject, ok := dict[id]
 	if !ok {
 		return "", errors.New("Id '" + id + "' was not available. Please make sure the reference exists.")
@@ -372,7 +380,7 @@ func (c *configImpl) GetFilePath() string {
 
 // GetFullQualifiedId returns the full qualified id of the config based on project, api and config id
 func (c *configImpl) GetFullQualifiedId() string {
-	return strings.Join([]string{c.GetProject(), c.GetApi().GetId(), c.GetId()}, string(os.PathSeparator))
+	return filepath.Join(c.GetProject(), c.GetApi().GetId(), c.GetId())
 }
 
 // NewConfig creates a new Config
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -228,17 +229,24 @@ func (c *configImpl) replaceDependencies(data map[string]map[string]string, dict
 }
 
 func (c *configImpl) parseDependency(dependency string, dict map[string]api.DynatraceEntity) (string, error) {
+	absolutePath := false
```
UnseenWizzard, line 232: can we not call this `absolutePath` ? These are still relative/absolute only from the project root, aren't they?

patrickpichler: Yeah but they are absolute from the project root. But yeah, if you have a better
word for it, I'm open for suggestions :+1:

UnseenWizzard:
Also doesn't this in part relate generally to how we deal with absolute paths, relative paths, single projects and the root project folder? 
(AKA is this fix "enough" or just a band-aid and we still have a bigger underlying issue?)

pkg/config/config.go
UnseenWizzard: `relativelyAbsolutePath` :P 
as we have general issues with absolute and relative path handling I'd prefer not adding confusion by naming things wrong

patrickpichler: What is the official name for such references? 
We were always talking about absolute paths, so this kinda burned into
my head. But if you tell me that we actually call it something else, then 
yeah sure :laughing:

patrickpichler:
I guess this is a first step into the right direction from my point of view. But I could
be completely wrong. 

I personally find it very confusing to have non consistent "paths" for references in our
configuration. 

One big issue which should be addressed sometime is to normalize the use of `/` and 
`\` for path separators. But this is a different topic.

pkg/config/config.go
UnseenWizzard: "official name" as in official in our concept? 
"officially" relative and absolute paths are pretty clear cut in their definitions, and I wouldn't want us to add confusing terminology to the already confusing (thus currently very brittle) handling of paths in monaco

I guess in this case it's `inAbsolutePathFormat` - but monaco does not use it (or even function) if it's an actual absolute path from os root /

UnseenWizzard:
> I personally find it very confusing to have non consistent "paths" for references in our
> configuration.

I agree, I just wanted to stress that we do have the underlying bug of not being able to load projects and dependencies from absolute paths. 
So I feel like moving now to having all references as `relativelyAbsolute`/ `absoluteFromProjectRoot` might be a step we need to revert once properly resolving that. 

---

What I don't fully understand: Does this fix a bug?
AKA is #126 a bug in how dependencies are resolved and this fixes that, or is #126 an error in the user's config and this change will improve the quality of the error they see?

patrickpichler:
This actually also fixes the issue that an absolute path as working directory breaks the tool.

We should really come up with a better name than relative/absolute path for those...
 @didiladi @tatomir146 any good idea?

patrickpichler:
After doing further tests i notice that this still has issues. 

One issue is that we mix path of project and project id. This causes references to fail with the
changes in this PR.

pkg/config/config.go
didiladi: I like `absoluteFromProjectRoot`! :+1:

patrickpichler: I renamed the variable to `absoluteFromProjectRoot`

patrickpichler:
I'm closing this PR, since it was more of a prototype anyway.

PR #144, opened 2021-01-29
Fix #66: Add SLO API to monaco

### LINKED ISSUE #66: Support new SLO API
The new Dynatrace SLO API started as a preview feature in version 204. Customers, who are part of the preview should be able to leverage the SLO API using Monitoring as Code. Furthermore, when the feature goes GA, we already support it in monaco.

Add support for the new SLO API and add a configuration used for the all-configs integration test.

didiladi:
Fix #66: Add SLO API to monaco

This commit adds support for the new SLO API to monaco.

This required some additional logic in regards to:
  * paging,
  * id determination,
  * and response parsing

Paging:

The SLO API is the first API, which supports paging. This required to
add paging support to `monaco`. While doing that, I realized that the
extensions API is already paginated - so I activated it there as well.

ID determination:

Previously, we mostly used a concrete type for unmarshalling the json
response to determine the id of a config on `POST`. However, the SLO API
leverages the `Location` header of the response to return the id of the
newly created config.

Response Parsing:

Upon adding support for the `Location` header, I realized that a
lot of other APIs use this header as well to return the id of the
created configuration on `POST`. That enabled me to simplify the
response handling.


didiladi:
Integration tests currently fail because of the missing token scope for the SLO read/write.

PR #149, opened 2021-02-01
support for mobile applications

kristofre:
Added support for the `applications/mobile` endpoint.
- Added endpoint definition in api.go
- Added test files in `integration-all-configs` project

Added a config type for the `applications/web`endpoint
- This already existed before as `application`, but now it is also there as `application-web`
- Added a warning message that with v2.0.0 the `application` config type will be deprecated if it is being used

For issue https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/issues/146

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -41,6 +41,7 @@ var apiMap = map[string]string{
 	// Environment API not Config API
 	"synthetic-monitor":  "/api/v1/synthetic/monitors",
 	"application":        "/api/config/v1/applications/web",
+	"application-mobile": "/api/config/v1/applications/mobile",
 	"app-detection-rule": "/api/config/v1/applicationDetectionRules",
 	"aws-credentials":    "/api/config/v1/aws/credentials",
 	// Early adopter API !
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -39,8 +39,10 @@ var apiMap = map[string]string{
 	"synthetic-location": "/api/v1/synthetic/locations",
 	// Early adopter API !
 	// Environment API not Config API
-	"synthetic-monitor":  "/api/v1/synthetic/monitors",
+	"synthetic-monitor": "/api/v1/synthetic/monitors",
+	// To be deprecated in v2.0.0, replaced by application-web
 	"application":        "/api/config/v1/applications/web",
+	"application-web":    "/api/config/v1/applications/web",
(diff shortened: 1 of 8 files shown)


pkg/api/api.go
```diff
@@ -41,6 +41,7 @@ var apiMap = map[string]string{
 	// Environment API not Config API
 	"synthetic-monitor":  "/api/v1/synthetic/monitors",
 	"application":        "/api/config/v1/applications/web",
```
didiladi, line 43: :crying_cat_face: I guess we missed the opportunity to call this `application-web`.

didiladi (commented):
Awesome! :rocket: Please also add the new API to the list of supported APIs in the README.md. Thanks!

pkg/api/api.go
kristofre: Didn't want to cause a breaking change in an intermediary release. Perhaps a candidate for v2.0.0?

kristofre:
> Awesome! 🚀 Please also add the new API to the list of supported APIs in the README.md. Thanks!

Done :-)!

pkg/api/api.go
didiladi: :+1: Yeah, for 2.0 we could deprecate `application` and print a warning it its still used - and introduce `application-web` additionally.

didiladi (dismissed):

pkg/api/api.go
kristofre: added application-web endpoint

didiladi (dismissed):

pkg/project/project.go
```diff
@@ -155,6 +155,12 @@ func (p *projectBuilder) processConfigSection(properties map[string]map[string]s
 
 		location = p.standardizeLocation(location, folderPath)
 
+		//Introduce deprecation warning message when using the config type "application"
+		util.Log.Debug("Config type: " + getConfigType(folderPath))
```
didiladi, line 159: It would be better to check the `api`, which is returned from `getExtendedInformationFromLocation`. Then you don't need to parse the API from the folder path.

UnseenWizzard:
Is this lacking a rebase to main, based on another PR, or purposely containing three CLI related commits by @patrickpichler ?

patrickpichler:
Long story :laughing: 
But those should not be in there, the branch is missing a rebase

didiladi (approved):

PR #150, opened 2021-02-01
Use map[]interface{} for json unmarshalling

### LINKED ISSUE #101: Use interface{} (or map[string]interface{}) instead of string for json content
@patrickpichler gave an awesome feedback [here](https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/pull/88#discussion_r545614848):

> Why do we unmarshal the json as a string?
> This seems a bit odd to me. Wouldn't something likeinterface{} be better?

Replace the json string with `interface{}` or `map[string]interface{}`. This makes the extraction of elements easier, and its more efficient. This makes the following possible:

```
var result map[string]interface{}
json.Unmarshal([]byte(empJson), &result)

stuff := result["stuff"].(map[string]interface{})
foo := stuff["foo"]
```

patrickpichler:
I guess this fixes #101 

As a nice side-effect it seems to fix the pretty printing for json errors. 
For whatever reason it was broken.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -31,7 +31,7 @@ import (
 //go:generate mockgen -source=config.go -destination=config_mock.go -package=config Config
 
 type Config interface {
-	GetConfigForEnvironment(environment environment.Environment, dict map[string]api.DynatraceEntity) (string, error)
+	GetConfigForEnvironment(environment environment.Environment, dict map[string]api.DynatraceEntity) (map[string]interface{}, error)
 	IsSkipDeployment(environment environment.Environment) bool
 	GetApi() api.Api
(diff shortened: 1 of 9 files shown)


didiladi:
:thinking: I don't understand the change in the json error pretty print. Does that mean that it broke in between / or even worse_ never worked despite being tested _a lot_?

I don't think that this just yet fixes #101 because we have a lot of places left, which treat and pass around the json as a string. Is the plan to tackle that as well?

patrickpichler:
I do not understand the json pretty print thing either... 
I tested it on main and it was broken... this fixes it.

patrickpichler:
As for the json thing, yes, I still plan on tackling this. 
This was just the first place i found it :laughing:

didiladi (dismissed):

didiladi (approved):

Issue #128, opened 2021-01-22
monaco init command for project initialisation 


kristofre:
Common feedback about Monaco is that people don't always know how to get started with the folder structure.

A `monaco init` command should be available that creates:

- Sample environments.yaml file
- Project directory
- For each supported config type, a folder should be created with a base yaml in there
- After the command executes it should print the exact command needed to apply that project

Dynatrace-Adam-Gardner:
Would empty files be OK inside the folders? If so, I have some code.

kristofre:
@Dynatrace-Adam-Gardner this is great stuff! For the yaml file, do you have the base structure there as well, something like this:
`config:
  - application: "application.json"

application:
  - name: "application-one"`

That would make it super easy.
If then that application.json also exists, all the user has to do is paste in the json body from the api explorer and they are good to go!

Dynatrace-Adam-Gardner:
Done, this now creates a YAML file with content like:

```
config:
- alerting-profile: alerting-profile.json
alerting-profile:
- name: alerting-profile-one
```

kristofre:
From the above it is not clear if it creates a sample `environments.yaml` as well?

Dynatrace-Adam-Gardner:
It doesn't yet, but I'll add it

Dynatrace-Adam-Gardner:
This is now complete. It generates files and folders for all APIs returned by the `api.NewApis()` which references the `apiMap` in `/pkg/api/api.go` so new APIs will automatically get folders and content.

One important note: You must delete folders that you're not using otherwise monaco will error.

Perhaps it's worth ignoring blank files? I'll start a new discussion for this.

PR #162, opened 2021-02-03
Feature/monacoinit

Dynatrace-Adam-Gardner:


Dynatrace-Adam-Gardner:
#128

Dynatrace-Adam-Gardner:
@patrickpichler I need a unit test for this. As this functionality creates folders and files on disk, how can I test that functionality without _actually_ creating the content? Is there another unit test sample I can use for inspiration?

pkg/initialise/initialise.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util/files"
+)
+
+//CreateTemplate Creates a blank set of monaco folders and files for each supported API
+func CreateTemplate(workingDir string) error {
+
+	workingDir = filepath.Clean(workingDir)
+	util.Log.Info("Initialising Monaco Demo Folders")
+	projectsFolder := workingDir + "/projects"
```
patrickpichler, line 17: Please use `filepath.Join` here, as this would not work on windows, as they use a different 
path separator `\`. 

(It would be best to replace all string concatenations with `filepath.Join`)

patrickpichler:
As for the unit tests: You could create a second `CreateTemplate` method, which takes the file creator as parameter. 
This method you then could test (there is this `NewInMemoryFileCreator()` method which doesn't 
persist to disk). 
The `CreateTemplate` method which exists right now, would then call the other method with an `NewDiskFileCreator`. 
Does this make sense?

patrickpichler:
On a side note, it seems like master got merged again into this branch.
It would be really nice if we could rebase instead of merge. 
(I have no idea how to do this with the github gui thinggy though :laughing: )

Dynatrace-Adam-Gardner:
> As for the unit tests: You could create a second `CreateTemplate` method, which takes the file creator as parameter.
> This method you then could test (there is this `NewInMemoryFileCreator()` method which doesn't
> persist to disk).
> The `CreateTemplate` method which exists right now, would then call the other method with an `NewDiskFileCreator`.
> Does this make sense?

Actually, all I'm doing is creating files and folders. Isn't this covered by the tests in `util.files.creator_test.go` ?

patrickpichler:
I guess in this case it is ok not to write a test. 

Or what do you say @didiladi @UnseenWizzard @tatomir146 ?

Dynatrace-Adam-Gardner:
Commit `4f94274` adds the `-c` or `--createSpecificAPI` flag to the `monaco init` command which will then only create dummy folders for those APIs

```
monaco.exe init --createSpecificAPI auto-tag,dashboard
```

Please note that I've copy pasted the `getAPIList` function from `downloads.go` which causes duplicate code. Perhaps we should pull this function out into Utils? It seems to be a useful general function.

cmd/monaco/main.go
```diff
@@ -336,3 +344,40 @@ func getDownloadCommand(fileReader util.FileReader) cli.Command {
...
+			&cli.StringFlag{
+				Name:    "createSpecificAPI",
+				Usage:   "Comma separated list of API's to create",
+				Aliases: []string{"c"},
+			},
+		},
+		Action: func(ctx *cli.Context) error {
+			var workingDir string
+
+			if ctx.Args().Present() {
```
patrickpichler, line 373: I guess we should handle the case somebody provides more than one argument. 
If you ask me, we should then return an error telling the user only to provide one.

pkg/initialise/initialise.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util/files"
+)
+
+//CreateTemplate Creates a blank set of monaco folders and files for each supported API
+func CreateTemplate(workingDir string, createSpecificAPI string) error {
+
+	util.Log.Info("Creating Config for APIs: %v", createSpecificAPI)
+
+	apiList, getAPIListErr := getAPIList(createSpecificAPI)
```
patrickpichler, line 19: Small nitpick: Normally in go you would name the error variable just `err`

pkg/initialise/initialise.go
```diff
@@ -0,0 +1,129 @@
...
+	if getAPIListErr != nil {
+		return getAPIListErr
+	}
+
+	workingDir = filepath.Clean(workingDir)
+	util.Log.Info("Initialising Monaco Demo Folders")
+	projectsFolder := filepath.Join(workingDir, "projects")
+	demoProjectFolder := filepath.Join(workingDir, "projects", "baseconfig")
```
patrickpichler, line 27: It might make sense to provide the initial project name also via an command
line argument.

pkg/initialise/initialise.go
```diff
@@ -0,0 +1,129 @@
...
+				*  - name: folderName-one
+				*
+				* For example:
+		        * config:
+		        * - alerting-profile: alerting-profile.json
+		        * alerting-profile:
+		        * - name: alerting-profile-one
+		*/
+		config.AddConfig(folderName, folderName+"-one")
+		fileCreator := files.NewDiskFileCreator()
```
patrickpichler, line 85: There should be no need to create a new `fileCreator` as you already use one
for creating folders.

pkg/initialise/initialise.go
Dynatrace-Adam-Gardner: I disagree. In this case, it's probably best to keep the commands as simple as possible. The folder can (and probably should) be renamed after the dummy content.

cmd/monaco/main.go
Dynatrace-Adam-Gardner: Happy to implement if only I knew how :)

pkg/initialise/initialise.go
patrickpichler: I would take it as an optional argument though. So by default call it `baseconfig`
but when i provide param `n` (or whatever) it i want to override it.

cmd/monaco/main.go
patrickpichler: Have a look at how it is done on line 263 :)

patrickpichler: Sorry I meant that you test if there are more than one argument present.

Having one argument to override the working dir is actually quite good.

pkg/initialise/initialise.go
```diff
@@ -0,0 +1,128 @@
...
+	demoProjectFolder := filepath.Join(workingDir, "projects", "baseconfig")
+
+	creator := files.NewDiskFileCreator()
+
+	// Create environments.yaml file
+	environmentsContent := `environment1:
+		    - name: "environment1"
+		    - env-url: "{{ .Env.ENVIRONMENT_ONE }}"
+		    - env-token-name: "{{ .Env.TOKEN_ENVIRONMENT_ONE }}"`
+	_, err = creator.CreateFile([]byte(environmentsContent), workingDir, "environments", ".yaml")
```
tatomir146, line 36: Will existing `environments.yaml` be overwritten?

pkg/initialise/initialise.go
```diff
@@ -0,0 +1,128 @@
...
+		util.Log.Error("Error creating top level projects folder. %v - %v", fullpath, err)
+		return err
+	}
+	util.Log.Info("Created Folder: %v", demoProjectFolder)
+
+	// For each allowed API, create the relevant folder
+
+	for folderName := range apiList {
+		configTypeFolderPath := filepath.Join(demoProjectFolder, folderName)
+		creator.CreateFolder(configTypeFolderPath)
```
tatomir146, line 63: If this fails, should there be a auto cleanup? We are are fine with partially created projects?

tatomir146 (commented):
> Please note that I've copy pasted the `getAPIList` function from `pkg/download/download.go` which causes duplicate code. Perhaps we should pull this function out into Utils? It seems to be a useful general function.

We should prevent duplicated code. I would vote for adding this method to the `api.go`. @didiladi @patrickpichler any thoughts on this?

cmd/monaco/main.go
```diff
@@ -336,3 +344,40 @@ func getDownloadCommand(fileReader util.FileReader) cli.Command {
...
+				return err
+			}
+
+			util.Log.Info("Dynatrace Monitoring as Code v" + version.MonitoringAsCode)
+
+			return nil
+		},
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:    "createSpecificAPI",
```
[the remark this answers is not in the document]
Dynatrace-Adam-Gardner: Good point. Decide whether `download-specific-api` will stay like that or also be `specific-api` as the same argument can be made with the `download` command. Whatever you decide for `download` I'll mirror for consistency with `init`.

cmd/monaco/main.go
```diff
@@ -336,3 +344,40 @@ func getDownloadCommand(fileReader util.FileReader) cli.Command {
...
+				Usage:   "Comma separated list of API's to create",
+				Aliases: []string{"c"},
```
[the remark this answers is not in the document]
Dynatrace-Adam-Gardner: `s` ?

cmd/monaco/main.go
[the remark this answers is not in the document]
Dynatrace-Adam-Gardner: Not sure where my previous comment went. Good point. Whatever you decide for `download`, let me know and I'll mirror with `init` for consistency.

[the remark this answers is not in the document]
tatomir146: Change is already suggested in #179.

cmd/monaco/main.go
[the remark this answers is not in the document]
tatomir146: `s` is also already reserved :( I think `a` would work?
```
			&cli.StringFlag{
				Name:    "specific-environment",
				Usage:   "Specific environment (from list) to deploy to",
				Aliases: []string{"s"},
			},
```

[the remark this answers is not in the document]
patrickpichler: Actually this should not be an issue, as those flags are command scoped.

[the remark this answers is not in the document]
patrickpichler: I guess the `continue-on-error` flag doesn't really make sense in the context of `init`.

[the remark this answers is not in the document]
didiladi: Why `-a`? :thinking: It would be cool to have a list of used flags and their aliases. I believe its confusing to the user if a given alias is used in two commands with two different meanings.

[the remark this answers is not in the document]
patrickpichler: But then we are limited to 24 flags :laughing:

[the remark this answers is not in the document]
patrickpichler: Actually @Dynatrace-Adam-Gardner the name should be in kebab case (I guess it is called that way :laughing: ) `create-specific-api`.

pkg/initialise/initialise.go
```diff
@@ -0,0 +1,128 @@
...
+		util.Log.Info("  Created File: %v/%v%v", configTypeFolderPath, folderName, ".yaml")
+		creator.CreateFile(nil, configTypeFolderPath, folderName, ".json")
+		util.Log.Info("  Created File: %v/%v%v", configTypeFolderPath, folderName, ".json")
+	}
+
+	return nil
+}
+
+//returns the list of API filter if the download specific flag is used, otherwise returns all the API's
+func getAPIList(createSpecificAPI string) (filterAPIList map[string]api.Api, err error) {
```
didiladi, line 102: As @tatomir146 said, please move this function to `api.go`, rename it to `GetAPIList` and use it here and in the download functionality.

pkg/initialise/initialise.go
```diff
@@ -0,0 +1,128 @@
...
+		/* Create file as "folderName.yaml"
+		 * eg. alerting-profile.yaml
+		 */
+		err := config.CreateYamlFile(creator, configTypeFolderPath, folderName)
+		if err != nil {
+			util.Log.Error("Error creating YAML file %v", err)
+		}
+
+		util.Log.Info("  Created File: %v/%v%v", configTypeFolderPath, folderName, ".yaml")
+		creator.CreateFile(nil, configTypeFolderPath, folderName, ".json")
```
didiladi, line 94: So, if I understand correctly, the json file is empty?

Not sure about the effort, but is it possible to fill it with the dummy content we have in the integration test resources (all-config-integration-test)? https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/tree/main/cmd/monaco/test-resources/integration-all-configs/project

Then we would automatically have a working dummy setup. What do you think?

didiladi (commented):
In general: Very nice idea and PR. :rocket: I like it!

cmd/monaco/main.go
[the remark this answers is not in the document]
didiladi: @patrickpichler yes, I know. But that would be consistent 24 flags :smile:. At least we should not use "commonly known flags" for different purposes. E.g.: 

`-v` verbose
`-e` environments
`-p` projects
`-c` config or cluster :thinking: 

ahh, and now I understand @tatomir146:
`-a` for api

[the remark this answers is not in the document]
patrickpichler: I like that idea :+1: 
Otherwise we end up in flag hell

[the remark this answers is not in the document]
tatomir146: @didiladi yeah, I think we should maybe use `-api` (not `specific` api). Let's proceed discussion here #182

pkg/initialise/initialise.go
Dynatrace-Adam-Gardner: Currently yes but the workflow here is to start from scratch so we assume nothing exists yet

pkg/initialise/initialise.go
Dynatrace-Adam-Gardner: The YAML file is preconfigured (for example):
```
config:
- auto-tag: auto-tag.json
auto-tag:
- name: auto-tag-one
```
But I think placing actually working content into the JSON is a bad idea - IMO people will undoubtedly push that config and we'll end up with misconfigured tenants.

I could add some text in their to make it clearer that they need to populate the file. For example: `Placeholder. Replace with your JSON content.`

didiladi: You are right, that would just be random configs, which would pollute to the environment of the user.

I really like the idea of the placeholder text - then its immediately clear to the user that the content of the json needs to be replaced.

pkg/initialise/initialise.go
Dynatrace-Adam-Gardner: Done with `b71614a`

pkg/initialise/initialise.go
Dynatrace-Adam-Gardner: Done in `b48d8e1`

patrickpichler:
@Dynatrace-Adam-Gardner what is the actual state of this?
Do you need help with anything?

patrickpichler:
Closed due to inactivity

PR #167, opened 2021-02-04
Added option to allow deployment to proceed even if error occurred

### LINKED ISSUE #127: Failing gracefully on upload error
Currently if for some reason (e.g. breaking API change, config in someone else's responsibility...) deployment of a single config (except Plugins) fails, the whole job is stopped. This makes it impossible to deploy new configs before solving a problem.

If config upload files (response other than 2xx) deployment should proceed and try to upload other configurations. If there are dependencies to the configuration failing to upload, they will also fail gracefully. 

**Additional context**
IMO this shouldn't be a standard upload case, but we could add support with a flag (--force, --follow-on-error or something similar)

tatomir146:
- resolved #127 
- supporting --continue-on-error (-c) parameter
- prints all errors on deployment summary
- dry-run reports all config problems now, not only the first one
- added support to validate single configuration in integration tests


cmd/monaco/continue_deployment_on_error_integration_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	RunIntegrationWithCleanup(t, allConfigsFolder, allConfigsEnvironmentsFile, "AllConfigs", func(fileReader util.FileReader) {
+
+		environments, errs := environment.LoadEnvironmentList("", allConfigsEnvironmentsFile, fileReader)
+		assert.Check(t, len(errs) == 0, "didn't expect errors loading test environments")
+
+		projects, err := project.LoadProjectsToDeploy("", api.NewApis(), allConfigsFolder, fileReader)
+		assert.NilError(t, err)
+
+		os.Setenv("NEW_CLI", "1")
```
patrickpichler, line 46: Maybe we should put the switch what cli to use in the `RunImpl` 
method definition. That way we do not need to set env variables.

pkg/deploy/deploy.go
```diff
@@ -137,6 +156,11 @@ func execute(environment environment.Environment, projects []project.Project, dr
 			}
 		}
 	}
+
+	if errors != nil {
```
patrickpichler, line 160: I guess we can simply return `errors` here, no need for that if.

pkg/deploy/deploy.go
```diff
@@ -127,7 +141,12 @@ func execute(environment environment.Environment, projects []project.Project, dr
 			}
 
 			if err != nil {
-				return err
+				// by default stop deployment on error
+				if !continueOnError && !dryRun {
```
patrickpichler, line 145: Small nitpick: It would be cool if we could turn this condition to an `||` 
without negations. This would be easier to read and grasp imho .

pkg/deploy/deploy.go
```diff
@@ -43,21 +43,35 @@ func Deploy(workingDir string, fileReader util.FileReader, environmentsFile stri
...
-			deploymentErrors[environment.GetId()] = err
+		errors := execute(environment, projects, dryRun, workingDir, continueOnError)
+		if errors != nil {
+			deploymentErrors[environment.GetId()] = errors
 		}
 	}
 
 	for environment, err := range deploymentErrors {
 		if dryRun {
 			util.Log.Error("Validation of %s failed with error %s\n", environment, err)
```
patrickpichler, line 54: I guess we have to change the text here as well. 
Maybe we should put the number of found errors also in the log.

pkg/deploy/deploy.go
```diff
@@ -43,21 +43,35 @@ func Deploy(workingDir string, fileReader util.FileReader, environmentsFile stri
...
-			util.PrettyPrintError(err)
+			for _, envError := range err {
+				util.PrettyPrintError(envError)
```
patrickpichler, line 56: I guess we now also have to change the `PrettyPrintError` function to have a 
fallback and print just the error. We might even consider to introduce a `PrettyPrintable`
interface and try to cast to this, instead of concrete types in there.

cmd/monaco/continue_deployment_on_error_integration_test.go
tatomir146: Yes, talked also to @didiladi and we should actually run tests with both cli versions. Having a switch would be helpful.

tatomir146: Leaving this as it s for now, we anyways need a good approach in running tests on both CLIs. We'll have discussion about this on the roadmap planning.

cmd/monaco/main.go
```diff
@@ -159,6 +159,7 @@ Examples:
 			ctx.String("specific-environment"),
 			ctx.String("project"),
 			ctx.Bool("dry-run"),
+			ctx.Bool("continue-on-error"),
```
patrickpichler, line 162: I just noticed the flag definition is missing in the flag array above, or am i wrong?

tatomir146: Actually intention was to support `continue-on-error` only with new_cli and this param is needed here (will anyways be false) or am I missing something?

patrickpichler: In that case we should explicitly write `false` in there if you ask me. 
Otherwise it looks like an error to me.

UnseenWizzard (commented):
only small comments

General question: We want to document that, or purposely leave it undocumented?

pkg/deploy/deploy.go
```diff
@@ -43,21 +43,33 @@ func Deploy(workingDir string, fileReader util.FileReader, environmentsFile stri
 	}
 
 	for _, environment := range environments {
-		err := execute(environment, projects, dryRun, workingDir)
-		if err != nil {
-			deploymentErrors[environment.GetId()] = err
+		errors := execute(environment, projects, dryRun, workingDir, continueOnError)
+		if errors != nil {
```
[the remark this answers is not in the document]
tatomir146: Actually, none of flag changes with new_cli are documented. But, as there is no reason to allow not support this option also with current cli, I have now backported it and updated documentation.

patrickpichler (approved):
LGTM

patrickpichler:
@tatomir146 could you please rebase instead of merge?

PR #169, opened 2021-02-08
Cope with faulty configs (name or id is null)

didiladi:
It can happen that you have a faulty config deployed in your
environment, which has no id or name set. These special cases were not
handled correctly before. Instead, `monaco` crashed when getting the
list of configs. This commit introduces a graceful handling of these
configs in monaco.

If the id is null -> Return an error
If the name is null -> Fill in the id as the name (as done by the
extensions API)

### DIFF
--- a/pkg/rest/config_upload.go
+++ b/pkg/rest/config_upload.go
@@ -232,7 +232,11 @@ func unmarshalJson(theApi api.Api, err error, resp Response, values []api.Value,
 		}
 
 		if available, array := isResultArrayAvailable(objmap, theApi); available {
-			values = append(values, translateGenericValues(array)...)
+			genericValues, err := translateGenericValues(array, theApi.GetId())
+			if err != nil {
+				return err, values
+			}
+			values = append(values, genericValues...)
 		}
 
 	} else {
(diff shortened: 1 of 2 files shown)


pkg/rest/config_upload.go
```diff
@@ -262,17 +266,41 @@ func isPaginatedResponse(jsonResponse map[string]interface{}) (paginated bool, p
...
+
+		if input["id"] == nil {
+			return values, fmt.Errorf("config of type %s was invalid: No id", configType)
+		}
+
+		// repair invalid configs - but log them
+		if input["name"] == nil {
+			jsonStr, err := json.Marshal(input)
+			if err != nil {
+				util.Log.Warn("Could not marshal %s", configType)
```
UnseenWizzard, line 285: This log is not particularly user friendly IMHO; could we log similarly to the log on line 289? 
```suggestion
				util.Log.Warn("Config of type %s was invalid. Ignoring it!", configType)
```

pkg/rest/config_upload.go
```diff
@@ -262,17 +266,41 @@ func isPaginatedResponse(jsonResponse map[string]interface{}) (paginated bool, p
...
+				continue
+			}
+
+			util.Log.Warn("Config of type %s was invalid: %s", configType, string(jsonStr))
```
UnseenWizzard, line 289:
```suggestion
			util.Log.Warn("Config of type %s was invalid. Auto-corrected to use ID as name!\nInvalid config: %s", configType, string(jsonStr))
```

UnseenWizzard: tried to make that more user friendly as well

UnseenWizzard (commented):
Great addition, but the logs won't be too helpful to end-users I'm afraid, suggested some improvements :)

patrickpichler (approved):

UnseenWizzard (approved):

PR #171, opened 2021-02-09
Release version 1.3.1

didiladi:
This commit extends the documentation for monaco version 1.3.1 and
sets its version. monaco 1.3.1 contains the following changes:

Bugfixes:
  * 2e0e88a: Cope with faulty configs (name or id is null) (#169)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,26 +1,10 @@
 # Monitoring as Code Tool - Release Notes
 
-- [Monitoring as Code Tool - Release Notes](#monitoring-as-code-tool---release-notes)
+- Versions:
   - [1.3.1](#131)
-    - [List of changes](#list-of-changes)
-      - [Bugfixes](#bugfixes)
   - [1.3.0](#130)
-    - [List of changes](#list-of-changes-1)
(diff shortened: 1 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -1,26 +1,36 @@
...
-  - [1.3.0](#130)
+  - [1.3.1](#131)
     - [List of changes](#list-of-changes)
+      - [Bugfixes](#bugfixes)
+  - [1.3.0](#130)
+    - [List of changes](#list-of-changes-1)
       - [Deprecation Warnings](#deprecation-warnings)
       - [Features](#features)
-      - [Bugfixes](#bugfixes)
+      - [Bugfixes](#bugfixes-1)
```
patrickpichler, line 11: Maybe we should introduce some `<a id=""/>` tags to link to them instead of
updating the whole index all the time.

didiladi: Good idea! Or we remove the subheadings from the toc.

patrickpichler: Would also work :laughing:

tatomir146 (approved):

PR #219, opened 2021-02-18
Add response log

### LINKED ISSUE #228: Add response logging
In order to better understand why a certain dyantrace api request
failed, it would be beneficial to also log the response we get from
the api. 

Since in the near future we are planning to call the api in parallel, 
the responses should go into its own file. In order to match them,
an unique id should be generated for the pair. 

The original discussion is in #194

patrickpichler:
Since it is sometimes also useful to see the response of certain
requests, there is now the possibility to specify a response log file
via the MONACO_RESPONSE_LOG env variable. It basically works exactly as
the request log.

The log output format has changed a bit and is now bsically the raw dump
of the outgoing requests/response.

This fixes #228


Dynatrace-Adam-Gardner:
Do these log files conform to an acceptable Dynatrace log file format? That way we can monitor and alert on failures in DT

patrickpichler:
The main use case for this file is to do some manual debugging.
Errors should be handled in the main monaco log.

pkg/util/logging.go
```diff
@@ -58,52 +62,146 @@ func SetupLogging(verbose bool) error {
...
 	}
 
+	requestLogFile.WriteString(fmt.Sprintf("Request-ID: %s\n", id))
+	requestLogFile.Write(dump)
+	requestLogFile.WriteString("\n=========================\n")
+
 	return requestLogFile.Sync()
 }
+
+func LogResponse(id string, response *http.Response, body []byte) error {
```
didiladi, line 164: Unused parameter.

pkg/util/logging.go
```diff
@@ -58,52 +62,146 @@ func SetupLogging(verbose bool) error {
...
+
+	var dumpBody = false
+
+	if contentTypes, ok := response.Header["Content-Type"]; ok {
+		contentType := contentTypes[len(contentTypes)-1]
+
+		dumpBody = shouldDumpBody(contentType)
+	}
+
+	dump, err := httputil.DumpResponse(response, dumpBody)
```
didiladi, line 177: Would that also print response headers?

didiladi (approved):
Cool, awesome stuff! :rocket: Just found some minor things.

pkg/util/logging.go
patrickpichler: Good catch :ok_hand:

pkg/util/logging.go
patrickpichler: Yes the headers are also logged.

didiladi (approved):

Issue #191, opened 2021-02-17
Missing license information in header


tatomir146:
**Describe the bug**
We are missing license information in header of deploy.go, download.go, api_util.go etc. We should check these in all files

didiladi:
Not cool! Is there a way to automatically enforce that?

tatomir146:
We are generally missing this in mock files:
```
grep -L -r "@license" --include "*.go" .                                                                                                           :(
./pkg/util/files_mock.go
./pkg/util/time_mock.go
./pkg/util/files/creator_mock.go
./pkg/util/files/creator_util.go
./pkg/util/files/creator.go
./pkg/util/files/creator_test.go
./pkg/config/config_mock.go
./pkg/deploy/deploy.go
./pkg/download/jsoncreator/json_creator_mock.go
./pkg/download/jsoncreator/json_creator.go
./pkg/download/jsoncreator/json_creator_test.go
./pkg/download/jsoncreator/json_creator_util.go
./pkg/download/download_test.go
./pkg/download/yamlcreator/yaml_creator_mock.go
./pkg/download/yamlcreator/yaml_creator.go
./pkg/download/yamlcreator/yaml_creator_test.go
./pkg/download/yamlcreator/yaml_creator_util.go
./pkg/download/download.go
./pkg/api/api_mock.go
./pkg/api/api_util.go
./pkg/rest/client_mock.go
./pkg/rest/client_util.go
```

patrickpichler:
We could fail the build by running the grep command from @tatomir146 and see if it is not empty.
We should ignore mock files though, as they are no longer checked in.

PR #230, opened 2021-02-19
Fail build on missing license headers

patrickpichler:
The lint command now also checks if license headers are missing. 
If that is the case, the build will fail.

There is also the new `add-license-headers` make target which adds
the license header to all files missing it.

This currently only works on linux.


patrickpichler:
Fixes #191

tools/license_header.txt
```diff
@@ -0,0 +1,13 @@
+@license
+Copyright {{.Year}} Dynatrace LLC
```
tatomir146, line 2: Nice!

tatomir146 (approved):

tatomir146 (approved):

PR #258, opened 2021-03-15
Feature/131/afero filesystem

diegorqc:
Implements Afero as the file system manager replacing the previous custom implementation, that was using separate readfile and writefile classes.

The FileManager implementation has the following features

- Allows to implement tests using a virtual filesystem using the method NewInMemoryFileManager(). This is composed of 2 layers, the first layer, reads files from the physical disk and loads them into the virtual filesystem. The sencond layers allows to modify (create, delete, rename) the file structure without changing the physicial files. 
- I moved the integration util that rewrites the config files into it's own file integration_test_config_rewrite.go. This file allows to modify the filesystem to create unique configs before running the actual integration, this allows to reuse the same dynatrace environment for multiple integration tests.
- This new feature would allow to create integration tests for the download fuctionality by changing the filemanager implementation to use  NewInMemoryFileManager()

Looking forward to your feedback


patrickpichler:
In general I like this PR, but I honestly question the use of the `FileManager` interface in general.
All of this is already covered by `afero.Fs` and we just add a tiny wrapper on top of it. Why
not simply get rid of it all and just use `afero.Fs`?

diegorqc:
I think the idea is to encapsulate the usage of afero. As golang evolves afero should not be needed for mocking the filesystem. In the latest [golang release](https://golang.org/doc/go1.16) there is a simple virtual filesytem implementation, but not as robust as afero right now. When we golang catches up with the full functionality it should be a simple change in a file to switch the implementation instead of having to modify lots of files with afero references.

patrickpichler:
If you think about it we try to abstract away an abstraction of the filesystem and I personally do not
see us winning much by having it.

patrickpichler:
Also by that reasoning `CreateEmptyFile` in `FileManger` should not return a afero 
related struct. 

But as I pointed out, I don't see the benefit of abstracting afero away. It provides some
useful methods. I do not want to create a wrapper for each of those methods.

diegorqc:
@patrickpichler  yes you are right if we go that route using afero directly I can make the changes to the code  just to double check @tatomir146  @didiladi  If you all agree I will make the changes tomorrow, afero will be used directly instead of the FileManager abstraction. I need to close finish this PR since this is related to another change that I have pending regarding download integration test and improvements

patrickpichler:
One more thing: Please use rebases instead of merging the main branch into your feature branch. 
That way we have a nice linear history.

pkg/util/files/manager.go
```diff
@@ -57,25 +60,48 @@ func (a *fileCreator) CreateFolder(path string) (fullpath string, err error) {
...
+	}
+	return dir, nil
+}
 
-//NewInMemoryFileCreator creates  a new instance of FileCreator
-func NewInMemoryFileCreator() FileCreator {
-	el := &fileCreator{}
-	el.fileManager = afero.NewMemMapFs()
+//NewInMemoryFileManager creates  a new instance of FileCreator that reads the files from disk and stores them in a virtual file system
+func NewInMemoryFileManager() FileManager {
```
didiladi, line 96: Nice! So simple and easy to understand - I like it! :1st_place_medal:

pkg/util/files_test.go
```diff
@@ -1,82 +0,0 @@
...
-
-func TestReplaceNameNotMatching(t *testing.T) {
-
-	assert.Equal(t, "management-zone", nameReplacingPostfixFunc("management-zone"))
-	assert.Equal(t, "config:", nameReplacingPostfixFunc("config:"))
-}
-
-func TestReplaceNameMatching(t *testing.T) {
-
-	assert.Equal(t, "- name: test_postfix", nameReplacingPostfixFunc("- name: test"))
```
didiladi, line 52: How do you know that the new `RewriteConfigNames` does what its supposed to do without tests similar to these? Or did I miss them in the review? :thinking:

didiladi (commented):
I'm also leaning toward @patrickpichler 's suggestion to remove the abstraction alltogether. I like the simple 4-line function to switch to a new in-memory js with afero! :+1:

cmd/monaco/integration_test_utils.go
```diff
@@ -163,14 +164,15 @@ func cleanupIntegrationTest(t *testing.T, envFile, suffix string, integrationTes
...
 
 	suffix := getTimestamp() + suffixTest
 	transformers := []func(string) string{getTransformerFunc(suffix)}
 
-	var integrationTestReader, err = util.NewInMemoryFileReader(configFolder, transformers)
-	assert.NilError(t, err)
-
-	testFunc(integrationTestReader)
-	cleanupIntegrationTest(t, envFile, suffix, integrationTestReader)
+	//var integrationTestReader, err = files.NewInMemoryFileManager(configFolder, transformers)
```
UnseenWizzard, line 173: please delete instead of commenting out

cmd/monaco/integration_test_utils.go
```diff
@@ -163,14 +164,15 @@ func cleanupIntegrationTest(t *testing.T, envFile, suffix string, integrationTes
...
+	var fileManager = files.NewInMemoryFileManager()
+	RewriteConfigNames(configFolder, fileManager, transformers)
```
UnseenWizzard, line 175: you have dropped checking for errors, but `RewriteConfigNames` may still return an error

pkg/util/files/manager.go
```diff
@@ -16,35 +16,38 @@ package files
...
 )
 
-//go:generate mockgen -source=creator.go -destination=creator_mock.go -package=files FileCreator
+//go:generate mockgen -source=manager.go -destination=manager_mock.go -package=files FileManager
 
-//FileCreator is an interface to encapsulate the file creation process. Is has 2 implementations
-//in memory and one that uses files on disk.
-type FileCreator interface {
+//FileManager is an interface to encapsulate the file read/creation process. It has 2 implementations
+type FileManager interface {
```
UnseenWizzard, line 29: @patrickpichler , @didiladi so you'd argue for dropping this in favour of using afero directly?

I generally agree, there is no reason to wrap an abstraction of the filesystem in another abstraction - I'm fine if we decide to go all afero for that and drop our own abstraction. 

However the methods in here do additional things as far as I can tell. 
They sanitize inputs, join filepaths, react to errors and return different outputs. 

Would that mean that you want to move the input validation/preparation and error handling into the calling code? 

If we need to do these things, I'd stick with these methods as they seem to be shared utilities more so than an abstraction layer. 
We could of course keep the methods and have callers pass in an `afero.Fs` - so that we could use afero methods directly where needed, instead of hiding it behind our FileManager - but retain the shared steps.

patrickpichler: Most of the methods are anyway thin wrappers around afero and do not bring any 
real benefit. I would drop all of them.

diegorqc: I'm closing this pull request and opening a new one with the Afero only implementation

Issue #131, opened 2021-01-25
User afero for files util


patrickpichler:
By merging #122 we now have a dependency on github.com/spf13/afero . 
Since it is a solid fs abstraction library it would make sense to replace our 
self written fs abstraction in `/util/files.go` with it.

PR #266, opened 2021-03-25
afero as filesystem

diegorqc:
Implements Afero as the file system manager replacing the previous custom implementation, that was using separate readfile and writefile classes.

Add afero.utils.go file to get a virtual filesystem for testing.


patrickpichler:
I guess this is the failing integration test run:
https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/runs/2199321921?check_suite_focus=true

Seems like you missed some `util.NewFileReader` calls in the integration tests.

diegorqc:
I had to make more changes to the implementation. There are parts of the code that required the use of the filesystem implicitly, for example template.ParseFiles takes a filepath and gets a template. Had to change this and other methods to consume the Filesystem. Also, Afero release 7 days ago a new implementation to be compatible with the new io.FS native golang filesystem, so I updated the object in all the methods that previously used Afero.Fs to use Afero.IOFS. Also I have added an end2end test that with a set of 2 simple configs (alerting-profiles, managementzones) executes the restore config cycle. (uploads the configs, download the configs, deletes the configs in the env, reapplies the configs to the cluster and finally clears the reapplied configs). I will be adding more complex scenarios, so Monaco can be used as a tool to easily migrate configs from one env to another one or to restore deleted configs easily. These end to end tests are part of this PR to make sure the tool keeps working with all these small changes being made across multiple files.

patrickpichler:
The issue with using `afero.IOFS` is that we are still on Golang `1.15`. 
From what I've seen `afero.IOFS` and `io.FS` need Golang `1.16`. It would be
nice to for now keep it as `afero.FS` and once we did the upgrade to Golang `1.16`
to change it. 

So for now, lets please focus on just using `afero.Fs`.

patrickpichler:
Actually forget what I just said :laughing: 
I've opened a PR upgrading to Golang `1.16`.

pkg/util/template.go
```diff
@@ -47,9 +49,9 @@ func NewTemplateFromString(name string, content string) (Template, error) {
 }
 
 // NewTemplate creates a new template for the given file
-func NewTemplate(fileName string) (Template, error) {
+func NewTemplate(fs afero.IOFS, fileName string) (Template, error) {
 
-	templ, err := template.ParseFiles(fileName)
+	templ, err := template.ParseFS(fs, fileName)
```
diegorqc, line 54: This allows to mock the template.Parse method using a virtual filesystem. Useful for integration test

diegorqc:
afero.IOFS (implements golang fs.FS) encapsulates afero.Fs to make it compatible with the golang fs.FS new filesystem. There are golang methods that consume the golang filesystem like template.ParseFS, and in order to use the afero virtual filesystem with those methods you will have to send an fs.FS implementation as a parameter. With the previous implementation, when using the virtual filesystem this method will fail, since it will try to look for files in the physical filesystem.  [link to template.text golang](https://golang.org/pkg/text/template/#Template.ParseFS)

patrickpichler:
This PR is blocked by #267

patrickpichler:
PR #267 just got merged, so please rebase your branch

cmd/monaco/main.go
```diff
@@ -212,7 +213,7 @@ Examples:
 
 	return app
 }
-func getDeployCommand(fileReader util.FileReader) cli.Command {
+func getDeployCommand(fileReader afero.IOFS) cli.Command {
```
patrickpichler, line 216: Nitpick: please rename all variables of type `afero.IOFS` to `fs`. 
`fileReader` isn't really the right name anymore.

pkg/download/download.go
```diff
@@ -96,13 +96,19 @@ func getAPIList(downloadSpecificAPI string) (filterAPIList map[string]api.Api, e
...
 	path := filepath.Join(basepath, projectName)
-	creator := files.NewDiskFileCreator()
 
 	util.Log.Info("Creating base project name %s", projectName)
-	fullpath, err := creator.CreateFolder(path)
+	exist, err := afero.Exists(fileManager.Fs, path)
+	if !exist && err == nil {
+		err = fileManager.Mkdir(path, 0777)
+	}
+	if err != nil {
```
patrickpichler, line 108: Please move this if before the one on line 105.
You can then drop the second condition for the if on 105.

pkg/download/download.go
```diff
@@ -145,7 +151,10 @@ func createConfigsFromAPI(api api.Api, token string, creator files.FileCreator,
 
 	subPath := filepath.Join(fullpath, api.GetId())
 
-	_, err = creator.CreateFolder(subPath)
+	exist, err := afero.Exists(fileManager.Fs, subPath)
+	if !exist && err == nil {
+		err = fileManager.Mkdir(subPath, 0777)
+	}
 	if err != nil {
```
patrickpichler, line 158: Same here as above

pkg/util/afero_utils.go
```diff
@@ -0,0 +1,48 @@
...
+	fs := afero.NewCopyOnWriteFs(baseLayer, afero.NewMemMapFs())
+	fsStdlib := afero.NewIOFS(fs)
+	return fsStdlib
+}
+
+//SanitizeName removes special characters, limits to max 254 characters in name, no special characters
+func SanitizeName(name string) string {
+	reg, err := regexp.Compile("[^a-zA-Z0-9-]+")
+	if err != nil {
+		log.Fatal(err)
```
patrickpichler, line 39: Please do not use `log.Fatal` here. 
Return an `error`, or at least use the logger from the `util` package.

patrickpichler:
It would be nice if the `fs` could always be the first argument to a function.
This makes it consistent and easier to read imho.

diegorqc:
Ok i will make the changes you mention and create a new branch with a single commit to squash the current commits, the commit history is getting a bit long for this functionality

patrickpichler:
There is no need to create a new branch. Just squash them and force push.
In this case it is fine, since you are the only one working on it anyway.

patrickpichler:
https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/runs/2234312841?check_suite_focus=true

`Incorrect Usage. flag provided but not defined: -environments` seems like we use `-environments` instead of 
`--environments` somewhere in the integration tests.

pkg/config/config.go
```diff
@@ -52,6 +53,7 @@ var dependencySuffixes = []string{".id", ".name"}
 const skipConfigDeploymentParameter = "skipDeployment"
 
 type configImpl struct {
+	fs                  afero.IOFS
```
patrickpichler, line 56: I would not store the in the config. 
We never actually need it.

patrickpichler:
https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/runs/2242973172?check_suite_focus=true

patrickpichler:
Strange seems like the integration tests are broken. I've looked through the integration tests on your
branch and i do not find any wrong flag issues.

diegorqc:
yes It's saying invalid value for flag --verbose, but this should be a valid value

patrickpichler:
It actually makes sense :man_facepalming: 
`--environments` is only available in the new cli and the integration tests
are executed without the `NEW_CLI` flag set. 

So all `--environments` have to be replaced with `-e`, or we set the `NEW_CLI` 
for all of them .

diegorqc:
I will set environments to -e, but also in the docs says that `monaco --project <project-folder> --environments <path-to-environment-yaml-file> [projects-root-folder]` could be used, this is the previous CLI right? because it's not using any commands

patrickpichler:
Actually you are right, no idea how I ended up with thinking this :laughing: 
So forget about the `NEW_CLI` thing.

patrickpichler:
This might be an issue in our integration tests setup.

patrickpichler (approved):
Fixes #131

PR #273, opened 2021-04-02
Fix absolute paths broken

patrickpichler:
By definition (1) golangs filesystem abstraction doesn't support
absolute paths. Since we need absolute paths to work, every use of
afero.IOFS has been replaced with just afero.Fs. In order to use golangs
abstraction some further research and prototyping is needed.

(1) https://go.googlesource.com/proposal/+/master/design/draft-iofs.md#file-name-syntax


UnseenWizzard:
we haven't released a version with this issue yet right?

patrickpichler:
Nope, last release was before the original change.

PR #282, opened 2021-04-13
fix: Paging in monaco was not working

### LINKED ISSUE #276: Paging does not work in monaco
**Describe the bug**
When having a lot of SLOs defined on the environment, `monaco` just grabs the first 10 configs when performing the GET (all) request to determine if a config is already defined. As a result, `monaco` performs a `POST` instead of a `PUT` for the SLO config.

**How to reproduce**
Define more than 10 SLO configs on the environment and apply another one. 

**Expected behavior**
The paging functionality in `monaco` should grab all pages if there are any, resulting in a list of all defined configs of the given type.

**Log Messages**

The following message is logged when running into the error:

```
[2021-04-12T08:33:53.319Z]     Response was: {"error":{"code":400,"message":"Validation failed for 1 Validators.","constraintViolations":[{"path":"builtin:monitoring.slo","message":"There are at least two SLOs with identical names configured. Please make SLO names unique first to save your changes.","parameterLocation":"PAYLOAD_BODY","location":null}]}}, responsible config: my-config.json
```

didiladi:
When having a lot of SLOs defined on the environment, monaco just
grabs the first 10 configs when performing the GET (all) request
to determine if a config is already defined. As a result, monaco
performs a POST instead of a PUT for the SLO config.

This commit fixes this behavior and fixes teh paging functionality.

Additionally, it extends the `all_configs_integration` test with a
check which applies the same configs again, to enforce a PUT
(update) of the just created configs.

fixes #276


pkg/rest/config_upload.go
```diff
@@ -194,20 +194,20 @@ func isApiDashboard(api api.Api) bool {
...
 		return nil, err
 	}
 
 	for {
-		var objmap map[string]interface{}
 
-		err, values = unmarshalJson(theApi, err, resp, values, objmap)
+		err, values, objmap := unmarshalJson(theApi, err, resp)
 		if err != nil {
 			return values, err
```
patrickpichler, line 208: Do you really want to return values here in case of an error?

pkg/rest/config_upload.go
```diff
@@ -221,63 +221,70 @@ func getExistingValuesFromEndpoint(client *http.Client, theApi api.Api, url stri
...
-
-		var jsonResp api.SyntheticLocationResponse
-		err = json.Unmarshal(resp.Body, &jsonResp)
-		if util.CheckError(err, "Cannot unmarshal API response for existing synthetic location") {
-			return err, values
-		}
-		values = translateSyntheticValues(jsonResp.Locations)
-
-	} else if theApi.GetId() == "synthetic-monitor" {
+func unmarshalJson(theApi api.Api, err error, resp Response) (error, []api.Value, map[string]interface{}) {
```
patrickpichler, line 227: It might make sense to move the whole `isPaginated` logic into this method,
and return some kind of `Pagnination` struct. That way we do not leak the 
`JSON` body to the outside and could add specialized pagination handling 
for different apis.

pkg/rest/config_upload.go
didiladi: Values is an empty array here and objmap is an empty map. I don't think go allows the use of `nil` here.

patrickpichler: Slices can be `nil`. So please return `nil` here to make it explicit that we do not return 
anything.

pkg/rest/config_upload.go
```diff
@@ -194,20 +194,20 @@ func isApiDashboard(api api.Api) bool {
 
 func getExistingValuesFromEndpoint(client *http.Client, theApi api.Api, url string, apiToken string) (values []api.Value, err error) {
 
-	values = make([]api.Value, 0)
+	existingValues := make([]api.Value, 0)
```
patrickpichler, line 197: Nitpick: Actually there is no need to initialize `existingValues` here, since we only use `append` 
on the slice anyway. 

We can simply use:
```go
var existingValues []api.Value
```

(feel free to ignore this :laughing:)

pkg/rest/config_upload.go
didiladi: Good input! Hmm, I'd rather see this method just doing the unmarshalling of a single response. But I could move the for loop in `getExistingValuesFromEndpoint` to its own fuction doing all of the pagination handling. Then this functionality would be encapsulated and extended if necessary in the future. Creating another abstraction for the `Pagination` is imo an overkill at this point in time. What do you think?

pkg/rest/config_upload.go
didiladi: Hmm, I did now know that append works correctly on an uninitialized slice. I'll try it.

pkg/rest/config_upload.go
didiladi: Ok, good input!

pkg/rest/config_upload.go
patrickpichler: I personally think it a bit strange that this method on the one hand parses `api.Values` 
and on the other hand actually return the raw json. Especially since some apis return 
JSON arrays. But I guess this is personal preference.

So I guess I'm fine with it :laughing:

patrickpichler (approved):
LGTM

Issue #275, opened 2021-04-12
Updating of configs for mobile application does not work


didiladi:
**Describe the bug**
When one tries to update the following config of a mobile application with monaco, the API returns an error.

The Config:
```
{
    "name": "{{ .name }}",
    "applicationType": "MOBILE_APPLICATION",
    "costControlUserSessionPercentage": 100,
    "apdexSettings": {
      "toleratedThreshold": 3000,
      "frustratingThreshold": 12000,
      "frustratedOnError": true
    },
    "optInModeEnabled": true,
    "sessionReplayOnCrashEnabled": true,
    "beaconEndpointType": "CLUSTER_ACTIVE_GATE"
}
```

The error:
```
{"code":400,"message":"Constraints violated!","constraintViolations":[{"path":"applicationType","message":"Payload does include parameters that must not be set!","parameterLocation":"PAYLOAD_BODY","location":null},{"path":"sessionReplayEnabled","message":"Payload does not include all required properties!","parameterLocation":"PAYLOAD_BODY","location":null}]}
``` 

**How to reproduce**
Just apply the config above two times. The first time (when monaco performs a POST), it works, but on the PUT it fails.

didiladi:
Here we actually see two errors:

**Error 1:**
```
{"path":"applicationType","message":"Payload does include parameters that must not be set!","parameterLocation":"PAYLOAD_BODY","location":null}
```

Looks like we need to remove the `applicationType` parameter when we perform the PUT. However, this can only be a temporary solution. A permanent one would be that the API handles this case in a graceful way. E.g. checking if the application type is about to change and then raise the 400, if the user tries to change it.

**Error 2:**
```
{"path":"sessionReplayEnabled","message":"Payload does not include all required properties!","parameterLocation":"PAYLOAD_BODY","location":null}
```

Was there a breaking change in the API? :thinking:

PR #289, opened 2021-04-15
fix: References on pre-existing SLOs do not work

### LINKED ISSUE #288: References on pre-existing SLOs do not work
**Describe the bug**
When referencing SLOs in other configs, `monaco` fails to resolve the dependency when the given SLO config is already present in the environment.

**How to reproduce**
* Create an SLO
* Reference an SLO in another config (e.g. a dashboard)
* Run `monaco` twice

**Additional context**
The problem is that we assume that the SLO API returns the `id` of the updated SLO (after doing a `PUT`) within the `Location` response header. However, this only happens in `POST`. For a `PUT`, we need to use the `id` we get from the `GET (all)` request.

didiladi:
The SLO API returns http status code `200` or `201` when updating an
existing config. Previously, `monaco` just listened for the `204` status
code on update. This commit changes the behavior and makes the response
checking more explicit.

fixes #288

### DIFF
--- a/pkg/rest/config_upload.go
+++ b/pkg/rest/config_upload.go
@@ -30,10 +30,10 @@ import (
 func upsertDynatraceObject(client *http.Client, fullUrl string, objectName string, theApi api.Api, payload []byte, apiToken string) (api.DynatraceEntity, error) {
 
 	existingObjectId, err := getObjectIdIfAlreadyExists(client, theApi, fullUrl, objectName, apiToken)
-	var dtEntity api.DynatraceEntity
 	if err != nil {
-		return dtEntity, err
+		return api.DynatraceEntity{}, err
 	}
+
 	var resp Response
 	path := fullUrl
 	body := payload
@@ -45,7 +45,9 @@ func upsertDynatraceObject(client *http.Client, fullUrl string, objectName strin
 		existingObjectId = objectName
 	}
 
-	if existingObjectId != "" {
+	isUpdate := existingObjectId != ""
+
+	if isUpdate {
 		path = joinUrl(fullUrl, existingObjectId)
 		// Updating a dashboard requires the ID to be contained in the JSON, so we just add it...
 		if isApiDashboard(theApi) {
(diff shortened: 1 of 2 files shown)


pkg/rest/config_upload.go
```diff
@@ -57,6 +59,18 @@ func upsertDynatraceObject(client *http.Client, fullUrl string, objectName strin
...
+
+		if success(resp) {
+			util.Log.Debug("\t\t\tUpdated existing object for %s (%s)", objectName, existingObjectId)
+			return api.DynatraceEntity{
+				Id:          existingObjectId,
+				Name:        objectName,
+				Description: "Updated existing object",
+			}, nil
+		} else {
+			return dtEntity, fmt.Errorf("Failed to update DT object %s (HTTP %d)!\n    Response was: %s", objectName, resp.StatusCode, string(resp.Body))
```
patrickpichler, line 71: Just to be consistent, could you please return `api.DynatraceEntity{}` here? 
(Also to make it more explicit)

pkg/rest/config_upload.go
```diff
@@ -90,17 +104,9 @@ func upsertDynatraceObject(client *http.Client, fullUrl string, objectName strin
...
-	}
-	if updateSuccess(resp) {
-		util.Log.Debug("\t\t\tUpdated existing object for %s (%s)", objectName, existingObjectId)
-		return api.DynatraceEntity{
-			Id:          existingObjectId,
-			Name:        objectName,
-			Description: "Updated existing object",
-		}, nil
+		if !success(resp) {
+			return dtEntity, fmt.Errorf("Failed to create DT object %s (HTTP %d)!\n    Response was: %s", objectName, resp.StatusCode, string(resp.Body))
```
patrickpichler, line 108: The same here

patrickpichler (dismissed):
All in all LGTM 

Just two small nitpicks.

pkg/rest/config_upload.go
UnseenWizzard: it will already be that from the var on line 33

But yes, we seem to use that very inconsistently in this method (and we're generally inconsistent e.g. on named return values)

Maybe a codestyle thing we should tackle separate from this?

UnseenWizzard (approved):

UnseenWizzard:
LGTM as well - I'd suggest we discuss @patrickpichler 's nitpick in general/removed from this bugfix, we're overall inconsistent with that.

pkg/rest/config_upload.go
patrickpichler: Yes, but if anybody would set a property on this var, we would return it without noticing ^^

I agree, that it is a code style topic, but to be at least consistent within the method, 
we should return a new empty struct.

didiladi: done

pkg/rest/config_upload.go
didiladi: done

didiladi:
I had to add another commit (d9e17ef) to this PR, to fix the integration tests which currently keep failing due to #275.

patrickpichler (approved):

PR #313, opened 2021-05-06
doc: Fix install section in README

patrickpichler:
The install section mentioned that we distribute monaco as a zip file,
which is not true. Hence the documentation has been adapted to use
executable instead of archive in the install section.


CruzanCaramele:
Thank you for taking care of this, looks better now

CruzanCaramele (dismissed):

CruzanCaramele (approved):

PR #325, opened 2021-05-21
add gh-pages documentation

CruzanCaramele:
moving documentation to GitHub pages


didiladi (dismissed):

.github/workflows/deployDocumentation.yml
```diff
@@ -0,0 +1,53 @@
+name: gh-documentation
+
+on:
+  pull_request:
+    branches: [documentation]
```
didiladi, line 5: shouldn't that be `main`, based on our latest discussione? :thinking:

CruzanCaramele: yes, i have changed it now 👍

.github/workflows/deployDocumentation.yml
```diff
@@ -0,0 +1,53 @@
+name: gh-documentation
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [documentation]
```
didiladi, line 7: It probably needs to be changed here as well, since we decided to have de documentation in `main`, right?

CruzanCaramele: @didiladi  I am leaving that on purpose because in the future if I push to the documentation branch, I want to see what the gh page looks like and make changes before making pull request to the main branch

didiladi: Ah, I see. How is this then working? Does a push to the `documentation` branch temporarily override the one coming from `main`, or is it published on some temporary location?

CruzanCaramele: yes it overrides the one coming from main temporarily

didiladi (approved):

PR #355, opened 2021-06-24
Update experimental-new-cli.md

zachBridges:
Adding a note about using --dry-run with the new CLI. 

The current docs has an example, `monaco -dry-run` which does not work with the new CLI (as it requires the user to specify the command, either `deploy` or `download`). 
It might also be useful to add a similar note to the "Validating Configuration" page as well.


patrickpichler (dismissed):

patrickpichler:
@zachBridges please update the commit message to match our
conventional commit standard.

zachBridges:
> @zachBridges please update the commit message to match our
> conventional commit standard.

All set. Let me know if there's anything else.

patrickpichler (approved):

Issue #305, opened 2021-04-28
Show warning on multiple detected entities with the same name


patrickpichler:
If for whatever reason there are multiple entities with the same name, 
monaco will use the id of the first it encounters. This behavior is fine, but
never the less we should log a warning in order to hint the user that he 
might want to fix that. 

I propose a simple warning telling the user the name, id and api of the 
duplicated entity.

PR #357, opened 2021-06-25
Print warning if two or mode configurations having same name are detected in Dynatrace.

tatomir146:
Implements #305

### DIFF
--- a/pkg/rest/config_upload.go
+++ b/pkg/rest/config_upload.go
@@ -188,13 +188,23 @@ func getObjectIdIfAlreadyExists(client *http.Client, api api.Api, url string, ob
 		return "", err
 	}
 
+	var configName = ""
+	var configsFound = 0
 	for i := 0; i < len(values); i++ {
 		value := values[i]
 		if value.Name == objectName {
-			return value.Id, nil
+			if configsFound == 0 {
+				configName = value.Id
+			}
+			configsFound++
+
 		}
 	}
-	return "", nil
+
+	if configsFound > 1 {
+		util.Log.Error("\t\t\tFound %d configs with same name: %s. Please delete duplicates.", configsFound, objectName)
+	}
+	return configName, nil
 }
 
 func isApiDashboard(api api.Api) bool {


pkg/rest/config_upload.go
```diff
@@ -188,13 +188,21 @@ func getObjectIdIfAlreadyExists(client *http.Client, api api.Api, url string, ob
...
 	}
 
+	var found = false
+	var configName = ""
 	for i := 0; i < len(values); i++ {
 		value := values[i]
 		if value.Name == objectName {
-			return value.Id, nil
+			if found {
+				util.Log.Warn("\t\t\tFound two or more configs having same name: %s. Please delete duplicates.", objectName)
```
UnseenWizzard, line 197: wondering if we'd have a benefit of skipping the early exit and actually counting the duplicates?
as is only the error case will exit earlier, and the normal case of no duplicates would iterate everything anyway

pkg/rest/config_upload.go
```diff
@@ -188,13 +188,21 @@ func getObjectIdIfAlreadyExists(client *http.Client, api api.Api, url string, ob
...
+				return configName, nil
+			} else {
+				found = true
+				configName = value.Id
+			}
 		}
 	}
-	return "", nil
+	return configName, nil
```
UnseenWizzard, line 205: did you notice any significant performance impact of now iterating through the complete values instead of exiting on the first match?

pkg/rest/config_upload.go
tatomir146: The idea was to save few ms there as it's not really a difference if there are 2 or 10 copies of same resource and also to print warning only once. If you think we should add counter, can do this. Haven't thought it would be useful as it's always the same resource, same warning over and over again?

pkg/rest/config_upload.go
tatomir146: No noticeable difference on cluster config, loading payload takes the most time but nothing changed there.

pkg/rest/config_upload.go
```diff
@@ -188,13 +188,21 @@ func getObjectIdIfAlreadyExists(client *http.Client, api api.Api, url string, ob
...
 
+	var found = false
+	var configName = ""
 	for i := 0; i < len(values); i++ {
 		value := values[i]
 		if value.Name == objectName {
-			return value.Id, nil
+			if found {
+				util.Log.Warn("\t\t\tFound two or more configs having same name: %s. Please delete duplicates.", objectName)
+				return configName, nil
```
patrickpichler, line 198: Do we maybe want to return an error here? 
We will only update the first config we find, but if there are multiple,
we cannot be sure it is the one the user desires to update.

tatomir146: This would change default behaviour of monaco, causing deployment to fail every time duplicated config is found. Maybe sticking with warning for now and changing behaviour in monaco 2.0 would be a way to go?

patrickpichler: It changes broken behavior I would argue, but you are right.
[This](https://xkcd.com/1172/) comes to mind :laughing:

tatomir146: Follow up ticket to change default behaviour with monaco 2.0: https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/issues/360

pkg/rest/config_upload.go
tatomir146: Counter is now added.

patrickpichler (approved):
LGTM

Small nitpick: 
There is no need for the `found` flag. The counter alone would be enough.

patrickpichler (approved):

patrickpichler (approved):

patrickpichler:
Fixes #360

Issue #394, opened 2021-07-21
[2.0-next] Introduce compound parameter type


patrickpichler:
There should be a new parameter type called `compound`. The main
use case for it is to allow the parameter value to consist of formatted 
string of other parameters. 

It should look something like this:

```yaml
configs:
 - id: overview
   config:
     name: "Overview"
     parameters:
       firstName: "Han"
       sureName: "Solo"
       fullName:
         type: compound
         format: "{{ .firstName }} {{ .sureName }}"
         references:
           - parameter: firstName
           - parameter: sureName
      
```

The value off the `fullName` parameters should then be `Han Solo`.

patrickpichler:
Will be done by @goberchtold

patrickpichler:
Fixed by #395

PR #395, opened 2021-07-26
feat: Introduce compound parameter type

goberchtold:
This creates a new parameter type, which is composed of other parameters
within the same config.
This parameter has a value `format`, which specifies a template string
for the actual value of the parameter, and a list of references that are
used by the parameter.

fixes: #394

### DIFF
--- a/pkg/config/v2/parameter/compound/compound.go
+++ b/pkg/config/v2/parameter/compound/compound.go
@@ -1,0 +1,180 @@
+// @license
+// Copyright 2021 Dynatrace LLC
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package compound
+
+import (
+	"bytes"
+	"fmt"
+	"text/template"
+
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/config/v2/coordinate"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/config/v2/errors"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/config/v2/parameter"
+	"github.com/dynatrace-oss/dynatrace-monitoring-as-code/pkg/util"
+)
+
+// CompoundParameterType specifies the type of the parameter used in config files
+const CompoundParameterType = "compound"
+
+var CompoundParameterSerde = parameter.ParameterSerDe{
+	Serializer:   writeCompoundParameter,
+	Deserializer: parseCompoundParameter,
+}
+
+type CompoundParameter struct {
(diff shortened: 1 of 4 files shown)


pkg/config/v2/parameter/compound/compound.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		return nil, createParameterParseError(context, "format is not a valid template: %v", err)
+	}
+
+	references, ok := context.Value["references"]
+	if !ok {
+		return nil, createParameterParseError(context, "missing property `references`")
+	}
+
+	referencedParameterSlice, ok1 := references.([]interface{})
```
patrickpichler, line 90: Nitpick: Any reason why it is called `ok1` and not just `ok`?

pkg/config/v2/parameter/compound/compound.go
```diff
@@ -0,0 +1,165 @@
...
+				Environment: context.Environment,
+			},
+			ParameterName: context.ParameterName,
+			Reason:        "unexpected type. parameter is not of type `CompoundParameter`",
+		}
+	}
+
+	result := make(map[string]interface{})
+
+	format := util.ToString(compoundParam.Format)
```
patrickpichler, line 119: Will this always work? 
It might be beneficial to simply store the raw format string 
we get in the parse function in the parameter struct.

pkg/config/v2/parameter/compound/compound.go
```diff
@@ -0,0 +1,165 @@
...
+		ParameterName: context.ParameterName,
+		Reason:        fmt.Sprintf(reason, a...),
+	}
+}
+
+func toParameterReferences(params []interface{}, coord coordinate.Coordinate) (paramRefs []parameter.ParameterReference) {
+	for _, param := range params {
+		paramRefs = append(paramRefs, parameter.ParameterReference{
+			Config:   coord,
+			Property: util.ToString(param),
```
patrickpichler, line 161: This is a bit odd if you ask me. It would be possible to 
use something like
```yaml
references:
 - name: test
   age: test
 - name: hello
```

which never makes sense. We might want to 
switch on the type of the param here and return an
error in case it is either a map or an array.

pkg/config/v2/parameter/compound/compound.go
goberchtold: Good point. As a matter of fact, the `ToString` would have never worked at all ^^

pkg/config/v2/parameter/compound/compound.go
goberchtold: That's true, but shouldn't we just go all the way then and just only allow strings in this place and fail in any other case?

patrickpichler: I would agree if the config file format was not YAML :laughing: 
Otherwise one would be required to always add quotes around
the list of references. 

Hence my suggestion of treating everything besides arrays and 
maps as strings.

pkg/config/v2/parameter/compound/compound.go
```diff
@@ -34,7 +34,8 @@ var CompoundParameterSerde = parameter.ParameterSerDe{
 }
 
 type CompoundParameter struct {
-	Format               *template.Template
+	format               *template.Template
+	rawFormatString      string
 	ReferencedParameters []parameter.ParameterReference
```
patrickpichler, line 39: Nitpick: this could also be private

patrickpichler (dismissed):

patrickpichler (dismissed):

patrickpichler (approved):

PR #396, opened 2021-07-26
Feature/fix integration tests

patrickpichler:


didiladi (dismissed):
I have to say that this PR is really confusingbecause its mixing a fix + a new feature for the integration tests. Would have been cool to have a meaningful PR description, which indicates what I should look for. Took me a while to figure out that the 2 commits needed to be reviewed separately.

cmd/monaco/v2/legacy/integration_test_utils.go
```diff
@@ -103,7 +103,7 @@ func FailOnAnyError(errors []error, errorMessage string) {
 }
 
 func getTimestamp() string {
-	return strconv.FormatInt(time.Now().UnixNano(), 10)
+	return time.Now().Format("200601020304")
```
goberchtold, line 106:
```suggestion
	return time.Now().Format("20060102150405")
```
Careful, '03' for the hour would format it in 12 hour format. Also, I would at least leave the seconds in

cmd/monaco/v2/legacy/skip_deployment_validation_test.go
```diff
@@ -85,10 +118,18 @@ func TestValidationSkipDeploymentWithOverridingFlagValue(t *testing.T) {
...
+		envvars.InstallOsBased()
+	}()
+
 	statusCode := runner.RunImpl([]string{
 		"monaco",
 		"deploy",
-		"-dry-run",
+		"--dry-run",
 		"--environments", skipDeploymentEnvironmentsFile,
 		"-project", "projectD",
```
goberchtold, line 134:
```suggestion
		"--project", "projectD",
```

cmd/monaco/v2/legacy/integration_test_utils.go
```diff
@@ -168,24 +168,39 @@ func cleanupIntegrationTest(t *testing.T, fs afero.Fs, envFile, suffix string) {
 // e.g. my-config_1605258980000_Suffix
 
 func RunLegacyIntegrationWithCleanup(t *testing.T, configFolder, envFile, suffixTest string, testFunc func(fs afero.Fs)) {
-	envvars.InstallFakeEnvironment(map[string]string{
+	configFolder, _ = filepath.Abs(configFolder)
+	envFile, _ = filepath.Abs(envFile)
+
+	envvars.InstallFakeOsOverrideEnvironment(map[string]string{
 		"CONFIG_V1": "1",
```
tatomir146, line 175: Why are we still working with old format config? Why not converting to the new format?

patrickpichler: Since we are still supporting the legacy mode, it makes sense
to test it via integration tests (also that is the reason they are located
in the `legacy` folder).

As soon as we drop this migration layer, we can drop those tests. 

The new integration tests will probably be just the converted ones.

tatomir146: But wouldn't it make sense to convert these tests in v2 right away? We can run these in parallel. v2 will be out main version so we need to have integration tests for it.

patrickpichler: I will add this with their own PR, since I wanted to get the old ones working.
We probably need to adjust our framework around our integration tests a bit.

cmd/monaco/v2/legacy/integration_test_utils.go
```diff
@@ -168,24 +168,39 @@ func cleanupIntegrationTest(t *testing.T, fs afero.Fs, envFile, suffix string) {
...
 	})
 
 	defer func() {
 		envvars.InstallOsBased()
 	}()
 
-	suffix := getTimestamp() + suffixTest
+	randomNumber := rand.Intn(100)
+
+	suffix := fmt.Sprintf("%s_%d_%s", getTimestamp(), randomNumber, suffixTest)
```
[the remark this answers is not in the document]
patrickpichler: The reason for this is to prevent accidental collisions with other
parallel runs of the integration tests. By appending a random number
between 0 and 100 this will become even more unlikely (this was the
reason btw, why we disabled multi os integration tests)

PR #409, opened 2021-08-16
feat #405: Introduce new logger

### LINKED ISSUE #405: [2.0-next] Replace lumber with log from stdlib
Lumber (https://github.com/jcelliott/lumber) seems to be pretty much dead in
the water. The last commit to master was from 2016. 

Golangs stdlib logger pretty much provides everything we need out of the box,
so we should get rid of lumber and use `log` instead.

goberchtold:
As the logging package we've been using is dead, here is a new simple very simple package for logging.

This logger contains 2 loggers in the background, one using the default logger (which would be logging to stdout), and another one that is set to log to a logging file with the current timestamp. For that to work you need to call the `log.SetupLogging()` function, which also handles the setup for the request and response logs (which were not touched other than being moved to their own file). _Those loggers are contained in a new struct._

After the setup has been called, you can use the usual functions for corresponding logging levels, e.g. `log.Info` or `log.Debug`, which are package wide functions, ~~so there's no logging object.~~ _that are calling an underlying default logger, similar to how the standard logger is working._

~~Verbose is also just an exported field of the package~~ _The log level can be set with a field in the logger struct_, which means theoretically the verbose status can be changed at runtime, which can still be changed if that is a concern.

Tldr: Usage wise not that much changed, you just replace `util.Log.Info` with `log.Info` etc., and the `SetupLogging` doesn't take the verbose flag anymore, instead ~~the field `log.Verbose` is set directly.~~ _the log level is changed with the `SetLevel` method of the logger struct, so to change the log level of the default logger: `log.Default().SetLevel(log.LevelDebug)`_

fixes #405


patrickpichler (approved):
LGTM :+1:

Issue #413, opened 2021-08-25
[2.0-next]  Allow deleting specific environment(s) with v2 delete command


LaSacre:
The version 2 of MONACO does not have a delete function at the moment and does not execute a delete with every deploy.
This feature will implement a delete command that uses the dev.yaml and the delete.yaml and can be executed seperat.
The full command is `monaco delete dev.yaml delete.yaml`

This gains the benefit that **"monaco v2"** can use the same delete.yaml but **"monaco v2"** will not automatically execute the delete command as in **"monaco v1"**. The user got the power to control when and how often the command is executed.

CruzanCaramele:
@LaSacre thank you for working on and implementing the delete functionality for upcoming V2
As part of the functionality, we should also have a flag to specify an environment or environments where this operation is to be performed

In order to maintain consistency with other monaco commands, this options (flag) should be called:
`--environment` and `-e` for shorthand.

UnseenWizzard:
Renamed to reflect the current state/open todo. 
Delete command exists for v2, but the enhancement of being able to limit it to specific environments like other commands is still todo

PR #415, opened 2021-09-01
Feature/add delete functionality for v2

LaSacre:
Add a delete function to the monaco version 2. This gives
you a separated delete function that delete
specified objects, specification is in the delete.yaml,
on defined environments, defined in manifest.yaml.


patrickpichler:
Fixes #413

PR #421, opened 2021-09-20
Next download feature 2.0

drqc:
Created the PR for Monaco download 2.0. Still implementing the unit and integration test but everyone can grab the code and start testing/ reviewing it. I have reuse most of the entities from the convert command. 

 Pending tasks:
- Performance improvements in the dependency resolution feature
- Testing for special config cleanup. I have tested it with management zone and alerting profiles and works fine, will need to test it with more configurations.
- Documentation
- Integration test 
- Unit test


pkg/downloader/downloader.go
```diff
@@ -0,0 +1,195 @@
...
+	if errors != nil {
+		return nil, errors
+	}
+	return result, nil
+}
+func GetDependencies() map[string][]string {
+	dependenciesPerEnvironment := make(map[string][]string)
+	return dependenciesPerEnvironment
+}
+func solveDependencies(projectId string, configs map[string][]configv2.Config) map[string][]configv2.Config {
```
drqc, line 74: This might be a good case for a parallel foreach implementation https://stackoverflow.com/questions/24238820/parallel-for-loop

pkg/downloader/downloader.go
```diff
@@ -0,0 +1,195 @@
...
+			}
+		}
+	}
+	return configs
+}
+func newDependencyConfig(projectId string, config configv2.Config, dependencyCoor coordinate.Coordinate) configv2.Config {
+	newContent := strings.ReplaceAll(config.Template.Content(), dependencyCoor.DynatraceId, dependencyCoor.Config)
+	idProperty := "id" //since all depends on id to relative configs
+	ref := refParam.New(dependencyCoor.Project, dependencyCoor.Api, dependencyCoor.Config, idProperty)
+	config.Template.UpdateContent(newContent)
```
UnseenWizzard, line 104: would personally rather create a new Template with the newContent here, than extending it with an `UpdateContent` 
generally it's not something that will be modified, so I don't think we should make it possible just for this use

cmd/monaco/v2/download/cmd.go
```diff
@@ -0,0 +1,169 @@
...
+		if !isAPI {
+			log.Error("Value %s is not a valid API name", cleanAPI)
+			isErr = true
+		} else {
+			filterAPI := availableApis[cleanAPI]
+			filterAPIList[cleanAPI] = filterAPI
+		}
+	}
+	if isErr {
+		return nil, fmt.Errorf("There were some errors in the API list provided")
```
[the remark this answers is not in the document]
tatomir146: For the upload functionality we default to fail but can overrule this with `-c` parameter. The option would be to use this flag also for download functionality, or as I would prefer, to do it as @UnseenWizzard described.

CruzanCaramele:
this is good and very important work in this pull request

the download feature and command options should maintain consistency with the **deploy** command options

The download feature should use the same manifest file as the deployment file to reference environments to download configurations from and get rid of the **--environment** flag

UnseenWizzard:
feature was integrated into v2 dev branch

PR #427, opened 2021-09-22
[WIP] Feature/387/add integration tests

didiladi:
This commit contains the following changes:
 * extend the manifest model to allow to define URLs as env vars
 * add `cmd/monaco/v2/all_configs_integration_test.go` - the first test


pkg/util/log/log.go
```diff
@@ -89,7 +89,7 @@ func (l *extendedLogger) Error(msg string, a ...interface{}) {
 
 // Warn logs the message with the prefix WARN.
 func (l *extendedLogger) Warn(msg string, a ...interface{}) {
-	doLog(l, LevelError, msg, a...)
+	doLog(l, LevelWarn, msg, a...)
```
tatomir146, line 92: Oh that's already part of #428 :)

.github/workflows/build-test.yml
```diff
@@ -50,6 +50,16 @@ jobs:
       - name: Unit Test
         run: make test
 
+      - name: Integration Test V1
```
tatomir146, line 53: Couldn't both v1 and new integration tests be executed in parallel?

Makefile
```diff
@@ -55,6 +55,10 @@ integration-test: build
 	@go test -tags=cleanup -v ./...
 	@go test -tags=integration -v ./...
 
+integration-test-v1: build
+	@go test -tags=cleanup -v ./...
+	@go test -tags=integration_v1 -v ./...
```
tatomir146, line 60: Add `integration_v1` to `PHONY`
btw in case v1 and v2 tests are executed in parallel, cleanup would interfere there causing tests to fail. Should cleanup be the part of integration test?

.github/workflows/build-test.yml
didiladi: Good idea! However, I'd rather tackle that one in a follow-up PR :)

cmd/monaco/v2/all_configs_integration_test.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// tests all configs for a single environment
+func TestIntegrationAllConfigs(t *testing.T) {
+
+	configFolder := "test-resources/integration-all-configs/"
+	manifest := configFolder + "manifest.yaml"
+	specificEnvironment := ""
+
+	RunIntegrationWithCleanup(t, configFolder, manifest, specificEnvironment, "AllConfigs", func(fs afero.Fs) {
```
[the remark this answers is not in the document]
didiladi: The cleanup is done by `RunIntegrationWithCleanup` after applying the configs.

Makefile
didiladi: > Add integration_v1 to PHONY

True, forgot about adding the target to `PHONY` :+1: 

> btw in case v1 and v2 tests are executed in parallel, cleanup would interfere there causing tests to fail. Should cleanup be the part of integration test?

You are right, better would be to have a dedicated `cleanup` target. I'll take this with me for the follow-up PR.

pkg/util/test_config_rewrite.go
```diff
@@ -20,9 +20,9 @@ package util
```
didiladi, line 20: Tag for `integration_v1` is missing here. Currently we just have:

```
//go:build integration
// +build integration
```

tatomir146:
`integration_v1` tags missing in https://github.com/didiladi/dynatrace-monitoring-as-code/blob/037d17a758de514f15ccf3b23a60e2854b85769c/pkg/util/test_utils.go#L1

tatomir146:
Let's also remove `application` config type as we are anyways going to deprecate it in 2.0

tatomir146:
Also moving cleanup to v2 package would make sense as we anyways want to keep it once v1 is not needed anymore

cmd/monaco/v1/all_configs_integration_test.go
```diff
@@ -17,7 +17,7 @@
  * limitations under the License.
  */
 
-package legacy
+package v1
```
tatomir146, line 20: Should we maybe run all intergration tests with `-v`, it would help to see which configs are actually deployed.

pkg/util/log/log.go
didiladi: fixed in bd132d8

pkg/util/test_config_rewrite.go
didiladi: done in f2d1e3b

didiladi:
> Let's also remove `application` config type as we are anyways going to deprecate it in 2.0

IMHO that change is better suited in #435.

didiladi:
> `integration_v1` tags missing in https://github.com/didiladi/dynatrace-monitoring-as-code/blob/037d17a758de514f15ccf3b23a60e2854b85769c/pkg/util/test_utils.go#L1

done in ba4acf5

didiladi:
> Also moving cleanup to v2 package would make sense as we anyways want to keep it once v1 is not needed anymore

done in c690cf6

cmd/monaco/v1/all_configs_integration_test.go
didiladi: Good idea! Done in 8cd53cc

cmd/monaco/v2/all_configs_integration_test.go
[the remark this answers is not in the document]
didiladi: @tatomir146 Thanks, there was indeed still a bug in the name reploacemtn logic for the integration tests (to make the names unique). It is now fixed in bcc08b0.

PR #430, opened 2021-09-23
fix: allow deployment of multiple projects and envs at once

### LINKED ISSUE #429: Unable to deploy multiple projects
`deploy -p="general, prod-saas" -e env -c manifest.yaml`
 ends up in an error, only one project considered

tatomir146:
closes #429

### DIFF
--- a/cmd/monaco/v2/deploy/deploy.go
+++ b/cmd/monaco/v2/deploy/deploy.go
@@ -34,8 +34,8 @@ import (
 	"github.com/spf13/afero"
 )
 
-func Deploy(fs afero.Fs, deploymentManifestPath string, specificEnvironment string,
-	specificProject string, dryRun, continueOnError bool) error {
+func Deploy(fs afero.Fs, deploymentManifestPath string, specificEnvironment []string,
+	specificProject []string, dryRun, continueOnError bool) error {
 
 	deploymentManifestPath = filepath.Clean(deploymentManifestPath)
 	deploymentManifestPath, err := filepath.Abs(deploymentManifestPath)
@@ -57,11 +57,12 @@ func Deploy(fs afero.Fs, deploymentManifestPath string, specificEnvironment stri
 
 	environments := manifest.GetEnvironmentsAsSlice()
 
-	if specificEnvironment != "" {
-		filtered, err := filterEnvironmentByName(environments, specificEnvironment)
+	if len(specificEnvironment) > 0 {
+		filtered, errs := filterEnvironmentsByName(environments, specificEnvironment)
 
-		if err != nil {
-			return err
+		if errs != nil {
+			util.PrintErrors(errs)
+			return errors.New("error while loading environments")
 		}
 
 		environments = filtered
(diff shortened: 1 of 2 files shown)


cmd/monaco/v2/deploy/deploy.go
```diff
@@ -87,7 +87,11 @@ func Deploy(fs afero.Fs, deploymentManifestPath string, specificEnvironment stri
 	}
 
 	if specificProject != "" {
-		filtered, err := filterProjectIdsByName(projects, []string{specificProject})
+		projectsMap := strings.Split(specificProject, ",")
```
patrickpichler, line 90: This is actually a slice and not a map. I would rather call this var `projectsToDeploy` or 
something like this.

cmd/monaco/v2/deploy/deploy.go
```diff
@@ -175,13 +179,13 @@ func printErrorReport(deploymentErrors []error) {
 	groupedConfigErrors := groupConfigErrors(configErrors)
 
 	for project, apiErrors := range groupedConfigErrors {
-		log.Error("==== Project `%s`\n", project)
+		log.Error("==== Project `%s`", project)
```
patrickpichler, line 182: Please undo those changes, as they are completely unrelated to the original bug.

patrickpichler (commented):
Overall I would use the string slice parameter type offered by cli instead of
doing the parsing ourselves. The syntax would change a bit though. Instead
of passing in a string separated by `,` you would then need to pass in multiple
`-p` flag values (which I personally would prefer).

cmd/monaco/v2/deploy/deploy.go
```diff
@@ -399,18 +418,21 @@ func getApiNames(apis map[string]api.Api) []string {
...
+	for _, name := range names {
+		found := false
+		for _, env := range environments {
+			if env.Name == name {
+				result = append(result, env)
+				found = true
+			}
+		}
+		if !found {
+			return nil, fmt.Errorf("no environment found `%s`", name)
```
patrickpichler, line 433: Could we maybe collect all potential errors here?
I guess it is better for the end user if they see in one
go, what envs do not exist.

patrickpichler: We might want to do the same for projects

cmd/monaco/v2/deploy/deploy.go
```diff
@@ -86,30 +86,49 @@ func Deploy(fs afero.Fs, deploymentManifestPath string, specificEnvironment stri
 		return errors.New("error while loading projects")
 	}
 
-	if specificProject != "" {
-		filtered, err := filterProjectIdsByName(projects, []string{specificProject})
+	err, projects = checkProjectsToDeploy(specificProject, projects, environmentNames)
```
patrickpichler, line 89: Since the method loads projects and not checks them,
could we rename it to `loadProjectsToDeploy` please?

cmd/monaco/v2/deploy/deploy.go
tatomir146: It's already done for projects 
https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/blob/af39771231aedecad2a4558da557123d95c2d7d4/cmd/monaco/v2/deploy/deploy.go#L323

tatomir146:
I guess this lint error `//TODO` will anyways be fixed by @UnseenWizzard with his PR.

patrickpichler (approved):

PR #432, opened 2021-09-27
Support RUM Metrics

Dynatrace-Adam-Gardner:
Add support and update docs pages for `calculated-metrics-rum`


didiladi:
I'll close this PR, because the work is continued in #502, which builds on the awesome work of @Dynatrace-Adam-Gardner. :muscle:

PR #462, opened 2021-10-25
feat(test): Add multi-project integration test

didiladi:
This commit adds the integration test `multi_project_integration_test.go`
which aims to test the deployment of multiple (possibly nested) projects
with monaco in config v2.

Sadly, I had to revert a renaming of the package `cmd/monaco/v2/` into
`cmd/monaco/main/` to prevent the following error when executing the
built binary of monaco:

```
./bin/monaco: line 1: syntax error near unexpected token `newline'
./bin/monaco: line 1: `!<arch>'
```


UnseenWizzard:
change addressed

PR #480, opened 2021-11-10
README.md: update with link to bill of materials

DavidPHirsch:
docs: update to README


CruzanCaramele:
@DavidPHirsch modify your commit message with the prefix **docs:**
So as to pass the commit compliance check

PR #492, opened 2021-11-26
fix: create filepath instead of trimming string

### LINKED ISSUE #463: Path resolution for dependencies in monaco 1.6 is not working across OS
When using the same Monaco project with the same folder structure the dependency resolution works differently as seen in the image, this blocks the reusability of monaco projects across different OS.

On the left of the image is Monaco exe running on windows, on the right running on Linux. Same project, same folder structure, same relative path reference.

This might be solved on monaco 2.0. I will start testing the same scenario with the next branch

image

tobigremmer-dt:
Fixes #463


tobigremmer-dt:
@diegorqc fyi

didiladi (commented):
@tobigremmer-dt Thanks for this PR! :100: The changes look good!

Can you please also add a test for the functionality?

UnseenWizzard:
closing this as stale - if you made any progress on the discussed changes/testing please just open a new PR. Thanks
(note this should be obsolete with different handling that will be introduced in v2)

PR #535, opened 2022-02-07
Feature/api plugin elastic

ffurlanetto:
Adding support for Elasticsearch extension configuration through Monaco


.github/workflows/build-test.yml
```diff
@@ -50,16 +50,6 @@ jobs:
       - name: Unit Test
         run: make test
 
-      - name: Integration Test
```
ffurlanetto, line 53: Integration test should be run as part of another job/action (otherwise it keeps failing at build time for forked projects)

Makefile
```diff
@@ -8,7 +8,7 @@ lint:
 ifeq ($(OS),Windows_NT)
 	@.\tools\check-format.cmd
 else
-	@go get github.com/google/addlicense
+	@go install github.com/google/addlicense
```
ffurlanetto, line 11: updated as getting deprecation notice

Makefile
```diff
@@ -25,7 +25,7 @@ else
 endif
 
 mocks:
-	@go get github.com/golang/mock/mockgen
+	@go get -d github.com/golang/mock/mockgen
```
ffurlanetto, line 28: updated as getting deprecation notice

UnseenWizzard:
Hi @ffurlanetto :wave: 

Sorry for the late reaction on this! 

For adding a new API please also add a sample configuration ([as described in documentation](https://dynatrace-oss.github.io/dynatrace-monitoring-as-code/Guides/add_new_api#add-a-new-api-to-monaco)) 

Please drop your other changes. 

- Monaco is currently still built on Go 1.16, `get` becomes deprecated with 1.17 (which you're probably using locally). There is an incoming change to 1.18 which will also address moving to `install`
- Another open PR addresses the issue of integration tests running for forks, please don't just drop these tests completely

PR #537, opened 2022-02-09
docs: update version 1.6.0 and next and add docs homepage

Giulia-dipietro:


CruzanCaramele:
removing the 2nd dot from `..environments_file.md` will fix the issue

Issue #349, opened 2021-06-22
Various timing issues between configs


didiladi:
**Describe the bug**
We currently have a timing issue between calculated service metrics and SLOs. When an SLO depends on a calc service metric, `monaco` correctly creates the calc service metric first. However, as soon as the SLO config is applied, the API validation kicks in with the following error:

```
[2021-06-22T07:48:01.047Z]     Response was: {"error":{"code":400,"message":"Validation failed for 1 Validators.","constraintViolations":[{"path":"builtin:monitoring.slo/0/metricNumerator","message":"Given property 'metricNumerator' with value: 'calc:service.<snip>' violates the following constraint: Metric selector for numerator is invalid..","parameterLocation":"PAYLOAD_BODY","location":null}]}}, responsible config:<snip>.json
```

It complains that it can't find the calculated service metric. This error just happens when the environment creates both configs for the first time.

**How to reproduce**
Add a SLO config depending on a calculated service metric:

```
{
  "enabled": true,
  "name": "{{ .name }}",
  "description": "Measures the ratio of service cally below {{ .milliseconds }} milliseconds. (# of service calls < {{ .milliseconds }} milliseconds / # of total service calls)",
  "status": "WARNING",
  "error": "NONE",
  "useRateMetric": false,
  "metricRate": "",
  "metricNumerator": "{{ .calculatedServiceMetricId }}",
  "metricDenominator": "builtin:service.requestCount.server",
  "target": {{ .target }},
  "warning": {{ .warning }},
  "evaluationType": "AGGREGATE",
  "timeframe": "-1h",
  "filter": "type(\"SERVICE\"), entityId(\"{{ .meId }}\")"
}
```

```
config:
  - responseTimeLessThan250ms: "response-time-less-than-X-ms.json"

responseTimeLessThan250ms:
  - name: "Response Time less than 250ms"
  - calculatedServiceMetricId: "/service/calculated-metrics-service/responseTimeLessThan250ms.id"
  - milliseconds: "250"
  - target: "95"
  - warning: "96"
  - meId: "<snip>"
```

**Expected behavior**
`monaco` should wait with applying the SLO until the calc service metris is fully present at the environment. Probably, some kind of retry mechanism is needed here.

kristofre:
The same behaviour exists with Applications and Synthetic Monitors. If in one monaco run you create an application and then reference it in a synthetic test, Monaco will fail stating the Application does not exist. A subsequent run would be successful.

[...]

PR #538, opened 2022-02-10
Feat/handle timing issues

UnseenWizzard:
This includes refactorings and some preparatory work from https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/pull/532, but removed the app-detection change until further internal discussion of the API. 

This fixes: 
* #349 
* partially #275 (stripping create-only) - additional required update properties can either be checked in the API doc, or should be returned in a better API error response tbh

### DIFF
--- a/cmd/monaco/test-resources/integration-all-configs/project/calculated-metrics-application-mobile/metrics-mobile-app.yaml
+++ b/cmd/monaco/test-resources/integration-all-configs/project/calculated-metrics-application-mobile/metrics-mobile-app.yaml
@@ -6,6 +6,4 @@ speed:
   - appId: "coming from override"
 
 speed.environment1:
-  # temporarily created by hand in the integration test environment
-  # will be fixed in https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/issues/275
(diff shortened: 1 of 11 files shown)


cmd/monaco/test-resources/integration-all-configs/project/calculated-metrics-application-mobile/metrics-mobile-app.yaml
```diff
@@ -6,6 +6,4 @@ speed:
   - appId: "coming from override"
 
 speed.environment1:
-  # temporarily created by hand in the integration test environment
-  # will be fixed in https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/issues/275
-  - appId: "MOBILE_APPLICATION-23395670912B4E78"
+  - appId: "/only-post/application-mobile/mobile-application.id"
```
didiladi, line 9: Since you also fixed #275 alongside this, you can get rid of the `only-post` project. This has the additional benefit that the update of mobile app configs would also be tested in the integration tests.

didiladi (approved):
Cool! :medal_sports: :rocket: 

Just found a minor thing for the integration tests, which can also be tackled in a follow-up. LGTM

PR #543, opened 2022-02-16
small edits for your consideration

StackScribe:
I'm not sure what line #32 is about.  It looks like it is between the line that discusses the -se flag and the code sample itself.  Nothing explains what the metadata is -- and metadata is the plural of metadatum so "metadatas" is probably not rignt.


didiladi:
@StackScribe you are right about line 32 in documentation/versioned_docs/version-1.7.0/commands/deploying-projects.md. This looks completely replaced. Maybe a leftover from the creation of the docs based on docusaurus @CruzanCaramele? I'll merge in this PR, because it contains valuable changes to the docs. Would be cool to get another PR for the misplaced line. Thanks! :)

PR #563, opened 2022-02-27
[CI] Refactor github actions to get better feedback about failing checks and limit which checks run

UnseenWizzard:
@patrickpichler would especially like your feedback on this "great way to spend a Sunday afternoon" of mine :D 

Main changes: 
- introduce gotestsum for more readable unit test output
- introduce action to collect and post test results on PR
- refactor release actions into a single one
- split documentation build and release actions
- only run documentation actions when docs actually changed
- only run Go build/test actions when monaco actually changed
- automatically apply run-integration-test label to PRs if monaco changed
- replace golangci-lint with reviewdog wrapped golangci-lint to get more visible lint results

Points for discussion: 
- Makefile changes
- The release note generation does not seem to have previously worked. I've not tested if it does now.
- Unit test results are suboptimal as they fail to show line numbers
- Tools and Actions where selected based on looking maintained and widely used - e.g. I found no directly fitting action to report go test results from go json reports

Some screenshots from test PRs in my fork: 

ci-lint
tests
tests2


UnseenWizzard:
Note: This change breaks the current strategy for required checks as the 'Build' stage is not actually performed on all PRs anymore :/ 

On the other hand documentation failing would currently not impact if a change can be merged (as long as build passed). 
If we want to keep a required check, we could move the logic of what to check into a single (big..) CI stage and require that. 

Or drop this requirement... there seems to be no way to configure github to just require all checks that run for a PR to pass. They need to be listed explicitly

.github/workflows/documentation-validate.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.4.0
+      - uses: actions/setup-node@v2.5.1
+        with:
+          node-version: '12.x'
+      - name: Build
+        working-directory: ./documentation
+        run: |
+          npm i
```
patrickpichler, line 21: You might want to use `npm ci` here.

Makefile
```diff
@@ -2,13 +2,18 @@ BINARY=monaco
 
 .PHONY: lint format mocks build install clean test integration-test test-package default add-license-headers
 
-default: build
+default: test
```
patrickpichler, line 5: Any reason you change the default target?

patrickpichler (commented):
Overall this looks good to me :ok_hand: 

The only part I'm not sure about, is the automated tagging of PRs with 
the integration test label. Right now we only ran integration tests, when 
a PR/branch seems to be close to done. This will increase the number
or integration tests runs and hence we might run into funky issues with
dynatrace :thinking:

UnseenWizzard:
won't fix/do

PR #579, opened 2022-03-19
ci: Build static monaco binary

### LINKED ISSUE #567: Static linking of monaco binary
When trying to use monaco in alpine or from scratch, additional packages (libc and others) must be installed.

It would be nice if the go binary would be linked statically.

patrickpichler:
In order to be libc agnostic, monaco is now build as a static binary.

fix #567

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,18 +33,27 @@ build: clean lint
 	@go build ./...
 	@go build -o ./bin/${BINARY} ./cmd/monaco
 
+build-release: clean lint
+	@echo Release build ${BINARY}
+	@ GOOS=windows GOARCH=amd64 go build -o ./build/${BINARY}-windows-amd64.exe ./cmd/monaco
+	@ GOOS=windows GOARCH=386   go build -o ./build/${BINARY}-windows-386.exe   ./cmd/monaco
+	@ GOOS=linux   GOARCH=amd64 go build -o ./build/${BINARY}-linux-amd64       ./cmd/monaco
+	@ GOOS=linux   GOARCH=386   go build -o ./build/${BINARY}-linux-386         ./cmd/monaco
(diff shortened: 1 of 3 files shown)


UnseenWizzard:
released binaries are actually still built with xgo in https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/blob/main/.github/workflows/release.yml#L31 and do not use the build cmd from the Makefile...

As that isn't actually needed for cross compiling since go 1.15 and xgo is archived the release built should be changed to just build using different OS/ARCH vars... 
As a quick solution for this the same tags and ld_flags could be passed to xgo as well though

patrickpichler:
I have dropped `xgo` from the release action and updated the `make build` goal to cross compile to all supported architectures.

UnseenWizzard (changes requested):

Makefile
```diff
@@ -29,21 +29,22 @@ mocks:
 	@go generate ./...
 
 build: clean lint
-	@echo Build ${BINARY}
-	@go build ./...
-	@go build -o ./bin/${BINARY} ./cmd/monaco
+	GOOS=windows GOARCH=amd64 CGO_ENABLED=0 go build -tags netgo -ldflags '-w -extldflags "-static"' -o ./bin/${EXECUTABLE}-windows-amd64.exe ./cmd/monaco
+	GOOS=windows GOARCH=386 CGO_ENABLED=0 go build -tags netgo -ldflags '-w -extldflags "-static"' -o ./bin/${EXECUTABLE}-windows-386.exe ./cmd/monaco
```
UnseenWizzard, line 33: the release build will miss these as it's uploading assets from a `build` folder, not `bin`, pls adapt one of them

Laubi:
This PR also fixes #702

PR #583, opened 2022-03-19
feat: Drop unecessary envvar module

patrickpichler:
Since GOs testing libary offers a way of setting env variables just for
the duration of a unit test, the envvar module was unecessary and has
hence been deleted.


UnseenWizzard:
this was integrated into v2 dev branch

PR #612, opened 2022-04-21
test: skip dependency transformation

### LINKED ISSUE #611: Integration tests: Dependencies are not properly resolved
**Describe the bug**
Some dependencies are not properly resolved. For example as part of integration tests:

I would expect this to work:

Screen Shot 2022-04-21 at 1 57 09 PM

I.e. "name" and "applicationId" references are resolved and template variables are substituted with actual values, i.e.:

Screen Shot 2022-04-21 at 2 04 04 PM

However, deploying this config results in:

Screen Shot 2022-04-21 at 2 00 57 PM

---

**UPDATE**

This is related to a transformer that runs as part of integration testing. For example, this line:
```
  - name: \"/project/application/application.name\"
```

Is substituted with:
```
  - name: \"/project/application/application.name_1650549231182295000AllConfigs\"
```

Which isn't recognized as a dependency anymore.

tobigremmer-dt:
Resolves #611

### DIFF
--- a/pkg/util/test_utils.go
+++ b/pkg/util/test_utils.go
@@ -20,10 +20,11 @@
 package util
 
 import (
-	"gotest.tools/assert"
 	"os"
 	"strings"
 	"testing"
+
+	"gotest.tools/assert"
 )
 
 func SetEnv(t *testing.T, key string, value string) {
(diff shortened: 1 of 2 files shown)


didiladi (changes requested):
Cool, this looks good on a first glance. :rocket: 

@tobigremmer-dt could you add a unit test (test `ReplaceName`), where we run into this problem?

tobigremmer-dt:
@didiladi Just want to make sure I understand you correctly. You want me to add a unit test for an integration test?

didiladi (approved):

PR #628, opened 2022-04-27
fix: do not ignore vars that have "=" in their value

ytsibizov-dt:
In some cases the value of an environment variable may have an equal sign (for example, within an URL).
Do not try to split in more than two parts.


didiladi (commented):
Cool! Thanks for tackling this! :1st_place_medal: From the code change, this looks good to me! :rocket:

Can you please add a unit test for this change?

didiladi (approved):

PR #633, opened 2022-05-04
Feature/601/legacy endpoints fixed 2

### LINKED ISSUE #601: Support for non "easy-to-add" endpoints
Endpoints that are not in a "easy-to-add" format (according to https://dynatrace-oss.github.io/dynatrace-monitoring-as-code/Guides/add_new_api) shall be supported.

tobigremmer-dt:
Resolves #601


pkg/download/download.go
```diff
@@ -120,17 +120,80 @@ func downloadConfigFromEnvironment(fs afero.Fs, environment environment.Environm
...
+			}
 		}
 	}
 	util.Log.Info("END downloading info %s", projectName)
 	return nil
 }
 
-func createConfigsFromAPI(fs afero.Fs, api api.Api, token string, fullpath string, client rest.DynatraceClient,
-	jcreator jsoncreator.JSONCreator, ycreator yamlcreator.YamlCreator) (err error) {
+func createConfigsFolder(
```
UnseenWizzard, line 142: as far as I can tell this is the same as done on line 207-208 for non legacy folder creation? 

assuming you pulled this into a method to avoid duplication, please also use the method there :)

UnseenWizzard, line 142: as far as I can tell this is the same as done on line 207-208 for non legacy folder creation? 

assuming you pulled this into a method to avoid duplication, please also use the method there :)

pkg/rest/client.go
```diff
@@ -139,8 +139,16 @@ func (d *dynatraceClientImpl) ReadByName(api Api, name string) (json []byte, err
 }
 
 func (d *dynatraceClientImpl) ReadById(api Api, id string) (json []byte, err error) {
-	fullUrl := api.GetUrlFromEnvironmentUrl(d.environmentUrl) + "/" + id
-	response, err := get(d.client, fullUrl, d.token)
+	url := ""
```
UnseenWizzard, line 142:
```suggestion
	url := ""
```
```suggestion
	var url string
```

UnseenWizzard, line 142:
```suggestion
	url := ""
```
```suggestion
	var url string
```

pkg/rest/config_upload.go
```diff
@@ -29,28 +29,38 @@ import (
...
 	body := payload
 	configType := theApi.GetId()
+	isLegacyApi := theApi.IsLegacyApi()
+	objectId := ""
+
+	// The calculated-metrics-log  and legacy APIs don't have a POST endpoint.
+	// To create a new log metric we need to use PUT which requires a metric key
+	// for which we can just take the objectName
+	if isLegacyApi || configType == "calculated-metrics-log" {
+		objectId = objectName
```
UnseenWizzard, line 50: That is actually not true is it? The legacy APIs do not require an identifier (unlike metrics). 

Your change to the update method actually drops the id again. 

I'd prefer if you move the building of the update API path out of the upload completely, then we have one less "if legacy do something else". 

---

Additionally you've changed the update logic for calculated-metrics-log!  **if** a config object already exists with the configured name, and an ID that is not equal to the name, you will create a new one rather than update. 
I'm not 100% sure if that is an issue right now as I don't know what will be set as ID when such configs are created in the UI - it might be fine with that also uses the name but it will be an issue if generates a UUID. Also I would not rely on any assumptions on how Dynatrace behaves internally in this case.

UnseenWizzard, line 50: That is actually not true is it? The legacy APIs do not require an identifier (unlike metrics). 

Your change to the update method actually drops the id again. 

I'd prefer if you move the building of the update API path out of the upload completely, then we have one less "if legacy do something else". 

---

Additionally you've changed the update logic for calculated-metrics-log!  **if** a config object already exists with the configured name, and an ID that is not equal to the name, you will create a new one rather than update. 
I'm not 100% sure if that is an issue right now as I don't know what will be set as ID when such configs are created in the UI - it might be fine with that also uses the name but it will be an issue if generates a UUID. Also I would not rely on any assumptions on how Dynatrace behaves internally in this case.

pkg/config/config.go
```diff
@@ -139,6 +141,42 @@ func (c *configImpl) IsSkipDeployment(environment environment.Environment) bool
 	return false
 }
 
+func (c *configImpl) IsSkippedDueToConfig(environment environment.Environment, dict map[string]api.DynatraceEntity) (bool, string, error) {
+	// Some APIs require additional validations based on their config
+	isHostsAutoUpdate := c.GetApi().IsHostsAutoUpdateApi()
+
+	if isHostsAutoUpdate {
```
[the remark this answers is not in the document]
tobigremmer-dt: You are understanding correctly. However, unfortunately, there is a mismatch between what the DT API returns as part of a GET response (empty "updateWindows.windows") and what the API expects as part of a POST/PUT body when a resource is created/updated.

Without this evaluation in place you would not be able to deploy a host auto update that was just downloaded.

pkg/config/config_test.go
```diff
@@ -42,7 +42,7 @@ var testDevEnvironment = environment.NewEnvironment("development", "Dev", "", "h
 var testHardeningEnvironment = environment.NewEnvironment("hardening", "Hardening", "", "https://url/to/hardening/environment", "HARDENING")
 var testProductionEnvironment = environment.NewEnvironment("prod-environment", "prod-environment", "production", "https://url/to/production/environment", "PRODUCTION")
 var testManagementZoneApi = api.NewStandardApi("management-zone", "/api/config/v1/managementZones")
-var testHostsAutoUpdateApi = api.NewLegacyApi("hosts-auto-update", "/api/config/v1/hosts/autoupdate")
+var testHostsAutoUpdateApi = api.NewSingleConfigurationApi("hosts-auto-update", "/api/config/v1/hosts/autoupdate")
```
UnseenWizzard, line 45: I think this can be removed as well, it was only used in `TestIsSkippedDueToConfig`

Issue #306, opened 2021-04-29
Missing Application Detection Rules


mark-dt:
**Describe the bug**
When downloading the configuration if an application has multiple detection rules then only one will be downloaded, I think the files are being overwritten.
In the YAML file for the detection rules we see 2 entries, so monaco is detecting them correctly but somehow overwriting one (?)

**How to reproduce**
- create multiple detection rules for same application
- download config
- check the detection rules for the application with multiple detection rules

**Expected behavior**
I would expect to see all application detection rules for all defined applications

**Log output**
N/A

**Environment (please complete the following information):**
 - OS: Win 10
- monaco 1.5.2

tatomir146:
@diegorqc can you please have a look here?

Issue #377, opened 2021-07-08
request-naming-service and monaco not working well together


heydenb:
We defined a Global request naming rule using the API and this works fine. But when putting it in our monaco configuration it is recreated every time monaco runs.
The reason for this is that this API does not take a name attribute. So monaco probably doesn't know if it already exists or not. Giving it an "id" is also not accepted.

tatomir146:
@heydenb can you please give us an example of the yaml and JSON config used in this case?

heydenb:
```
config:
  - schedulerJobRequestNaming: "scheduler-job-request-naming.json"
  - rmiProxyRequestNaming: "rmiproxy-request-naming.json"

schedulerJobRequestNaming:
  - name: "Scheduler Job Request Naming Rule"

rmiProxyRequestNaming:
  - name: "RMIProxy Request Naming Rule"
  - requestAttrProxyService: "projects/scheduler/request-attributes/rmiProxyCallService.name"
  - requestAttrProxyMethod: "projects/scheduler/request-attributes/rmiProxyCallMethod.name"
```

and

```
{
    "name": "{{.name}}",
    "order": "1",
    "enabled": true,
    "namingPattern": "{RequestAttribute:SchedulerJobName}",
    "conditions": [
        {
            "attribute": "SERVICE_REQUEST_ATTRIBUTE",
            "comparisonInfo": {
              "type": "STRING_REQUEST_ATTRIBUTE",
              "comparison": "EXISTS",
              "requestAttribute": "SchedulerJobName",
              "negate": false,
              "isCaseSensitive": false
            }
        },
        {
            "attribute": "CUSTOMSERVICE_CLASS",
            "comparisonInfo": {
            "type": "STRING",
            "comparison": "CONTAINS",
            "value": "JobViaRest",
            "negate": false,
            "isCaseSensitive": false
            }
        }
    ],
    "placeholders": []
  }
```

afri33:
It's not a solution, but it is maybe a small workaround/helpful in some cases. I could also make updates to an existing rule if name and namingPattern are the same. Example:

```
config:
    - demo-rule: "request-naming-service.json"
  
demo-rule:
    - name: "/demo/v1/<COUNTRY>/demo"
    - namingPattern: "/demo/v1/<COUNTRY>/demo"
```

[...]

Issue #254, opened 2021-03-12
Application with multiple detection rules fails with "Given application detection rule duplicates rule"


miha-anzur:
Hello 
During the upload of downloaded data with monaco following error is reported

2021-03-12 11:13:35 ERROR Deployment to perform failed with error Failed to upsert DT object Web admin tools (HTTP 400)!
    Response was: {"error":{"code":400,"message":"Constraints violated.","constraintViolations":[{"path":"","message":"Given application detection rule duplicates rule 57edee13-13ce-4202-bfbe-8b68cd9674f7","parameterLocation":"PAYLOAD_BODY","location":null}]}}, responsible config: perform/app-detection-rule/Webadmintools.json

Create application with multiple detection rules.
Execute:
monaco download -e environment.yaml
-- data stored to folder perform

Execute:
monaco -e environment.yaml -v perform/

tatomir146:
@diegorqc maybe you can help here?

diegorqc:
I will need more details about the yaml file and json file downloaded, but at first this looks similar to #179 specifically the comment from Kristof. Also I noticed that the  [Dynatrace API for app detection rules](https://www.dynatrace.com/support/help/dynatrace-api/configuration-api/rum/application-detection-configuration/get-all/) has 3 possible values when getting the configs. Only 2 of those values include a name, so this might be something to consider, since we are using the name to identify the config

mdicss:
See also: https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/discussions/293#discussioncomment-619795

PR #654, opened 2022-05-25
Bugfix/642/unique names conflict

### LINKED ISSUE #642: Dashboard with same name break download/deploy functionality
**Describe the bug**
If multiple dashboards exist that have the same name, download and deploy does not work correctly.

1) A single <dashboard-name>.json is overwritten, every time a dashboard is downloaded with the same name
2) Single <dashboard-name>.json is deployed (probably last instance in list?) which breaks all backup use cases.

For example, an environment has 2 dashboards called "Applications", 4 dashboards called "Config". All instances are detected but only single Config.json and single Applications.json is downloaded. Also, instances are merged into single identifier (i.e. "Applications" and "Config") which have multiple list items with key "name":

```
config:
- Applications: Applications
- Config: Config.json
- Config: Config.json
- Applications: Applications
- Config: Config.json
- Config: Config.json
 
Applications:
- name: 'Applications'
- name: 'Applications'

Config:
- name: Config
- name: Config
- name: Config
- name: Config
```

This issue was mentioned in #179 already. However, I feel this issue requires a dedicated (and timely) fix as this breaks functionality.

tobigremmer-dt:
Resolves #642 and #377


pkg/download/yamlcreator/yaml_creator.go
```diff
@@ -26,24 +27,53 @@ import (
 
 //YamlCreator implements method to create the yaml configuration file
 type YamlCreator interface {
+	ReadYamlFile(fs afero.Fs, configSubPath string, apiId string) error
 	CreateYamlFile(fs afero.Fs, path string, name string) error
```
UnseenWizzard, line 31: nitpick: Now that you've introduced reading/updating I'd rename this method to `WriteYamlFile` to fit what it's really doing - creating or overwriting

pkg/download/yamlcreator/yaml_creator.go
```diff
@@ -58,19 +88,176 @@ func (yc *YamlConfig) AddConfig(name string, rawName string) {
...
+}
+
+func (yc *YamlConfig) parseConfigDetails(unmarshaledData map[string]map[string]string) error {
+	for configId := range unmarshaledData {
+		if configId != "config" {
+			// As of May 2022, download does not support project structure
+			// Fallback to environment unique id
+			environmentUniqueConfigId := yc.EnvironmentName
+
+			entityUuid, err := yc.generateUuidFromConfigId(environmentUniqueConfigId, configId)
```
UnseenWizzard, line 218: In the case that the previously downloaded config is one with a unique name, the generated UUID will never match anything that Dynatrace actually returns. I don't see a need to generate any UUIDs here.

If I'm not missing something in the rest of the yaml_creator logic, I think you could replace this with: 
`if IsUuid(configId) then {use it as DetailConfig.Id} else {leave DetailConfig.Id blank}`

pkg/util/uuid_generator.go
```diff
@@ -33,3 +35,29 @@ func GenerateUuidFromName(name string) (string, error) {
...
+		return false
+	}
+}
+
+// GenerateUuidFromConfigId takes the unique project identifier within an environment, a config id and
+// generates a valid UUID based on provided information
+func GenerateUuidFromConfigId(projectUniqueId string, configId string) (string, error) {
+	// Return if configId is UUID
+	isUuid := IsUuid(configId)
+	if isUuid {
```
UnseenWizzard, line 54: I am not a fan of this method either returning OR generating a UUID, especially with the name and documentation not telling me! (and it being perfectly valid to generate uuids from uuids)
I've actually stumbled to understand the download/parse logic due to this behaviour making the downloaded non-unique-name ids work. 

I would prefer using this function only for generating UUIDs and doing the `IsUuid` check explicitly where needed. 
If my understanding/comment on` yaml_creator.go/parseConfigDetails` line 218 is correct I think the only place you will have to do the check and generate a UUID if it is not, is on deployment

pkg/deploy/deploy.go
```diff
@@ -31,11 +31,94 @@ import (
 	"github.com/spf13/afero"
 )
 
-func Deploy(workingDir string, fs afero.Fs, environmentsFile string,
-	specificEnvironment string, proj string, dryRun bool, continueOnError bool) error {
-	environments, errors := environment.LoadEnvironmentList(specificEnvironment, environmentsFile, fs)
+type Deploy struct {
```
UnseenWizzard, line 34: Honestly I didn't yet get to fully review this file due to the amount of changes you're introducing by turning this from simple functions into an Object, but: I think you could achieve the test-ability without turning this into an instantiated Deploy struct by keeping the function type definitions you might need for mocking/dep inversion and passing them into the methods that need them.

I see no _real_ need for the big structural and conceptual change introduced here - but the other comments in this review are more important that this one.

If I find the time I'll check out your fork and see if I can make a proposal how that would look.

pkg/delete/delete.go
```diff
@@ -61,16 +61,22 @@ func LoadConfigsToDelete(fs afero.Fs, apis map[string]api.Api, path string) (con
 			return configs, err
 		}
 
-		apiName, validConfig := apis[configType]
+		apiIface, validConfig := apis[configType]
```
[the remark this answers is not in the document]
tobigremmer-dt: Good idea. However, I actually is the interface (api.Api).

pkg/project/project.go
```diff
@@ -285,3 +292,21 @@ func (p *projectImpl) HasDependencyOn(project Project) bool {
 	}
 	return false
 }
+
+// GenerateConfigUuid generates a unique id for this config, considering environment and project structure
+func (p *projectImpl) GenerateConfigUuid(configId string) (string, error) {
```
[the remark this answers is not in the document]
tobigremmer-dt: Well, it kind of is. The UUID is generated for that very specific environment and project.

pkg/download/yamlcreator/yaml_creator.go
tobigremmer-dt: If configId is no UUID, a new UUID is generated based on configId (unique within yaml file) + environment + project. This generated UUID is then used to create the config entity.

pkg/util/uuid_generator.go
tobigremmer-dt: I think there's a misunderstanding. A UUID is always generated. In case of a download, the yaml file's configId (+ environment + project) is translated into a UUID. This UUID is matched with whatever UUID is returned from Dynatrace. With this information, Monaco knows which config instance to update.

UnseenWizzard:
note: possibly also resolves #306 and #254

pkg/download/download.go
```diff
@@ -163,9 +180,32 @@ func createConfigsFromSingleConfigurationAPI(
...
+	err = ycreator.ReadYamlFile(fs, subPath, apiId)
+	if err != nil {
+		util.Log.Debug("No previously downloaded configs found.")
+	}
+
 	idVal := api.NewIdValue()
 
-	name, cleanName, filter, err := jcreator.CreateJSONConfig(fs, client, api, idVal, subPath)
+	// At this point all API specific substitutions have been made (e.g. reports name = dashboard id)
+	configId, err := getConfigId(idVal.Id, idVal.Name, api)
```
UnseenWizzard, line 193: It took me way to long to realize in the diff view that this is in the single configuration API method, where you will always only have one config. 
Please just use and sanitize the API ID directly - creating the 'blank' ID Value and parsing it like a real value is just useless operations - and with the incorrect copy pasted comment it's confusing/will possibly make others think there's an error in here in the future.

~~I don't understand what this does/is meant to do tbh.~~

~~`api.NewIdValue()` as I read it will create a struct containing the api's ID as Id and Name value. (for some reason)
Your getConfigId will sanitize this down to a string usable as a filename on all OSs - in either case sanitizing the API ID though. So e.g. this will be `dasboard-v2`~~

~~You comment makes me think that the content of idVal is expected to be something else than what api.NewIdValue() returns? :confused:~~

pkg/download/download.go
```diff
@@ -218,11 +260,27 @@ func createConfigsFromAPI(
 	}
 
 	for _, val := range values {
-		util.Log.Debug("getting detail %s", val)
+		util.Log.Debug("Getting details for id %s, name %s", val.Id, val.Name)
 		cont++
 		util.Log.Debug("REQUEST counter %v", cont)
 
-		_, cleanName, filter, err := jcreator.CreateJSONConfig(fs, client, api, val, subPath)
+		// At this point all API specific substitutions have been made (e.g. reports name = dashboard id)
```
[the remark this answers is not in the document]
tobigremmer-dt: Moved comment to correct line. Name is determined from existing value in config yaml OR downloaded UUID. `getNameForConfig` therefore obsolete.

UnseenWizzard (approved):
Thanks for all the work on this feature and the comments @tobigremmer-dt!
As you're on holiday I took the liberty to add a small change myself in https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/pull/678/commits/13c3ef973f5734da01523e1ec99062422cd16a84 

That needed me to open a new PR as I am not allowed to push to your fork (seems that could be allowed via a PR setting: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)

UnseenWizzard:
Approved but superseded by https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/pull/678 which adds one commit simplifying error collection in deployment / when continuing on error

Issue #179, opened 2021-02-15
[Grouper/Epic] Download Functionality problems and improvements


tatomir146:
Used internal env to test download functionality.

**Describe the bug**
Various things which should be improved by download functionality:
- remove "metadata" from JSON, no need to have this 
- syntheticLocations are empty `config: []`
- yaml formatting - no new lines used, making yaml files hard readable
- yaml: quoting values makes file better readable in IDEs
- main.go: downloadSpecificAPI should be either renamed to `download-specific-api` to keep consistency or even better to `specific-api` as it's a part of download command
- main.go: `p` alias used for `downloadSpecificApi` > not really intuitiv 
- main.go: help for download functionality tells `Yaml file containing environment to deploy to` - `deploy` should be renamed to `download`
- download.go: multiple `enviroments` typos
- IDs should be stripped eg: alerting-profile "mzId": "-6221876622438591000" 
- verbose prints out e.g. nil as string  `DEBUG getting detail {... Cl_ES %!s(*string=<nil>)}`
- built in extensions are also downloaded: ruxit.python.redis:slave_state:unsynchronized Redis Slave unsynchronized
- if there are multiple dashboards with same name then something like this is generated:
```
config:
- Home: Home.json
- Home: Home.json
- Home: Home.json
- Home: Home.json
Home:
- name: Home
- name: Home
- name: Home
- name: Home
```

kristofre:
Not sure if I agree with 

> remove "metadata" from JSON, no need to have this

Usecase is as follows: what if I download from environment/cluster X (let's say non-prod) that is on version a and I want to apply that config to environment/cluster Y (production) that is on version b and there is a difference in API behaviour. The meta data of the version would allow us to process that properly

diegorqc:
about the yaml formatting, we will have to look for an different parser from golang struct to yaml, the default one is quite strict in the convention for formatting https://github.com/go-yaml/yaml/issues/556. We will have to create custom formatters to get quoting and line breaks for certain blocks.

tatomir146:
> Not sure if I agree with
> 
> > remove "metadata" from JSON, no need to have this
> 
> Usecase is as follows: what if I download from environment/cluster X (let's say non-prod) that is on version a and I want to apply that config to environment/cluster Y (production) that is on version b and there is a difference in API behaviour. The meta data of the version would allow us to process that properly

[...]

PR #670, opened 2022-06-22
fix(download): Correct debug logging of config being downloaded

UnseenWizzard:
Rather than trying to log the Value object as a string, when logging that a config is downloaded, log the name and id.

ref: #179

### DIFF
--- a/pkg/download/download.go
+++ b/pkg/download/download.go
@@ -209,7 +209,7 @@ func createConfigsFromAPI(
 		return err
 	}
 	for _, val := range values {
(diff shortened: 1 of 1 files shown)


Laubi (approved):

UnseenWizzard:
Closing as this was already addressed in another PR, now logging Id and Name in a slightly different but equivalent way
https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/blob/main/pkg/download/download.go#L267

Issue #666, opened 2022-06-21
Download creates empty synthetic-locations config if only PUBLIC locations exists


UnseenWizzard:
**Describe the bug**
monaco download will create an empty synthetic-locations config folder, if only PUBLIC locations exist. 

This is due the download of individual configurations checking and ignoring PUBLIC locations and only downloading manually created PRIVATE ones, while the yaml creation does not check if there's actually only downloaded locations. 

This will result in the creation of:
```
synthetic-location/
    synthetic-location.yaml
```

with `synthetic-location.yaml` containing only: 
```
config: []
```

**How to reproduce**
- download synthetic-location from an environment without private locations 
  (monaco download -e {env yaml} -p synthetic-location

**Expected behavior**
No synthetic-location config folder is created at all.

---

**Implementation Hints/Notes**
This is located in `pkg/download`. 

Specifially in `download.go:createConfigsFromAPI() ` the config folder is created, config files are written with jsoncreator.go (or not) and the yaml is written with yamlcreator.go. 

While this method has an early exit if no data is returned at all from the API, further sorting like "is a synthetic location PUBLIC at all" happens in jsoncreator.go, without download actutally noticing that no files are written. That needs to change.

PR #676, opened 2022-07-12
fix: download-creates-folders-for-empty-configurations

shschulze:
monaco download will create an empty synthetic-locations config folder, if only PUBLIC locations exist.

after downloading an api it checks, wether it is empty and deletes deletes the folder if needed

fixes: #666

### DIFF
--- a/pkg/download/download.go
+++ b/pkg/download/download.go
@@ -83,13 +83,11 @@ func getAPIList(downloadSpecificAPI string) (filterAPIList map[string]api.Api, e
 
 		return filterAPIList, nil
 	}
-
 	requestedApis := strings.Split(downloadSpecificAPI, ",")
 	isErr := false
 	for _, id := range requestedApis {
 		cleanAPI := strings.TrimSpace(id)
 		isAPI := api.IsApi(cleanAPI)
-
 		if !isAPI {
 			util.Log.Error("Value %s is not a valid API name", cleanAPI)
 			isErr = true
@@ -253,6 +251,7 @@ func createConfigsFromAPI(
 		util.Log.Error("error creating folder for api %v %v", api.GetId(), err)
 		return err
 	}
+	emptyfolder := true
 
 	apiId := api.GetId()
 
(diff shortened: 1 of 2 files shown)


pkg/download/download.go
```diff
@@ -220,13 +221,22 @@ func createConfigsFromAPI(
...
 		}
+		emptyfolder = false
 		ycreator.AddConfig(cleanName, name)
 	}
 
-	err = ycreator.CreateYamlFile(fs, subPath, api.GetId())
-	if err != nil {
-		util.Log.Error("error creating config api yaml file: %v", err)
-		return err
+	if emptyfolder {
```
UnseenWizzard, line 228: Could you add a test for the logic please? 

There is some limited test setup in `download_test.go` that should give some pointers on how you might setup that test. 

In general see 
* https://dynatrace-oss.github.io/dynatrace-monitoring-as-code/Get-started/intro
* https://gobyexample.com/testing
on Go testing :)

pkg/download/download.go
```diff
@@ -220,13 +221,22 @@ func createConfigsFromAPI(
 		if filter {
 			continue
 		}
+		emptyfolder = false
```
UnseenWizzard, line 224: Does that actually work? 
The problem with e.g. PUBLIC synthetic locations is at the API returns them (so there will be values), but `CreateJSONConfig` filters them out and does not any files to the folder. 

The code looks like `emptyfolder` will we false in that case?

shschulze: discussed in weekly

Laubi (dismissed):

UnseenWizzard (approved):

pkg/download/download_test.go
```diff
@@ -114,3 +120,36 @@ func TestGetAPIList(t *testing.T) {
...
+
+	apiMock.EXPECT().
+		IsSingleConfigurationApi().Return(false).AnyTimes()
+
+	jcreator.EXPECT().
+		CreateJSONConfig(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).
+		Return("demo.json", "demo", true, nil) // every value is filtered
+
+	ycreator.EXPECT().
+		CreateYamlFile(gomock.Any(), gomock.Any(), gomock.Any()).
```
UnseenWizzard, line 146: I think `CreateYamlFile` has changed to `WriteYamlFile` in the latest version. Please rebase and update the mock usage :)

PR #677, opened 2022-07-12
feat: Add support for elasticsearch-extension

Laubi:
Adds support for the elasticsearch-extension.
Additionally changes the makefile to not add the add-license dependency, but to install it locally.


pkg/api/api.go
```diff
@@ -48,6 +48,10 @@ var apiMap = map[string]apiInput{
 		apiPath:                      "/api/config/v1/extensions",
 		propertyNameOfGetAllResponse: "extensions",
 	},
+	"extension-elasticsearch": {
```
UnseenWizzard, line 51: @Laubi please also add a test configuration to the integration tests and document that this is now supported. 
See https://dynatrace-oss.github.io/dynatrace-monitoring-as-code/Guides/add_new_api#add-a-new-api-to-monaco (points 2. & 3.)

UnseenWizzard (changes requested):

UnseenWizzard (approved):

Issue #669, opened 2022-06-21
Download is not removing non-standard identifiers - application-web


UnseenWizzard:
**Describe the bug**
Download functionality is currently only removing the basic toplevel "id" key from downloaded configurations. 
Several configurations have additional/different identifiers that need to be stripped before they can be re-uploaded. 
Application Web configs contain an "identifier" instead of an "id" which needs to be removed (at least on new creation, on update it may be present)

**How to reproduce**
- download any application-web
- delete the application-web
- try to deploy the downloaded application-web again
  - This will be rejected by DT with a 400 " Payload must not have an id in a POST request" 
- remove the `identifier` property and rerun monaco deploy
- this will work

**Sample Log output**
```
2022-06-21 14:22:04 INFO  Dynatrace Monitoring as Code v1.7.0
2022-06-21 14:22:04 DEBUG Reading projects...
2022-06-21 14:22:04 DEBUG   project - ojq
2022-06-21 14:22:04 DEBUG Processing file: ojq/application-web/application-web.yaml
2022-06-21 14:22:04 DEBUG 		ojq/application-web/BORGApplication1644085193155050783AllConfigs
2022-06-21 14:22:04 DEBUG Sorting projects...
2022-06-21 14:22:04 INFO  Executing projects in this order: 
2022-06-21 14:22:04 INFO  	1: ojq (1 configs)
2022-06-21 14:22:04 INFO  Processing environment ojq...
2022-06-21 14:22:04 INFO  	Processing project ojq...
2022-06-21 14:22:04 DEBUG 		Deploying configs in this order: 
2022-06-21 14:22:04 DEBUG 			1: ojq/application-web/BORGApplication1644085193155050783AllConfigs.json
2022-06-21 14:22:04 DEBUG 		Applying config `BORG Application_1644085193155050783AllConfigs` using ojq/application-web/BORGApplication1644085193155050783AllConfigs.json
2022-06-21 14:22:05 INFO  Deployment summary:
2022-06-21 14:22:05 ERROR Deployment to ojq failed with error!
2022-06-21 14:22:05 ERROR 	Failed to create DT object BORG Application_1644085193155050783AllConfigs (HTTP 400)!
[...]

Issue #668, opened 2022-06-21
Download is not removing non-standard identifiers - custom-service


UnseenWizzard:
**Describe the bug**
Download functionality is currently only removing the basic toplevel "id" key from downloaded configurations. 
Several configurations have additional/different identifiers that need to be stripped before they can be re-uploaded. 
custom-services configs contain `rules` and `methodRules` with their own `id` properties. 

If a config is pre-exisiting (and thus its rules and method rules exist) these ids being present is fine. 
If a config is new/re-created after delete they can not be updated, as Dynatrace fails to find the rules by ID which it would create if no ID was defined. 

Note that rules and rules.methodRules are both arrays, so several entries with id might be present.

**How to reproduce**
- download any custom-service
- delete the custom-service
- try to deploy the downloaded custom service again
  - This will fail with `Dependency of config {NAME} was not available`.
- remove any `rules.id` and `rules.methodRules.id` properties and rerun monaco deploy
- this will work

**Expected behavior**
Download strips these IDs as well and is able to both update and newly create a downloaded custom-service configuration.

PR #683, opened 2022-07-15
Bugfix: download is not removing non standard identifiers #668 & #669 

shschulze:
- added one helper function to remove a key with specified path in  json_creator
fixes: #668 
fixes: #669

### DIFF
--- a/pkg/download/jsoncreator/json_creator.go
+++ b/pkg/download/jsoncreator/json_creator.go
@@ -112,11 +112,12 @@ func processJSONFile(data map[string]interface{}, id string) ([]byte, error) {
 
 //replaceKeyProperties replaces name or displayname for each config
 func replaceKeyProperties(dat map[string]interface{}) map[string]interface{} {
-	//removes id field
-	delete(dat, "id")
 
-	// Removes metadata field
-	delete(dat, "metadata")
+	dat = removeKey(dat, []string{"metadata"})
+	dat = removeKey(dat, []string{"id"})
+	dat = removeKey(dat, []string{"identifier"})
+	dat = removeKey(dat, []string{"rules", "id"})
+	dat = removeKey(dat, []string{"rules", "methodRules", "id"})
 
 	if dat["name"] != nil {
 		dat["name"] = "{{.name}}"
@@ -127,6 +128,36 @@ func replaceKeyProperties(dat map[string]interface{}) map[string]interface{} {
 	//for reports
 	if dat["dashboardId"] != nil {
 		dat["dashboardId"] = "{{.name}}"
+	}
+	return dat
+}
+
+// removes key with specified path
+func removeKey(dat map[string]interface{}, key []string) map[string]interface{} {
+	if len(key) == 0 || dat == nil {
+		//noting todo
+		return dat
+	}
+	if len(key) == 1 {
+		delete(dat, key[0])
+		return dat
+	}
+	if dat[key[0]] == nil {
+		// no field: nothing to do
+		return dat
+	}
+	if field, ok := dat[key[0]].(map[string]interface{}); ok {
+		dat[key[0]] = removeKey(field, key[1:])
+		return dat
+	}
+
+	if arrayOfFields, ok := dat[key[0]].([]interface{}); ok {
+		for i := range arrayOfFields {
+			if field, ok := arrayOfFields[i].(map[string]interface{}); ok {
+				arrayOfFields[i] = removeKey(field, key[1:])
+			}
+		}
+		dat[key[0]] = arrayOfFields
 	}
 	return dat
 }
--- a/pkg/download/jsoncreator/json_creator_test.go
+++ b/pkg/download/jsoncreator/json_creator_test.go
@@ -135,3 +135,120 @@ func TestProcessJSONFile(t *testing.T) {
 	_, err = processJSONFile(sample, "testId")
 	assert.NilError(t, err)
 }
+
+func TestReplaceKeyProperties(t *testing.T) {
+	data := map[string]interface{}{
+		"metadata":    true,
+		"id":          12,
+		"identifier":  "id",
+		"noRemove":    "aaa",
+		"name":        "replace",
+		"displayName": "replace",
+		"rules": []interface{}{
+			map[string]interface{}{
+				"2":           12,
+				"id":          "random string",
+				"methodRules": map[string]interface{}{"bool": false, "id": "asdf"},
+				"bool":        false,
+			},
+			map[string]interface{}{
+				"2":           12,
+				"id":          "random string",
+				"methodRules": map[string]interface{}{"id": "asdf"},
+				"bool":        false,
+			},
+		},
+	}
+	data = replaceKeyProperties(data)
+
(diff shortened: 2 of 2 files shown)


pkg/download/jsoncreator/json_creator.go
```diff
@@ -101,7 +101,7 @@ func getDetailFromAPI(client rest.DynatraceClient, api api.Api, entityId string)
 func processJSONFile(data map[string]interface{}, id string) ([]byte, error) {
 	data = replaceKeyProperties(data)
 
-	jsonfile, err := json.MarshalIndent(data, "", " ")
+	jsonfile, err := json.MarshalIndent(data, "", "    ")
```
UnseenWizzard, line 104: Please remove this change. 
This was purposely not fixed yet, as it would mean all already downloaded files changing for users keeping them under version control. 
As such this was kept for a 2.x release, where things change substantially anyway (https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/issues/640)

pkg/download/jsoncreator/json_creator.go
```diff
@@ -112,11 +112,12 @@ func processJSONFile(data map[string]interface{}, id string) ([]byte, error) {
 
 //replaceKeyProperties replaces name or displayname for each config
 func replaceKeyProperties(dat map[string]interface{}) map[string]interface{} {
-	//removes id field
-	delete(dat, "id")
 
-	// Removes metadata field
-	delete(dat, "metadata")
+	dat = removeKey(dat, []string{"metadata"}) // Removes metadata field
```
UnseenWizzard, line 116: nitpick: you could remove the comments, it's somewhat obvious what this line does

pkg/download/jsoncreator/json_creator.go
```diff
@@ -131,6 +132,33 @@ func replaceKeyProperties(dat map[string]interface{}) map[string]interface{} {
 	return dat
 }
 
+// removes key with specified path
+func removeKey(dat map[string]interface{}, key []string) map[string]interface{} {
+	if len(key) == 0 {
+		return dat
+	} else if len(key) == 1 {
```
UnseenWizzard, line 139: as you fail early here and return out of your ifs you can skip the else and improve readability

```go
if len(key) == 0 {
    return dat
}

if len(key) == 1 {
    delete(dat, key[0])
    return dat
}

[the logic in the normal case]
```

pkg/download/jsoncreator/json_creator.go
```diff
@@ -131,6 +132,33 @@ func replaceKeyProperties(dat map[string]interface{}) map[string]interface{} {
 	return dat
 }
 
+// removes key with specified path
+func removeKey(dat map[string]interface{}, key []string) map[string]interface{} {
```
UnseenWizzard, line 136: :exclamation: Add tests for this method please

pkg/download/jsoncreator/json_creator.go
```diff
@@ -131,6 +132,33 @@ func replaceKeyProperties(dat map[string]interface{}) map[string]interface{} {
...
+			dat[key[0]] = removeKey(field, key[1:])
+		} else if arrayOfFields, ok := dat[key[0]].([]interface{}); ok {
+			for i := range arrayOfFields {
+				if field, ok := arrayOfFields[i].(map[string]interface{}); ok {
+					arrayOfFields[i] = removeKey(field, key[1:])
+				}
+			}
+			dat[key[0]] = arrayOfFields
+		} else {
+			util.Log.Error("unexpected type")
```
UnseenWizzard, line 156:
```suggestion
			util.Log.Error("Unable to parse configuration to remove identifier property %v. Correct configuration manually if necessary.", key[0])
```

Or something similar would be more helpful to a user encountering this error :)

pkg/download/jsoncreator/json_creator.go
```diff
@@ -131,6 +132,33 @@ func replaceKeyProperties(dat map[string]interface{}) map[string]interface{} {
 	return dat
 }
 
+// removes key with specified path
+func removeKey(dat map[string]interface{}, key []string) map[string]interface{} {
+	if len(key) == 0 {
```
UnseenWizzard, line 137: I still believe you could improve the readability of this a lot, if you remove some of the nesting / else block and just return early out of ifs. 

E.g. the method could look like this, turning all the initial checks into a (for me) much more readable sequence than the else blocks do:
```go

	if len(key) == 0 {
		// invalid input, nothing to remove
		return dat
	}

	if dat[key[0]] == nil {
		// no field of this name, nothing to do
		return dat
	}

	if len(key) == 1 {
		delete(dat, key[0])
		return dat
	}

	if field, ok := dat[key[0]].(map[string]interface{}); ok {
		dat[key[0]] = removeKey(field, key[1:])
	} else if arrayOfFields, ok := dat[key[0]].([]interface{}); ok {
		for i := range arrayOfFields {
			if field, ok := arrayOfFields[i].(map[string]interface{}); ok {
				arrayOfFields[i] = removeKey(field, key[1:])
			}
		}
		dat[key[0]] = arrayOfFields
	} else {
		util.Log.Error("Unable to parse configuration to remove identifier property %v. Correct configuration manually if necessary.", key[0])
	}

	return dat
```

(you could - and if I was writing this I would - drive this further and do the same with the last if/else-if/else block. But it doesn't help readability as much as the initial three IMHO)

pkg/download/jsoncreator/json_creator_test.go
```diff
@@ -135,3 +135,89 @@ func TestProcessJSONFile(t *testing.T) {
...
+			map[string]interface{}{
+				"a": 12,
+				"b": "random string",
+				"c": map[string]interface{}{"id": "asdf"},
+				"d": false,
+			},
+		},
+	}
+
+	data = removeKey(data, []string{"invalid"})
```
UnseenWizzard, line 208: nitpick: I'd personally split these into individuation test cases. This way it's easy to spot what assumption broke if one of them fails from a (good) test name alone. That's especially helpful in CI output if you quickly want to see what's wrong without searching for the detailed output of the failed test

pkg/download/jsoncreator/json_creator_test.go
```diff
@@ -135,3 +135,89 @@ func TestProcessJSONFile(t *testing.T) {
...
+	data = removeKey(data, []string{"id"})
+	assert.Assert(t, data["id"] == nil, "id should be removed")
+
+	data = removeKey(data, []string{"REC", "a"})
+	assert.Assert(t, data["REC"].([]interface{})[0].(map[string]interface{})["a"] == nil, "must be removed")
+
+	data = removeKey(data, []string{"REC", "rec2"})
+	assert.Assert(t, data["REC"].([]interface{})[0].(map[string]interface{})["rec2"] == nil, "must be removed")
+
+}
```
UnseenWizzard, line 223: Please also add a test for the error case of the input data not conforming to the assumption of remove key. 
So a nested property that's neither an array nor string map, triggering the last else block and leaving the data untouched

shschulze: Removed the error case, as it was not necessary

Issue #178, opened 2021-02-15
Download is not removing non-standard indentifiers - calculated-metrics-service


anton-freyberg:
**Describe the bug**
when trying to deploy calculated service metrics I downloaded from an environment, I have to change all entityIDs to "null", since these cannot be prefixed when deploying. 

This is the error I got:     Response was: {"error":{"code":400,"message":"Constraints violated.","constraintViolations":[{"path":"entityId","message":"Not allowed to define a service metric with an entityId","parameterLocation":"PAYLOAD_BODY","location":null}]}}, responsible config: project\calculated-metrics-service\MoneynewResponsetimeinservicebooking.json

**How to reproduce**
 I just downloaded the whole environment and exported the calculated service metrics in the folder below 
[project.zip](https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/files/5982291/project.zip)
Then I deleted a metric in the environment and tried to deploy the newly created project

**Expected behavior**
By deploying the project above, I was expecting to deploy the metric I deleted in the UI

**Log output**
If applicable, add log snippet to help explain your problem.
[20210215-153855.log](https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/files/5982305/20210215-153855.log)

**Environment (please complete the following information):**
 - OS: Win 10
 - Tool version [e.g. 1.3.1]

**Additional context**

tatomir146:
@diegorqc are the ID's stripped from configuration after download? I don't see anything in documentation.

anton-freyberg:
So this does not seem to be an isolated case. Same with the application (or application-web) folders.  Identifiers are kept and produce errors. In the case of  application-web, I tried to deploy some config and this was the error:

 2021-02-15 17:45:33 ERROR Deployment to integration failed with error Failed to upsert DT object Wohnbaubank OBK (HTTP 400)!
    Response was: {"error":{"code":400,"message":"Constraints violated.","constraintViolations":[{"path":"id","message":"Payload must not have an id in a POST request.","parameterLocation":"PAYLOAD_BODY","location":null}]}}, responsible config: deploy\application-web\WohnbaubankOBK.json
2021-02-15 17:45:33 ERROR Errors during deployment! Check log!

diegorqc:
[...]

PR #690, opened 2022-07-20
fix(download): Remove non standard identifier `entityId` during download #178

shschulze:
removing `entityId` from JSON
fixes: #178


pkg/download/jsoncreator/json_creator_test.go
```diff
@@ -163,6 +164,7 @@ func TestReplaceKeyProperties(t *testing.T) {
 
 	assert.Assert(t, data["metadata"] == nil, "metadata should be removed")
 	assert.Assert(t, data["id"] == nil, "id should be removed")
+	assert.Assert(t, data["entityId"] == nil, "id should be removed")
```
Laubi, line 167: I would write here (for all) what id should be removed. Here it would be 'entityId', in the field below 'identifier'

PR #691, opened 2022-07-20
feat: Allow newlines in variables

Laubi:
Up to now, it was impossible to use newlines in YAML configurations directly; you had to escape them. 
With this change, it is allowed to use all features yaml gives us to include newlines directly in the yaml variable definition.
Look at the dashboard example for insights on how you can use it.

### DIFF
--- a/documentation/docs/configuration/yaml_config.md
+++ b/documentation/docs/configuration/yaml_config.md
@@ -240,3 +240,17 @@ of an alerting profile from the env var `ALERTING_PROFILE_VALUE`.
 ```
 
 > :warning: Values you pass into a configuration as environment variables must not contain the `=` character.
+
+### Newlines in variables
+
+All line breaks in variables are escaped.
+
+```yaml
+development:
+    - name: "Dev"
+    - example1: "This is \\n already escaped" 
+    - example2: "This will \n be escaped"
+    - text: |
+        This will also
+        be escaped
+```
--- a/pkg/util/template_test.go
+++ b/pkg/util/template_test.go
@@ -108,3 +108,61 @@ func getTemplateTestPropertiesClashingWithEnvVars() map[string]string {
 
 	return m
 }
+
+func TestEscapeNewlineCharacters(t *testing.T) {
+
+	p := map[string]interface{}{
+		"string without newline": "just some string",
+		"string with newline":    "some\nstring",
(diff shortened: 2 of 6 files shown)


documentation/docs/configuration/yaml_config.md
```diff
@@ -240,3 +240,17 @@ of an alerting profile from the env var `ALERTING_PROFILE_VALUE`.
 ```
 
 > :warning: Values you pass into a configuration as environment variables must not contain the `=` character.
+
+### Newlines in variables
+
+All line breaks in variables are escaped.
```
UnseenWizzard, line 246: Do we support all types of line breaks in yaml now (https://yaml-multiline.info/) or a subset? (TIL there's a lot more options to do multi lines..)
If all would be nice to add a few tests with > multilines and quoted and plain multilines. 
If not please describe explicitly what multilines are okay to use :)

pkg/util/template_test.go
```diff
@@ -108,3 +108,60 @@ func getTemplateTestPropertiesClashingWithEnvVars() map[string]string {
...
+
+	empty := map[string]interface{}{}
+
+	assert.DeepEqual(t, escapeNewlineCharacters(empty), empty)
+}
+
+func TestEscapeNewline(t *testing.T) {
+	assert.Equal(t, escapeNewlines("String without newline"), `String without newline`)
+	assert.Equal(t, escapeNewlines("String with one\nnewline"), `String with one\nnewline`)
+	assert.Equal(t, escapeNewlines("String with one windows\r\nnewline"), "String with one windows\r\\nnewline")
```
UnseenWizzard, line 163: can you please add a test for already escaped newlines in the input? 
` "String with already escaped \\n newline" `

documentation/docs/configuration/yaml_config.md
Laubi: As long as the yaml parser supports it, we support it. The string we get from yaml is processed. Our code has no extra logic to check how the user defined the text. 
Not sure if it makes sense to include all options yaml gives us. To do this would include at least nine checks.

If we want to do this, we should add a unit-test for the parser itself and verify input & output.

UnseenWizzard (dismissed):

UnseenWizzard (approved):

Issue #640, opened 2022-05-13
Unusual indentation in downloaded config files


tobigremmer-dt:
**Describe the bug**
When configuration is downloaded. the json files are created with 1 space indentation. Although there is no official standard, 2 or 4 whitespaces became a best practice.

The Monaco download functionality shall adapt this best practice.

PR #693, opened 2022-07-20
style: changed json-indent to 4 Spaces

shschulze:
changed the indent of json files

fixes: #640


Laubi:
Should we maybe use tabs instead? With that, users may be able to select the width themself. I, for example, prefer two spaces instead of 4. 
@UnseenWizzard

Laubi:
I just saw that the .editorconfig defines 2 spaces as width. Maybe we should stay consistent with that.

Issue #586, opened 2022-03-29
Error deleting resource when resource name contains slash character


kasarpolis:
We use slashes in our management zone names.  Monaco throws an error when attempting to delete them.

Example:
delete.yaml 

```
delete:
- "management-zone/mz-name-part1/mz-name-part2"
```

https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/blob/70e6b4b6d8c4b53d3f6a5e3b36e59fde34ddfbb3/pkg/delete/delete.go#L89

UnseenWizzard:
As the delete yaml always expects entries as `type/name` this should be simple to fix by splitting on the first `/` only

PR #706, opened 2022-08-01
fix(download): Deleting configs with slashes in the name does no longer error

Laubi:
With this change, slashes in names are allowed during deletion. The first slash is used to slit between API & name, and all following slashes are ignored.

Tested with the following delete.yaml:
```yaml
delete:
  - conditional-naming-processgroup//a
  - conditional-naming-processgroup/a/b/c/
  - conditional-naming-processgroup/a/b
```

Fixes: #586


UnseenWizzard (approved):
:+1: lgtm

PR #709, opened 2022-08-02
fix: filter the mime types correctly for logging

Laubi:
Before this fix, all responses and requests containing a charset were ignored, and all content types starting with text/.

### DIFF
--- a/pkg/util/logging.go
+++ b/pkg/util/logging.go
@@ -134,13 +134,7 @@ func LogRequest(id string, request *http.Request) error {
 		return nil
 	}
 
-	var dumpBody = false
-
-	if contentTypes, ok := request.Header["Content-Type"]; ok {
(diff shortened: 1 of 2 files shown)


UnseenWizzard (dismissed):

pkg/util/logging.go
```diff
@@ -203,15 +203,15 @@ func LogResponse(id string, response *http.Response) error {
 }
 
 func shouldDumpBody(contentType string) bool {
-	if strings.HasPrefix("text/", contentType) {
+	if strings.HasPrefix(contentType, "text/") {
```
UnseenWizzard, line 206: :sweat_smile: nice catch :+1:

Laubi:
Added some additional tests, so that Headers::get is used in the tests

UnseenWizzard (approved):

PR #714, opened 2022-08-16
Fix/upgrade docusaurus

UnseenWizzard:


CruzanCaramele (approved):
I Checked out and tested the branch locally, it looks good 👍

PR #726, opened 2022-09-01
Bugfix: Downloading tenant config fails with monaco v1.8.2 and update documentation for Monaco v.1.8.3

Laubi:
[fix: Do not abort download after the first download error.](https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/commit/6e00168f206add5ab9ad3bf78d82a9bcba3300b6) 

If a download error occurred, for example if a token permission was not available, the error would be reported and the download stopped. This commit removes the return of the error, so that it is logged, but not propagated.
Fixes https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/issues/725


UnseenWizzard (approved):
'fix'/undoing error handling looks reasonable to me. 

In v2 downloads we should reconsider proper error handling though - if we want 'fully functional' downloaded configurations included dependencies between configs, missing configs will be an error that can't be ignored.
Maybe making failing on error the default and introducing a "continue on error" option as for deploy.

PR #732, opened 2022-09-07
fix: More leniently handle Dynatrace errors that are ususally caused …

UnseenWizzard:
…by newly created applications not being found yet

Trigger retries in more general cases of server errors (5xx) being reported for known APIs. Also unit test retry decision logic and related code.


pkg/rest/config_upload_test.go
```diff
@@ -160,3 +171,177 @@ func TestIsApiDashboard(t *testing.T) {
 	isFalse := isApiDashboard(testReportsApi)
 	assert.Equal(t, false, isFalse)
 }
+
+func TestIsSuccess(t *testing.T) {
```
Laubi, line 175: Please remove this empty test, as you test the functionality itself below

Laubi (approved):

Issue #738, opened 2022-09-15
Documentation links are broken


dzoech:
**Describe the bug**
When following a Google link to the documentation, I got redirected to `On https://dynatrace-oss.github.io/dynatrace-monitoring-as-code/next/configuration/configuration_structure/`. Please not the slash at the end.

This causes all links, e.g. to `Supported configuration types`, to break as they point to `https://dynatrace-oss.github.io/dynatrace-monitoring-as-code/next/configuration/configuration/configTypes_tokenPermissions` in this case. Please not the `/configuration/configuration`.

UnseenWizzard:
Thanks for bringing this up @dzoech! 

I've opened a PR to fix this (and likely a few more links)

PR #739, opened 2022-09-15
fix(docs): Configure all links as relative or absolute from content root

UnseenWizzard:
We had several styles of (relative) links between pages, which occaisonally caused issues. For details on docusaurus linking see: https://docusaurus.io/docs/markdown-features/links

fixes: #738


CruzanCaramele (approved):
@UnseenWizzard  The links look good and working now
I checked out your branch and tested it

Issue #741, opened 2022-09-16
Double Quotes in configuration properties can lead to JSON templating failing


UnseenWizzard:
As quotes are currently not escaped before placing config properties from a yaml into a json, 
it is possible to define configuration properties in a config yaml that lead to invalid JSONs on templating. 

Specifically this happens as YAML allows: 
```
- key: This is a perfectly fine "string value"
```

Which on templating a json of the form:
```
{
  "value" : " {{ .key }}"
}
```

Results in the invalid output:
```
{
  "value" : "This is a perfectly fine "string value""
}
```

Which should instead be: 
```
{
  "value" : "This is a perfectly fine \"string value\""
}
```

PR #742, opened 2022-09-16
Fix/non escaped quotes

UnseenWizzard:
see  #741

### DIFF
--- a/cmd/monaco/special_character_in_config_integration_test.go
+++ b/cmd/monaco/special_character_in_config_integration_test.go
@@ -19,22 +19,27 @@
 
 package main
 
-/* Commented out because of https://github.com/dynatrace-oss/dynatrace-monitoring-as-code/issues/121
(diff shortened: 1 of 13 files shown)


UnseenWizzard:
@Laubi please take a close look, as it replaces your newline escape code. 
Did I miss anything? Any edge cases you can still think off?

UnseenWizzard:
The escaping of double quotes currently breaks usecases where the JSON only becomes valid after configuration properties are placed in it. 

e.g. the e2e test case placing a property of `"key": ' "email A", "email B" '` into a json list:

```
"list": [ {{ .key }} ] 
```

which without escaping produces valid json of: 
```
"list": [ "email A", "email B" ] 
```
but with the fix becomes invalid
```
"list": [ \"email A\", \"email B\" ] 
```

Laubi (approved):

PR #758, opened 2022-10-25
fix: Project loading should ignore hidden directories

Laubi:
Up to now, hidden directories were ignored iff they were in the 'top' level of the loaded project. If a hidden directory was below that, it was not. This fix fixes:
  1) hidden directories can no longer be projects
  2) mixed projects with hidden directories does no longer error


UnseenWizzard (approved):
:rocket:

PR #773, opened 2022-11-09
fix(download): Skip plugin-definitions of ad-metrics

Laubi:
Anomaly-detection-metrics that start with dynatrace.* or ruxit.* are metrics that plugins define - we need to skip those during download.

### DIFF
--- a/documentation/docs/commands/downloading-configuration.md
+++ b/documentation/docs/commands/downloading-configuration.md
@@ -32,6 +32,19 @@ To download specific APIs only, use `--specific-api` to pass a list of API value
 monaco download --specific-api alerting-profile,dashboard --environments=my-environment.yaml
 ```
 
+# Skipped configurations
+
+Some configurations on Dynatrace are predefined and must not be downloaded and reuploaded on the same, or another environment.
+
+Those skipped configurations are:
+
+| API                         | Details                                                                           |
+|-----------------------------|-----------------------------------------------------------------------------------|
+| `dashboard`/`dashboard-v2`  | Dashboards that have the owner `Dynatrace` set, or dashboards, which are a preset |
+| `anomaly-detection-metrics` | Configurations that start with `dynatrace.` or `ruxit.`                           |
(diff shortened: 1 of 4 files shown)


pkg/download/download.go
```diff
@@ -318,6 +321,18 @@ func createConfigsFromAPI(
 	return nil
 }
 
+func filterConfigsNotToDownload(a api.Api, values []api.Value) []api.Value {
+	result := make([]api.Value, 0, len(values))
+
+	for _, v := range values {
+		if !(a.GetId() == "anomaly-detection-metrics" && (strings.HasPrefix(v.Id, "dynatrace.") || strings.HasPrefix(v.Id, "ruxit."))) {
+			result = append(result, v)
+		}
```
UnseenWizzard, line 330: Could you please include some (debug) logging when a config is filtered out? 

And add a mention to the documentation that these are filtered and that any configs that where manually created with internal prefixes should be renamed - as we've seen when discussing it there's at least some chance that someone has created such a config via the API

Laubi:
I've added an additional commit to fix the actual problem as well.

documentation/docs/commands/downloading-configuration.md
```diff
@@ -32,6 +32,19 @@ To download specific APIs only, use `--specific-api` to pass a list of API value
...
 
+# Skipped configurations
+
+Some configurations on Dynatrace are predefined and must not be downloaded and reuploaded on the same, or another environment.
+
+Those skipped configurations are:
+
+| API                         | Details                                                         |
+|-----------------------------|-----------------------------------------------------------------|
+| `dashbaord`/`dashboard-v2`  | Dashboards that have the owner 'Dynatrace' set, or are a preset |
```
UnseenWizzard, line 43:
```suggestion
| `dashboard`/`dashboard-v2`  | Dashboards that have the owner 'Dynatrace' set, or are a preset |
```

UnseenWizzard (dismissed):

UnseenWizzard (approved):

PR #798, opened 2023-01-31
Fix/remove outdated log mon configs

UnseenWizzard:
this removes or replaces and use of calculated-metrics-log, which is a legacy Log monitoring v1 API and no longer works since we upgraded the E2E environments to the latest log monitoring version


cmd/monaco/integration_test_utils.go
```diff
@@ -138,9 +138,15 @@ func cleanupIntegrationTest(t *testing.T, fs afero.Fs, envFile, suffix string) {
...
+			}
 
 			values, err := client.List(api)
-			assert.NilError(t, err)
+			if err != nil {
+				t.Logf("Failed to cleanup any test configs of type %q: %v", api.GetId(), err)
+			}
 
 			for _, value := range values {
 				// For the calculated-metrics-log API, the suffix is part of the ID, not name
```
Laubi, line 152: Since we skip `calculated-metrics-log` now, we could simplify this code now?

UnseenWizzard: We could drop the OR case in the case that checks for the suffix in the ID - in the v2 version of this, I've actually done this, but have neglected it here it seems

Laubi (approved):

PR #808, opened 2023-02-14
Feature: Allow deployment/deletion with environment-groups

Laubi:
This change allows users to specify an environment group instead of only environments to deploy to.
For example, an environment 'staging' could contain hundreds of environments. To not need to specify each one, this change adds the possibility to specify the `--group` (`-g`) flag to deploy all environments within one group.

This flag is currently mutually exclusive with `--environment` (`-e`).

A new restriction introduced in the `manifest.yaml`: `environmentGroups[].name` must be unique.

image
image
image
image
image
image

### DIFF
--- a/cmd/monaco/deploy/deploy.go
+++ b/cmd/monaco/deploy/deploy.go
@@ -312,16 +312,6 @@ func groupEnvironmentConfigErrors(errors []configError.DetailedConfigError) Grou
 	return groupErrors
 }
 
-func toEnvironmentNames(environments []manifest.EnvironmentDefinition) []string {
-	result := make([]string, 0, len(environments))
-
-	for _, env := range environments {
-		result = append(result, env.Name)
-	}
-
-	return result
-}
-
 func filterProjectsByName(projects []project.Project, names []string) ([]string, error) {
 	var result []string
 
@@ -400,16 +390,6 @@ func toProjectMap(projects []project.Project) map[string]project.Project {
 
 func containsName(names []string, name string) bool {
 	return slices.Contains(names, name)
-}
-
-func toEnvironmentMap(environments []manifest.EnvironmentDefinition) map[string]manifest.EnvironmentDefinition {
-	result := make(map[string]manifest.EnvironmentDefinition)
-
-	for _, env := range environments {
-		result[env.Name] = env
-	}
-
-	return result
 }
 
 func getClient(environment manifest.EnvironmentDefinition, dryRun bool) (client.Client, error) {
--- a/cmd/monaco/deploy/deploy.go
+++ b/cmd/monaco/deploy/deploy.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/deploy"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/util/maps"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/util/slices"
 	"path/filepath"
 	"strings"
@@ -34,7 +35,7 @@ import (
 	"github.com/spf13/afero"
 )
 
-func Deploy(fs afero.Fs, deploymentManifestPath string, specificEnvironments []string,
+func Deploy(fs afero.Fs, deploymentManifestPath string, specificEnvironments []string, environmentGroup string,
 	specificProject []string, dryRun, continueOnError bool) error {
 
 	deploymentManifestPath = filepath.Clean(deploymentManifestPath)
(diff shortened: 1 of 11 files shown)


pkg/manifest/manifest_loader.go
```diff
@@ -220,12 +220,19 @@ func toEnvironments(context *ManifestLoaderContext, groups []group) (map[string]
 	var errors []error
 	environments := make(map[string]EnvironmentDefinition)
 
+	groupNames := make(map[string]bool, len(groups))
+
 	for i, group := range groups {
 		if group.Name == "" {
 			errors = append(errors, newManifestLoaderError(context.ManifestPath, fmt.Sprintf("missing group name on index `%d`", i)))
-			continue
```
Laubi, line 226: I removed this continue as any error would fail the manifest-loading.
With this removed, we can collect all errors immediately instead of showing this single error to the user and maybe showing more after it is fixed. Wdyt?

cmd/monaco/deploy/deploy_test.go
```diff
@@ -191,6 +191,6 @@ func TestDeploy_ReportsErrorForBrokenV2Config(t *testing.T) {
 	manifestPath, _ := filepath.Abs("manifest.yaml")
 	_ = afero.WriteFile(testFs, manifestPath, []byte("manifestVersion: 1.0\nprojects:\n- name: project\nenvironmentGroups:\n- name: default\n  environments:\n  - name: environment1\n    url:\n      type: environment\n      value: ENV_URL\n    token:\n      name: ENV_TOKEN\n"), 0644)
 
-	err := Deploy(testFs, "manifest.yaml", []string{}, []string{}, true, false)
+	err := Deploy(testFs, "manifest.yaml", []string{}, "", []string{}, true, false)
 	assert.ErrorContains(t, err, "error while loading projects")
 }
```
UnseenWizzard, line 196: no new tests on this level?

cmd/monaco/runner/runner.go
```diff
@@ -203,17 +209,20 @@ func getDeleteCommand(fs afero.Fs) (deleteCmd *cobra.Command) {
...
 			}
 
-			return delete.Delete(fs, manifestName, deleteFile, environment)
+			return delete.Delete(fs, manifestName, deleteFile, environments, group)
 		},
 		ValidArgsFunction: completion.DeleteCompletion,
 	}
 
-	deleteCmd.Flags().StringSliceVarP(&environment, "environment", "e", make([]string, 0), "Deletes configuration only for specified envs. If not set, delete will be executed on all environments defined in manifest.")
+	deleteCmd.Flags().StringVarP(&group, "group", "g", "", "Specify the environmentGroup that should be used for deletion. This flag is mutually exclusive with '--environment'. If this flag is specified, all environments within the specified group will be used for configuration-detection.")
```
UnseenWizzard, line 217:
```suggestion
	deleteCmd.Flags().StringVarP(&group, "group", "g", "", "Specify the environmentGroup that should be used for deletion. This flag is mutually exclusive with '--environment'. If this flag is specified, configuration will be deleted from all environments within the specified group.")
```

pkg/manifest/manifest_loader.go
```diff
@@ -220,12 +220,19 @@ func toEnvironments(context *ManifestLoaderContext, groups []group) (map[string]
 	var errors []error
 	environments := make(map[string]EnvironmentDefinition)
 
+	groupNames := make(map[string]bool, len(groups))
```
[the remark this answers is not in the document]
Laubi: I am aware, but are you sure this is an argument in this code? What performance gains are we talking about? I would argue, absolutely neglectable ones. 
What do we gain in using bools? Clearer code (struct{}{} is not clear), and and easier time checking existences.

pkg/manifest/manifest_loader_test.go
```diff
@@ -692,3 +692,189 @@ func Test_validateManifestVersion(t *testing.T) {
...
+		},
+		{
+			name: "unknown url type",
+			manifestContent: `
+manifestVersion: 1.0
+projects: [{name: a}]
+environmentGroups: [{name: b, environments: [{name: c, url: {value: d, type: f}, token: {name: e}}]}]
+`,
+			errsContain: []string{"f is not a valid"},
+		},
```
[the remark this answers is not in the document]
Laubi: Could be, those tests are also the first tests including any of those checks :)

pkg/manifest/manifest_loader.go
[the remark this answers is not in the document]
UnseenWizzard: > I would argue, absolutely neglectable ones.

I agree, that's what I said in the nitpick ;) 

Personally, I don't find the access much more unreadable though

pkg/manifest/manifest_loader_test.go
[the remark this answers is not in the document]
Laubi: We have a test :816

pkg/manifest/manifest_loader.go
Laubi: It would not 'impact' the behavior, but only some error messages, e.g. if the token is not set, the name cannot be printed

pkg/manifest/manifest_loader.go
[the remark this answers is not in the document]
Laubi: I prefer `if map[key] {}` instead of `if _, f := map[key]; f {` for such simple cases

cmd/monaco/deploy/deploy_test.go
UnseenWizzard: tbh I did not think about what, just wanted to trigger you to think if tests are needed here :wink: 
I didn't take a further look than the change set, and saw a few added/changed conditionals in deploy.go

cmd/monaco/deploy/deploy.go
```diff
@@ -55,13 +56,26 @@ func Deploy(fs afero.Fs, deploymentManifestPath string, specificEnvironments []s
...
+	}
+
+	if len(specificEnvironments) > 0 {
+		environments, err = environments.FilterByNames(specificEnvironments)
+		if err != nil {
+			return fmt.Errorf("failed to filter environments: %w", err)
+		}
 	}
 
-	environmentMap := toEnvironmentMap(environments)
```
UnseenWizzard, line 63: seems these methods could now be removed?

Laubi: Yes, removed

cmd/monaco/deploy/deploy_test.go
UnseenWizzard: as discussed elsewhere, let's look into proper automated testing for environment deployments and overwrite on it's own in a followup

PR #810, opened 2023-02-14
fix: fixed module name in intellij run configurations

warber:


warber:
already covered in https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/809

PR #818, opened 2023-02-15
feat: Implement Version Check for Settings 2.0 Upload on older DT servers

warber:
Added functionality to `DynatraceClient` to be aware of the dynatrace server version:
It is possible to add the following option when creating a `DynatraceClient`:

You can choose between determining the server version automatically during construction of the `DynatraceClient`:

`NewDynatraceClient("http://xxxxx.com/","xxxx",WithAutoServerVersion())` or

determine the version on your own and pass it using:
`NewDynatraceClient("http://xxxxx.com/","xxxx",WithServerVersion(specificVersion))` 

(useful for testing, or situations where it is not convinient to do network calls during construction of the object.

The `download` CLI command now includes determining the Dynatrace server version and a check whether it is bellow
_"1.262.0"_ in which case we print out a warning. Execution of the command is still considered to be successful and returns with exit code 0.

_Example:_
```
$ monaco download direct https://xxxxxx.com TOKEN
2023/02/15 10:09:23 INFO  Dynatrace Configuration as Code v2.x
2023/02/15 10:09:23 WARN  Uploading Settings 2.0 objects to the same environment is not possible due to your cluster version being below 1.262.0, which Monaco does not support for reliably updating downloaded settings. Please refer to the documentation for guidance.
[....]
```

The `deploy` CLI command now includes determining the Dynatrace server version and a check whether it is bellow
_"1.262.0"_ in which case we print out a warning. Execution of the command is still considered to be successful and returns with exit code 0.

_Example:_
```
$ monaco deploy ./path/to/manifes.yaml
[...]
2023/02/15 10:34:22 WARN  Unable to update Settings 2.0 object of schema "xxxxxxxxxx" and object id "xxxxxxxxxxxxxxx" on Dynatrace tenant with a version < 1.262.0
2023/02/15 10:34:22 INFO  Deployment finished without errors

```

Note: I've intentionally skipped doing anything for `delete` command as it is mentioned in the ticket, as I could not remember the reason for it. Pls shout at me if it's needed/should be added.

### DIFF
--- a/cmd/monaco/download/download_command.go
+++ b/cmd/monaco/download/download_command.go
@@ -16,6 +16,10 @@ package download
 
 import (
 	"fmt"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/client"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/util"
+	"net/http"
+	"os"
 
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
@@ -105,6 +109,9 @@ func GetDownloadConfigsCommand(fs afero.Fs, command Command, downloadCmd *cobra.
 			return nil
 		},
 		ValidArgsFunction: completion.DownloadDirectCompletion,
+		PreRun: func(cmd *cobra.Command, args []string) {
+			printUploadToSameEnvironmentWarning(args[0], os.Getenv(args[1]))
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			url := args[0]
 			tokenEnvVar := args[1]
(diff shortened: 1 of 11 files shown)


pkg/util/version_utils.go
```diff
@@ -49,10 +55,23 @@ func (v Version) GreaterThan(other Version) bool {
...
 
+// SmallerThan determines whether this version is smaller than the given version
 func (v Version) SmallerThan(other Version) bool {
 	return other.GreaterThan(v)
 }
 
+// Valid returns whether this version is valid or not.
+// A version is considered to be "invalid" if it is equal to "0.0.0" or
+// contains a negative number, e.e. "0.-1.2"
+func (v Version) Valid() bool {
```
Laubi, line 66: Shouldn't this be InValid? 
According to your tests, Version{-1, 1, 1} would be valid

warber: fixedup - never happend :D

Laubi (approved):

cmd/monaco/download/download_command.go
```diff
@@ -253,3 +260,19 @@ func setupSharedFlags(cmd *cobra.Command, project, outputFolder *string, forceOv
 		log.Fatal("failed to setup CLI %v", err)
 	}
 }
+
+// printUploadToSameEnvironmentWarning function may display a warning message on the console,
+// notifying the user that downloaded objects cannot be uploaded to the same environment.
+// It verifies the version of the tenant and, depending on the result, it may or may not display the warning.
+func printUploadToSameEnvironmentWarning(environmentURL, token string) {
```
jskelin, line 267: for me dominant function is `check`, not `printWarning`.
That's why I would go with a name similar to `checkVersionCompatibility`

warber: disagree, the actual purpose is to warn the user by printing something on the screen. How it is done is described in the comment.

jskelin (approved):

warber:
@Laubi I've changed the warn message a bit according to our shourt discussion on slack. pls take a look https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/2f620316150b130631f5f60007481923653b5f4c/cmd/monaco/download/download_command.go#L274

Laubi (approved):

PR #819, opened 2023-02-15
fix: Extend error handling for entity downloads

UnseenWizzard:
additional changes:
- Make sure logs appear when threads are started
- Not when threads are created
- Make the download process faster for entities
- Make the download process more error-resilient
- Make the download process explain some errors when still not resilient
- Set concurrent env. var to empty to test default
- Add Entities Config Type
- Update tests with new object definitions
- correct small json encoding bug for entities lists
- add more unit testing
- sanitize directories: Make sure there are no semi-colon on Windows


dcryans:
Pushed changes and answered to questions in the previous PR for context:
https://github.com/Dynatrace/internal-monitoring-as-code/pull/86

internal/environment/get_env_value.go
```diff
@@ -0,0 +1,94 @@
...
+	"os"
+	"strconv"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
+)
+
+const (
+	DefaultEnvValueInt         = 0
+	ConcurrentRequestsEnvKey   = "CONCURRENT_REQUESTS"
+	DefaultConcurrentDownloads = 50
```
UnseenWizzard, line 27: with you changes this does not need to be an exported constant anymore. Neither does the Default zero value.
You could make them private to the package, or IMO just write them into the map definition and default return path directly (which would just mean you have assert actual values in the tests)

pkg/rest/retry.go
```diff
@@ -73,7 +75,7 @@ func GetWithRetry(client *http.Client, url string, apiToken string, settings Ret
 	} else {
 		additionalMessage := ""
 		if resp.StatusCode == 403 {
-			concurrentDownloadLimit := ConcurrentRequestLimitFromEnv(false)
+			concurrentDownloadLimit := environment.GetEnvValueInt(environment.ConcurrentRequestsEnvKey)
```
UnseenWizzard, line 78: As discussed, do you think this could move into download? 
On any other (non-parallel) use of GetWithRetry - e.g. during deployment this wrongly suggest that a 403 could mean rate-limiting, and changing the env var would actually not make any difference at all

pkg/client/client.go
```diff
@@ -751,33 +691,34 @@ func (d *DynatraceClient) DeleteSettings(objectID string) error {
...
+func isRetryOnEmptyResponse(receivedCount int, emptyResponseRetryCount int, resp rest.Response) (bool, int, error) {
 	if receivedCount == 0 {
-		if *emptyResponseRetryCount < emptyResponseRetryMax {
-			*emptyResponseRetryCount++
-			throttleTooManyRequests("Received empty array response, retrying with same nextPageKey (HTTP: %d) ", resp.StatusCode)
-			return true, nil
+		if emptyResponseRetryCount < emptyResponseRetryMax {
+			emptyResponseRetryCount++
+			rateLimitStrategy := rest.CreateRateLimitStrategy()
+			rateLimitStrategy.ThrottleCallAfterError("Received empty array response, retrying with same nextPageKey (HTTP: %d) ", resp.StatusCode)
```
UnseenWizzard, line 700: hmm.. I actually meant that we could possibly move the logic of these methods into a new implementation of RateLimitStrategy that is then used for the entity download.

If that is not possible (didn't check how the code would look, where the strategy is created right now), I would actually suggest you stick to just the sleep as part of the client. 
I would then move this code into entities_client at least though, as it's entity specific and client.go is already huge

internal/environment/get_env_value.go
dcryans: right, asserting in the tests will fix my concern with using the map

pkg/client/client.go
dcryans: I'll see what can be done

pkg/client/client.go
```diff
@@ -443,10 +519,10 @@ func (d *DynatraceClient) ListSettings(schemaId string, opts ListSettingsOptions
 			}
 		}
 
-		return nil
+		return len(parsed.Items), len(result), nil
 	}
 
-	err := d.ListPaginated(pathSettingsObjects, params, addToResult)
+	_, err := d.ListPaginated(pathSettingsObjects, params, schemaId, addToResult)
```
UnseenWizzard, line 525: just noticed this in a discussion about another PR: `ListPaginated` does not need to/should not be exported either. 
Could you please turn this into `listPaginated` when you address the other functions that don't need to be exported? Thanks!

pkg/client/entities_client.go
```diff
@@ -122,3 +127,77 @@ func isInvalidReflectionValue(value reflect.Value) bool {
 		return false
 	}
 }
+
+func GenListEntitiesParams(entityType string, entitiesType EntitiesType, ignoreProperties []string) url.Values {
```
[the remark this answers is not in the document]
dcryans: Done

internal/environment/get_env_value.go
dcryans: Done

pkg/client/client.go
dcryans: Done

pkg/rest/retry.go
dcryans: Change Done
I still think this change was too expensive compared to the risk you are mentioning
I hope that, in the future, I will not have to make the choice of not helping users with their errors to avoid delaying my project.

pkg/client/client.go
dcryans: BTW, it is not Entities Specific, well I don't believe so
I'm pretty sure it could happen for Settings as well, given an big enough load on the cluster

dcryans: If it did happen without my code present, you couldn't notice, as an empty response would not trigger errors

dcryans: I extracted the shared throttling code from rateLimiter to an internal source
This will safeguard the integrity of the rateLimiter's original purpose
While allowing me to throttle calls made after specific errors

If I don't do this, then I'm creating unnecessary 403s, which will force users to lower their concurrent requests, and restart their lengthy extractions after waiting for 5-10minutes

The real problem is that the rateLimiter only acts after we have busted the limit, it doesn't limit the rate at all.
A real rateLimiter would limit the global amount of calls 'per minute' for all threads, making that we never bust the limit, or at least try not to as we are not the only users of the clusters

Then I wouldn't need to add individual throttling for all of my error handling
And we could use as many ConcurrentRequests as we want because they would be limited anyways

dcryans:
@UnseenWizzard 
Rebase seems to have gone a bit crazy, I see tons of commits that are duplicated
Ok if I create a new branch and 'manually squash' my changes onto it?
Unless you have another trick in your sleave?

PR #821, opened 2023-02-16
refactor: simplify filtering in api

jskelin:


UnseenWizzard:
Seems you missed to update a unit test to a changed signature
`Error: cmd/monaco/download/download_test.go:248:19: assignment mismatch: 2 variables but getApisToDownload returns 1 value`

cmd/monaco/download/download_configs.go
```diff
@@ -145,10 +158,10 @@ func downloadConfigs(apis api.ApiMap, opts downloadOptions) (project.ConfigsPerT
 
 	c = client.LimitClientParallelRequests(c, opts.concurrentDownloadLimit)
 
-	apisToDownload, errors := getApisToDownload(apis, opts.specificAPIs)
-	if len(errors) > 0 {
-		err = PrintAndFormatErrors(errors, "failed to load apis")
-		return nil, err
+	apisToDownload := getApisToDownload(apis, opts.specificAPIs)
+	if len(apisToDownload) == 0 {
+		util.PrintError(fmt.Errorf("no APIs to download"))
```
UnseenWizzard, line 163: as you already fmt an error message here, I would also just return this error.
Also if the returned error from this method is printed further up the chain, there might not be a need for PrintError on this level at all.

cmd/monaco/download/download_configs.go
```diff
@@ -185,38 +198,26 @@ func downloadConfigs(apis api.ApiMap, opts downloadOptions) (project.ConfigsPerT
...
+	} else {
+		return apis.Filter(skipDownloadFilter, removeDeprecatedEndpoints)
 	}
+}
 
-	if len(apisToDownload) == 0 {
-		errors = append(errors, fmt.Errorf("no APIs to download"))
+func skipDownloadFilter(api api.Api) bool {
+	if api.ShouldSkipDownload() {
+		log.Info("APIs that won't be downloaded and need manual creation: '%v'.", api.GetId())
```
UnseenWizzard, line 211:
```suggestion
		log.Info("API can not be downloaded and needs manual creation: '%v'.", api.GetId())
```
previously this logged the full list in one line. now you log it per API, which seems fine, but the message should change a bit IMO

cmd/monaco/download/download_configs.go
```diff
@@ -185,38 +198,26 @@ func downloadConfigs(apis api.ApiMap, opts downloadOptions) (project.ConfigsPerT
...
-
-	return apisToDownload, errors
+	return false
 }
 
-func deprecatedEndpointFilter(api api.Api) bool {
-	return api.DeprecatedBy() != ""
+func removeDeprecatedEndpoints(api api.Api) bool {
+	if api.DeprecatedBy() != "" {
+		log.Warn("API '%s' is deprecated by '%s' and will not be downloaded", api.GetId(), api.DeprecatedBy())
```
UnseenWizzard, line 219: FYI: you could use %q format to automatically print the string surrounded by quotes
```suggestion
		log.Warn("API %q is deprecated by %q and will not be downloaded", api.GetId(), api.DeprecatedBy())
```

pkg/api/api_test.go
```diff
@@ -43,182 +40,43 @@ func TestCreateApis(t *testing.T) {
 	apis := make(map[string]Api)
 	apis = NewApis()
```
UnseenWizzard, line 41: you can just do apis := NewApis()

pkg/api/api_test.go
```diff
@@ -43,182 +40,43 @@ func TestCreateApis(t *testing.T) {
...
-	notification, ok := apis["notification"]
-	assert.Assert(t, ok, "Expected `notification` key in KnownApis")
-	assert.Equal(t, notification.GetUrl(testDevEnvironment.GetEnvironmentUrl()), "https://url/to/dev/environment/api/config/v1/notifications", "Expected to get `notification` API url")
+	assert.Contains(t, apis, "notification", "Expected `notification` key in KnownApis")
+	assert.Equal(t, "https://url/to/dev/environment/api/config/v1/notifications", apis["notification"].GetUrl(testDevEnvironment.GetEnvironmentUrl()), "Expected to get `notification` API url")
 }
 
 func TestCreateApisResultsInError(t *testing.T) {
 	apis := make(map[string]Api)
 	apis = NewApis()
```
UnseenWizzard, line 49: you can just do apis := NewApis()

pkg/api/filter.go
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package api
+
+// Filter return true iff specific api needs to be filtered/ removed from list
+type Filter func(api Api) bool
+
+// Filter apply all passed filters and return new filtered array
+func (m ApiMap) Filter(filters ...Filter) ApiMap {
+	apis := make(ApiMap)
```
UnseenWizzard, line 24: nit/question: indentation looks kinda off in this file? like it's a level too deep. 
Did you run `make format`?

UnseenWizzard (changes requested):
(setting 'request changes' as otherwise it stays in review required filter)

cmd/monaco/download/download_configs.go
jskelin: tnx

pkg/api/filter.go
jskelin: it is done now

cmd/monaco/download/download_configs.go
jskelin: But it is more clear if result is just returned from the sub function, and than on the upper lever check result and make decision. 
It will also be easier for further refactoring

UnseenWizzard: this comment still stands - did you have a reason not to address it? :)

jskelin: furthermore, I would even go without error at this place.
That kind of error should happen before entering pkg command in validation phase of preparing data. and this should work fine if empty slice is passed.

UnseenWizzard: I don't quite understand your argument, maybe mine wasn't worded well either.
 
Yes this should absolutely return an error.
I meant instead of 

```
util.PrintError(fmt.Errorf("..."))
return fmt.Errorf("...")
```

do e.g. 
```
err := fmt.Errorf("...")
util.PrintError(err)
return fmt.Errorf(err)
```

or if the method that ends up calling this already prints the error when handling it (which it probably does, as your top-level error handling boils down to "tell the user"?) just do:
```
return fmt.Errorf(err)
```

UnseenWizzard (approved):

PR #823, opened 2023-02-16
Ci/explicit workflow permissions

UnseenWizzard:
This sets explicit minimal permissions for all workflows

### DIFF
--- a/.github/workflows/end-to-end-test.yml
+++ b/.github/workflows/end-to-end-test.yml
@@ -47,7 +47,7 @@ jobs:
         if: github.event.action == 'labeled' && github.event.label.name == env.E2E_TEST_LABEL
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.event.pull_request.head.ref }}
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: 🏁 Build release binaries
         run: make build-release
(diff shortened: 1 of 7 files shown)


.github/workflows/end-to-end-test.yml
```diff
@@ -47,7 +47,7 @@ jobs:
         if: github.event.action == 'labeled' && github.event.label.name == env.E2E_TEST_LABEL
         uses: actions/checkout@v3
         with:
-          ref: ${{ github.event.pull_request.head.ref }}
+          ref: ${{ github.event.pull_request.head.sha }}
```
warber, line 50: intended? i don't quite know what this is doing

warber (approved):

.github/workflows/end-to-end-test.yml
UnseenWizzard: yes intended :) See the commit message for that one. 
In short rather than checking out the HEAD reference it's the exact head sha at the time of the labeled event. 
So for this PR the E2E build checked out 464dc05153d1a368ffc67e4de62414404f5a39e5

UnseenWizzard:
merging diregarding the E2E check which has failed due to a non-cleaned up configuriation

.github/workflows/end-to-end-test.yml
warber: thx

PR #833, opened 2023-02-21
chore: cleanup packages

warber:
follow up of https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/832
* group id related code into `internal/idutils`
* renamed `internal/environment` package to `internal/featureflags`

### DIFF
--- a/cmd/monaco/download/download_command.go
+++ b/cmd/monaco/download/download_command.go
@@ -16,7 +16,7 @@ package download
 
 import (
 	"fmt"
-	utilEnv "github.com/dynatrace/dynatrace-configuration-as-code/internal/environment"
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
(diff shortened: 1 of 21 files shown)


cmd/monaco/download/download_command.go
```diff
@@ -48,7 +48,7 @@ To download entities, use download entities`,
 
 	GetDownloadConfigsCommand(fs, command, downloadCmd)
 
-	if utilEnv.FeatureFlagEnabled("MONACO_FEAT_ENTITIES") {
+	if featureflags.FeatureFlagEnabled("MONACO_FEAT_ENTITIES") {
```
UnseenWizzard, line 51: nitpick: just reading this, this could probably be called as 
```suggestion
	if featureflag.Enabled("MONACO_FEAT_ENTITIES") {
```
and tell me everything I need to know with a simpler package and function name

UnseenWizzard (approved):

cmd/monaco/download/download_command.go
warber: i also thought about exactly naming it like that first. :D 
my gut feeling tells me that `featureFlag.IsFeatureEnabled()` is the most common.

PR #834, opened 2023-02-22
Fix: Enhance settings upsert logic to work with current enhancements of the settings API

warber:
Changes in upsert logic:

* get settings2.0 object with matching origin Object ID instead of external ID, if a setting obj was found and tenant version is < 1.262 we print the warning
* Deleted PUT operation on settings endpoint as it is not needed. POST endpoint now handles both, creating and updating settings objects

Further:
* deleted settings download filters that are not needed anymore with recent changes in settings fw
* Adjusted methods names of ConfigClient interface to contain reflect the fact that it is manipulating Configs as these interfaces are embedded into a bigger one and its clearer what each method does

btw: I don't know why legacy integration tests fail on GH. I ran them from my local machine, and it was running fine.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -35,42 +35,42 @@ import (
 // ConfigClient is responsible for the classic Dynatrace configs. For settings objects, the [SettingsClient] is responsible.
 // Each config endpoint is described by an [Api] object to describe endpoints, structure, and behavior.
 type ConfigClient interface {
-	// List lists the available configs for an API.
+	// ListConfigs lists the available configs for an API.
 	// It calls the underlying GET endpoint of the API. E.g. for alerting profiles this would be:
 	//    GET <environment-url>/api/config/v1/alertingProfiles
 	// The result is expressed using a list of Value (id and name tuples).
-	List(a Api) (values []Value, err error)
+	ListConfigs(a Api) (values []Value, err error)
 
-	// ReadById reads a Dynatrace config identified by id from the given API.
+	// ReadConfigById reads a Dynatrace config identified by id from the given API.
 	// It calls the underlying GET endpoint for the API. E.g. for alerting profiles this would be:
 	//    GET <environment-url>/api/config/v1/alertingProfiles/<id> ... to get the alerting profile
-	ReadById(a Api, id string) (json []byte, err error)
+	ReadConfigById(a Api, id string) (json []byte, err error)
 
-	// UpsertByName creates a given Dynatrace config if it doesn't exist and updates it otherwise using its name.
+	// UpsertConfigByName creates a given Dynatrace config if it doesn't exist and updates it otherwise using its name.
 	// It calls the underlying GET, POST, and PUT endpoints for the API. E.g. for alerting profiles this would be:
 	//    GET <environment-url>/api/config/v1/alertingProfiles ... to check if the config is already available
 	//    POST <environment-url>/api/config/v1/alertingProfiles ... afterwards, if the config is not yet available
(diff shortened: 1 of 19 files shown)


pkg/client/client.go
```diff
@@ -293,13 +293,13 @@ func isNewDynatraceTokenFormat(token string) bool {
...
 
-	// list all settings with matching external ID
+	//list all settings with matching origin object ID
 	settings, err := d.ListSettings(obj.SchemaId, ListSettingsOptions{
 		// we don't care about the actual value setting
 		DiscardValue: true,
 		// only consider objects with an external id that matches the expected one
 		Filter: func(o DownloadSettingsObject) bool {
-			return o.ExternalId == externalId
+			return o.ObjectId == obj.OriginObjectId
```
UnseenWizzard, line 302: wait, so we're fulling letting the Settings framework handle the "upsert" logic? 

does making this call even make sense anymore then? If it's just used for the old version + existing object check maybe we can just implement a GET by object id rather than iterating?

warber: yes, it's completely handled by the settings fw.
yes, actually a get would be also possible. didn't thought about that :+1:

UnseenWizzard: As is we load and iterate over potentially hundreds of settings just to find a specific one. 
I would really suggest we replace it with just a get by objectId

warber: just saw that the current implementation of `DynatraceClient` does not have a method to Get a single object. Are you OK with doing this "optimization" in a follow-up ? :)

warber:
@UnseenWizzard I've addressed your comment (GET single setting object instead of iterating) 
Addidionally, I've done a little cleanup --> renaming the methods of ConfigClient to have the name "config(s)" in it to make it more clear what they are targeting, since they are embedded into a larger interface with the other interfaces for settings and entities....

pkg/client/client.go
```diff
@@ -291,98 +297,57 @@ func isNewDynatraceTokenFormat(token string) bool {
...
-		}
-
-		log.Debug("\tCreated object %s (%s) with externalId %s", obj.Id, obj.SchemaId, externalId)
-		return entity, nil
-	} else {
-		payload, err := buildPutRequestPayload(obj, externalId)
-		if err != nil {
-			return DynatraceEntity{}, fmt.Errorf("failed to build settings object for upsert: %w", err)
+		if fetchedSettingObj != nil {
+			log.Warn("Unable to update Settings 2.0 object of schema %q and object id %q on Dynatrace tenant with a version < 1.262.0", obj.SchemaId, obj.OriginObjectId)
```
UnseenWizzard, line 312: Only just noticed 'Dynatrace environment' rather than tenant please.

pkg/client/client.go
```diff
@@ -291,98 +297,57 @@ func isNewDynatraceTokenFormat(token string) bool {
...
+	if err != nil {
+		return DynatraceEntity{}, fmt.Errorf("failed to upsert dynatrace obj: %w", err)
+	}
 
-		entity, err := parsePutResponse(resp)
-		if err != nil {
-			return DynatraceEntity{}, fmt.Errorf("failed to parse response: %w", err)
-		}
+	if !success(resp) {
+		return DynatraceEntity{}, fmt.Errorf("failed to update settings object with externalId %s (HTTP %d)!\n\tResponse was: %s", externalId, resp.StatusCode, string(resp.Body))
```
UnseenWizzard, line 341: `failed to upsert` rather than "update"

UnseenWizzard (commented):
looks good to me, comments are just about wording in user facing messages

UnseenWizzard (approved):

PR #836, opened 2023-02-23
Improve dependency resolution & topology sort performance

Laubi:


cmd/monaco/download/download.go
```diff
@@ -141,12 +141,14 @@ func writeConfigs(downloadedConfigs project.ConfigsPerType, opts downloadOptions
 		return err
 	}
 
+	log.Info("Searching for circular dependencies")
```
warber, line 144: I would move this and line 148 to DEBUG

Laubi: I have a perfect reason I added it and that it's an info message.

After downloading, we log "Downloaded configurations written to '...'" and then *nothing* happens. But the program also does not end. Very confusing for any user.
With that info line, we indicate that we're still looking doing something, and that there's a reason the program's not done yet

warber: makes sense :)

warber (approved):

PR #844, opened 2023-02-27
Fix/setup client correctly in e2e tests

UnseenWizzard:


cmd/monaco/integrationtest/v1/integration_test_utils.go
```diff
@@ -56,7 +56,8 @@ func AssertAllConfigsAvailable(t *testing.T, fs afero.Fs, manifestFile string, m
 	projects := loadProjects(t, fs, manifestFile, mani)
 
 	for _, e := range environments {
-		client := clientFromEnvDef(t, e)
+		c, err := client.CreateClientForEnvironment(e)
```
UnseenWizzard, line 59: @warber I have reused the method you added to the client, but to be honest I'm not happy that it exists. 
Only for this convenience method we now have a dependency from `client` to `manifest` that is otherwise not needed. Without this, `client` would only depend on `api` and `rest` which seems more 'correct'

I can see how the method is helpful but before I started using it here, it was only used once.
I didn't want to change it in this PR, but IMO this method should at least move out of `client` or just be reverted to callers being responsible of supplying URL & token correctly. 
Wdyt?

warber: agree, this is bs :+1:

UnseenWizzard: Taking a look at removing this I see a way to refactor the whole v1/v2 e2e test utils a bit and reuse more things. 
I would suggest we merge this for now, and I open an extra PR for that later today

warber (approved):

Laubi (approved):

PR #847, opened 2023-02-28
chore: Cleanup v1 environment interface

warber:
Just one of many baby steps for refactoring.
* Removed unnecessary public Environment interface for environments v1.
* Unexported stuff
* moved Environment v1 convert package


pkg/environment/v1/environment.go
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package environment
+package v1
```
UnseenWizzard, line 17: I think this was always just `environment`, because no v2 environment package exists - they're just part of the manifest in v2. 
Do you plan to move v2 environment related things to the package in the future? Or just added v1 to align with other packages?

pkg/converter/converter.go
```diff
@@ -106,21 +97,22 @@ func newReferenceParserError(projectId string, config configV1.Config, parameter
...
 }
 
 var (
 	_ configErrors.ConfigError = (*ConvertConfigError)(nil)
-	_ configErrors.ConfigError = (*ReferenceParserError)(nil)
+	_ configErrors.ConfigError = (*ReferenceParseError)(nil)
 )
 
-func Convert(context ConverterContext, environments map[string]environmentV1.Environment,
+// Convert takes a map of v1 environments and converts it to v2 manifest
```
UnseenWizzard, line 114:
```suggestion
// Convert takes a map of v1 environments and converts it to v2 manifest
```

```suggestion
// Convert takes v1 environments and projects and converts them into a v2 manifest and projects
```

UnseenWizzard (dismissed):

pkg/environment/v1/environment.go
warber: >  they're just part of the manifest in 
Actually, no it is not used in Manifest v2. that's why i moved it to v1. like it was done for e.g. project v1.

UnseenWizzard: > no it is not used in Manifest v2

I meant to say that the concept of `environment` in v2 is part of the manifest directly. 

So unlike `config` and `project` there is no v2 code for the `environment` package, which is why there was no sub-package so far. 
If there's no intention to change that, the package could just become `v1environment` instead of adding a sub-package with a non-explicit import name

warber: I've moved environment stuff into "v1environment" package inside convert

pkg/api/api_test.go
```diff
@@ -19,13 +19,13 @@
...
 import (
-	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/environment/v1"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/converter/v1environment"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
-var testDevEnvironment = v1.NewEnvironment("development", "Dev", "", "https://url/to/dev/environment", "DEV")
+var testDevEnvironment = v1environment.NewEnvironmentV1("development", "Dev", "", "https://url/to/dev/environment", "DEV")
```
UnseenWizzard, line 28: stupid question, why does the current api_test need a v1 environment?

UnseenWizzard (approved):

pkg/api/api_test.go
UnseenWizzard: okay, I just looked... it is used twice for: `.GetUrl(testDevEnvironment.GetEnvironmentUrl()),` :sweat_smile: 
You could replace this with a string

PR #850, opened 2023-03-01
feat: Allow defining the environment-type in the manifest

Laubi:


UnseenWizzard (approved):

pkg/manifest/manifest_loader.go
```diff
@@ -304,6 +310,19 @@ func extractUrlType(config environment) (UrlType, error) {
...
 
+func parseEnvironmentType(context *ManifestLoaderContext, config environment, g string) (EnvironmentType, error) {
+	switch strings.ToLower(config.Type) {
+	case "":
+		fallthrough
+	case "classic":
+		return Classic, nil
+	case "platform":
+		return Platform, nil
+	}
```
warber, line 321:
```suggestion
switch strings.ToLower(config.Type) {
case "", "classic":
	return Classic, nil
case "platform":
	return Platform, nil

```

pkg/manifest/manifest_writer.go
```diff
@@ -128,6 +130,17 @@ func toWriteableEnvironmentGroups(environments map[string]EnvironmentDefinition)
...
 
+func getType(env EnvironmentDefinition) string {
+	switch env.Type {
+	case Classic:
+		return "classic"
+	case Platform:
+		return "platform"
+	}
+
+	panic(fmt.Sprintf("Unexpected environment type %q in environment %q.", env.Type, env.Name))
```
warber, line 141: i think that's not the right place to panic

UnseenWizzard: I missed that! Sry. Agree with @warber if we have an error case here we should return and handle errors as everywhere else - even if it's a "this should never happen" case like I think this one is

PR #851, opened 2023-03-01
Feat: Prepare `DynatraceClient` to support OAuth authorization

warber:
* PR contains changes that allow passing an `http.Client` that hides the authorization mechanism to the `DynatraceClient`.
Example Usage:

```golang
client.NewDynatraceClient(client.NewTokenAuthClient(environment.Token), envURL, client.WithAutoServerVersion())
``` 
or
```golang
client.NewDynatraceClient(client.NewOAuthClient(oauthCredentials), envURL, client.WithAutoServerVersion())
```

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -213,13 +213,6 @@ func WithRetrySettings(retrySettings rest.RetrySettings) func(*DynatraceClient) 
 	}
 }
 
-// WithHTTPClient sets the http client to be used by the DynatraceClient
-func WithHTTPClient(client *http.Client) func(dynatraceClient *DynatraceClient) {
-	return func(d *DynatraceClient) {
-		d.client = client
-	}
-}
-
 // WithServerVersion sets the Dynatrace version of the Dynatrace server/tenant the client will be interacting with
 func WithServerVersion(serverVersion version.Version) func(client *DynatraceClient) {
 	return func(d *DynatraceClient) {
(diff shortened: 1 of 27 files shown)


pkg/client/client.go
```diff
@@ -251,15 +307,13 @@ func NewDynatraceClient(environmentURL string, token string, opts ...func(dynatr
 		log.Warn("You are using an insecure connection (%s). Consider switching to HTTPS.", parsedUrl.Scheme)
 	}
 
-	if !isNewDynatraceTokenFormat(token) {
-		log.Warn("You used an old token format. Please consider switching to the new 1.205+ token format.")
-		log.Warn("More information: https://www.dynatrace.com/support/help/dynatrace-api/basics/dynatrace-api-authentication")
+	if httpClient == nil {
+		httpClient = &http.Client{}
```
UnseenWizzard, line 311: does that case still happen and should we even support it? Or rather error out if no client was given?

UnseenWizzard (approved):

pkg/client/client.go
warber: I don't have a strong opinion on that.
I think it is easy to handle so let's keep it?

UnseenWizzard (approved):

pkg/client/client.go
UnseenWizzard: fine by me, was just wondering

Laubi (approved):

PR #858, opened 2023-03-03
Fix: Handle duplicate config IDs when converting v1 config

warber:


pkg/project/v1/project.go
```diff
@@ -239,8 +235,24 @@ func (p *projectBuilder) sortConfigsAccordingToDependencies() error {
 	return err
 }
 
+func (p *projectBuilder) dedupConfigIDs() {
```
jskelin, line 238: i would go with a different name, e.g. `resolveDuplicateIDs`

jskelin (approved):

pkg/project/v1/project.go
warber: done

jskelin (dismissed):

jskelin (approved):

PR #860, opened 2023-03-06
Refactor/remove interface from api

jskelin:


Laubi:
Two major points from my side (nitpicks come later)
* I'd change a lot of occurrences from *API to 'only' API, as no copies will be made most of the time since they are in a slice or map
* I'd still call the ApiMap ApiMap, not APIs. APIs sound more like a slice, not a Map, what it is in the end (and used as such, which makes for me the important difference. If it would be used only with methods, I would not care, but the object is used as a map *everywhere*)

I also need to have a deeper look

pkg/api/apis.go
```diff
@@ -0,0 +1,117 @@
...
+func getAPIs(fromApiInputs APIs) APIs {
+	apis := make(APIs)
+	for id, a := range fromApiInputs {
+		apis[id] = newAPI(id, a)
+		//a.copy().setID(id).applyRules(nameOfGetAllResponse, singleConfigurationApi)
+		//newAPI(*a, setId(id), nameOfGetAllResponse, singleConfigurationApi) //order of applying roles is important!!!
+	}
+
+	return apis
+}
```
warber, line 43: I think to have a "read-only" view of APIs, we can change the type from `map[string]*API` to `map[string]API`

pkg/api/api.go
```diff
@@ -62,154 +28,77 @@ type apiImpl struct {
...
 
-// Contains return true iff requested API is part of particular ApiMap
-func (m ApiMap) Contains(api string) bool {
-	_, ok := m[api]
-	return ok
+// newAPI create new API as a copy of template with a given name
+func newAPI(name string, template *API) *API {
+	n := *template // if API structure is contains pointers or substructure, perform deep copy!!!
+	n.id = name
+	n.applyRules(nameOfGetAllResponse, singleConfigurationApi) //order of applying roles is important!!!
```
warber, line 78: It's unclear to me what are "rulses" that need to be applied?
as it seems you just check for different combination of set values on the API values.
Couldnt that just be hardcoded in the actual configEndpoints values?

jskelin:
> naming `APIs` is same approach sas `type Environments map[string]EnvironmentDefinition` and it leads to the point where inner implementation is going to be closed and exposed only via well defined functions and methods.

Laubi:
> naming `APIs` is same approach sas `type Environments map[string]EnvironmentDefinition` and it leads to the point where inner implementation is going to be closed and exposed only via well-defined functions and methods.

Those are bad, and a future refactoring of mine would be to get rid of them. 
The map[] is really not useful and usually results in duplicate information. For APIs, where we have a lot and need to look it up often, it makes sense. 
But for the naming, I'd still go with ApiMap.
`Environments` does not sound like a `map[string]Env`, nor does `Projects`. Sound more like `Env[]` and `Project[]`.

pkg/project/v1/project_loader_test.go
```diff
@@ -25,9 +25,9 @@ import (
 	"strings"
 	"testing"
 
-	"gotest.tools/assert"
-
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/api"
+	"github.com/stretchr/testify/assert"
+	assert2 "gotest.tools/assert"
```
Laubi, line 30: We discussed this at some point that we want to use `testify` in all new files (and switch it slowly in old ones), but don't want to mix them.

pkg/project/v1/project_loader.go
```diff
@@ -122,7 +122,18 @@ func getAllProjectFoldersRecursively(fs afero.Fs, availableApis api.ApiMap, path
...
+// folders with API in path are not valid projects
+func containsApiName(apis api.APIs, path string) bool {
+	for api := range apis {
+		if strings.Contains(path, api) {
+			return true
+		}
+	}
+
+	return false
+}
```
Laubi, line 135: nitpick: newline.
I am surprised gofmt does not change this

pkg/download/classic/filter_test.go
```diff
@@ -99,38 +100,38 @@ func TestShouldConfigBeSkipped(t *testing.T) {
...
 	t.Run("anomaly-detection-metrics - ids should not be skipped", func(t *testing.T) {
-		assert.False(t, apiFilters["anomaly-detection-metrics"].shouldBeSkippedPreDownload(api.Value{
+		assert.False(t, apiFilters["anomaly-detection-metrics"].shouldBeSkippedPreDownload(client.Value{
 			Id: "b836ff25-24e3-496d-8dce-d94110815ab5",
 		}))
 	})
 
 	t.Run("anomaly-detection-metrics - random strings should not be skipped", func(t *testing.T) {
-		assert.False(t, apiFilters["anomaly-detection-metrics"].shouldBeSkippedPreDownload(api.Value{
+		assert.False(t, apiFilters["anomaly-detection-metrics"].shouldBeSkippedPreDownload(client.Value{
```
Laubi, line 134: Not happy with the name `Value,` but also not part of this pr. We might find a better name for it at some point.

pkg/deploy/deploy.go
```diff
@@ -95,10 +95,10 @@ func getWordsForLogging(isDryRun bool) (action, verb string) {
 	return "Deploying", "deploy"
 }
 
-func deployConfig(client client.ConfigClient, apis api.ApiMap, entityMap *entityMap, conf *config.Config) (parameter.ResolvedEntity, []error) {
+func deployConfig(configClient client.ConfigClient, apis api.APIs, entityMap *entityMap, conf *config.Config) (parameter.ResolvedEntity, []error) {
 
 	apiToDeploy := apis[conf.Coordinate.Type]
-	if apiToDeploy == nil {
+	if apiToDeploy == (api.API{}) {
```
Laubi, line 101: Instead of a nil check, use the found check in the map lookup

pkg/api/endpointsV1.go
```diff
@@ -0,0 +1,349 @@
...
+	{
+		ID:                  "geo-ip-address-mappings",
+		URLPath:             "/api/config/v1/geographicRegions/ipAddressMappings",
+		SingleConfiguration: true,
+	},
+}
+
+// GetV2Name returns the ID of APIs in v2 - replacing deprecated APIs with their new version and dropping the -v2 marker
+// from APIs introducing the breaking change of handling non-unique-names. This is used in v1 -> v2 conversion
+func GetV2Name(forV1Api API) string {
```
Laubi, line 343: Should be called GetV2Id then, right?

pkg/api/apis.go
```diff
@@ -0,0 +1,102 @@
...
+func NewAPIs() APIs {
+	return newAPIs(configEndpoints)
+}
+
+func NewV1APIs() APIs {
+	return newAPIs(configEndpointsV1)
+}
+
+func newAPIs(t []API) APIs {
+	apis := make(APIs)
```
Laubi, line 34: You can initialize this already with len(t)

pkg/api/apis.go
```diff
@@ -0,0 +1,102 @@
...
+
+// Contains return true iff requested API is part APIs set
+func (apis APIs) Contains(api string) bool {
+	_, ok := apis[api]
+	return ok
+}
+
+// Filter apply all passed filters and return new filtered array
+func (apis APIs) Filter(filters ...Filter) APIs {
+	ret := make(APIs)
```
Laubi, line 49: Argumentable: Could also pre-initialize. One larger initialization at the start might be better than smaller ones later

pkg/api/apis.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+	}
+	return ret
+}
+
+// Filter return true iff specific api needs to be filtered/ removed from list
+type Filter func(api API) bool
+
+// NoFilter is dummy filter that do nothing.
+func NoFilter(API) bool {
```
Laubi, line 69: Can be private

pkg/api/apis.go
```diff
@@ -0,0 +1,102 @@
...
+	return false
+}
+
+// RetainByName leave ony given APIs. If api is not provided, nothing is removed.
```
Laubi, line 73: Please update the docs. This is just wrong

Laubi (changes requested):

Laubi:
Please also add documentation to all public functions/types that you added. For this PR I would argue to keep it to all changes in the internal/api package as it's the main focus

pkg/project/v1/project_loader_test.go
jskelin: Yes indeed we did. And I am doing it whenever is possible...
Here is the case that one uses some wierd stuff things and is not so clear what is happen here:
```
	// NOT test-resources/hidden-directories/.logs
	// NOT test-resources/hidden-directories/project2/.logs
	assert2.DeepEqual(t, projects, []string{
		"test-resources/hidden-directories/project1",
		"test-resources/hidden-directories/project2/subproject",
	}, cmpopts.SortSlices(func(a, b string) bool { return strings.Compare(a, b) < 0 }))
```

So I changed everything else, and this one left for future....

pkg/api/apis.go
jskelin: does thant optimization really makes any benefit here?
Downside is that it makes code less readable

pkg/api/apis.go
jskelin: optimization agan....
This is not performance critical. and when will become an issue, than we are going to find better solution

Laubi (approved):

PR #883, opened 2023-03-13
feat: check cluster generation download/deploy 

warber:
WIth this PR we do a basic call to the version endpoint of the configured environment to verify that credentials (oauth or basic token) are correct. Based on the manifest environment type field it is distinguished whether to call the platform gen version endpoint or the classic gen version endpoint. If a call fails for a single environment the error is propagated back to main and the program stops.

### DIFF
--- a/cmd/monaco/cmdutils/cmdutils.go
+++ b/cmd/monaco/cmdutils/cmdutils.go
@@ -18,7 +18,6 @@ package cmdutils
 
 import (
 	"fmt"
-	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/client"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/manifest"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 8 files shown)


warber:
TODO: the path to the 3rd gen version endpoint would need to be ajusted to `/metadata/..` once this is available.

cmd/monaco/cmdutils/cmdutils.go
```diff
@@ -26,3 +32,35 @@ func SilenceUsageCommand() func(cmd *cobra.Command, args []string) {
...
+		}
+
+		// Assume 3rd gen cluster an check version endpoint
+		if env.Type == manifest.Platform {
+			oauthCredentials := client.OauthCredentials{
+				ClientID:     env.Auth.OAuth.ClientId.Value,
+				ClientSecret: env.Auth.OAuth.ClientSecret.Value,
+				TokenURL:     "https://sso-dev.dynatracelabs.com/sso/oauth2/token",
+			}
+			if _, err := client.GetDynatraceVersion2ndGen(client.NewOAuthClient(oauthCredentials), env.Url.Value); err != nil {
```
UnseenWizzard, line 59: I think you meant to call the 3rd gen method here?
```suggestion
			if _, err := client.GetDynatraceVersion3rdGen(client.NewOAuthClient(oauthCredentials), env.Url.Value); err != nil {
```

cmd/monaco/cmdutils/cmdutils.go
```diff
@@ -26,3 +32,35 @@ func SilenceUsageCommand() func(cmd *cobra.Command, args []string) {
...
+// in order to verify that the user has configured the environments correctly.
+// Depending on the configured environment "type" the function tries to call the version endpoint of either
+// 2nd gen cluster (classic) or 3rd gen cluster (platform). The function will return an error as soon as
+// it receives an error from calling the version endpoint of an environment
+func VerifyClusterGen(envs manifest.Environments) error {
+	for _, env := range envs {
+		// Assume 2nd gen cluster and check version endpoint
+		if env.Type == manifest.Classic {
+			if _, err := client.GetDynatraceVersion2ndGen(client.NewTokenAuthClient(env.Auth.Token.Value), env.Url.Value); err != nil {
+				log.Error("Could not verify Dynatrace cluster generation of environment %q (%q). Please check the configured Auth credentials in the manifest", env.Name, env.Url)
```
UnseenWizzard, line 46: I think it might be helpful to include a link to the documentation section about environments here.

cmd/monaco/download/download_command.go
```diff
@@ -111,7 +111,14 @@ func getDownloadConfigsCommand(fs afero.Fs, command Command, downloadCmd *cobra.
 		},
 		ValidArgsFunction: completion.DownloadDirectCompletion,
 		PreRun: func(cmd *cobra.Command, args []string) {
-			printUploadToSameEnvironmentWarning(args[0], os.Getenv(args[1]))
+			serverVersion, err := client.GetDynatraceVersion2ndGen(client.NewTokenAuthClient(os.Getenv(args[1])), args[0])
```
UnseenWizzard, line 114: I just realized that we'll need to re-think direct download with OAuth :grimacing:

cmd/monaco/cmdutils/cmdutils.go
warber: yes I just wanted to test your review skills :D 
I will also forgot to unit test this func

cmd/monaco/download/download_command.go
```diff
@@ -265,15 +272,30 @@ func setupSharedFlags(cmd *cobra.Command, project, outputFolder *string, forceOv
...
+	var err error
+	if env.Type == manifest.Classic {
+		serverVersion, err = client.GetDynatraceVersion2ndGen(client.NewTokenAuthClient(env.Auth.Token.Value), env.Url.Value)
+	} else {
+		credentials := client.OauthCredentials{
+			ClientID:     env.Auth.OAuth.ClientId.Value,
+			ClientSecret: env.Auth.OAuth.ClientSecret.Value,
+			TokenURL:     "https://sso-dev.dynatracelabs.com/sso/oauth2/token",
+		}
+		serverVersion, err = client.GetDynatraceVersion2ndGen(client.NewOAuthClient(credentials), env.Url.Value)
```
UnseenWizzard, line 286: same as the comment in deploy, am I missing something or should this be calling the 3rdGen function?

cmd/monaco/download/download_configs.go
```diff
@@ -69,7 +70,12 @@ func (d DefaultCommand) DownloadConfigsBasedOnManifest(fs afero.Fs, cmdOptions m
 		return fmt.Errorf("environment %q was not available in manifest %q", cmdOptions.specificEnvironmentName, cmdOptions.manifestFile)
 	}
 
-	printUploadToSameEnvironmentWarning(env.Url.Value, env.Auth.Token.Value)
+	err := cmdutils.VerifyClusterGen(manifest.Environments{env.Name: env})
```
UnseenWizzard, line 73: Why does this happen both in the `_command` and here? Could we move it to one place? 
I assume the call here is because we haven't loaded the manifest yet before - so maybe the version check for `direct` can also move in here?

UnseenWizzard (changes requested):
A few comments that I think should be addressed. 

Also as you mentioned the Platform version URL is expected to change to `platform/metadata/v1/environment-api-info/ ` - this should be verified and changed before merging

cmd/monaco/download/download_configs.go
warber: your comment doesn't make sense to me because _[...]move in here[...]_ refers to a place inside `DownloadConfigsBasedOnManifest`
atm this check is only performed when dealing with manifests.

cmd/monaco/download/download_command.go
warber: yep

pkg/client/version_check.go
```diff
@@ -29,10 +29,19 @@ type ApiVersionObject struct {
...
-	versionUrl := environmentUrl + versionPath
+func GetDynatraceVersion3rdGen(client *http.Client, environmentURL string) (version.Version, error) {
+	return GetDynatraceVersion(client, environmentURL, versionPath3rdGen)
+}
+
+func GetDynatraceVersion2ndGen(client *http.Client, environmentURL string) (version.Version, error) {
+	return GetDynatraceVersion(client, environmentURL, versionPath2ndGen)
+}
+
+func GetDynatraceVersion(client *http.Client, environmentUrl string, apiPath string) (version.Version, error) {
```
jskelin, line 43: I would put this function as private

And I would add a unit test

cmd/monaco/download/download_command.go
warber: *facepalm* yeah copy paste err

cmd/monaco/download/download_configs.go
UnseenWizzard: With "in here" i meant from the `download_command` to the `download_configs` file (which IIRC also contains the functions for direct download?)
Currently the (simpler) version check happens as part of cobra validation in one case, and in or logic in the _cofings.go file in the other - which is strange to me (but no huge issue)

Fine for now, but let's revisit when we've decided how we should deal with oAuth for direct downloads.

UnseenWizzard (approved):

PR #892, opened 2023-03-15
Feat: Enable explicit specification of SSO OAuth token_endpoint in manifest.yaml

jskelin:
Some customers have their own SSO OAuth systems connected with Dynatrace. Or like Dynatrace, have more than one OAuth authorization system. For that purpose, it must be possible to specify OAuth `tokenEndpoint` address. 

With this PR new section `tokenEndpoint` inside `oAuth` will be defined (see example). If `tokenEndpoint` is not provided, the default one `https://sso.dynatrace.com/sso/oauth2/token` will be used.

``` yaml
environmentGroups:
  - environments:
    - auth:
        oAuth:
          clientId:
            name: ENV_CLIENT_ID
          clientSecret:
            name: ENV_CLIENT_SECRET
          tokenEndpoint:
            value: "https://sso.dynatrace.com:443/sso/oauth2/token"
```

### DIFF
--- a/pkg/manifest/manifest_loader.go
+++ b/pkg/manifest/manifest_loader.go
@@ -59,13 +59,6 @@ type manifestLoaderError struct {
 	Reason       string
 }
 
-func newManifestLoaderError(manifest string, reason string) manifestLoaderError {
-	return manifestLoaderError{
-		ManifestPath: manifest,
-		Reason:       reason,
-	}
-}
-
 func (e manifestLoaderError) Error() string {
 	return fmt.Sprintf("%s: %s", e.ManifestPath, e.Reason)
 }
@@ -78,7 +71,7 @@ type environmentLoaderError struct {
 
 func newManifestEnvironmentLoaderError(manifest string, group string, env string, reason string) environmentLoaderError {
 	return environmentLoaderError{
-		manifestLoaderError: newManifestLoaderError(manifest, reason),
+		manifestLoaderError: manifestLoaderError{manifest, reason},
 		Group:               group,
 		Environment:         env,
 	}
@@ -95,7 +88,7 @@ type projectLoaderError struct {
 
 func newManifestProjectLoaderError(manifest string, project string, reason string) projectLoaderError {
 	return projectLoaderError{
-		manifestLoaderError: newManifestLoaderError(manifest, reason),
+		manifestLoaderError: manifestLoaderError{manifest, reason},
 		Project:             project,
 	}
 }
@@ -110,7 +103,7 @@ func LoadManifest(context *ManifestLoaderContext) (Manifest, []error) {
 	log.Debug("Loading manifest %q. Restrictions: groups=%q, environments=%q", context.ManifestPath, context.Groups, context.Environments)
 
 	if !files.IsYamlFileExtension(manifestPath) {
-		return Manifest{}, []error{newManifestLoaderError(context.ManifestPath, "manifest file is not a yaml")}
+		return Manifest{}, []error{manifestLoaderError{context.ManifestPath, "manifest file is not a yaml"}}
 	}
 
 	exists, err := files.DoesFileExist(context.Fs, manifestPath)
@@ -120,13 +113,13 @@ func LoadManifest(context *ManifestLoaderContext) (Manifest, []error) {
 	}
 
 	if !exists {
-		return Manifest{}, []error{newManifestLoaderError(context.ManifestPath, "specified manifest file is either no file or does not exist")}
+		return Manifest{}, []error{manifestLoaderError{context.ManifestPath, "specified manifest file is either no file or does not exist"}}
 	}
 
 	data, err := afero.ReadFile(context.Fs, manifestPath)
 
 	if err != nil {
-		return Manifest{}, []error{newManifestLoaderError(context.ManifestPath, fmt.Sprintf("error while reading the manifest: %s", err))}
+		return Manifest{}, []error{manifestLoaderError{context.ManifestPath, fmt.Sprintf("error while reading the manifest: %s", err)}}
 	}
 
 	return parseManifest(context, data)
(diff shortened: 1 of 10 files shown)


pkg/manifest/manifest_persitence.go
```diff
@@ -39,8 +39,9 @@ type authSecret struct {
 }
 
 type oAuth struct {
-	ClientID     authSecret `yaml:"clientId"`
-	ClientSecret authSecret `yaml:"clientSecret"`
+	ClientID      authSecret `yaml:"clientId"`
+	ClientSecret  authSecret `yaml:"clientSecret"`
+	TokenEndpoint url        `yaml:"token_endpoint,omitempty"`
```
UnseenWizzard, line 44: please make that `tokenEndpoint` , so we consistently use the same case in the YAML

jskelin: done

pkg/manifest/manifest_loader_test.go
```diff
@@ -22,15 +22,18 @@ import (
...
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/spf13/afero"
+	"gopkg.in/yaml.v2"
 	"math"
 	"reflect"
 	"testing"
 
-	"gotest.tools/assert"
+	"github.com/stretchr/testify/assert"
+	assert2 "gotest.tools/assert"
```
jskelin, line 36: Two complex cases left to migration on `github.com/stretchr/testify/assert`

``` go
assert2.DeepEqual(t, test.expectedManifest, mani, cmp.AllowUnexported(EnvironmentDefinition{}))
```

and 
``` go
assert2.DeepEqual(t, got, tt.want, cmpopts.SortSlices(func(a, b ProjectDefinition) bool { return a.Name < b.Name }))
```

Laubi:
You have overlooked the code that we also need to write it in manifest_writer.go

pkg/manifest/manifest_loader_test.go
Laubi, line 1: In this file, the test "TestLoadManifest" is the most important test, as it currently covers the actual loading with most code covered quite nicely.

Please add tests to that method that test if the tokenEndpoint.type is env, value, or invalid. Also a case where type=environment and the env-variable name can not be looked up or is empty

pkg/manifest/manifest_loader_test.go
Laubi: Since we agreed that we do either all or nothing in a file, I checked how we can do this, and it's really trivial to replace the code.
Simply use assert.Equal: `assert.Equal(t, got, tt.want)` and `assert.Equal(t, test.expectedManifest, mani)`

Please fix the code only to use testify

pkg/manifest/manifest_loader.go
```diff
@@ -248,39 +221,65 @@ func parseOAuth(a oAuth) (OAuth, error) {
...
+		ClientId:      clientID,
+		ClientSecret:  clientSecret,
+		TokenEndpoint: urlDef,
 	}, nil
 }
 
-func parseManifestFile(context *ManifestLoaderContext, data []byte) (manifest, []error) {
-	var m manifest
-	var errs []error
+func getManifestYAML(context *ManifestLoaderContext) (manifest, error) {
```
Laubi, line 240: The idea was to report all issues we can find at once to a user. This change now reports only one error, the user fixes it, and then the next error and so on. Why did you change it?

Also the name does not fit - I think parseManifest was quite good.

jskelin: `getManifestYAML` just checks does file exists and is it a readable and if can, call unmarshall function

Logic from `parseManifestFile` is part of `verifyManifestYAML` which returns `[]error`. 

`getManifestYAML` is renamed to `readManifestYAML`

jskelin:
> You have overlooked the code that we also need to write it in manifest_writer.go

Added writer and its tests

internal/oauth2/endpoints/endpoints.go
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+package endpoints
+
+import (
+	"golang.org/x/oauth2"
+)
+
+var Dynatrace = oauth2.Endpoint{
```
UnseenWizzard, line 23: nitpick: I don't think there is any need for this to be an `oauth2.Endpoint`? 
If I'm not mistaken we only ever pull the TokenURL string out when loading and use it as string everywhere - so I think we could simply remove the unnecessary dependency on oauth2 and define a string in here.

pkg/manifest/manifest_loader.go
```diff
@@ -248,39 +221,65 @@ func parseOAuth(a oAuth) (OAuth, error) {
 		return OAuth{}, fmt.Errorf("failed to parse ClientSecret: %w", err)
 	}
 
+	var urlDef UrlDefinition
+	if a.TokenEndpoint == nil {
+		urlDef = UrlDefinition{
+			Value: endpoints.Dynatrace.TokenURL,
```
UnseenWizzard, line 227: I think here you need to set `Type: Default` for the logic in the writer to work

internal/oauth2/endpoints/endpoints.go
UnseenWizzard: fine to keep, as it's standard practice in other libs/packages :+1:

pkg/manifest/manifest_loader.go
jskelin: added

UnseenWizzard (approved):
Looks good from my point of view. 

Please still run e2e tests as well. 

IMO we could merge this and change main if decide on Monday to move responsibility for the default URL to the `client` rather than `manifest` loading. 
Unless you think it's easier to change while still in the PR/never introducing the change to the manifest.

PR #895, opened 2023-03-15
Fix: Enhance cluster generation checks

warber:
The current solution implemented with https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/883 has a very bad user experience. Especially the error logs are bad and confuse the user.

This PR tries to improve on that and introduces
* enhanced logging
* GetDynatraceClassicURL to determine the classic gen URL
* RespErr returned from GetVersion and GetDynatraceClassicURL
* explicit handling of 403 in cluster gen checks
* better log messages
* deduplication of error logs

Additionally, it turned out that the version endpoint of the platform will be removed. for the sake of still verifying platform tenants, we make a call to: `/platform/core/v1/environment-api-info`

:exclamation: this is only currently available and will need to be changed to `/platform/metadata/v1/classic-environment-domain` once available

Further getting a 403 Return code from the endpoints is also taken into consideration which basically means that something is wrong with the token, and does not necessarily mean that you are targeting the wrong environment type.

Example output in case the user configured a classic environment but gave an URL to a platform tenant:
image

### DIFF
--- a/cmd/monaco/cmdutils/cmdutils.go
+++ b/cmd/monaco/cmdutils/cmdutils.go
@@ -17,10 +17,12 @@
 package cmdutils
 
 import (
-	"fmt"
+	"errors"
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/client"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/manifest"
 	"github.com/spf13/cobra"
+	"net/http"
 )
 
 // SilenceUsageCommand gives back a command that is just configured to skip printing of usage info.
@@ -32,30 +34,60 @@ func SilenceUsageCommand() func(cmd *cobra.Command, args []string) {
 	}
 }
 
-// VerifyClusterGen takes a manifestEnvironments map and tries to call the version endpoint of each environment
-// in order to verify that the user has configured the environments correctly.
-// Depending on the configured environment "type" the function tries to call the version endpoint of either
-// classic gen or platform gen. The function will return an error as soon as
-// it receives an error from calling the version endpoint of an environment
-func VerifyClusterGen(envs manifest.Environments) error {
+// VerifyEnvironmentGeneration takes a manifestEnvironments map and tries to verify that each environment can be reached
+// using the configured credentials
+func VerifyEnvironmentGeneration(envs manifest.Environments) bool {
 	for _, env := range envs {
 		switch env.Type {
 		case manifest.Classic:
-			if _, err := client.GetDynatraceVersion(client.NewTokenAuthClient(env.Auth.Token.Value), client.Environment{URL: env.Url.Value, Type: client.Classic}); err != nil {
-				return fmt.Errorf("could not verify Dynatrace cluster generation of environment %q (%q). Please check the configured Auth credentials in the manifest", env.Name, env.Url)
-			}
+			return isClassicEnvironment(env)
 		case manifest.Platform:
-			oauthCredentials := client.OauthCredentials{
-				ClientID:     env.Auth.OAuth.ClientId.Value,
-				ClientSecret: env.Auth.OAuth.ClientSecret.Value,
-				TokenURL:     env.Auth.OAuth.TokenEndpoint.Value,
-			}
-			if _, err := client.GetDynatraceVersion(client.NewOAuthClient(oauthCredentials), client.Environment{URL: env.Url.Value, Type: client.Platform}); err != nil {
-				return fmt.Errorf("could not verify Dynatrace cluster generation of environment %q (%q). Please check the configured Auth credentials in the manifest", env.Name, env.Url)
-			}
+			return isPlatformEnvironment(env)
 		default:
-			return fmt.Errorf("invalid environment type")
+			log.Error("Could not authorize against the environment with name %q (%s). Unknown environment type.", env.Name, env.Url.Value)
+			return false
 		}
 	}
-	return nil
+	return true
+}
+
+func isClassicEnvironment(env manifest.EnvironmentDefinition) bool {
+	if _, err := client.GetDynatraceVersion(client.NewTokenAuthClient(env.Auth.Token.Value), env.Url.Value); err != nil {
+		var respErr client.RespError
+		if errors.As(err, &respErr) {
+			log.Error("Could not authorize against the environment with name %q (%s) using token authorization.", env.Name, env.Url.Value)
+			if respErr.StatusCode != http.StatusForbidden {
+				log.Error("Please verify that this environment is a Dynatrace Classic environment.")
+			} else {
+				log.Error(err.Error())
+			}
+		} else {
+			log.Error("Could not connect to environment %q (%s)", env.Name, env.Url.Value)
+		}
+		return false
+	}
+	return true
+}
+
+func isPlatformEnvironment(env manifest.EnvironmentDefinition) bool {
+	oauthCredentials := client.OauthCredentials{
+		ClientID:     env.Auth.OAuth.ClientId.Value,
+		ClientSecret: env.Auth.OAuth.ClientSecret.Value,
+		TokenURL:     env.Auth.OAuth.TokenEndpoint.Value,
+	}
+	if _, err := client.GetDynatraceClassicURL(client.NewOAuthClient(oauthCredentials), env.Url.Value); err != nil {
+		var respErr client.RespError
+		if errors.As(err, &respErr) {
+			log.Error("Could not authorize against the environment with name %q (%s) using oAuth authorization.", env.Name, env.Url.Value)
+			if respErr.StatusCode != http.StatusForbidden {
+				log.Error("Please verify that this environment is a Dynatrace Platform environment.")
+			} else {
+				log.Error(err.Error())
+			}
+		} else {
+			log.Error("Could not connect to environment %q (%s)", env.Name, env.Url.Value)
+		}
+		return false
+	}
+	return true
 }
--- a/cmd/monaco/cmdutils/cmdutils_test.go
+++ b/cmd/monaco/cmdutils/cmdutils_test.go
@@ -75,8 +75,8 @@ func TestVerifyClusterGen(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if err := VerifyClusterGen(tt.args.envs); (err != nil) != tt.wantErr {
-				t.Errorf("VerifyClusterGen() error = %v, wantErr %v", err, tt.wantErr)
+			if ok := VerifyEnvironmentGeneration(tt.args.envs); ok == tt.wantErr {
+				t.Errorf("VerifyEnvironmentGeneration() error = %v, wantErr %v", ok, tt.wantErr)
 			}
 		})
 	}
@@ -88,7 +88,7 @@ func TestVerifyClusterGen(t *testing.T) {
 		}))
 		defer server.Close()
 
-		err := VerifyClusterGen(manifest.Environments{
+		ok := VerifyEnvironmentGeneration(manifest.Environments{
 			"env": manifest.EnvironmentDefinition{
 				Name: "env",
 				Type: manifest.Classic,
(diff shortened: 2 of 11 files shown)


cmd/monaco/cmdutils/cmdutils.go
```diff
@@ -53,11 +55,14 @@ func VerifyClusterGen(envs manifest.Environments) error {
 				TokenURL:     ssoTokenURL,
 			}
 			if _, err := client.GetDynatraceVersion(client.NewOAuthClient(oauthCredentials), client.Environment{URL: env.Url.Value, Type: client.Platform}); err != nil {
-				return fmt.Errorf("could not verify Dynatrace cluster generation of environment %q (%q). Please check the configured Auth credentials in the manifest", env.Name, env.Url)
+				log.Error("Could not authorize against the environment with name %q (%s) using token authorization. "+
```
UnseenWizzard, line 58:
```suggestion
				log.Error("Could not authorize against the environment with name %q (%s) using oAuth authorization. "+
```

cmd/monaco/cmdutils/cmdutils.go
```diff
@@ -53,11 +55,14 @@ func VerifyClusterGen(envs manifest.Environments) error {
...
+					"Please verify that this environment is a classic generation cluster.", env.Name, env.Url.Value)
```
UnseenWizzard, line 59:
```suggestion
					"Please verify that this environment is a Dynatrace Platform environment.", env.Name, env.Url.Value)
```

cmd/monaco/cmdutils/cmdutils.go
```diff
@@ -39,12 +39,14 @@ var ssoTokenURL = "https://sso.dynatrace.com/sso/oauth2/token" //nolint:gosec
...
 // it receives an error from calling the version endpoint of an environment
-func VerifyClusterGen(envs manifest.Environments) error {
+func VerifyClusterGen(envs manifest.Environments) bool {
 	for _, env := range envs {
 		switch env.Type {
 		case manifest.Classic:
 			if _, err := client.GetDynatraceVersion(client.NewTokenAuthClient(env.Auth.Token.Value), client.Environment{URL: env.Url.Value, Type: client.Classic}); err != nil {
-				return fmt.Errorf("could not verify Dynatrace cluster generation of environment %q (%q). Please check the configured Auth credentials in the manifest", env.Name, env.Url)
+				log.Error("Could not authorize against the environment with name %q (%s) using token authorization. "+
+					"Please verify that this environment is a classic generation cluster.", env.Name, env.Url.Value)
```
UnseenWizzard, line 48:
```suggestion
					"Please verify that this environment is a classic Dynatrace environment.", env.Name, env.Url.Value)
```

We're not interacting with the cluster, but an environment/tenant - saying environment twice sounds a bit strange, but I couldn't come up with any better suggestion

cmd/monaco/cmdutils/cmdutils_test.go
```diff
@@ -75,7 +75,7 @@ func TestVerifyClusterGen(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if err := VerifyClusterGen(tt.args.envs); (err != nil) != tt.wantErr {
+			if err := VerifyClusterGen(tt.args.envs); err == tt.wantErr {
```
UnseenWizzard, line 78: nit: it's not really an Error anymore. Maybe name this ok like you did below and changing the test cases to always define the wanted true/false result?

cmd/monaco/download/download_configs.go
```diff
@@ -70,9 +70,9 @@ func (d DefaultCommand) DownloadConfigsBasedOnManifest(fs afero.Fs, cmdOptions m
 		return fmt.Errorf("environment %q was not available in manifest %q", cmdOptions.specificEnvironmentName, cmdOptions.manifestFile)
 	}
 
-	err := cmdutils.VerifyClusterGen(manifest.Environments{env.Name: env})
-	if err != nil {
-		return err
+	ok := cmdutils.VerifyClusterGen(manifest.Environments{env.Name: env})
+	if !ok {
+		return fmt.Errorf("unable to verify Dynatrace cluster gneration")
```
UnseenWizzard, line 75:
```suggestion
		return fmt.Errorf("unable to verify Dynatrace environment generation")
```

cmd/monaco/runner/runner.go
```diff
@@ -37,7 +37,6 @@ func Run() int {
 	rootCmd := BuildCli(afero.NewOsFs())
 
 	if err := rootCmd.Execute(); err != nil {
```
UnseenWizzard, line 39: how does that impact the logs? IIRC logging how execution failed as the very last line of log was added intentionally

UnseenWizzard (changes requested):

cmd/monaco/cmdutils/cmdutils.go
```diff
@@ -34,30 +36,56 @@ func SilenceUsageCommand() func(cmd *cobra.Command, args []string) {
...
+		if errors.As(err, &respErr) {
+			log.Error("Could not authorize against the environment with name %q (%s) using token authorization.", env.Name, env.Url.Value)
+			if respErr.StatusCode != http.StatusForbidden {
+				log.Error("Please verify that this environment is a Dynatrace Classic environment.")
+			} else {
+				log.Error(err.Error())
 			}
-			if _, err := client.GetDynatraceVersion(client.NewOAuthClient(oauthCredentials), client.Environment{URL: env.Url.Value, Type: client.Platform}); err != nil {
-				return fmt.Errorf("could not verify Dynatrace cluster generation of environment %q (%q). Please check the configured Auth credentials in the manifest", env.Name, env.Url)
+		}
```
UnseenWizzard, line 66: We should probably also log something if the error is not a Response Error? 
E.g. I think that could happen when the URL is just completely wrong/there's no server replying?

Maybe a more general "Could not connect to environment" and printing the error.
Currently we swallow the actual error if it's not a RespErr

cmd/monaco/deploy/deploy.go
```diff
@@ -136,14 +136,12 @@ func loadManifest(fs afero.Fs, manifestPath string, groups []string, environment
 	return &m, nil
 }
 
-func verifyClusterGen(environments manifest.Environments, dryRun bool) error {
+func verifyClusterGen(environments manifest.Environments, dryRun bool) bool {
```
UnseenWizzard, line 139: nit 
```suggestion
func verifyEnvironmentGen(environments manifest.Environments, dryRun bool) bool {
```

cmd/monaco/download/download_configs.go
```diff
@@ -70,9 +70,9 @@ func (d DefaultCommand) DownloadConfigsBasedOnManifest(fs afero.Fs, cmdOptions m
 		return fmt.Errorf("environment %q was not available in manifest %q", cmdOptions.specificEnvironmentName, cmdOptions.manifestFile)
 	}
 
-	err := cmdutils.VerifyClusterGen(manifest.Environments{env.Name: env})
-	if err != nil {
-		return err
+	ok := cmdutils.VerifyEnvironmentGeneration(manifest.Environments{env.Name: env})
+	if !ok {
+		return fmt.Errorf("unable to verify Dynatrace cluster gneration")
```
UnseenWizzard, line 75:
```suggestion
		return fmt.Errorf("unable to verify Dynatrace environment generation")
```

pkg/client/metadata.go
```diff
@@ -0,0 +1,57 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return "", RespError{
+			Err:        fmt.Errorf("failed to query classic environment URL: (HTTP %v) %v", resp.StatusCode, string(resp.Body)),
+			StatusCode: resp.StatusCode,
+		}
+	}
+
+	var jsonResp classicEnvURL
+	if err := json.Unmarshal(resp.Body, &jsonResp); err != nil {
+		return "", RespError{
```
UnseenWizzard, line 51: I'd return that as a plain error, status code will be 2xx at this point

pkg/client/metadata.go
```diff
@@ -0,0 +1,57 @@
...
+// environment
+func GetDynatraceClassicURL(client *http.Client, environmentURL string) (string, error) {
+	endpointURL := environmentURL + classicEnvironmentDomainPath
+
+	resp, err := rest.Get(client, endpointURL)
+	if err != nil {
+		return "", fmt.Errorf("failed to query classic environment url %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
```
UnseenWizzard, line 42:
```suggestion
	if !resp.IsSuccess() {
```

would cover all 2xx codes

pkg/client/version_check.go
```diff
@@ -29,61 +29,45 @@ type ApiVersionObject struct {
...
-		return version.Version{}, fmt.Errorf("usupported environment type")
-	}
+func GetDynatraceVersion(client *http.Client, environmentURL string) (version.Version, error) {
+	versionURL := environmentURL + versionPathClassic
+
 	resp, err := rest.Get(client, versionURL)
 	if err != nil {
 		return version.Version{}, fmt.Errorf("failed to query version of Dynatrace environment: %w", err)
 	}
 	if resp.StatusCode != http.StatusOK {
```
UnseenWizzard, line 42:
```suggestion
	if !resp.IsSuccess() {
```

just saw it here as well

cmd/monaco/cmdutils/cmdutils.go
warber: an error in this case is already ultimately printed to the user ... looks smth like that:
```
2023/03/17 08:53:02 ERROR HTTP Request failed with Error: Get "https://WRONG.dev.apps.dydnatracelabs.com/api/v1/config/clusterversion": dial tcp: lookup WRONG.dev.apps.dydnatracelabs.com on 1.2.3.4 no such host
Error: unable to verify Dynatrace environment generation
```
I'll try to enhance it a bit

pkg/client/client.go
```diff
@@ -237,17 +237,12 @@ func WithAutoServerVersion() func(client *DynatraceClient) {
 		var serverVersion version.Version
 		var err error
```
jskelin, line 238: it can be more elegant

``` go
		var err error
		if _, ok := d.client.Transport.(*oauth2.Transport); ok {
			// for platform enabled tenants there is no dedicated version endpoint
			// so this call would need to be "redirected" to the second gen URL, which do not currently resolve
			d.serverVersion = version.UnknownVersion
		} else if serverVersion, err = GetDynatraceVersion(d.client, d.environmentUrl); err != nil {
			log.Error("Unable to determine Dynatrace server version: %v", err)
			d.serverVersion = version.UnknownVersion
		} else {
			d.serverVersion = serverVersion
		}
```

UnseenWizzard (approved):

cmd/monaco/cmdutils/cmdutils.go
```diff
@@ -32,30 +34,60 @@ func SilenceUsageCommand() func(cmd *cobra.Command, args []string) {
...
 
-// VerifyClusterGen takes a manifestEnvironments map and tries to call the version endpoint of each environment
-// in order to verify that the user has configured the environments correctly.
-// Depending on the configured environment "type" the function tries to call the version endpoint of either
-// classic gen or platform gen. The function will return an error as soon as
-// it receives an error from calling the version endpoint of an environment
-func VerifyClusterGen(envs manifest.Environments) error {
+// VerifyEnvironmentGeneration takes a manifestEnvironments map and tries to verify that each environment can be reached
+// using the configured credentials
+func VerifyEnvironmentGeneration(envs manifest.Environments) bool {
```
jskelin, line 39: more conceptual question/thinking

Why print error message inside. I would go with returning error, than caller can decide what to do with error (e.g. which error lever to use for printing...)

jskelin (approved):

cmd/monaco/cmdutils/cmdutils.go
warber: > more conceptual question/thinking
> 
> Why print error message inside. I would go with returning error, than caller can decide what to do with error (e.g. which error lever to use for printing...)

in general i agree with that, you need to do a bit of refactoring, and give the caller more information what went wrong, so that the caller is able to log stuff, tho

PR #898, opened 2023-03-16
⚠️ Remove [config].token and [config].type (breaking change)

Laubi:
This PR introduces two breaking changes:
1. Removal of `[config].token`. The alternative `[config].auth.token` has to be used.
2. Romoval of `[config].type`. We can resolve the value by looking at the rest of the environment-configuration, so it has no replacement.

Example configuration:
```yaml
manifestVersion: 1.0

projects: 
- name: project

environmentGroups:
- name: environment-group
  environments:
  - name: environment-name
    url:
      value: "https://example.com"
    auth:
      token:
        name: TOKEN_ENV_VAR
```

### DIFF
--- a/pkg/manifest/manifest_loader.go
+++ b/pkg/manifest/manifest_loader.go
@@ -393,61 +393,34 @@ func shouldSkipEnv(context *ManifestLoaderContext, group group, env environment)
 }
 
 func parseEnvironment(context *ManifestLoaderContext, config environment, group string) (EnvironmentDefinition, []error) {
-	var errors []error
+	var errs []error
 
-	envType, err := parseEnvironmentType(context, config, group)
+	auth, err := parseAuth(config.Auth)
 	if err != nil {
-		errors = append(errors, err)
-	}
-
-	a, err := parseCredentials(config, envType)
-	if err != nil {
-		errors = append(errors, newManifestEnvironmentLoaderError(context.ManifestPath, group, config.Name, err.Error()))
+		errs = append(errs, newManifestEnvironmentLoaderError(context.ManifestPath, group, config.Name, fmt.Sprintf("failed to parse auth section: %s", err)))
 	}
 
 	urlDef, err := parseURLDefinition(config.URL)
 	if err != nil {
-		errors = append(errors, newManifestEnvironmentLoaderError(context.ManifestPath, group, config.Name, err.Error()))
+		errs = append(errs, newManifestEnvironmentLoaderError(context.ManifestPath, group, config.Name, err.Error()))
 	}
 
-	if len(errors) > 0 {
-		return EnvironmentDefinition{}, errors
+	if len(errs) > 0 {
+		return EnvironmentDefinition{}, errs
+	}
+
+	envType := Platform
+	if auth.OAuth == (OAuth{}) {
+		envType = Classic
 	}
 
 	return EnvironmentDefinition{
 		Name:  config.Name,
 		Type:  envType,
 		URL:   urlDef,
-		Auth:  a,
+		Auth:  auth,
 		Group: group,
 	}, nil
-}
-
-func parseCredentials(config environment, envType EnvironmentType) (Auth, error) {
-	if config.Token == nil && config.Auth == nil {
-		return Auth{}, errors.New("'auth' property missing")
-	}
-
-	if config.Token != nil && config.Auth != nil {
-		return Auth{}, errors.New("both 'auth' and 'token' are present")
-	}
-
-	if config.Token != nil {
-		log.Warn("Environment %s: Field 'token' is deprecated, use 'auth.token' instead.", config.Name)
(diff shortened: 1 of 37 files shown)


pkg/manifest/manifest_writer.go
```diff
@@ -23,12 +23,12 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-type ManifestWriterContext struct {
+type WriterContext struct {
```
UnseenWizzard, line 26: Advanced commit nitpick: You don't actually document anything in this commit ;) I guess the msg is copied from the LoaderContext commit

UnseenWizzard (approved):
lgtm

jskelin (dismissed):

pkg/manifest/manifest_loader.go
```diff
@@ -298,7 +298,7 @@ func validateManifestVersion(manifestVersion string) error {
 	return nil
 }
 
-func toEnvironments(context *ManifestLoaderContext, groups []group) (map[string]EnvironmentDefinition, []error) {
+func toEnvironments(context *ManifestLoaderContext, groups []group) (map[string]EnvironmentDefinition, []error) { // nolint:gocognit
```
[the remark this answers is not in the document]
Laubi: As stated in the commit, this would only make the code more complex and less understandable, not solve anything.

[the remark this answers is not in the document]
jskelin: extracting some of complexity in separate function is reducing complexity. 

prototipe of solution:

``` go
func toEnvironments(context *LoaderContext, groups []group) (map[string]EnvironmentDefinition, []error) {
	var errors []error

	for _, e := range groupEnvironmentNameUniquenessCheck(groups) {
		errors = append(errors, manifestLoaderError{context.ManifestPath, e.Error()})
	}

	for _, e := range environmentNameUniquenessCheck(groups) {
		errors = append(errors, manifestLoaderError{context.ManifestPath, e.Error()})
	}

	environments := make(map[string]EnvironmentDefinition)

	for _, group := range groups {
		for _, env := range group.Environments {

			// skip loading if environments is not empty, the environments does not contain the env name, or the group should not be included
			if shouldSkipEnv(context, group, env) {
				log.Debug("skipping loading of environment %q", env.Name)
				continue
			}

			parsedEnv, configErrors := parseEnvironment(context, env, group.Name)

			if configErrors != nil {
				errors = append(errors, configErrors...)
				continue
			}

			environments[parsedEnv.Name] = parsedEnv
		}
	}

	if errors != nil {
		return nil, errors
	}

	return environments, nil
}

func groupEnvironmentNameUniquenessCheck(gr []group) []error {
	var err []error
	knownNames := make(map[string]interface{}, len(gr))
	for i, g := range gr {
		if g.Name == "" {
			err = append(err, fmt.Errorf("missing group name on index `%d`", i))
		} else if _, ok := knownNames[g.Name]; ok {
			err = append(err, fmt.Errorf("duplicated group name %q", g.Name))
		} else {
			knownNames[g.Name] = nil
		}
	}
	return err
}

func environmentNameUniquenessCheck(gr []group) []error {
	var err []error
	knownNames := make(map[string]interface{}, len(gr))
	for _, g := range gr {
		for j, env := range g.Environments {
			if env.Name == "" {
				err = append(err, fmt.Errorf("missing environment name in group %q on index `%d`", g.Name, j))
				continue
			}

			if _, ok := knownNames[env.Name]; ok {
				err = append(err, fmt.Errorf("duplicated environment name %q", env.Name))
				continue
			}

			knownNames[g.Name] = nil
		}
	}

	return err
}
```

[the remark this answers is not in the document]
Laubi: This does not reduce complexity; it introduces a quite higher one. Your code is less understandable by a human than the code above. Yes, the linter no longer complains, but it adds unnecessary indirections which are more difficult.

As stated above, when the linter error occurred, I pair-programmed the part, and together we decided that such a refactoring as you did would only increase complexity, not reduce it.

Will keep the code as is.

pkg/manifest/manifest_writer.go
Laubi: Thanks, the commit was missing code :)

pkg/manifest/manifest_writer.go
```diff
@@ -23,12 +23,16 @@ import (
...
 )
 
-type ManifestWriterContext struct {
-	Fs           afero.Fs
+// WriterContext holds all information for [WriteManifest]
+type WriterContext struct {
+	// Fs holds the abstraction of the file system.
+	Fs afero.Fs
+
+	// ManifestPath holds the path from where the manifest should be loaded.
```
UnseenWizzard, line 31: is that really the path a manifest was loaded from, for **Writer** or is it where the manifest will be written to?

Laubi: good catch :d

PR #899, opened 2023-03-16
ci: Rename docker container image to dynatrace-configuration-as-code

UnseenWizzard:
Change the image name from 'monitoring' to 'configuration'.
Also, drop the extra 'monaco' that was added after the registry url
(previous image name was actually `monaco/dynatrace-monitoring-as-code`)

Additionally, the container image push function in the release pipeline is modified, to prepare it for being called for different registries than the current internal one.

### DIFF
--- a/.ci/releasePipeline.groovy
+++ b/.ci/releasePipeline.groovy
@@ -245,8 +245,8 @@ void createContainerAndPushToStorage(Map args = [version: null]) {
                 script {
                     try {
                         sh 'docker login --username $username --password $password $registry'
-                        sh 'DOCKER_BUILDKIT=1 make docker-container OUTPUT=./build/docker/monaco CONTAINER_NAME=$registry/monaco/dynatrace-monitoring-as-code VERSION=$version'
-                        sh 'docker push $registry/monaco/dynatrace-monitoring-as-code:$version'
+                        sh 'DOCKER_BUILDKIT=1 make docker-container OUTPUT=./build/docker/monaco CONTAINER_NAME=$registry/dynatrace-configuration-as-code VERSION=$version'
+                        sh 'docker push $registry/dynatrace-configuration-as-code:$version'
                     } finally {
                         sh 'docker logout $registry'
                     }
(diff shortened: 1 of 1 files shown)


.ci/releasePipeline.groovy
```diff
@@ -235,18 +235,18 @@ void signWithSignService(Map args = [source: null, version: null, destDir: '.',
 
 }
 
-void createContainerAndPushToStorage(Map args = [version: null]) {
+void createContainerAndPushToStorage(Map args = [version: null, registrySecretsPath: null]) {
     stage('📦 Release Container Image') {
         withEnv(["version=${args.version}"]) {
-            withVault(vaultSecrets: [[path        : 'keptn-jenkins/monaco/registry-deploy',
+            withVault(vaultSecrets: [[path        : "${args.registrySecretsPath}",
```
jskelin, line 241: I don't see benefit in this.
On the other hand, whit this change we are splitting information that are deeply tight one with another

UnseenWizzard: The benefit will be re-using the function with a different registry configuration (like Docker hub) by supplying a different secret of the same structure.
As such, I don't see the registry config (path, user, pw) to be coupled to the docker build/push at all - they're just input.

jskelin: That would imply that address of container would be same on every  registry storage... And we don't know that. For me it is premature optimization.
Either way, it is not so important. Approved

jskelin (approved):

.ci/releasePipeline.groovy
UnseenWizzard: > that address of container would be same on every registry storage... And we don't know that.

The way we currently configure it, it is. 
The "registry_path" as currently configured is actually the full repository path.

An actually generalized way would e.g. be to split that properly into defining registry URL and repository name.
And we might have to do that anyway if Docker Hub behaves differently on login than the internal registry.

I will merge this and we change it if needed when we implement push to Docker hub

PR #902, opened 2023-03-17
fix: Log underlying error when Dynatrace generation check fails

UnseenWizzard:
If the DT generation validation failed with an expected (non HTTP code) error, we previously swalled the actual error completely, making this case impossible to debug based on the logs alone. Thus the error is now included in the log.


jskelin (approved):
lgtm

UnseenWizzard:
rebased to incorporate the Id->ID, Url->URL refactor. Please reapprove :)

warber (approved):

PR #904, opened 2023-03-17
Make `const Absent URLType = iota`

jskelin:
As `URLType` is int for which zero value 0, it is more logically that the constant for unprovided information have value 0.

#### Make `const Absent URLType = iota`
This constant is used to carry information about the origin of the data. With this in mind, if a field `URLDefinition.Type` is zero value, it is more likely that the content of `URLDefinition.Value` is generated inside the application (and not provided by the user).

### DIFF
--- a/pkg/manifest/manifest_writer.go
+++ b/pkg/manifest/manifest_writer.go
@@ -139,6 +139,7 @@ func getAuth(env EnvironmentDefinition) auth {
 
 	var te *url
 	switch env.Auth.OAuth.TokenEndpoint.Type {
+	case Absent:
 	case ValueURLType:
 		te = &url{
 			Value: env.Auth.OAuth.TokenEndpoint.Value,
@@ -148,8 +149,6 @@ func getAuth(env EnvironmentDefinition) auth {
 			Type:  urlTypeEnvironment,
 			Value: env.Auth.OAuth.TokenEndpoint.Name,
 		}
-	case Absent:
-		te = nil
 	}
 
 	return auth{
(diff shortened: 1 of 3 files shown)


pkg/manifest/manifest_loader.go
```diff
@@ -225,7 +225,6 @@ func parseOAuth(a oAuth) (OAuth, error) {
 	if a.TokenEndpoint == nil {
 		urlDef = URLDefinition{
 			Value: endpoints.Dynatrace.TokenURL,
-			Type:  Absent,
```
UnseenWizzard, line 228: nit: I would still define it here, to be explicit

pkg/manifest/manifest_writer.go
```diff
@@ -139,6 +139,7 @@ func getAuth(env EnvironmentDefinition) auth {
 
 	var te *url
 	switch env.Auth.OAuth.TokenEndpoint.Type {
+	case Absent:
```
UnseenWizzard, line 142: won't that change result in writing out the default as a value after all? 
If the field was absent from, we now treat it like a value and write whatever URL is in it, rather than not writing it

UnseenWizzard (changes requested):

pkg/manifest/manifest_writer.go
Laubi: Not in go, you'd have to use the `fallthrough`. 
Not sure why we need this change anyway

Laubi (changes requested):
I would argue against this change.
We discussed making this field optional by making the field `TokenURL` optional. 

If we want to test the client, I would rather make the field explicit by always setting it and fail if it's not set.

pkg/manifest/manifest_loader.go
jskelin: done

PR #905, opened 2023-03-20
refactor: Move loading of default TokenURL to client

UnseenWizzard:
We decided that the responsibility of loading the default Dynatrace SSO token URL should be encapsulated in the Dynatrace Client, rather than manifest loading.
This is achieved by having the manifest treat the token URL as a pointer, and loading the default URL when nothing was given to the client on instantiation.

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -300,6 +300,8 @@ func (t *TokenAuthTransport) setHeader(key, value string) {
 	t.header.Set(key, value)
 }
 
+var defaultOAuthTokenURL = endpoints.Dynatrace.TokenURL
+
 // NewTokenAuthClient creates a new HTTP client that supports token based authorization
 func NewTokenAuthClient(token string) *http.Client {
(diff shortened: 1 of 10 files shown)


pkg/client/client.go
```diff
@@ -298,13 +299,25 @@ func NewTokenAuthClient(token string) *http.Client {
 
 // NewOAuthClient creates a new HTTP client that supports OAuth2 client credentials based authorization
 func NewOAuthClient(oauthConfig OauthCredentials) *http.Client {
+
+	config := getClientCredentialsConfig(oauthConfig)
+	return config.Client(context.TODO())
+}
+
+func getClientCredentialsConfig(oauthConfig OauthCredentials) clientcredentials.Config {
```
UnseenWizzard, line 307: discussion/todo: try reverting this and testing on an actual HTTP client integration level, this is safer to changes

Laubi (approved):

PR #911, opened 2023-03-21
fix: Don't filter apis on settings only download

UnseenWizzard:
This refactoring splits API and Settings downloads  - and with that fixes the issue of presenting the user with useless information about deprecated API types while they were only downloading Settings anyway.

Previously API and Settings download logic was intertwined in the doDownload method,
leading to somewhat difficult to read code and doing some API related logic (filtering),
and printing unrelated messages, even when just Settings were being downloaded.

Split API and Settings downloads into seperate methods, each taking care of downloading
either all or specific types.

To simplify reading of the conditions under which APIs or Settings are being downloaded
at all, simple boolean methods with speaking names are introduced.

* **the actual fix happens in the last commit** and is achieved by refactoring/restructuring existing logic
*  the first three commits just introduce some changes related to tests that are added to make the refactoring safer

* users are no longer presented with API deprecation warnings if they are downloading Settings

### DIFF
--- a/cmd/monaco/download/download_configs.go
+++ b/cmd/monaco/download/download_configs.go
@@ -179,42 +179,60 @@ func downloadConfigs(apis api.APIs, opts downloadConfigsOptions) (project.Config
 
 	c = client.LimitClientParallelRequests(c, opts.concurrentDownloadLimit)
 
-	apisToDownload := getApisToDownload(apis, opts.specificAPIs)
+	configObjects := make(project.ConfigsPerType)
+
+	if shouldDownloadClassicConfigs(opts) {
+		classicCfgs, err := downloadClassicConfigs(c, apis, opts.specificAPIs, opts.projectName)
+		if err != nil {
+			return nil, err
+		}
+		maps.Copy(configObjects, classicCfgs)
+	}
+
+	if shouldDownloadSettings(opts) {
+		settingsObjects := downloadSettings(c, opts.specificSchemas, opts.projectName)
+		maps.Copy(configObjects, settingsObjects)
+	}
+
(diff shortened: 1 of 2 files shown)


cmd/monaco/download/download_configs.go
```diff
@@ -179,42 +179,60 @@ func downloadConfigs(apis api.APIs, opts downloadConfigsOptions) (project.Config
...
 	}
 
-	// return specific download objects
-	if len(opts.specificSchemas) > 0 || len(opts.specificAPIs) > 0 {
-		return configObjects, nil
+	return configObjects, nil
+}
+
+// shouldDownloadAPIs returns true unless onlySettings or specificSchemas but no specificAPIs are defined
+func shouldDownloadAPIs(opts downloadConfigsOptions) bool {
```
warber, line 201: let's rename this to shouldDownloadClassicConfigs

cmd/monaco/download/download_configs_test.go
```diff
@@ -0,0 +1,373 @@
...
+	}{
+		{
+			name: "Default opts: downloads Configs and Settings",
+			givenOpts: downloadConfigsOptions{
+				specificAPIs:    nil,
+				specificSchemas: nil,
+				onlyAPIs:        false,
+				onlySettings:    false,
+			},
+			expectedBehaviour: func(c *client.MockClient) {
```
warber, line 108: nice approach :)

cmd/monaco/download/download_configs.go
UnseenWizzard: rebased and amended to rename this

warber (approved):

PR #918, opened 2023-03-24
feat: introduce `MONACO_FEAT_VERIFY_ENV_TYPE` feature flag

warber:
In the current implementation we make a basic check against the DT environment, to verify that the user is targeting the correct "type" of the environment (classic vs platform). This check can put the user into the situation that he/she cannot use monaco at all if the check does not work anymore (for whatever reason)

This PR introduces a Feature flag `MONACO_FEAT_VERIFY_ENV_TYPE` tha can be used to explicitly turn off that check, by
setting it to `0`.

User is able to set `MONACO_FEAT_VERIFY_ENV_TYPE` to `0` to disable that check.


internal/featureflags/featureflags_test.go
jskelin, line 1: I wouldn't test each FF separately by name (e.g. TestFeatDangerousCommands, TestFeatVerifyEnvType)  , rather create one for specific test case!

I see that next test cases miss:
* env variable is empty (`export TEST`)
* env variable is not set
* env variable is set to unknown value (`export TEST=something`)

jskelin (changes requested):

internal/featureflags/featureflags.go
```diff
@@ -0,0 +1,75 @@
...
+func FeatDangerousCommands() FeatureFlag {
+	return FeatureFlag{
+		envName:        "MONACO_ENABLE_DANGEROUS_COMMANDS",
+		defaultEnabled: false,
+	}
+}
+
+// FeatVerifyEnvType returns the feature flag that tells whether the environment check
+// at the beginning of execution is enabled or not
+func FeatVerifyEnvType() FeatureFlag {
```
Laubi, line 70: What about calling a 'new' function here that already resolves the value?

internal/featureflags/featureflags.go
```diff
@@ -0,0 +1,75 @@
...
+// FeatEntities returns the feature flag that tells whether Dynatrace Entities download/matching is enabled or not
+func FeatEntities() FeatureFlag {
+	return FeatureFlag{
+		envName:        "MONACO_FEAT_ENTITIES",
+		defaultEnabled: false,
+	}
+}
+
+// FeatDangerousCommands returns the feature flag that tells whether dangerous commands for the CLI are enabled or not
+func FeatDangerousCommands() FeatureFlag {
```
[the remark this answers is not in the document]
Laubi: I think 'featureflags' is the better name here, as it's for flags for features, not features themselves.

Though I'd also drop the 'Feat' part

internal/featureflags/featureflags.go
```diff
@@ -0,0 +1,75 @@
...
+
+// FeatureFlag represents a command line switch to turn certain features
+// ON or OFF. Values are read from environment variables defined by
+// the feature flag. The feature flag can have default values that are used
+// when the resp. environment variable does not exist
+// Feature flags are considered to be "enabled" if their resp. environment variable
+// is set (to some value, usually "1").
+// Feature flags are considered to be "disabled" if their resp. environment variable
+// is set to "0" explicitly.
+type FeatureFlag struct {
```
Laubi, line 29: Should we maybe check for better values here?
1, true for enabled
0, false for disabled

```
// Feature flags are considered to be "disabled" if their resp. environment variable
// is set to "0" explicitly.
```

so setting MONAOC_FF=false would enable the ff currently

internal/featureflags/featureflags.go
```diff
@@ -0,0 +1,75 @@
...
+	// envName is the environment variable name
+	// that is used to read the value from
+	envName string
+	// defaultEnabled states whether this feature flag
+	// is enabled or disabled by default
+	defaultEnabled bool
+}
+
+// Eval evaluates the feature flag
+func (ff FeatureFlag) Eval() bool {
```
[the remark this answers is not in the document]
warber: yeah why not :)

internal/featureflags/featureflags.go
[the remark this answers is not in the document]
warber: i also think `featureflags` is just fine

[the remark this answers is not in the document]
warber: will drop the Feat from the funcs tho :)

internal/featureflags/featureflags_test.go
warber: I'll add a general test, but i will leave the tests for the specific FF as they are testing the Default values

internal/featureflags/featureflags.go
warber: I don't see a reason for that

internal/featureflags/featureflags.go
warber: open to discussing that.
for me it is fine, just continued with the semantics that where already there.

Laubi: Question is if we should panic whether the value is completely invalid. Imo would be ok.

warber: > https://pkg.go.dev/strconv#ParseBool
> I think we should use it, as it contains the current use-case, but also good other ones

looks good, let's do it like that :)

internal/featureflags/featureflags.go
```diff
@@ -0,0 +1,85 @@
...
+	envName string
+	// defaultEnabled states whether this feature flag
+	// is enabled or disabled by default
+	defaultEnabled bool
+}
+
+// New creates a new FeatureFlag
+// envName is the environment variable the feature flag is loading the values from when evaluated
+// defaultEnabled defines whether the feature flag is enabled or not by default
+func New(envName string, defaultEnabled bool) FeatureFlag {
```
jskelin, line 37: With this is possible in code to do something like:
``` go
featureflags.New("MONACO_FEAT_NEW", false).Enable()
```
That allows us to easily define new FF anyware. But as a price, our FF are going to be spreaded all over the project!

internal/featureflags/featureflags_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+package featureflags_test
+
+import (
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/featureflags"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestFeatureFlag(t *testing.T) {
```
jskelin, line 27: missing case
```go
	t.Setenv("MONACO_FEAT_TEST_FLAG", "0")
	assert.False(t, ff.Enabled())
```

jskelin (dismissed):
in general ok

internal/featureflags/featureflags.go
warber: technically you could do that before as well.
we could think of unexporting the featureflag type (and not introducing New)
this way we force the developer to introduce feature flags only via exported funcs..

jskelin (dismissed):

PR #921, opened 2023-03-28
feat: Make config.Config.Type an interface

Laubi:
Until now, config.Type has been a struct that contains all possible fields. To check what type it is, several isX methods were implemented. This is convenient for a few types, but with more and more unmaintainable.

This change makes config.Type an interface that all types can implement. This allows us to add more types and only change a few places where it is necessary to distinguish the types. One decision was to force the users of the method 'deploy' to cast and check the type on their own. One alternative would have been to add generics to the config object. To accomplish this, we would have to change all methods where the config.Config type is used, and even then we are not sure if this would really work. Another alternative would have been to add a method 'getData' to the Type interface. But this accomplishes basically the same, as the return value of this function would have to be checked and parsed too.

### DIFF
--- a/pkg/config/v2/config_writer.go
+++ b/pkg/config/v2/config_writer.go
@@ -261,50 +261,52 @@ func toTopLevelConfigDefinition(context *serializerContext, configs []Config) (t
 }
 
 func extractConfigType(context *serializerContext, config Config) (typeDefinition, error) {
-	if config.Type.IsSettings() {
-		return getConfigTypeSettings(config, context)
-	} else if config.Type.IsEntities() {
-		return getConfigTypeEntities(config)
-	} else {
-		return getConfigTypeApi(config)
-	}
 
+	switch t := config.Type.(type) {
+	case SettingsType:
+		serializedScope, err := getScope(context, config)
+		if err != nil {
+			return typeDefinition{}, err
+		}
+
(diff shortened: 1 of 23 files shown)


pkg/config/v2/config_writer.go
```diff
@@ -261,50 +261,44 @@ func toTopLevelConfigDefinition(context *serializerContext, configs []Config) (t
...
-	} else if config.Type.IsEntities() {
-		return getConfigTypeEntities(config)
-	} else {
-		return getConfigTypeApi(config)
-	}
 
-}
+	switch t := config.Type.(type) {
+	case SettingsType:
+		scopeParam, found := config.Parameters[ScopeParameter]
```
UnseenWizzard, line 267: nitpick: This switch might be a bit easier to read if at least getConfigTypeSettings was still an extra method (e.g. one just taking context, parameters and type t)

Laubi: I extracted the part of getting the scope into its own method, IMO that makes the most sense.

warber:
@Laubi I know this is already merged but could it be that we don't need the the `func ID() TypeId` at all?
You are checking for the actual type using type assertions anyway (which is good). `ID()` is just used for kind of pretty print the type in messages.

PR #926, opened 2023-03-29
Add OAuth support for `download direct`

jskelin:
New DT platform requires OAuth authentication. To be able to download configuration without manifest from platform, new flags (`--oauth-client-id` and `--oauth-client-secret`) are added.

With this changes, redesign of `download` command has been done. 

With this changes all arguments of  `download` command are removed. With that, previous `monaco download manifest environment ...`  can be achieved with next `monaco download --manifest --environment`, while `monaco download direct <url> ...` with `monaco download --url --token ...`.

### DIFF
--- a/cmd/monaco/download/download_command.go
+++ b/cmd/monaco/download/download_command.go
@@ -40,7 +40,7 @@ func GetDownloadCommand(fs afero.Fs, command Command) (downloadCmd *cobra.Comman
 	var onlySettings bool
 	var token string
 	var oAuthClientID, oAuthClientSecret string
-	var manifest string
+	var manifestFile string
 	var url string
 
 	downloadCmd = &cobra.Command{
@@ -67,7 +67,7 @@ Either downloading based on an existing manifest, or define an URL pointing to a
 				}
 
 				options := directDownloadCmdOptions{
-					environmentUrl: url,
+					environmentURL: url,
 					auth: auth{
 						token:        token,
 						clientID:     oAuthClientID,
@@ -93,7 +93,7 @@ Either downloading based on an existing manifest, or define an URL pointing to a
 					return errors.New("missing --environment/-e flag")
 				}
 				options := manifestDownloadOptions{
-					manifestFile:            manifest,
+					manifestFile:            manifestFile,
 					specificEnvironmentName: specificEnvironment,
 					downloadCmdOptions: downloadCmdOptions{
 						sharedDownloadCmdOptions: sharedDownloadCmdOptions{
@@ -117,7 +117,7 @@ Either downloading based on an existing manifest, or define an URL pointing to a
 	downloadCmd.Flags().StringVar(&token, "token", "", "Token secret to connect to DT server")
 	downloadCmd.Flags().StringVar(&oAuthClientID, "oauth-client-id", "", "OAuth client ID is used to connect to DT server via OAuth (mandatory for OAuth access type)")
 	downloadCmd.Flags().StringVar(&oAuthClientSecret, "oauth-client-secret", "", "OAuth client secret is used to connect to DT server via OAuth (mandatory for OAuth access type)")
-	downloadCmd.Flags().StringVarP(&manifest, "manifest", "m", "manifest.yaml", "Path to the manifest.yaml file to be read")
+	downloadCmd.Flags().StringVarP(&manifestFile, "manifest", "m", "manifest.yaml", "Path to the manifest.yaml file to be read")
 	downloadCmd.Flags().StringVarP(&specificEnvironment, "environment", "e", "", "Specify a concrete environment that shall be downloaded (only usable with --manifest)")
 	downloadCmd.Flags().StringVarP(&url, "url", "u", "", "URL to the dynatrace environment from which to download configuration from")
 	if featureflags.Entities().Enabled() {
--- a/cmd/monaco/download/download_command.go
+++ b/cmd/monaco/download/download_command.go
@@ -16,6 +16,7 @@ package download
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
@@ -42,11 +43,11 @@ Either downloading based on an existing manifest, or by defining environment URL
 		Example: `- monaco download manifest manifest.yaml some_environment_from_manifest
 - monaco download direct https://environment.live.dynatrace.com API_TOKEN_ENV_VAR_NAME`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return fmt.Errorf("'direct' or 'manifest' sub-command is required")
+			return fmt.Errorf("\"direct\" or \"manifest\" sub-command is required") //TODO: this is become obsolete
 		},
 	}
 
-	getDownloadConfigsCommand(fs, command, downloadCmd)
+	addDownloadConfigsCommand(fs, command, downloadCmd)
 
 	if featureflags.Entities().Enabled() {
 		getDownloadEntitiesCommand(fs, command, downloadCmd)
@@ -55,13 +56,14 @@ Either downloading based on an existing manifest, or by defining environment URL
 	return downloadCmd
 }
 
-func getDownloadConfigsCommand(fs afero.Fs, command Command, downloadCmd *cobra.Command) {
+func addDownloadConfigsCommand(fs afero.Fs, command Command, downloadCmd *cobra.Command) {
 	var project, outputFolder string
 	var forceOverwrite bool
 	var specificApis []string
 	var specificSettings []string
 	var onlyAPIs bool
 	var onlySettings bool
+	var token string
 
 	manifestDownloadCmd := &cobra.Command{
 		Use:     "manifest [manifest file] [environment to download]",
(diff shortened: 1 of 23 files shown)


cmd/monaco/download/download_command.go
```diff
@@ -16,140 +16,115 @@ package download
...
 
-	directDownloadCmd := &cobra.Command{
-		Use:     "direct [URL] [TOKEN_NAME]",
-		Aliases: []string{"d"},
-		Short:   "Download configuration from a Dynatrace environment specified on the command line",
-		Example: `monaco download direct https://environment.live.dynatrace.com API_TOKEN_ENV_VAR_NAME`,
-		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) != 2 || args[0] == "" || args[1] == "" {
-				return fmt.Errorf(`url and token have to be provided as positional argument`)
+		Example: `- monaco download --manifest manifest.yaml --token API_TOKEN_ENV_VAR_NAME
```
UnseenWizzard, line 53:
```suggestion
		Example: `- monaco download --manifest manifest.yaml --environment dev-env
```

cmd/monaco/download/download_command.go
```diff
@@ -16,140 +16,115 @@ package download
...
-			options := directDownloadOptions{
-				environmentUrl: url,
-				envVarName:     tokenEnvVar,
-				downloadCommandOptions: downloadCommandOptions{
-					downloadCommandOptionsShared: downloadCommandOptionsShared{
-						projectName:    project,
-						outputFolder:   outputFolder,
-						forceOverwrite: forceOverwrite,
+			if url != "" {
+				manifestFile = ""
```
UnseenWizzard, line 64: it's not necessary to do this is it? the content isn't used anyway in this logical branch

cmd/monaco/download/download_command.go
```diff
@@ -16,140 +16,115 @@ package download
...
+						specificSchemas: specificSettings,
+						onlyAPIs:        onlyAPIs,
+						onlySettings:    onlySettings,
+					},
+				}
+
+				return command.DownloadConfigs(fs, options)
+
+			} else {
+				if specificEnvironment == "" {
```
UnseenWizzard, line 89: I think we could achieve this check simply by defining a 

```
downloadCmd.MarkFlagsRequiredTogether("manifest", "environment")
```

cmd/monaco/download/download_command.go
```diff
@@ -16,140 +16,115 @@ package download
...
+	setupSharedConfigsFlags(downloadCmd, &project, &outputFolder, &forceOverwrite, &specificApis, &specificSettings, &onlyAPIs, &onlySettings)
+	downloadCmd.Flags().StringVar(&token, "token", "", "Token secret to connect to DT server")
+	downloadCmd.Flags().StringVar(&oAuthClientID, "oauth-client-id", "", "OAuth client ID is used to connect to DT server via OAuth (mandatory for OAuth access type)")
+	downloadCmd.Flags().StringVar(&oAuthClientSecret, "oauth-client-secret", "", "OAuth client secret is used to connect to DT server via OAuth (mandatory for OAuth access type)")
+	downloadCmd.Flags().StringVarP(&manifestFile, "manifest", "m", "manifest.yaml", "Path to the manifest.yaml file to be read")
+	downloadCmd.Flags().StringVarP(&specificEnvironment, "environment", "e", "", "Specify a concrete environment that shall be downloaded (only usable with --manifest)")
+	downloadCmd.Flags().StringVar(&url, "url", "", "URL to the dynatrace environment from which to download configuration from")
+	downloadCmd.MarkFlagsMutuallyExclusive("url", "manifest")
```
UnseenWizzard, line 120: either as manual `PreRun` logic or here via `MarkFlagsMutuallyExclusive` we also need to exclude: 

manifest + either oAuth flag

manifest + token should be made impossible by the fact that URL and token are required together, but URL and manifest are exclusive. 
Cobra might give a better error message if token + manifest are already invalid though

cmd/monaco/download/download_command_test.go
```diff
@@ -711,59 +826,27 @@ func TestValidCommands(t *testing.T) {
...
-			commandMock := createDownloadCommandMock(t)
-
-			cmd := GetDownloadCommand(afero.NewOsFs(), commandMock)
-			cmd.SetArgs(strings.Split(test.args, " "))
-			cmd.SetOut(io.Discard) // skip output to ensure that the error message contains the error, not the help message
-			err := cmd.Execute()
-
-			// for all test cases there should be at least an error
-			assert.Assert(t, err != nil, "there should be an error")
+func TestEntitiesDownloadEnabled(t *testing.T) {
```
UnseenWizzard, line 842: If I understand the code correctly, the test name is wrong.

This tests a case when Entities download is NOT enabled, verifying that it can't  actually be used.

```suggestion
func TestEntitiesDownloadUnavailableByDefault(t *testing.T) {
```

or something like that

cmd/monaco/download/download_configs.go
```diff
@@ -43,12 +43,53 @@ type downloadCommandOptions struct {
...
+			errors = append(errors, err)
+		} else {
+			retVal.OAuth.ClientSecret = v
+		}
+	}
+	return &retVal, errors
+}
+
+func readEnvVariable(envVar string) (manifest.AuthSecret, error) {
+	var context string
```
UnseenWizzard, line 80: nit: should this be named `content`?

pkg/manifest/manifest.go
```diff
@@ -54,24 +42,22 @@ type OAuth struct {
 func (o OAuth) GetTokenEndpointValue() string {
 	if o.TokenEndpoint == nil {
 		return ""
+	} else {
```
UnseenWizzard, line 45: nit: why introduce the `else` branch here?
seems especially strange as this PR also drops an else from [download_configs.go#getApisToDownload](https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/926/files#diff-3f23ca622bf717266f5e6efcbe7a9c319ffdba2f19a76ed66d1893676d8e5e83R313)

maybe once change is by @jskelin and one by @warber though? How want to deal with else vs early-exit, might a good thing to discuss for coding standards :sweat_smile:

pkg/manifest/manifest_loader_test.go
```diff
@@ -508,11 +508,6 @@ func TestVerifyManifestYAML(t *testing.T) {
 			given:    given{},
 			expected: expected{errorMessage: "no `environmentGroups` defined"},
 		},
-		{
```
UnseenWizzard, line 511: why is this test removed?

UnseenWizzard:
With this change, we left out the Entites download right? 
So it is still split in manifest/direct subcommand and downloading entities from a Platform environment currently only works when using a manifest?

Fine for getting this change into the 2.0 release as it's hidden/unused at the moment, but we must not forget to adapt it in a followup. FYI @dcryans

cmd/monaco/download/download_command.go
jskelin: this will be part of refactoring (as separate PR or one of last commits in this one)

cmd/monaco/download/download_configs.go
jskelin: done

pkg/manifest/manifest.go
jskelin: on places like this usually I would use elvis operator (?:)
I'll return it back

pkg/manifest/manifest_loader_test.go
jskelin: duplication of lines 496-500

cmd/monaco/integrationtest/v2/config_restore_e2e_test.go
```diff
@@ -253,33 +263,23 @@ func preparation_uploadConfigs(t *testing.T, fs afero.Fs, suffixTest string, con
...
-		parameters = []string{
-			"download",
-			"direct",
-			os.Getenv("URL_ENVIRONMENT_1"),
-			"TOKEN_ENVIRONMENT_1",
-			"--verbose",
-			"--output-folder", downloadFolder,
-		}
+	parameters := []string{"download", "--verbose", "--output-folder", downloadFolder}
+	if apisToDownload != "all" {
```
UnseenWizzard, line 282: nit: could drop the nested if and just do 
```suggestion
	if apisToDownload != "all" && apisToDownload != "" {
```

cmd/monaco/download/download_command.go
jskelin: The whole help section is changed. please take a look

cmd/monaco/download/download_command.go
jskelin: done i preRunE

cmd/monaco/download/download_command_test.go
jskelin: Anything related with endities was ignored and left as it was :))

I would to divide entity related code to separate files

cmd/monaco/download/download_command.go
```diff
@@ -30,101 +30,92 @@ import (
...
+		log.Fatal("failed to setup CLI %v", err)
+	}
+
+	return cmd
+}
+
+func preRunChecks(f downloadCmdOptions) error {
+	switch {
+	case f.environmentURL != "" && f.manifestFile != "manifest.yaml":
+		return errors.New("\"url\" and \"token\" are mutually exclusive")
```
UnseenWizzard, line 97:
```suggestion
		return errors.New("\"url\" and \"manifest\" are mutually exclusive")
```

cmd/monaco/download/download_command.go
```diff
@@ -30,101 +30,92 @@ import (
...
+	case f.environmentURL != "" && f.specificEnvironmentName != "":
+		return errors.New("")
```
UnseenWizzard, line 99:
```suggestion
		return errors.New("\"environment\" is specific to manifest-based download and incompatible with direct download from "\url"\")
```

cmd/monaco/integrationtest/v2/config_restore_e2e_test.go
jskelin: there is 2 nested if branching here...

cmd/monaco/download/download_command.go
```diff
@@ -30,101 +30,92 @@ import (
...
+	case f.environmentURL != "":
+		switch {
+		case f.token == "":
+			return errors.New("if \"url\" is set, \"token\" also must be set")
+		case (f.clientID == "") != (f.clientSecret == ""):
+			return errors.New("\"oauth-client-id\" and \"oauth-client-secret\" always come in pair")
```
UnseenWizzard, line 105:
```suggestion
			return errors.New("\"oauth-client-id\" and \"oauth-client-secret\" must always be set together")
```

UnseenWizzard: actually I wonder if the base cobra `RequiredTogether` wouldn't give us a better message for this - IIRC it tells you which one you're missing

UnseenWizzard (commented):
small comments re: user facing messages

cmd/monaco/download/download_command.go
```diff
@@ -16,140 +16,115 @@ package download
...
-						forceOverwrite: forceOverwrite,
-					},
-					specificAPIs:    specificApis,
-					specificSchemas: specificSettings,
-					onlyAPIs:        onlyAPIs,
-					onlySettings:    onlySettings,
-				},
-			}
+	downloadCmd = &cobra.Command{
+		Use:   "download --manifest <manifest.yaml> --token <API_TOKEN_ENV_VAR_NAME>",
```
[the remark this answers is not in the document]
jskelin: covered in examples.
I have a feeling that setting flag usage here is wrong. by default cobra add `[flags]` at end of this line

cmd/monaco/download/download_command.go
jskelin: done

cmd/monaco/download/download_command.go
jskelin: It would be nice if there is an way to say order of the rules. E.g. for setting that `manifest` and `token` are mutually exclusive, and that `token` and `url` needs to come together, in case of the call:
```
monaco download --manifest myManifest.yaml --token TOKEN...
``` 
error message will be like: `token` and `url` must be both set. 

For `... --manifest .. --url --token` error will report that `manifest` and `url` are forbidden combination.


That is by my opinion misleading

PR #929, opened 2023-03-30
feat: Allow purge of settings objects

Laubi:
This change adds purging all settings objects from one or more environments.

If settings should not be deleted, an opt-out flag is introduced to skip them.

### DIFF
--- a/pkg/delete/delete.go
+++ b/pkg/delete/delete.go
@@ -170,6 +170,7 @@ func DeleteAllConfigs(client client.ConfigClient, apis api.APIs) (errors []error
 		log.Info("Deleting %d configs of type %s...", len(values), api.ID)
 
 		for _, v := range values {
+			log.Debug("Deleting config %s/%s", api.ID, v.Id)
 			// TODO(improvement): this could be improved by filtering for default configs the same way as Download does
 			err := client.DeleteConfigById(api, v.Id)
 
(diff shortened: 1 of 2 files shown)


cmd/monaco/purge/command.go
```diff
@@ -48,13 +50,14 @@ func GetPurgeCommand(fs afero.Fs) (purgeCmd *cobra.Command) {
...
 
-			return purge(fs, manifestName, environment, specificApis)
+			return purge(fs, manifestName, environment, specificApis, !skipSettings)
 		},
 		ValidArgsFunction: completion.PurgeCompletion,
 	}
 
 	purgeCmd.Flags().StringSliceVarP(&environment, "environment", "e", make([]string, 0), "Deletes configuration only for specified envs. If not set, delete will be executed on all environments defined in manifest.")
 	purgeCmd.Flags().StringSliceVarP(&specificApis, "api", "a", make([]string, 0), "One or more specific APIs to delete from (flag can be repeated or value defined as comma-separated list)")
+	purgeCmd.Flags().BoolVar(&skipSettings, "skip-settings", false, "Settings are not deleted if this flag is set.")
```
UnseenWizzard, line 60: if we add a skip-settings, should we add skip-configs as well? (I'm not sure this skip is even needed for purge tbh)

UnseenWizzard (approved):
lgtm. just small comments

pkg/delete/delete.go
```diff
@@ -181,3 +182,40 @@ func DeleteAllConfigs(client client.ConfigClient, apis api.APIs) (errors []error
...
+
+	for _, s := range schemaIds {
+		log.Info("Collecting configs of type %s...", s)
+		settings, err := c.ListSettings(s, client.ListSettingsOptions{DiscardValue: true})
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+
+		log.Info("Deleting %d configs of type %s...", len(settings), s)
```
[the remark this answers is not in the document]
Laubi: It's configs, settings, and in the near future, even more. I'd just stick with configs for an overall word

cmd/monaco/purge/command.go
Laubi: Removed the flag altogether

UnseenWizzard (approved):

PR #931, opened 2023-03-31
refactor: Remove `limitingClient` decorator and use the limiter directly

warber:
This PR changes the usage of the `conurrency.Limiter` component to be used by the client instead of as a decorator.
Given that we want to get rid of the `client.Client` interface (and potentially all other interfaces there) it's easier to use that component inside directly inside the implementation of the resp. client instead of having the need to create a decorator for each and every caller side interface.

--

--


cmd/monaco/download/download_entities.go
```diff
@@ -108,15 +105,15 @@ func (d DefaultCommand) DownloadEntities(fs afero.Fs, cmdOptions entitiesDirectD
...
+			outputFolder:           cmdOptions.outputFolder,
+			projectName:            cmdOptions.projectName,
+			forceOverwriteManifest: cmdOptions.forceOverwrite,
 		},
 		specificEntitiesTypes: cmdOptions.specificEntitiesTypes,
 	}
 
-	dtClient, err := client.NewClassicClient(cmdOptions.environmentUrl, token)
+	dtClient, err := client.NewClassicClient(cmdOptions.environmentUrl,
+		token, client.WithClientRequestLimiter(concurrency.NewLimiter(environment.GetEnvValueIntLog(environment.ConcurrentRequestsEnvKey))))
```
Laubi, line 116: Should we just hide the limiter somewhere inside instead of exposing it to consumers?

UnseenWizzard: I agree with @Laubi - I don't think there's any case where we don't use the limiter anymore, so like the version check we should see it as a default (maybe introduce "default options"?) so we don't forget about it the next time we construct a client?

warber: You are probably right - I will change the code so that the client uses a limiter with its values read from the environment variable in any case by default.

Also, I was thinking about not exposing the limiter at all, But then I had the feeling that this component is basically designed for reuse (from any differnt package).  However, I just see that It is not so obvious to reuse it, e.g.:

```
type MyFancyComponent struct {
 limiter concurrency.Limiter
}
```

will panic as Limiter does not provide usable zero values, i.e. you need to remember to construct it with `NewLimiter`.
This can be fixed with guards of course.

cmd/monaco/cmdutils/cmdutils.go
```diff
@@ -40,19 +40,19 @@ func SilenceUsageCommand() func(cmd *cobra.Command, args []string) {
 // CreateDTClient is driven by data given through a manifest.EnvironmentDefinition to create an appropriate client.Client.
 //
 // In case when flag dryRun is true this factory returns the client.DummyClient.
-func CreateDTClient(env manifest.EnvironmentDefinition, dryRun bool) (client.Client, error) {
+func CreateDTClient(env manifest.EnvironmentDefinition, dryRun bool, opts ...func(dynatraceClient *client.DynatraceClient)) (client.Client, error) {
```
jskelin, line 43: why not put dryRun also as optional fucntion parameter

warber: it's evaluated in the factory function, if it's set we are passing back ka DummyClient. So it doesn't make sense to pass it as a client option

warber: For now `CreateDTClient` decides based on information what client implementation to hand back.
`DryRun` is one of the parameters this decision is based on. It has nothing to do with the internals of a specific client implementation. So your point still doesn't make sense to me, sorry.

warber: I would even say it's the opposite of what you say what would be more consistent: passing the limiter parameters as required (non optional) parameter.

pkg/client/client.go
```diff
@@ -230,6 +235,14 @@ var (
 	_ Client         = (*DynatraceClient)(nil)
 )
 
+// WithClientRequestLimiter specifies that a specifies the limiter to be used for
+// limiting parallel client requests
+func WithClientRequestLimiter(limiter *concurrency.Limiter) func(client *DynatraceClient) {
```
jskelin, line 240: with the face that `concurrency.Limiter` is from internal package, is it going to be possible to use this as a library?

warber: inside our code-base, yes. from another module, no

jskelin (dismissed):

jskelin (dismissed):

UnseenWizzard (approved):

PR #933, opened 2023-04-03
Silence help message on download error

jskelin:
After the last changes on the download command, help messages will be printed after every unsuccessful execution. With this PR help messages for unsuccessful execution are going to be suppressed.

### DIFF
--- a/cmd/monaco/download/download_command.go
+++ b/cmd/monaco/download/download_command.go
@@ -50,6 +50,8 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
 			return preRunChecks(f)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
+
 			if f.environmentURL != "" {
 				f.manifestFile = ""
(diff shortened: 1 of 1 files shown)


cmd/monaco/download/download_command.go
```diff
@@ -50,6 +50,8 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
 			return preRunChecks(f)
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
```
UnseenWizzard, line 53: Probably makes no functional difference, so I'd merge this and address it after 2.0 release: 
In other commands we set this up in the PreRun function - would it make sense to silence this in the pre-run as well? Or should we ruther not silence other commands in the pre-run, but on run? 
See the usages of cmd.SilenceUsageCommand

UnseenWizzard (approved):

cmd/monaco/download/download_command.go
jskelin: Tried to put as another command, but did not work. Another command doesn't use PreRunE, probably that's the reason.

PR #939, opened 2023-04-04
fix(download): Resolve Managment Zone references by numeric ID

UnseenWizzard:
Management zones are always referenced in configurations via their legacy numeric ID for backwards compatibility reasons.

To still allow download code to correctly build reference parameters to management zone Settings configurations, which are only known by their object ID, code is introduced that leverages knowledge about how the numeric ID is formed internally and transforms a given object ID into a numeric ID.

### DIFF
--- a/internal/idutils/numeric_id.go
+++ b/internal/idutils/numeric_id.go
@@ -1,0 +1,126 @@
+/*
+ * @license
+ * Copyright 2023 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


jskelin (dismissed):

internal/idutils/numeric_id.go
```diff
@@ -0,0 +1,123 @@
...
+	val.SetBytes(u[:])             // turn the 16 UUID bytes into a 128bit number
+	val.SetBytes(val.Bytes()[0:8]) // keep the most significant 64bits (8byte)
+
+	numID := int(val.Uint64())
+
+	return numID
+}
+
+// getLegacyNumericId implements the Dynatrace logic for transforming a "legacy" random UUID to a numeric ID
+func getLegacyNumericId(u uuid.UUID) int {
```
warber, line 68:
```suggestion
func getLegacyNumericID(u uuid.UUID) int {
```

Laubi (approved):

UnseenWizzard:
closed in favor of #946

PR #945, opened 2023-04-05
Improve Autocompletion for the download command

Laubi:
I'd like to show a small demo, but I don't want to dox my ENV vars (even though its only the names)


cmd/monaco/download/download_command.go
```diff
@@ -85,7 +85,16 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
 		getDownloadEntitiesCommand(fs, command, cmd)
 	}
 
-	err := cmd.RegisterFlagCompletionFunc("api", completion.AllAvailableApis)
+	err := errors.Join(
```
jskelin, line 88: I am totally for `errors.Join`!
But it have side effect on our pretty print. 
e.g.
```
2023/04/06 08:45:37 ERROR the content of the environment variable "bla" is not set
2023/04/06 08:45:37 ERROR the content of the environment variable "adf" is not set
2023/04/06 08:45:37 ERROR the content of the environment variable "adsf" is not set
```
vs
```
2023/04/06 08:45:15 ERROR the content of the environment variable "bla" is not set
the content of the environment variable "adf" is not set
the content of the environment variable "adsf" is not set
```

jskelin: This is probably not an issue here, but in general...

cmd/monaco/convert/command.go
```diff
@@ -19,7 +19,7 @@ package convert
 import (
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/cmd/monaco/cmdutils"
-	"github.com/dynatrace/dynatrace-configuration-as-code/cmd/monaco/runner/completion"
```
jskelin, line 22: I wouldn't move it, It is mora a util while other packages on that level are business.

cmd/monaco/download/download_command.go
Laubi: As discussed offline: errors.Join is, overall not the way to go. Here is an exception to the rule, as this code fails before the cli is basically started. That should never happen as we also have build targets to start the cli and check that no errors occur.

cmd/monaco/convert/command.go
Laubi: Yes it is a util, but equally it does not belong below the runner. 

I thought about moving it to internal/completion, but we should keep all CLI-related stuff inside the cmd package

jskelin: Even when I see the content of this file, all exposed functions should go closer to caller (in same file), and unexposed reused functions are utils.

But this is some another refactoring...

jskelin (approved):

jskelin (approved):

jskelin (approved):

PR #946, opened 2023-04-05
fix : Correctly resolve dependencies when downloading Settings 2.0 Management Zones

UnseenWizzard:
 fix(download): Resolve Managment Zone references by numeric ID

Management zones are always referenced in configurations via their numeric
ID instead of a Config UUID or Settings objectID for backwards compatibility
reasons.

To still allow download code to correctly build reference parameters to
management zone Settings configurations, which are only known by their objectID,
code is introduced that leverages knowledge about how the numeric ID is formed
internally and transforms a given object ID into a numeric ID.

The implementation matches the logic of the Dynatrace implementation for
this as closely as possible.

This version contains the base change matching Dynatrace numeric ID generation, and simplifies our code in the second commit.

References to ManagementZones will be correctly resolved on download

### DIFF
--- a/internal/idutils/numeric_id.go
+++ b/internal/idutils/numeric_id.go
@@ -1,0 +1,126 @@
+/*
+ * @license
+ * Copyright 2023 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package idutils
+
+import (
+	"encoding/base64"
+	"fmt"
+	"github.com/google/uuid"
+	"math/big"
+	"regexp"
+)
+
+var uuidRegex = regexp.MustCompile(".*?([0-9a-fA-F]{8}\\b-[0-9a-fA-F]{4}\\b-[0-9a-fA-F]{4}\\b-[0-9a-fA-F]{4}\\b-[0-9a-fA-F]{12}).*?")
+
+// GetNumericIDForObjectID parses the Settings Object ID of a Dynatrace Management Zone (only object with numeric IDs)
+// into a numeric identifier. To achieve this is replicates the en-/decoding logic used in Dynatrace as closely as possible.
+func GetNumericIDForObjectID(objectID string) (int, error) {
+	decodedObjectID, err := base64.RawURLEncoding.DecodeString(objectID)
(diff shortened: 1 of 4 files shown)


internal/idutils/numeric_id.go
```diff
@@ -0,0 +1,77 @@
...
+	uuidString := string(matches[1])
+
+	u, err := uuid.Parse(uuidString)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse UUID %q: %w", uuidString, err)
+	}
+
+	if u.Variant() == uuid.RFC4122 && u.Version() == 4 {
+		return getLegacyNumericId(u)
+	} else {
```
warber, line 51: else branch can be ommitted

internal/idutils/numeric_id.go
```diff
@@ -0,0 +1,76 @@
...
+// into a numeric identifier. To achieve this it replicates the en-/decoding logic used in Dynatrace as closely as possible.
+func GetNumericIDForObjectID(objectID string) (int, error) {
+	decodedObjectID, err := base64.RawURLEncoding.DecodeString(objectID)
+	if err != nil {
+		return 0, fmt.Errorf("failed to decode objectID %q: %w", objectID, err)
+	}
+
+	matches := uuidRegex.FindSubmatch(decodedObjectID)
+	if len(matches) != 2 {
+		return 0, fmt.Errorf("failed to read UUID from decoded objectID %q: expected regex match for contained UUID but got %v", decodedObjectID, matches)
```
jskelin, line 39: Is it useful to include `objectID` in err message?

pkg/download/dependency_resolution.go
```diff
@@ -132,8 +133,19 @@ func collectConfigsById(configs project.ConfigsPerType) map[string]config.Config
...
+			if conf.OriginObjectId != "" && conf.Coordinate.Type == "builtin:management-zones" {
+				// resolve Management Zone Settings by Numeric ID as well
+				numID, err := idutils.GetNumericIDForObjectID(conf.OriginObjectId)
+				if err != nil {
+					log.Warn("Failed to decode numeric ID of config %q, auto-resolved references may be incomplete: %v", conf.Coordinate, err)
+				} else {
+					strId := fmt.Sprintf("%d", numID)
+					configsById[strId] = conf
+				}
+			}
```
warber, line 148: This part seems to be not covered by any tests.
If it's easy to provide a test case for it pls do it, otherwise I guess it's fine

UnseenWizzard: Added some tests, but didn't find a reasonable way to test two of the error cases - they should never happen if the previous checks have passed

warber: I was more like thinking of adding specific test cases to `dependency_resolution_test.go`

UnseenWizzard: ah... right, apparently reading is not my strong suite, missed on what you had actually commented 😅 

Yes, I will use the known IDs used in the idutils test cases for some dependency cases as well 👍

UnseenWizzard: test extended & commits cleaned up

PR #947, opened 2023-04-05
Move parameter-resolving and Rendering of a config into the Deploy function

jskelin:
Move parameter-resolving and Rendering of a config into the Deploy function

Both functions "deploySettings" and "deployConfig" currently do parameter resolving and rendering. Automation resources will need it as well.
The goal of this story is to move this functionality one level up into the "DeployConfigs" function. With that, all future configs will work in the future as well.

### DIFF
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -44,18 +44,11 @@ func DeployConfigs(client dtclient.Client, apis api.APIs, sortedConfigs []config
 	for i := range sortedConfigs {
 		c := &sortedConfigs[i] // avoid implicit memory aliasing (gosec G601)
 
-		if c.Skip {
-			log.Info("\tSkipping deployment of config %s", c.Coordinate)
-		}
-
-		logAction, logVerb := getWordsForLogging(opts.DryRun)
-		log.Info("\t%s config %s", logAction, c.Coordinate)
-
 		entity, deploymentErrors := deploy(client, apis, entityMap, c)
 
 		if deploymentErrors != nil {
 			for _, err := range deploymentErrors {
-				errors = append(errors, fmt.Errorf("failed to %s config %s: %w", logVerb, c.Coordinate, err))
+				errors = append(errors, fmt.Errorf("failed to deploy config %s: %w", c.Coordinate, err))
 			}
 
 			if !opts.ContinueOnErr && !opts.DryRun {
@@ -71,6 +64,7 @@ func DeployConfigs(client dtclient.Client, apis api.APIs, sortedConfigs []config
 
 func deploy(client dtclient.Client, apis api.APIs, em *entityMap, c *config.Config) (*parameter.ResolvedEntity, []error) {
 	if c.Skip {
+		log.Info("\tSkipping deployment of config %s", c.Coordinate)
 		return &parameter.ResolvedEntity{EntityName: c.Coordinate.ConfigId, Coordinate: c.Coordinate, Properties: parameter.Properties{}, Skip: true}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/deploy/deploy.go
```diff
@@ -62,20 +62,30 @@ func DeployConfigs(client client.Client, apis api.APIs, sortedConfigs []config.C
...
-			entity, deploymentErrors = deploySetting(client, entityMap, &c)
+		properties, errs := resolveProperties(&c, entityMap.get())
+		if len(errs) > 0 {
+			entity = parameter.ResolvedEntity{}
+			deploymentErrors = append(deploymentErrors, errs...)
+		}
 
-		case config.ClassicApiType:
-			entity, deploymentErrors = deployConfig(client, apis, entityMap, &c)
+		renderedConfig, err := c.Render(properties)
```
Laubi, line 71: If resolveProperties fails, rendering should not be done.

pkg/deploy/deploy.go
```diff
@@ -62,20 +62,30 @@ func DeployConfigs(client client.Client, apis api.APIs, sortedConfigs []config.C
...
+			deploymentErrors = append(deploymentErrors, err)
+		}
 
-		default:
-			deploymentErrors = []error{fmt.Errorf("unknown config-type (ID: %q)", c.Type.ID())}
+		if deploymentErrors == nil {
+			switch t := c.Type.(type) {
+			case config.EntityType:
+				log.Debug("Entity are not deployable, skipping entity type: %s", t.EntitiesType)
+				continue
```
Laubi, line 81: \minor: Please add empty newlines between cases. I think it is way easier to distinguish cases.

pkg/deploy/deploy.go
```diff
@@ -103,36 +113,24 @@ func getWordsForLogging(isDryRun bool) (action, verb string) {
 	return "Deploying", "deploy"
 }
 
-func deployConfig(configClient client.ConfigClient, apis api.APIs, entityMap *entityMap, conf *config.Config) (parameter.ResolvedEntity, []error) {
-
-	t, ok := conf.Type.(config.ClassicApiType)
+func deployConfig(configClient client.ConfigClient, apis api.APIs, entityMap *entityMap, c *config.Config, properties parameter.Properties, renderedConfig string) (parameter.ResolvedEntity, []error) {
+	t, ok := c.Type.(config.ClassicApiType)
```
Laubi, line 117: Why rename `conf` to `c`? Imo `conf` was a better var name than `c`. It's more descriptive and does not clash with any package. And I have a harder time reviewing because it's in the same commit

Laubi (changes requested):

Laubi:
Commit-nitpicks: 
1. Please rename the 'sanitize' stuff Prs to 'renamed variables', and add a small description why, for example, package name collisions, etc.

2. Please also add a description to your second commit about why it is changing; a short sentence would be nice.

Also, for the future: Renaming and functional change (conf -> c) should definitely be their own commit. It makes reviewing really hard if there are two things done which have nothing to do with each other.

pkg/deploy/deploy.go
jskelin: One was conf, another was c. I'll bring it back to conf

pkg/deploy/deploy.go
```diff
@@ -41,42 +41,17 @@ func DeployConfigs(client dtclient.Client, apis api.APIs, sortedConfigs []config
 	entityMap := newEntityMap(apis)
 	var errors []error
 
-	for _, c := range sortedConfigs {
-		c := c // to avoid implicit memory aliasing (gosec G601)
+	for i := range sortedConfigs {
+		c := &sortedConfigs[i] // avoid implicit memory aliasing (gosec G601)
 
 		if c.Skip {
```
Laubi, line 47: This is now duplicated - the depoy method as well as the DeployConfigs logic now do check if Skip is set. Imo, the code, as it was, was fine. 
Now we have two places to worry. 

With this change, we no do
1) Info log when skipping
2) Printing we do validation or deployment (we aren't, we skip the config!)
3) Return an empty object if we skip

The code is not optimal, but I suggest moving the continue back in this method. Skipping really means no deployment. I like moving the deploy functionality in its own method, but skipped configs should never reach it.

pkg/deploy/deploy.go
```diff
@@ -86,14 +61,46 @@ func DeployConfigs(client dtclient.Client, apis api.APIs, sortedConfigs []config
 			if !opts.ContinueOnErr && !opts.DryRun {
 				return errors
 			}
-		} else {
-			entityMap.put(entity.Coordinate, entity)
+		} else if entity != nil {
```
Laubi, line 64: If the skip above is changed to what it was before, this will make more sense again.
If an error occurs, we should not care about the entity.

pkg/deploy/deploy.go
jskelin: The code is far from optimal, on that we agree :))

I'll move the logging info in `deploy` function, as well as logging for "real" deployment..

Bullet 3) I do not understand.
Functionality with the refactoring stays the same as before. Only problem is with logging (on info level). But that was also before, e.g. for `case config.EntityType` first prints on the info level that something is deployed, and after that on debug level that it is not.

pkg/deploy/deploy.go
jskelin: For this case
```
		case config.EntityType:
			log.Debug("Entity are not deployable, skipping entity type: %s", t.EntitiesType)
			continue
```
there is no error, and nothing will be put in `entityMap`.

Laubi (approved):

PR #948, opened 2023-04-06
fix: generate sha sum correctly on jenkins

jskelin:
Currently a file with sha sum generated during the build process doesn't comply to the standard. With this changes this is fix.

### DIFF
--- a/.ci/releasePipeline.groovy
+++ b/.ci/releasePipeline.groovy
@@ -174,7 +174,7 @@ void signWinBinaries(Map args = [source: null, version: null, destDir: null, pro
 
 void createShaSum(Map args = [source: null, dest: null]) {
     stage('Compute SHA sum') {
-        sh "sha256sum ${args.source} | sed 's, .*/, ,' > ${args.dest}"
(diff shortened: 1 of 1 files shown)


UnseenWizzard (approved):

.ci/releasePipeline.groovy
```diff
@@ -174,7 +174,7 @@ void signWinBinaries(Map args = [source: null, version: null, destDir: null, pro
 
 void createShaSum(Map args = [source: null, dest: null]) {
     stage('Compute SHA sum') {
-        sh "sha256sum ${args.source} | sed 's, .*/, ,' > ${args.dest}"
+        sh "sha256sum ${args.source} | sed 's, .*/,  ,' > ${args.dest}"
```
Laubi, line 177: Why is this double space needed here? What was the exact problem and why does this fix it?
The PR description is lacking in why it was broken and why this space fixes it. I have no clue.

jskelin: c/p from https://unix.stackexchange.com/a/310401/481532
Binary mode is indicated with ‘*’, text mode with ‘ ’ (space).

First space is separator, second indicates that it was read in text mode. Probably the binary mode would be better, but this works well

PR #950, opened 2023-04-07
Fix: Resolving classic environment URL does not work

warber:
The response payload of `/metadata/v1/classic-environment-domain` changed to conatin the field "domain" instead of "endpoint" which causes the resolution of the classic environment URL to fail.

### DIFF
--- a/pkg/client/metadata.go
+++ b/pkg/client/metadata.go
@@ -29,7 +29,18 @@ const classicEnvironmentDomainPath = "/platform/metadata/v1/classic-environment-
 const deprecatedClassicEnvDomainPath = "/platform/core/v1/environment-api-info"
 
 type classicEnvURL struct {
+	// Domain is the URL of the classic environment
+	Domain string `json:"domain"`
+	// Endpoint is the URL of the classic environment
+	// Note: newer environments return the classic environment URL in the Domain field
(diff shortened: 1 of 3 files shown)


pkg/client/metadata.go
```diff
@@ -29,7 +29,7 @@ const classicEnvironmentDomainPath = "/platform/metadata/v1/classic-environment-
 const deprecatedClassicEnvDomainPath = "/platform/core/v1/environment-api-info"
 
 type classicEnvURL struct {
-	Endpoint string `json:"endpoint"`
+	Domain string `json:"domain"`
```
UnseenWizzard, line 32: With what version does that change? 
And I guess it doesn't change for both the new and the depracted endpoint? 

I think we'd be best of, if we tried to read both Domain and Endpoint in the JSON and take what's set

UnseenWizzard (changes requested):

pkg/client/metadata.go
warber: done

warber: it's covered in TestGetDynatraceClassicEnvironmentFallsBackToDeprecatedPath

UnseenWizzard (approved):

PR #961, opened 2023-04-19
fix(deploy): Use numeric IDs for Management Zone Settings instead of ObjectID

UnseenWizzard:
As Management Zones are still mostly referenced via their numeric ID in other Settings, their
ID needs to be resolved down to the numeric ID rather than Settings Object ID in all cases.

Without this change, their object ID would potentially be templated into other configurations that
reference a management zone's ID.

This feature is behind a new 'MONACO_FEAT_USE_MZ_NUMERIC_ID' flag, which is active by default.

Testing on latest version DT environments, it was found that non-deprecated APIs (like dashboards)
already support MZ references via Settings Object ID, but this may be different for older envs.
Thus the resolving to numeric ID when deploying happens by default but can be turned off if needed
/in the future.

-

Depending on DT environment version nothing might change, or currently failing references to MZ IDs start working

### DIFF
--- a/pkg/client/dtclient/dummy_client.go
+++ b/pkg/client/dtclient/dummy_client.go
@@ -17,6 +17,7 @@
 package dtclient
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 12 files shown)


pkg/client/dtclient/dummy_client.go
```diff
@@ -246,8 +247,17 @@ func (c *DummyClient) ConfigExistsByName(a api.API, name string) (exists bool, i
 }
 
 func (c *DummyClient) UpsertSettings(obj SettingsObject) (DynatraceEntity, error) {
+
+	id := obj.Id
+
+	// to ensure decoding of Management Zone Numeric IDs works for dry-runs the dummy client needs to produce a fake but validly formated objectID
+	if obj.SchemaId == "builtin:management-zones" {
```
UnseenWizzard, line 254: note: Theoretically I think we could always do this and save the `if`, but I wasn't sure which option is better - be explicit about the MZ handling or keeping the dummy client simpler

pkg/client/dtclient/dummy_client.go
```diff
@@ -263,7 +273,7 @@ func (c *DummyClient) ListSettings(_ string, _ ListSettingsOptions) ([]DownloadS
 	return make([]DownloadSettingsObject, 0), nil
 }
 
-func (l *DummyClient) DeleteSettings(_ string) error {
+func (c *DummyClient) DeleteSettings(_ string) error {
```
UnseenWizzard, line 276: just brings the receiver names in line for all of the methods, don't know why this one was different

warber (approved):

PR #962, opened 2023-04-19
fix: Project folder is overwritten on monaco download

warber:
This PR adds code to randomize the project folder name _in case it already exists._ The project folder name is randomized using a **timestamp**, similar to what is done for the manifest file. (`project-name_TIMESTAMP`)
Tested with `--url` as well as `--manifest` option the output looks like:

image

* When downloading using `--url` or `--mananifest` using the same output folder (`-o`), the folder names are randomized if the folder already exists

### DIFF
--- a/pkg/download/download_writer.go
+++ b/pkg/download/download_writer.go
@@ -30,12 +30,12 @@ import (
 )
 
 type WriterContext struct {
-	EnvironmentUrl         string
-	ProjectToWrite         project.Project
-	Auth                   manifest.Auth
-	OutputFolder           string
-	ForceOverwriteManifest bool
-	timestampString        string
+	EnvironmentUrl  string
+	ProjectToWrite  project.Project
+	Auth            manifest.Auth
+	OutputFolder    string
+	ForceOverwrite  bool
+	timestampString string
 }
 
 func (c WriterContext) GetOutputFolderFilePath() string {
@@ -107,7 +107,7 @@ func getManifestFileName(fs afero.Fs, writerContext WriterContext) string {
 		return manifestFileName
 	}
 
-	if writerContext.ForceOverwriteManifest {
+	if writerContext.ForceOverwrite {
 		log.Info("Overwriting existing manifest.yaml in download target folder.")
 		return manifestFileName
 	}
@@ -125,7 +125,7 @@ func getProjectFolderName(fs afero.Fs, writerContext WriterContext) string {
 		return writerContext.ProjectToWrite.Id
 	}
 
-	if writerContext.ForceOverwriteManifest {
+	if writerContext.ForceOverwrite {
 		log.Info("Overwriting existing pojrect folder named %q in %q.", projectFolderName, outputFolder)
 		return projectFolderName
 	}
--- a/pkg/download/download_writer.go
+++ b/pkg/download/download_writer.go
@@ -53,11 +53,32 @@ func WriteToDisk(fs afero.Fs, writerContext WriterContext) error {
 }
 
 func writeToDisk(fs afero.Fs, writerContext WriterContext) error {
-
 	log.Debug("Preparing downloaded data for persisting")
 
-	manifestName := getManifestFilePath(fs, writerContext)
-	m := createManifest(writerContext)
+	manifestFileName := getManifestFileName(fs, writerContext)
+	projectFolderName := getProjectFolderName(fs, writerContext)
+
+	projectDefinition := manifest.ProjectDefinitionByProjectID{
+		writerContext.ProjectToWrite.Id: {
+			Name: writerContext.ProjectToWrite.Id,
+			Path: projectFolderName,
+		},
+	}
+
+	manifest := manifest.Manifest{
+		Projects: projectDefinition,
+		Environments: map[string]manifest.EnvironmentDefinition{
+			writerContext.ProjectToWrite.Id: {
+				Name: writerContext.ProjectToWrite.Id,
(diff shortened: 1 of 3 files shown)


cmd/monaco/download/download_configs.go
```diff
@@ -109,10 +109,6 @@ func (d DefaultCommand) DownloadConfigsBasedOnManifest(fs afero.Fs, cmdOptions d
 
 	printUploadToSameEnvironmentWarning(env)
 
-	if !cmdOptions.forceOverwrite {
```
UnseenWizzard, line 112: note that this was appending the environment name to the project, so `project_environment` not `project_project` as you described in the PR description. 

We shortly discussed on Monday to drop this (@Laubi may have created an extra bug/story?).
The only use I can see is if we allow downloading from several environments in the manifest we'd need different project names to identity each env. But we currently don't do that anyway

pkg/download/download_writer.go
```diff
@@ -53,21 +53,42 @@ func WriteToDisk(fs afero.Fs, writerContext WriterContext) error {
...
 func writeToDisk(fs afero.Fs, writerContext WriterContext) error {
-
 	log.Debug("Preparing downloaded data for persisting")
 
-	manifestName := getManifestFilePath(fs, writerContext)
-	m := createManifest(writerContext)
+	manifestFilePath := getManifestFileName(fs, writerContext)
+	projectFolderPath := getProjectFolderName(fs, writerContext)
+
+	projectDefinition := manifest.ProjectDefinitionByProjectID{
```
UnseenWizzard, line 61: any deeper reason to pull the manifest creation out of its own named function?

pkg/download/download_writer.go
```diff
@@ -53,21 +53,42 @@ func WriteToDisk(fs afero.Fs, writerContext WriterContext) error {
...
+	}
 
 	outputFolder := writerContext.GetOutputFolderFilePath()
 
 	log.Debug("Persisting downloaded configurations")
 	errs := writer.WriteToDisk(&writer.WriterContext{
 		Fs:              fs,
 		OutputDir:       outputFolder,
-		ManifestName:    manifestName,
+		ManifestName:    manifestFilePath,
```
UnseenWizzard, line 89: nit: should the `WriterContext.ManifestName` also be renamed? If it's meant to be filepath, `ManifestName` is the wrong name for it

pkg/download/download_writer.go
```diff
@@ -78,43 +99,38 @@ func writeToDisk(fs afero.Fs, writerContext WriterContext) error {
...
-				Name: wc.ProjectToWrite.Id,
-				URL: manifest.URLDefinition{
-					Type:  manifest.ValueURLType,
-					Value: wc.EnvironmentUrl,
-				},
-				Group: "default",
-				Auth:  wc.Auth,
-			},
-		},
+	if writerContext.ForceOverwriteManifest {
```
UnseenWizzard, line 128: nit: we should probably rename this to just `ForceOverwrite`

cmd/monaco/download/download_configs.go
warber: yep that's BS and was a wrong assumption from my side, will drop that commit

pkg/download/download_writer.go
warber: phu, no really **strong** argument.. I just didn't want to have the `getProjectFolderName()` call inside the `createManifest()` func and the other `getManifestFileName()` outside.... but both outside kind of outside consecutively. So you see immediately that the manifest file name and project folder name are kind of not static.

also `createManifest()` was basically just constructing that struct.. there is not much complexity to hide...

pkg/download/download_writer.go
warber: actually..... it's just the name of a file I guess :D so `manifestFilePath` is wrong and should be called `manifestFileName` same for `projectFolderPath` which is also just the name of the folder.

warber: that brings me to the quick conclusion that "ProjectDefinition.Path" should be renamed to "ProjectDefinition.FolderName" :D it's confusing :D

warber: updated the PR pls take a look.
e2e tests are also green now :)

pkg/download/download_writer.go
UnseenWizzard: makes sense :) Was really just wondering, not asking to put it back into it's own function

UnseenWizzard (approved):

UnseenWizzard:
For completeness/following up on this in the future, could you please update the PR description @warber ?
The `project_envrionment` thing was not dropped for manifests downloads now. 
And if I'm not mistaken, the project name will not always be randomized for direct download? The flag defaults to `project` and we just check if such a folder exists regardless of doing manifest or direct download

warber:
> And if I'm not mistaken, the project name will not always be randomized for direct download? The flag defaults to project and we just check if such a folder exists regardless of doing manifest or direct download
@UnseenWizzard  exactly.
if you run `monaco download --url -o some-folder` the first time it will just create a `manifest.yaml` and a `project` folder.
When you run the same command a second time it will create a `manifest_timestamp.yaml` and a `project_timestamp.yaml` folder. unless you provide the `-f` flag, then manifest as well as project folder will be overwritten...

For `monaco download --manifest -o some-folder`: the first time it will create a `manifest.yaml` and a `project_environment` folder. the second time it will create a `manifest_timestamp.yaml` and `project_enviroment_timestamp` folder......

PR #963, opened 2023-04-19
fix: Monaco 2.0 timing - required resource not created yet

warber:
This PR fixes the issue of deploying calculated metrics with dependencies to applications that are not ready yet.
Our current heuristics to catch such errors did not fire for that case.


UnseenWizzard (approved):
As discussed directly: It doesn't feel amazing to just keep adding cases to this, but let's discuss how to do this better after this issue is fixed/released

PR #968, opened 2023-04-21
Refactor: Unify `Downloader`s interface

warber:
Currently, we use `settings.Downloader` or `config.Downloader`, ... to handle downloading of the resp. configuration(s) from DT.
However, while they are very similar, they do not share a common interface. This means that adding yet another downloader (e.g. for automation resources) requires special logic as they cannot be treated uniformly.
This PR introduces a simple contract of how a downloader should be interfaced with: 

```
type Downloader[T v2.Type] interface {
	Download(projectName string, specificConfigs ...T) (projectv2.ConfigsPerType, error)
}
``` 

It would be easily possible but I did **not** touch the code/interface for entity download. as this is handled completely on its own in a separate command.

Further, the interface is (imo) not perfect, e.g. the `projecName` is only passed because it needs to be set on the returned config values. This could be done differently. However, that can be changed anytime. The more important aspect is, that they **do have** the same interface, however that may look like

### DIFF
--- a/cmd/monaco/download/downloaders.go
+++ b/cmd/monaco/download/downloaders.go
@@ -1,0 +1,41 @@
+/*
+ * @license
+ * Copyright 2023 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package download
+
+import (
+	v2 "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/download"
+)
+
+type downloaders []interface{}
+
+func (d downloaders) Classic() download.Downloader[v2.ClassicApiType] {
+	return getDownloader[v2.ClassicApiType](d)
+}
+
+func (d downloaders) Settings() download.Downloader[v2.SettingsType] {
+	return getDownloader[v2.SettingsType](d)
+}
+
+func getDownloader[T v2.Type](d downloaders) download.Downloader[T] {
+	for _, downloader := range d {
+		if dl, ok := downloader.(download.Downloader[T]); ok {
+			return dl
+		}
+	}
+	panic("No downloader implementation found")
+}
--- a/pkg/download/downloader.go
+++ b/pkg/download/downloader.go
@@ -1,0 +1,31 @@
+/*
+ * @license
+ * Copyright 2023 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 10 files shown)


cmd/monaco/download/downloaders.go
```diff
@@ -0,0 +1,44 @@
...
+
+// Downloaders is a map of downloaders per type of configuration
+type Downloaders map[v2.TypeId]Downloader
+
+// For returns the downloader for a given type of configuration or
+// panics if there is none
+func (d Downloaders) For(t v2.TypeId) Downloader {
+	if dl, ok := d[t]; !ok {
+		panic("No downloader implementation found for type " + t)
+	} else {
```
Laubi, line 41: Personally, I am not a fan of unnecessary 'else' branches.
I would just skip the keyword if they are not necessary. 
Absolut nitpick, no need to fix it here

cmd/monaco/download/downloaders.go
```diff
@@ -0,0 +1,44 @@
...
+	projectv2 "github.com/dynatrace/dynatrace-configuration-as-code/pkg/project/v2"
+)
+
+// Downloader represents a component that is responsible for downloading configuration for a given project from Dynatrace
+type Downloader interface {
+
+	// Download downloads configurations from a Dynatrace environment.
+	// If only projectName is given, it will download all configuration.
+	// If additionally specific configuration names/types are given, then it will only download those
+	Download(projectName string, specificConfigs ...string) (projectv2.ConfigsPerType, error)
```
Laubi, line 30: Would it make sense to use the `...config.Type` here instead? E.g. ClassicApiType, SettingsType, and others in the future 🤔
Sure we would need to map the strings to the correct object, but it could be clearer and keep more information in case we need it in the future

pkg/download/settings/settings.go
```diff
@@ -64,41 +66,41 @@ func NewSettingsDownloader(client dtclient.SettingsClient, opts ...func(*Downloa
...
-func DownloadAll(client dtclient.SettingsClient, projectName string) v2.ConfigsPerType {
-	return NewSettingsDownloader(client).DownloadAll(projectName)
-}
-
-// Download downloads all settings 2.0 objects for the given schema IDs and a given project
-// The returned value is a map of settings 2.0 objects with the schema ID as keys
-func (d *Downloader) Download(schemaIDs []string, projectName string) v2.ConfigsPerType {
-	return d.download(schemaIDs, projectName)
+func (d *Downloader) Download(projectName string, schemaIDs ...string) (v2.ConfigsPerType, error) {
+	if schemaIDs == nil || len(schemaIDs) == 0 {
```
Laubi, line 70: nitpick:  `len(schemaIDs) == 0`  is enough here as `len(nil) = 0`

cmd/monaco/download/downloaders.go
Laubi, line 1: I would argue that the interface `Downloader` does not belong into the `cmd` pkg, but rather the `pkg/download` pkg.

cmd/monaco/download/downloaders.go
warber: :thinking:  i like that idea

Laubi: What I ment was the other side, we need to map the api-string to the ApiType, but also nothing too bad

cmd/monaco/download/downloaders.go
warber: Yeah, I would be a bit worried if there a fans of unnecessary "else" branches out there :-O :smiley:

Laubi (approved):

pkg/download/downloader.go
```diff
@@ -0,0 +1,31 @@
...
+
+package download
+
+import (
+	v2 "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2"
+	projectv2 "github.com/dynatrace/dynatrace-configuration-as-code/pkg/project/v2"
+)
+
+// Downloader represents a component that is responsible for downloading configuration for a given project from Dynatrace
+type Downloader[T v2.Type] interface {
```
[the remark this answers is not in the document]
warber: yeah, I had it in `cmd/download` first, but after review comment of @Laubi  I've moved it to `pkg/download`. I didn't thought too much about it tho tbh.

PR #972, opened 2023-04-25
feat(download): Allow switching between fast and basic dependency resolvers

UnseenWizzard:

It was found that the currently used 'fast' dependency resolution is using excessive memory.

As a quick fix for customers running into OOM errors in some cases, this change makes it possible to switch back to the old naive implementation which is more CPU intensive and thus slower, but has a much lower memory footprint.

Control of wich implementation is used is done via a new MONACO_FEAT_USE_FAST_DEPENDENCY_RESOLUTION feature flag.

This is somewhat of a band-aid fix, until a resolution algorithm/implemementation is found that combines acceptable memory and CPU usage.

Per default nothing changes, but the new Feature Flag allows users to switch to the slow but memory friendly dependency resolution.


pkg/download/dependency_resolution/resolver/basic_dep_resolver.go
```diff
@@ -0,0 +1,81 @@
...
+
+func BasicResolver(configsById map[string]config.Config) basicResolver {
+	return basicResolver{
+		configsById: configsById,
+	}
+}
+
+func (r basicResolver) ResolveDependencyReferences(configToBeUpdated *config.Config) {
+	resolveScope(configToBeUpdated, r.configsById)
+	resolveTemplateB(configToBeUpdated, r.configsById)
```
warber, line 40:
```suggestion
	resolveTemplate(configToBeUpdated, r.configsById)
```

UnseenWizzard: I've lazily put both resolvers in the same package for now, so while they share the `resolveScope` they each implement a`resolveTemplate` which would overlap.

I'm not really happy with the package structure, but I didn't want to introduce several packages with a single file each

warber: ah ok looked like a typo (left-over) from something to experiment with... anyway , ok i guess :man_shrugging:

warber (approved):

pkg/download/dependency_resolution/resolver/basic_dep_resolver.go
UnseenWizzard: I've renamed it to something more intentionally different :)

Laubi (approved):

PR #973, opened 2023-04-25
feat: Download Automation API resources

warber:
This PR adds a `Downloader` that makes use of the automation API client for downloading automation resources.
Feature flag for automation resources needs to be enabled.

This PR does not include special handling of escaping templating inside workflow payloads.

If the user enables the feature flag for automation resources, monaco will go and try to download all (supported types) of automation resources (workflows, business calendars, scheduling rules)

image
image

### DIFF
--- a/cmd/monaco/download/downloaders.go
+++ b/cmd/monaco/download/downloaders.go
@@ -44,7 +44,7 @@ func makeDownloaders(options downloadConfigsOptions) (downloaders, error) {
 		ClientID:     options.auth.OAuth.ClientID.Value,
 		ClientSecret: options.auth.OAuth.ClientSecret.Value,
 		TokenURL:     options.auth.OAuth.GetTokenEndpointValue(),
-	}))
+	}), automation.WithClientRequestLimiter(concurrency.NewLimiter(environment.GetEnvValueIntLog(environment.ConcurrentRequestsEnvKey))))
 
 	return downloaders{settings.NewDownloader(dtClient), classic.NewDownloader(dtClient), dlautomation.NewDownloader(autClient)}, nil
 }
--- a/cmd/monaco/download/downloaders.go
+++ b/cmd/monaco/download/downloaders.go
@@ -39,14 +39,18 @@ func makeDownloaders(options downloadConfigsOptions) (downloaders, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	autClient := automation.NewClient(options.environmentURL, client.NewOAuthClient(context.TODO(), client.OauthCredentials{
-		ClientID:     options.auth.OAuth.ClientID.Value,
-		ClientSecret: options.auth.OAuth.ClientSecret.Value,
-		TokenURL:     options.auth.OAuth.GetTokenEndpointValue(),
-	}), automation.WithClientRequestLimiter(concurrency.NewLimiter(environment.GetEnvValueIntLog(environment.ConcurrentRequestsEnvKey))))
-
-	return downloaders{settings.NewDownloader(dtClient), classic.NewDownloader(dtClient), dlautomation.NewDownloader(autClient)}, nil
+	var automationDownloader download.Downloader[v2.AutomationType] = dlautomation.NoopAutomationDownloader{}
+	if options.auth.OAuth != nil {
+		autClient := automation.NewClient(options.environmentURL, client.NewOAuthClient(context.TODO(), client.OauthCredentials{
+			ClientID:     options.auth.OAuth.ClientID.Value,
+			ClientSecret: options.auth.OAuth.ClientSecret.Value,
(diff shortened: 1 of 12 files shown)


pkg/download/automation/automation.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/client/automation"
+	config "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/coordinate"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/parameter"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/parameter/value"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/template"
+	v2 "github.com/dynatrace/dynatrace-configuration-as-code/pkg/project/v2"
+)
+
+var automationTypesToResources = map[config.AutomationType]automation.ResourceType{
```
warber, line 32: pkg/config and pkg/download/automation do not share the same type for resources, so we need a mapping

UnseenWizzard: this mapping is from pkg/config -> pkg/client/automation though right? As they're used right in the client calls

nitpick/maybe-bad suggestion: maybe importing `pkg/client/automation` as `automationClient` would make the distinction to this automation pkg clearer

pkg/download/automation/automation.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		// nothing to do if there were no configs
+		// continue with next resource
+		if len(response.Results) == 0 {
+			continue
+		}
+		log.Info("Downloaded %d objects for automation resource %s", len(response.Results), string(at.Resource))
+
+		// convert and collect all objects into a slice of configs.Config values to return
```
UnseenWizzard, line 80: nit: I'm not a big fan of these "general description comments" (specifically this one and line 59) - if the functionality is self-contained and needs a description, I'd prefer if you'd pull it into a function with a speaking name. (IMO might make sense here, for line 59 I'd personally drop the comment, downloading is exactly what I expect the automation.Download method to do)

pkg/download/automation/automation.go
```diff
@@ -0,0 +1,111 @@
...
+		var configs []config.Config
+		for _, obj := range response.Results {
+			configId := obj.Id
+			var content string
+			if indentedData, err := json.MarshalIndent(json.RawMessage(obj.Data), "", "  "); err == nil {
```
UnseenWizzard, line 85: nit: should we pull that marshal call into an internal util? I think every downloader uses it, and they all define 2 space indent and would have to change together if this ever changes

cmd/monaco/download/downloaders.go
```diff
@@ -17,12 +17,38 @@
...
 type downloaders []interface{}
 
+func makeDownloaders(options downloadConfigsOptions) (downloaders, error) {
+	dtClient, err := dynatrace.CreateClient(options.environmentURL, options.auth, false,
+		dtclient.WithClientRequestLimiter(concurrency.NewLimiter(environment.GetEnvValueIntLog(environment.ConcurrentRequestsEnvKey))))
+	if err != nil {
+		return nil, err
+	}
+
+	autClient := automation.NewClient(options.environmentURL, client.NewOAuthClient(context.TODO(), client.OauthCredentials{
```
UnseenWizzard, line 43: There is no concurrency limiter on the automation client?

UnseenWizzard (commented):
Please look into handling errors the same way as other downloaders (I might be wrong in thinking they are more lenient though!), and the use / introduction of the same Limiting for concurrent calls as the dtClient (unless handled differently in automation client, didn't yet look past the changes in this PR)

Otherwise all nitpicks :)

pkg/download/automation/automation.go
```diff
@@ -0,0 +1,111 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		// nothing to do if there were no configs
+		// continue with next resource
+		if len(response.Results) == 0 {
+			continue
+		}
+		log.Info("Downloaded %d objects for automation resource %s", len(response.Results), string(at.Resource))
```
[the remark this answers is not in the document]
UnseenWizzard: Personally as a user I'd like to get an indication that it tried to download but there was nothing - but I may be wrong about this being helpful

cmd/monaco/download/downloaders.go
warber: i will just add it , its easy

warber:
@UnseenWizzard I've adressed all your comments in one commit. pls take a look :v:

PR #960, opened 2023-04-19
Feat/automation/deploy logic

jskelin:

PR #974, opened 2023-04-26
feat: Upsert automation API resources

warber:
This PR adds the necessary code to be able to deploy (upsert) Automation Resources using the automation client

Parts of the logic for this PR were taken from [here](https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/960) and incorporated into this PR.

Further, also the e2e test resources are included in this PR.

Given a downloaded automation config, e.g. workflow:
image

The user is now able to deploy it to the dt environment:

image


pkg/deploy/automation.go
```diff
@@ -0,0 +1,82 @@
...
+		resp, err = client.Upsert(automation.SchedulingRules, id, []byte(renderedConfig))
+	default:
+		err = fmt.Errorf("unkonwn rsource type %q", t.Resource)
+	}
+
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	if resp != nil {
```
UnseenWizzard, line 70: if the response was `nil` but we got no error from the Upsert, shouldn't we explicitly return an error from this method?

As is, we'll just return nil and any previous errors (line 81) in that case

pkg/deploy/automation_test.go
```diff
@@ -0,0 +1,125 @@
...
+	config "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/coordinate"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/parameter"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/project/v2/topologysort"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestDeployAotmationShouldFailRenderTemplate(t *testing.T) {
```
UnseenWizzard, line 33: This fails because the config's `Type` is a classic Config API, not because of the "faulty template", right? 
There is no rendering happening inside the `deployAutomation` call - it already receives the rendered template (in this case as empty string)

pkg/deploy/clientset.go
```diff
@@ -0,0 +1,90 @@
...
+		dtClient:  dtClient,
+		autClient: autClient,
+	}
+}
+
+func CreateClientSet(url string, auth manifest.Auth, dryRun bool, opts ...func(dynatraceClient *dtclient.DynatraceClient)) (*clientSet, error) {
+	var dtClient dtclient.Client
+	var autClient automationClient
+	var err error
+	if dryRun {
```
UnseenWizzard, line 58: this `if` is redundant to the `case dryRun` on line 64, isn't it?

pkg/deploy/clientset.go
```diff
@@ -0,0 +1,90 @@
...
+			dtClient:  &dtclient.DummyClient{},
+			autClient: &dummyAutomationClient{}}, nil
+	}
+	switch {
+	case dryRun:
+		dtClient = &dtclient.DummyClient{}
+		autClient = &dummyAutomationClient{}
+	case auth.OAuth == nil:
+		dtClient, err = dtclient.NewClassicClient(url, auth.Token.Value, opts...)
+		autClient = &dummyAutomationClient{}
```
UnseenWizzard, line 69: what happens if the `autClient` is nil in the clientSet in this case?

The base branch introduced early validation whether the configs match the environment (https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/960/files#diff-3616826a0b04f4d0646286809b7d1f28c7e2dfce0d31091ef554181709888060R123-R127) 
I think this is missing from this version? 

We should tell the user as early as possible if their config can't be deployed to the current environment (considering whether a config is being deployed or skipped of course).
And if we do that, I don't think we should ever get into cases where deployment of Workflows to a Classic env would even be attempted and the `autClient` is being called?

pkg/deploy/clientset.go
```diff
@@ -0,0 +1,90 @@
...
+	case auth.OAuth != nil:
+		oauthCredentials := client.OauthCredentials{
+			ClientID:     auth.OAuth.ClientID.Value,
+			ClientSecret: auth.OAuth.ClientSecret.Value,
+			TokenURL:     auth.OAuth.GetTokenEndpointValue(),
+		}
+		dtClient, err = dtclient.NewPlatformClient(url, auth.Token.Value, oauthCredentials, opts...)
+		autClient = automation.NewClient(url, client.NewOAuthClient(context.TODO(), oauthCredentials))
+	default:
+		return nil, fmt.Errorf("unable to create authorizing HTTP Client for environment %s - no oauth credentials given", url)
```
UnseenWizzard, line 79: I think this message is at least partially wrong - "no oauth credentials given" would be the `case auth.OAuth == nil` - if we reach the default case something else went wrong

pkg/deploy/deploy.go
```diff
@@ -87,65 +87,19 @@ func deploy(client dtclient.Client, apis api.APIs, em *entityMap, c *config.Conf
 
 	case config.SettingsType:
 		log.Info("\tDeploying config %s", c.Coordinate)
-		return deploySetting(client, properties, renderedConfig, c)
+		return deploySetting(clientSet.settings(), properties, renderedConfig, c)
 
 	case config.ClassicApiType:
-		log.Info("\tDeploying config %s", c.Coordinate)
-		return deployConfig(client, apis, em, properties, renderedConfig, c)
+		return deployConfig(clientSet.classic(), apis, em, properties, renderedConfig, c)
```
UnseenWizzard, line 93: This one is missing the info log that the other cases have - I think that was accidental, because it had it before.

UnseenWizzard, line 93: nit: should we rename `deployConfig` to `deployClassicConfig` ? 
It could clear up the naming overlap with `config` used for "monaco configuration object" (e.g. in the logs), and be in line with the "classic" naming of the ApiType, client and deploy/classic.go

pkg/deploy/deploy.go
```diff
@@ -87,65 +87,19 @@ func deploy(client dtclient.Client, apis api.APIs, em *entityMap, c *config.Conf
...
-
-	return &parameter.ResolvedEntity{
-		EntityName: entity.Name,
-		Coordinate: conf.Coordinate,
-		Properties: properties,
-		Skip:       false,
-	}, nil
-}
-
 func upsertNonUniqueNameConfig(client dtclient.ConfigClient, apiToDeploy api.API, conf *config.Config, configName string, renderedConfig string) (dtclient.DynatraceEntity, error) {
```
UnseenWizzard, line 103: this method is Config API / ClassicAPIType specific -> it should also move to the new deploy/classic.go file :)

pkg/deploy/deploy.go
```diff
@@ -160,52 +114,6 @@ func upsertNonUniqueNameConfig(client dtclient.ConfigClient, apiToDeploy api.API
...
-	return &parameter.ResolvedEntity{
-		EntityName: name,
-		Coordinate: c.Coordinate,
-		Properties: properties,
-		Skip:       false,
-	}, nil
-
-}
-
 func getEntityID(c *config.Config, e dtclient.DynatraceEntity) (string, error) {
```
UnseenWizzard, line 117: this method and `extractScope` are Settings specific -> they should move to deploy/settings.go as well

pkg/deploy/clientset.go
warber: incorportated check from other PR

cmd/monaco/deploy/deploy.go
```diff
@@ -417,6 +402,32 @@ func toProjectMap(projects []project.Project) map[string]project.Project {
...
+		if err := checkConfigsForEnvironment(envs[envName], cfgs); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func checkConfigsForEnvironment(env manifest.EnvironmentDefinition, cfgs []config.Config) error {
+	for i := range cfgs {
+		if onlyAvailableOnPlatform(&cfgs[i]) && !platformEnvironment(env) && !cfgs[i].Skip {
```
UnseenWizzard, line 266: advanced nit: assuming that the "onlyAvailableOnPlatform" check will grow in complexity in the future, you could lead with the `!cfgs[i].Skip` and save on the more costly checks if the config is skipped anyway

cmd/monaco/integrationtest/cleanup.go
```diff
@@ -34,6 +34,7 @@ import (
 // Deletes all configs that end with a test suffix and any Settings created by the given manifest
 func CleanupIntegrationTest(t *testing.T, fs afero.Fs, manifestPath string, loadedManifest manifest.Manifest, suffix string) {
 
+	t.Setenv("MONACO_FEAT_AUTOMATION_RESOURCES", "1") // i don't know why we need to set it here (again)
```
UnseenWizzard, line 37: just wondering, if this env var isn't set on cleanup anymore, isn't the test suffix one that is used in some names missing as well?

UnseenWizzard (approved):

UnseenWizzard (changes requested):
just saw that the e2e test configs still all define an originObjectId -> please remove some of them to also include the case of generating a new ID in the e2e tests

cmd/monaco/integrationtest/cleanup.go
warber: nevermind, I just saw the issue was setting the env variable too late :man_facepalming:

cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/scheduling-rule/config.yaml
```diff
@@ -0,0 +1,91 @@
...
+        configId: 355eadf6-d803-4c71-97a3-7ae95b5c7d74
+        configType: business-calendar
+        property: id
+        type: reference
+    template: d71c9f00-60f6-4172-9fbe-0656a93cbceb.json
+    skip:
+      type: environment
+      name: SKIP_AUTOMATION_CONFIGS
+      default: true
+    originObjectId: d71c9f00-60f6-4172-9fbe-0656a93cbceb
```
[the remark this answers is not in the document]
warber: As agreed, let's do that after implementing proper cleanup of automation resources in integration tests. which will happen when implementing the "delete automation resources" story.

PR #977, opened 2023-04-26
feat(deploy): Log number of projects, environments and (on debug) configs to deploy

UnseenWizzard:
feat(deploy): Log number of projects, environments and (on debug) configs to deploy

It is sometimes helpful to get a quick overview of the number of projects and configurations
that are being deployed by just looking at the logs.

As collecting the number of configs requires counting them by running through all configs
per project, environment and type, this is limited to only happen when debug logging is
active.

- A simple way to check if debug is even currently active is added to our logger in it's own commit.
- Counting configs could be done elsewhere, e.g. while loading or sorting - this was just a quick addition on the side, if we think this is helpful but should be done more efficiently without extra iteration I'm happy to add counts to existing methods (which then get a bit more complex by keeping a count they don't need.

- INFO logs about projects and environments will now include the number of them
- DEBUG logs will include the total number of configs

### DIFF
--- a/cmd/monaco/deploy/deploy.go
+++ b/cmd/monaco/deploy/deploy.go
@@ -191,14 +191,30 @@ func sortConfigs(projects []project.Project, environmentNames []string) (project
 }
 
 func logProjectsInfo(projects []project.Project) {
-	log.Info("Projects to be deployed:")
+	log.Info("Projects to be deployed (%d):", len(projects))
 	for _, p := range projects {
 		log.Info("  - %s", p)
 	}
+
+	if log.DebugEnabled() {
+		logConfigInfo(projects)
+	}
+}
+
+func logConfigInfo(projects []project.Project) {
+	cfgCount := 0
+	for _, p := range projects {
+		for _, cfgsPerTypePerEnv := range p.Configs {
+			for _, cfgsPerType := range cfgsPerTypePerEnv {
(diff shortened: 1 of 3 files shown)


cmd/monaco/deploy/deploy.go
```diff
@@ -191,14 +191,26 @@ func sortConfigs(projects []project.Project, environmentNames []string) (project
 }
 
 func logProjectsInfo(projects []project.Project) {
-	log.Info("Projects to be deployed:")
+	log.Info("Projects to be deployed (%d):", len(projects))
 	for _, p := range projects {
 		log.Info("  - %s", p)
 	}
+
+	if log.DebugEnabled() {
```
Laubi, line 199: `Run Static Code Analysis / lint (pull_request) Successful in 1m`
How can it be successful, if here are obvious errors 🤔 

SA9003: empty branch (staticcheck)

UnseenWizzard: ....

UnseenWizzard: fixed the issue.

I think the reporting error is an problem of the review dog action we use, which wraps the linter and posts on the PR - but it looks like this one runs "successfully" regardless of the linter's status

Laubi (approved):

PR #978, opened 2023-04-26
Add References e2e tests

Laubi:

Adds e2e tests for references between configs & settings in all permutations


cmd/monaco/integrationtest/v2/all_configs_integration_test.go
```diff
@@ -45,7 +45,7 @@ func runAllConfigsTest(t *testing.T, specificEnvironment string) {
 	configFolder := "test-resources/integration-all-configs/"
 	manifest := configFolder + "manifest.yaml"
 
-	RunIntegrationWithCleanup(t, configFolder, manifest, specificEnvironment, "AllConfigs", func(fs afero.Fs) {
+	RunIntegrationWithCleanup(t, configFolder, manifest, specificEnvironment, "AllConfigs", func(fs afero.Fs, _ TestContext) {
```
UnseenWizzard, line 48: If you've added the context for the sake of simplifying the signature if new things are added, why add it only as an input to the test function? 

Why not make it the base input of the integration test method itself and also pass it to the test function? 
Currently it just duplicates the suffix and I explicitly have to pass it twice if needed.

why not something like:
```go
type TestContext struct {
   folder string
   manifest manifest.Manifest
   specificEnv string
   uniqueSuffix string
}

RunIntegrationWithCleanup(t, ctx, func(fs, ctx)
```

Or in future steps even replacing the whole "WithX" logic from the context? e.g. if an fs is set -> onGivenFs; cleanup: true/false, ...

UnseenWizzard: nevermind, just saw you did just that in the last commit from 45min ago :D nice 👌

UnseenWizzard: I just saw you partially do that with the latest commit introducing the test options struct.
Two questions to that still: 
* why make it different to the context passed to the test function? Seems extra effort if future test funcs need more details, and IMO not worth the separation of concerns here.
* why not pull the test options up to the exported method level and replace the named specific methods completely?

cmd/monaco/integrationtest/v2/references_test.go
```diff
@@ -0,0 +1,245 @@
...
+				managementZone := findConfig(t, confsPerType, "management-zone", "zone_"+ctx.suffix)
+				profile := findSetting(t, confsPerType, "builtin:alerting.profile", "profile_"+ctx.suffix, "name")
+				notification := findSetting(t, confsPerType, "builtin:problem.notifications", "notification_"+ctx.suffix, "displayName")
+
+				assertRefParamFromTo(t, profile, managementZone)
+				assertRefParamFromTo(t, notification, profile)
+			},
+		},
+		{
+			project: "config-with-mngt-zone",
```
UnseenWizzard, line 92: nit: `classic-with-settings-mngt-zone` would better fit the naming scheme you've applied to the other cases and make it clearer what this case does

cmd/monaco/integrationtest/v2/references_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+				RunIntegrationWithCleanupOnGivenFs(t, fs, configFolder, manifestFile, env, testName, func(fs afero.Fs, ctx TestContext) {
+
+					// upsert
+					cmd := runner.BuildCli(fs)
+					cmd.SetArgs([]string{"deploy", "-v", manifestFile, "--environment", env, "--project", proj})
+					err := cmd.Execute()
+					assert.Nil(t, err, "create: did not expect error")
+
+					// update just to be sure
```
UnseenWizzard, line 127: is that step really needed? If the tests fail without it, something's wrong with upserting.
If it's to ensure that the API has registered and replies with everything that was just created, why not just sleep instead?

cmd/monaco/integrationtest/v2/references_test.go
```diff
@@ -0,0 +1,245 @@
...
+				notification := findConfig(t, confsPerType, "notification", "notification_"+ctx.suffix)
+
+				assertRefParamFromTo(t, profile, managementZone)
+				assertRefParamFromTo(t, notification, profile)
+			},
+		},
+	}
+
+	for _, env := range envs {
+		for _, tc := range tests {
```
UnseenWizzard, line 109: nit: any reason to deviate from the usual `tt` naming of the test case var? (other than `tc` actually sounding more correct, just off idiom

cmd/monaco/integrationtest/v2/references_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+					tc.validate(t, ctx, confsPerType)
+				})
+			})
+		}
+	}
+}
+
+func assertRefParamFromTo(t *testing.T, from config.Config, to config.Config) {
+	name := paramName(to.Coordinate.Type, to.Coordinate.ConfigId)
```
UnseenWizzard, line 175: relying exactly on how the parameter reference ID is formated on download seems brittle.
Couldn't you achieve virtually the same with just a 

```go
assert.Contains(t, from.References(), to.Coordinate)
```

? (leaving out the check whether it really references the ID rather than another property, but I'm not sure that level of details is even needed for this E2E test)

cmd/monaco/integrationtest/v2/all_configs_integration_test.go
Laubi: I wanted to keep the changes rather small if possible, so that the PR is easy to review.

I had a local change where I also inlined and renamed all runIntegrationX tests since they do not need to be exported.
I am happy to propose a follow-up PR that does exactly that, inlining and calling only one leftover runIntegration method.

However, some points that should be discussed:
W_hy I used options instead of WithX_: I had a local poc with the WithX, but our approach to the test cases is not nice for that. WithX would need to be the last argument (varadic), but then we would have a huge method in between. Refactoring the method into a var and passing it would be an option, but I do not like it personally. An options struct is here IMO the better alternative.

_Why different context_: I will not die on this hill, but the separation between initialization and execution. The suffix we pass in is not the suffix used in the test. We can discuss this offline if it makes sense to combine those.

_Why not pull the test options up to the exported method level and replace the named specific methods completely_: Stated above, smaller PR. Happy to create a new one that refactors exactly that.

cmd/monaco/integrationtest/v2/references_test.go
Laubi: Good observation, like it

cmd/monaco/integrationtest/v2/all_configs_integration_test.go
UnseenWizzard: Makes sense to me, thanks for the extensive answer :) 

Let's discuss in the team, if and how exactly we continue refactoring 👍

UnseenWizzard (approved):

PR #979, opened 2023-04-27
ci: Publish containers on docker-hub

Laubi:
This change introduces releasing containers on docker-hub, and also setting the latest-tag.

First of all, we try to find what kind of release this is, either full-, rc-, or dev-release. If it is neither, we fail the build to restrict the possible tags.

Then we release monaco to docker for full- and rc-releases. Additionally, we publish the latest tag for those releases.

Furthermore, we push the latest-tag also for our internal releases, but only if it's not a dev-release.


.ci/releasePipeline.groovy
```diff
@@ -112,7 +130,22 @@ pipeline {
...
                                             case "container":
-                                                createContainerAndPushToStorage(version: version, registrySecretsPath: 'keptn-jenkins/monaco/registry-deploy')
+                                                // docker hub, only full releases and rc-releases
+                                                if (!isDevRelease) {
+                                                    createContainerAndPushToStorage(version: version, registrySecretsPath: 'keptn-jenkins/monaco/dockerhub-deploy', registry: 'Docker')
+
+                                                    // push latest to full releases and rc-releases
+                                                    createContainerAndPushToStorage(version: 'latest', registrySecretsPath: 'keptn-jenkins/monaco/dockerhub-deploy', registry: 'Docker')
+                                                } else {
+                                                    echo "Skipping release to docker"
```
UnseenWizzard, line 140:
```suggestion
                                                    echo "Skipping release of ${version} to docker hub"
```
maybe more informative in the build logs

.ci/releasePipeline.groovy
```diff
@@ -26,11 +32,23 @@ pipeline {
...
+                            isDevRelease = devRegex.matcher(version).matches()
+
                             echo "version= ${version}"
+                            echo "isFullRelease=${isFullRelease}"
+                            echo "isRcRelease=${isRcRelease}"
+                            echo "isDevRelease=${isDevRelease}"
                             echo "GitHub releaseId= ${releaseId}"
+
+                            if (!isFullRelease && !isRcRelease && !isDevRelease) {
+                                error('Given version tag is not a valid tag. Use v1.2.3, v1.2.3-rc*, or v1.2.3-dev*')
```
UnseenWizzard, line 50: did you try this case out with a test tag?

.ci/releasePipeline.groovy
Laubi: The release is printed in other properties, but we can add it here as well.

.ci/releasePipeline.groovy
Laubi: Yes, and it aborts as expected

.ci/releasePipeline.groovy
```diff
@@ -112,7 +130,22 @@ pipeline {
...
                                             case "container":
-                                                createContainerAndPushToStorage(version: version, registrySecretsPath: 'keptn-jenkins/monaco/registry-deploy')
+                                                // docker hub, only full releases and rc-releases
+                                                if (!isDevRelease) {
+                                                    createContainerAndPushToStorage(version: version, registrySecretsPath: 'keptn-jenkins/monaco/dockerhub-deploy', registry: 'Docker')
+
+                                                    // push latest to full releases and rc-releases
+                                                    createContainerAndPushToStorage(version: 'latest', registrySecretsPath: 'keptn-jenkins/monaco/dockerhub-deploy', registry: 'Docker')
+                                                } else {
+                                                    echo "Skipping release of ${version} to docker"
```
UnseenWizzard, line 140: my improvement point was actually more about "docker hub" rather than "docker", also logging the version was just a bonus :D

UnseenWizzard (approved):

PR #980, opened 2023-04-27
feat(download): Sanitize downloaded Automation configurations

UnseenWizzard:
Like existing types, Automation payloads contain data that is not needed on upload and just clutters config-as-code files - thus a simple sanitization is added. As these APIs are fewer and more consisten than Classic Config APIs, the full fledged sanitizer code from their download is not reused.

Additionally the 'title' field which all currently supported Automation types contain, is extracted as the config's 'name' if it is found.

To make this resilient, any failure in extraction is just treated as before and will persist the payload as is.

### DIFF
--- a/pkg/download/automation/automation.go
+++ b/pkg/download/automation/automation.go
@@ -17,6 +17,8 @@
 package automation
 
 import (
+	"encoding/json"
+	"fmt"
 	jsonutils "github.com/dynatrace/dynatrace-configuration-as-code/internal/json"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
 	automationClient "github.com/dynatrace/dynatrace-configuration-as-code/pkg/client/automation"
@@ -74,10 +76,12 @@ func (d *Downloader) Download(projectName string, automationTypes ...config.Auto
 
 		var configs []config.Config
 		for _, obj := range response.Results {
+
 			configId := obj.Id
-			content, _ := jsonutils.MarshalIndent(obj.Data)
+			t, configName := createTemplateFromRawJSON(obj, string(at.Resource))
+
 			c := config.Config{
-				Template: template.NewDownloadTemplate(configId, configId, string(content)),
+				Template: t,
 				Coordinate: coordinate.Coordinate{
 					Project:  projectName,
 					Type:     string(at.Resource),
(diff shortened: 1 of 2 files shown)


pkg/download/automation/automation.go
```diff
@@ -101,6 +105,41 @@ func (d *Downloader) Download(projectName string, automationTypes ...config.Auto
 type NoopAutomationDownloader struct {
 }
 
+func createTemplateFromRawJSON(obj automationClient.Response) (t template.Template, extractedName string) {
+	configId := obj.Id
+
+	var data map[string]interface{}
+	err := json.Unmarshal(obj.Data, &data)
+	if err != nil {
+		log.Warn("Failed to sanitize downloaded JSON for config %v - template may need manual cleanup: %v", configId, err)
```
Laubi, line 114: nitpick: Might be beneficial to show what type the config has

pkg/download/automation/automation.go
```diff
@@ -101,6 +105,41 @@ func (d *Downloader) Download(projectName string, automationTypes ...config.Auto
...
+	// remove properties not necessary for upload
+	delete(data, "id")
+	delete(data, "modificationInfo")
+	delete(data, "lastExecution")
+
+	// extract 'title' as name
+	configName := configId
+	if title, ok := data["title"]; ok {
+		configName = fmt.Sprintf("%v", title)
+		data["title"] = "{{.name}}"
```
Laubi, line 127: Shouldn't this happen anyway? If it's not there, should we still set it?

pkg/download/automation/automation.go
```diff
@@ -101,6 +105,41 @@ func (d *Downloader) Download(projectName string, automationTypes ...config.Auto
...
+	}
+
+	var content []byte
+	if modifiedJson, err := json.Marshal(data); err == nil {
+		content = modifiedJson
+	} else {
+		log.Warn("Failed to sanitize downloaded JSON for config %v - template may need manual cleanup: %v", configId, err)
+		content = obj.Data
+	}
+	content, _ = jsonutils.MarshalIndent(content)
```
Laubi, line 137: Why do we even return the error if we don't use it? It's only used in tests :/

Laubi (approved):

pkg/download/automation/automation.go
UnseenWizzard: Hmm.. I think on download it would be most helpful to tell the user the actual filepath - but I'm not sure we can fully do this at this point. I'll take a look though

UnseenWizzard: added the type

pkg/download/automation/automation.go
UnseenWizzard: won't change as discussed

pkg/download/automation/automation.go
UnseenWizzard: Possibly removing the error return for MarshalIndent (or reacting to it in places where we're not sure marshalling will work) to be done in unrelated followup PR

warber (approved):

Issue #982, opened 2023-05-02
Change user output with term `upsert` to `insert or update`


johannes-b:
**Describe the bug**
While the term `upsert` is very technical, it may cause confusion when used in output messages. For example: 

```
2023/04/28 14:08:01 INFO  Projects to be deployed:
2023/04/28 14:08:01 INFO    - srg-config-as-code
2023/04/28 14:08:01 INFO  Environments to deploy to:
2023/04/28 14:08:01 INFO    - pvy27067
2023/04/28 14:08:01 INFO  Deploying configurations to environment `pvy27067`...
2023/04/28 14:08:01 INFO        Deploying config srg-config-as-code:app:dynatrace.site.reliability.guardian:guardians:srg
2023/04/28 14:08:02 WARN  Failed to upsert config "srg". Waiting for 5s before retrying...
2023/04/28 14:08:07 WARN  Failed to upsert config "srg". Waiting for 5s before retrying...
2023/04/28 14:08:12 WARN  Failed to upsert config "srg". Waiting for 5s before retrying...
```

Note: Definition of term: 
```
The term upsert is a portmanteau – a combination of the words “update” and “insert.” In the context of relational databases, an upsert is a database operation that will update an existing row if a specified value already exists in a table, and insert a new row if the specified value doesn't already exist.
```

**Expected behavior**
Please change the output messages from `upsert` to `insert or update`.

**Additional context**

UnseenWizzard:
wording changed as of 2.0.3

PR #984, opened 2023-05-02
chore: Don't use "upsert" term in error messages

warber:
#982
PR contains also a bit of cleanup/housekeeping.
Also, I've decided to get rid of "SendWithRetryWithInitialTry" and made it the default behavior.

Error logs/messages do not contain the term "upsert"

### DIFF
--- a/pkg/client/dtclient/client.go
+++ b/pkg/client/dtclient/client.go
@@ -402,18 +402,18 @@ func (d *DynatraceClient) upsertSettings(obj SettingsObject) (DynatraceEntity, e
 	}
 	payload, err := buildPostRequestPayload(obj, externalId)
 	if err != nil {
-		return DynatraceEntity{}, fmt.Errorf("failed to build settings object for upsert: %w", err)
+		return DynatraceEntity{}, fmt.Errorf("failed to build settings object: %w", err)
 	}
 
 	requestUrl := d.environmentURL + d.settingsObjectAPIPath
 
 	resp, err := rest.SendWithRetryWithInitialTry(d.client, rest.Post, obj.Id, requestUrl, payload, d.retrySettings.Normal)
 	if err != nil {
-		return DynatraceEntity{}, fmt.Errorf("failed to upsert dynatrace obj: %w", err)
+		return DynatraceEntity{}, fmt.Errorf("failed to create or update dynatrace obj: %w", err)
 	}
 
 	if !success(resp) {
-		return DynatraceEntity{}, fmt.Errorf("failed to upsert settings object with externalId %s (HTTP %d)!\n\tResponse was: %s", externalId, resp.StatusCode, string(resp.Body))
+		return DynatraceEntity{}, fmt.Errorf("failed to create or update settings object with externalId %s (HTTP %d)!\n\tResponse was: %s", externalId, resp.StatusCode, string(resp.Body))
 	}
 
 	entity, err := parsePostResponse(resp)
--- a/pkg/rest/retry.go
+++ b/pkg/rest/retry.go
@@ -79,7 +79,7 @@ func GetWithRetry(client *http.Client, url string, settings RetrySetting) (resp 
 func SendWithRetry(client *http.Client, sendWithBody SendRequestWithBody, objectName string, path string, body []byte, setting RetrySetting) (resp Response, err error) {
 
 	for i := 0; i < setting.MaxRetries; i++ {
-		log.Warn("Failed to upsert config %q. Waiting for %s before retrying...", objectName, setting.WaitTime)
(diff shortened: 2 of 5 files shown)


pkg/client/dtclient/client.go
```diff
@@ -402,18 +402,18 @@ func (d *DynatraceClient) upsertSettings(obj SettingsObject) (DynatraceEntity, e
...
-		return DynatraceEntity{}, fmt.Errorf("failed to build settings object for upsert: %w", err)
+		return DynatraceEntity{}, fmt.Errorf("failed to build settings object: %w", err)
 	}
 
 	requestUrl := d.environmentURL + d.settingsObjectAPIPath
 
 	resp, err := rest.SendWithRetryWithInitialTry(d.client, rest.Post, obj.Id, requestUrl, payload, d.retrySettings.Normal)
 	if err != nil {
-		return DynatraceEntity{}, fmt.Errorf("failed to upsert dynatrace obj: %w", err)
+		return DynatraceEntity{}, fmt.Errorf("failed to create dynatrace obj: %w", err)
```
UnseenWizzard, line 412: This is a "create or update" though, unlike the Config APIs where we can differentiate between create and update and log them differently. 

"Upsert" actually reads better to me than "create/update", but I think that would be thing to log after the change

pkg/rest/retry.go
```diff
@@ -77,6 +77,10 @@ func GetWithRetry(client *http.Client, url string, settings RetrySetting) (resp
 
 // SendWithRetry will retry to call sendWithBody for a given number of times, waiting a give duration between calls
 func SendWithRetry(client *http.Client, sendWithBody SendRequestWithBody, objectName string, path string, body []byte, setting RetrySetting) (resp Response, err error) {
+	resp, err = sendWithBody(client, path, body)
```
UnseenWizzard, line 80: IIRC the Config Client calls to this happen after a first try failed (and a special retry case occurred)
While the Settings client goes straight to send with retry. 

Just changing this without modifying the handling for Config client errors, means we do an extra send/retry in their case right? 
I think that's visible from your change to the request_test assert as well, which expects +1 call. 

I don't think we should just do that, it kinda breaks the max retries by now always doing +1 "retry" call for Configs as they had an original API call before.

pkg/rest/retry.go
```diff
@@ -60,7 +60,7 @@ func GetWithRetry(client *http.Client, url string, settings RetrySetting) (resp
 	}
 
 	for i := 0; i < settings.MaxRetries; i++ {
-		log.Warn("Retrying failed GET request %s with error (HTTP %d)", url, resp.StatusCode)
+		log.Warn("Failed to get config %s with error (HTTP %d). Waiting for %s before retrying...", url, resp.StatusCode, settings.WaitTime)
```
UnseenWizzard, line 63: nit: These are not logging the `config` but the `url` - I think logging 

> Failed to get config https://my-end.live.dynatrace.com/api/something/something/uuid ..

reads a bit strange. 

Note that the Send one you're attempting to unify with, actually logs the `objectName`- maybe that's possible here as well. Maybe both the name and API call could be helpful also

UnseenWizzard (changes requested):
I think some of the logging changes make the log more misleading than it was before, please take a look at the comments

warber:
@UnseenWizzard I've dropped the questionable commits, pls rereview

pkg/client/dtclient/client.go
```diff
@@ -402,18 +402,18 @@ func (d *DynatraceClient) upsertSettings(obj SettingsObject) (DynatraceEntity, e
...
 
 	resp, err := rest.SendWithRetryWithInitialTry(d.client, rest.Post, obj.Id, requestUrl, payload, d.retrySettings.Normal)
 	if err != nil {
-		return DynatraceEntity{}, fmt.Errorf("failed to upsert dynatrace obj: %w", err)
+		return DynatraceEntity{}, fmt.Errorf("failed to create or update dynatrace obj: %w", err)
 	}
 
 	if !success(resp) {
-		return DynatraceEntity{}, fmt.Errorf("failed to upsert settings object with externalId %s (HTTP %d)!\n\tResponse was: %s", externalId, resp.StatusCode, string(resp.Body))
+		return DynatraceEntity{}, fmt.Errorf("failed to create settings object with externalId %s (HTTP %d)!\n\tResponse was: %s", externalId, resp.StatusCode, string(resp.Body))
```
UnseenWizzard, line 416:
```suggestion
		return DynatraceEntity{}, fmt.Errorf("failed to create or update settings object with externalId %s (HTTP %d)!\n\tResponse was: %s", externalId, resp.StatusCode, string(resp.Body))
```

Shouldn't this also be create or update @warber ?

UnseenWizzard (approved):
Thanks for addressing the multiple rounds of comments! 🚀

PR #987, opened 2023-05-03
ci: Don't release dev versions

Laubi:
With this change, we no longer create releases for all tags, but only for rc- and full versions. Additionally, the binaries are no longer uploaded to GitHub in those cases.


UnseenWizzard (dismissed):

UnseenWizzard:
looks good to me @Laubi , took the liberty to hit rebase already

Laubi:
I updated the commit by moving the release-id echo a bit above. Now we can see the debug information of the release-id again during the build

UnseenWizzard (approved):

PR #993, opened 2023-05-04
feat: Create custom random function

Laubi:
Benchmark information:
```
goos: windows
goarch: amd64
pkg: github.com/dynatrace/dynatrace-configuration-as-code/internal/rand
cpu: 11th Gen Intel(R) Core(TM) i9-11950H @ 2.60GHz
BenchmarkInt
BenchmarkInt-16                  3422638               343.5 ns/op            82 B/op          4 allocs/op
BenchmarkIntLegacy
BenchmarkIntLegacy-16             120304              9712 ns/op            5376 B/op          1 allocs/op
PASS

Process finished with the exit code 0
```

Code for the given benchmark:
```
func BenchmarkInt(b *testing.B) {
	b.ReportAllocs()
	for n := 1; n < b.N+1; n++ {
		Int(int64(n))
	}
}
func BenchmarkIntLegacy(b *testing.B) {
	b.ReportAllocs()
	for n := 1; n < b.N+1; n++ {
		newRand := rand.New(rand.NewSource(time.Now().UnixNano()))
		newRand.Int63n(int64(n))
	}
}
```

The reason for the more complicated algorithm being the winner is quite simple: We are initializing a new random source every time we want to create a new random value. This is way havier than just creating secure numbers

### DIFF
--- a/internal/throttle/throttle_requests.go
+++ b/internal/throttle/throttle_requests.go
@@ -18,10 +18,10 @@ package throttle
 
 import (
 	"fmt"
-	"math/rand"
 	"time"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/rand"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/timeutils"
 )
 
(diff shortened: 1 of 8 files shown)


internal/throttle/throttle_requests.go
```diff
@@ -42,14 +42,16 @@ func ThrottleCallAfterError(backoffMultiplier int, message string, a ...any) {
...
-	newRand := rand.New(rand.NewSource(time.Now().UnixNano())) //nolint:gosec
 
 	if backoffMultiplier < 1 {
 		backoffMultiplier = 1
 	}
 
-	addedWaitMillis := newRand.Int63n(MinWaitDuration.Nanoseconds()) //nolint:gosec
+	addedWaitMillis, err := rand.Int(MinWaitDuration.Nanoseconds())
+	if err != nil {
+		log.Error("Failed to generate random gitter. Falling back to use fixed value. Error: %s", err)
```
UnseenWizzard, line 52:
```suggestion
		log.Warn("Failed to generate random gitter. Falling back to use fixed value. Error: %s", err)
```
nit: as it resolves it self

internal/throttle/throttle_requests.go
```diff
@@ -42,14 +42,16 @@ func ThrottleCallAfterError(backoffMultiplier int, message string, a ...any) {
...
+		addedWaitMillis = 42
```
UnseenWizzard, line 53: nit: Why not just add zero millis in this exceptional case? Then we sleep for the defined `MinWaitDuration`.
While 42 is obviously the right answer, just not adding anything would IMO fit the rest of the logic better

UnseenWizzard (dismissed):

warber:
just curious, but what's the issue with the current solution and //nolint ?

Laubi:
@warber The reason is that we don't only have `//nolint`, but also `//nosemgrep`. Both in one line don't work, and in different lines it also errors.

UnseenWizzard (approved):

PR #996, opened 2023-05-09
Feat/escape jinja templates

jskelin:
Workflow definition can also accept Jinja templates. To avoid mixing of resolutions, workflow templates need to be escaped. With these changes, now it is possible to write `\{\{` and `\}\}`, which, as a result, won't be resolved during template preparation. 
Strings `\{\{` and `\}\}` will be replaced with propper jinnja sequence `{{` and `}}` before sending to Dynatrace instance.

### DIFF
--- a/internal/template/template_utils.go
+++ b/internal/template/template_utils.go
@@ -116,3 +116,20 @@ func escapeCharactersForJson(rawString string) (string, error) {
 func escapeNewlines(rawString string) string {
 	return strings.ReplaceAll(rawString, "\n", `\n`)
 }
+
+// EscapeJinjaTemplates replaces each occurrence of "{{" with "\{\{" and
+// each occurrence of "}}" with "\}\}"
+
+func EscapeJinjaTemplates(str string) string {
+	str = strings.ReplaceAll(str, "{{", "\\{\\{")
+	str = strings.ReplaceAll(str, "}}", "\\}\\}")
+	return str
+}
+
+// EscapeJinjaTemplates replaces each occurrence of "\{\{" with \{{" and
+// each occurrence of "\}\}" with "}}"
+func UnescapeJinjaTemplates(str string) string {
+	str = strings.ReplaceAll(str, "\\{\\{", "{{")
+	str = strings.ReplaceAll(str, "\\}\\}", "}}")
+	return str
+}
--- a/internal/template/template_utils.go
+++ b/internal/template/template_utils.go
@@ -17,6 +17,7 @@
 package template
 
 import (
+	"bytes"
 	"encoding/json"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/regex"
@@ -118,15 +119,15 @@ func escapeNewlines(rawString string) string {
 }
 
 // EscapeJinjaTemplates replaces each occurrence of "{{" with "\{\{" and each occurrence of "}}" with "\}\}"
-func EscapeJinjaTemplates(str string) string {
-	str = strings.ReplaceAll(str, `{{`, `\{\{`)
-	str = strings.ReplaceAll(str, "}}", "\\}\\}")
-	return str
+func EscapeJinjaTemplates(src []byte) []byte {
+	src = bytes.ReplaceAll(src, []byte(`{{`), []byte(`\{\{`))
+	src = bytes.ReplaceAll(src, []byte("}}"), []byte("\\}\\}"))
+	return src
 }
 
 // UnescapeJinjaTemplates replaces each occurrence of "\{\{" with \{{" and each occurrence of "\}\}" with "}}"
-func UnescapeJinjaTemplates(str string) string {
-	str = strings.ReplaceAll(str, "\\{\\{", "{{")
-	str = strings.ReplaceAll(str, "\\}\\}", "}}")
+func UnescapeJinjaTemplates(str []byte) []byte {
+	str = bytes.ReplaceAll(str, []byte("\\{\\{"), []byte("{{"))
+	str = bytes.ReplaceAll(str, []byte("\\}\\}"), []byte("}}"))
 	return str
 }
--- a/internal/template/template_utils.go
+++ b/internal/template/template_utils.go
@@ -121,13 +121,13 @@ func escapeNewlines(rawString string) string {
 // EscapeJinjaTemplates replaces each occurrence of "{{" with "\{\{" and each occurrence of "}}" with "\}\}"
 func EscapeJinjaTemplates(src []byte) []byte {
 	src = bytes.ReplaceAll(src, []byte(`{{`), []byte(`\{\{`))
-	src = bytes.ReplaceAll(src, []byte("}}"), []byte("\\}\\}"))
+	src = bytes.ReplaceAll(src, []byte(`}}`), []byte(`\}\}`))
 	return src
 }
 
 // UnescapeJinjaTemplates replaces each occurrence of "\{\{" with \{{" and each occurrence of "\}\}" with "}}"
 func UnescapeJinjaTemplates(str []byte) []byte {
-	str = bytes.ReplaceAll(str, []byte("\\{\\{"), []byte("{{"))
-	str = bytes.ReplaceAll(str, []byte("\\}\\}"), []byte("}}"))
+	str = bytes.ReplaceAll(str, []byte(`\{\{`), []byte(`{{`))
+	str = bytes.ReplaceAll(str, []byte(`\}\}`), []byte(`}}`))
 	return str
 }
--- a/internal/template/template_utils.go
+++ b/internal/template/template_utils.go
@@ -125,7 +125,7 @@ func EscapeJinjaTemplates(src []byte) []byte {
 	return src
 }
 
-// UnescapeJinjaTemplates replaces each occurrence of "\{\{" with \{{" and each occurrence of "\}\}" with "}}"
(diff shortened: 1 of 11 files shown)


pkg/deploy/automation.go
```diff
@@ -20,21 +20,16 @@ import (
...
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/idutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/template"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/client/automation"
 	config "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/parameter"
 )
 
-//go:generate mockgen -source=automation.go -destination=automation_mock.go -package=deploy automationClient
+//go:generate mockgen -source=automation.go -destination=automation_mock_client.go -package=deploy automationClient
```
UnseenWizzard, line 29: Is there any reason for renaming the destination file & adding a new gitignore entry?
Why can't this continue the convention of mocks ending in _mock.go?

internal/template/template_utils.go
```diff
@@ -116,3 +117,17 @@ func escapeCharactersForJson(rawString string) (string, error) {
 func escapeNewlines(rawString string) string {
 	return strings.ReplaceAll(rawString, "\n", `\n`)
 }
+
+// EscapeJinjaTemplates replaces each occurrence of "{{" with "\{\{" and each occurrence of "}}" with "\}\}"
+func EscapeJinjaTemplates(src []byte) []byte {
+	src = bytes.ReplaceAll(src, []byte(`{{`), []byte(`\{\{`))
```
UnseenWizzard, line 123: nit: using a raw string here and regular strings and escaped escape chars for all other cases is a bit strange, any reason?

pkg/deploy/automation_dummy_client.go
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package deploy
+
+import "github.com/dynatrace/dynatrace-configuration-as-code/pkg/client/automation"
+
+type dummyAutomationClient struct {
+}
+
+func (c *dummyAutomationClient) Upsert(_ automation.ResourceType, id string, _ []byte) (*automation.Response, error) {
```
UnseenWizzard, line 24: I thought an Automation Dummy already exists? @warber didn't you add one for validation?

pkg/download/automation/automation.go
```diff
@@ -135,11 +134,14 @@ func createTemplateFromRawJSON(obj automationClient.Response, configType string)
 		content = obj.Data
 	}
 	content = jsonutils.MarshalIndent(content)
+	if configType == "workflow" {
```
UnseenWizzard, line 137: I'm not sure we need to or should limit this to workflows - are we sure that templating is (and will) only be possible for workflows, not other automation resources?

pkg/deploy/automation_dummy_client.go
warber: yes i think @jskelin just don't want to have it in `automation.go`

UnseenWizzard: Ah! Thanks, I missed the removal from `automation.go`, I thought it was in it's own file already

pkg/deploy/automation.go
jskelin: dummy client for automation is moved to `automation_dummy_client.go`. This was done to match the naming as they're in the same package. 

If it is necessary to keep the current convention that mocks generated files, have _moch.go suffix, and it'll be returned.

pkg/config/v2/config.go
```diff
@@ -91,6 +91,10 @@ func (EntityType) ID() TypeId {
 // AutomationResource defines which resource is an AutomationType
 type AutomationResource string
 
+func (a *AutomationResource) String() string {
```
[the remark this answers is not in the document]
jskelin: changed

pkg/download/automation/automation.go
jskelin: I also think that it would be better to apply this to all templates that are produced by the automation interface. Since they have used it already, it can be expected to be expanded to other resources.

pkg/config/v2/config.go
```diff
@@ -91,6 +91,10 @@ func (EntityType) ID() TypeId {
 // AutomationResource defines which resource is an AutomationType
 type AutomationResource string
 
+func (a AutomationResource) String() string {
```
UnseenWizzard, line 94: I still don't understand why we need this. Especially now that it's not a pointer receiver anymore. 
Printing an AutomationResource value should just print it as the string it is anyway?

pkg/deploy/automation.go
UnseenWizzard: It's not necessary, but personally, I some times remove all mocks as `**/*_mock.go` when switching between branches that have added/removed certain mocks. 

So them following the same convention is helpful

internal/template/template_utils.go
UnseenWizzard: I meant that this one is a back-tick raw string, while the ones on lines 124, 130, 131 are regular strings

```go
src = bytes.ReplaceAll(src, []byte(`{{`), []byte(`\{\{`))
src = bytes.ReplaceAll(src, []byte("}}"), []byte("\\}\\}"))
[...]
str = bytes.ReplaceAll(str, []byte("\\{\\{"), []byte("{{"))
str = bytes.ReplaceAll(str, []byte("\\}\\}"), []byte("}}"))
```

you could probably use raw strings everywhere and skip the escaped-escape :) 


```go
src = bytes.ReplaceAll(src, []byte(`{{`), []byte(`\{\{`))
src = bytes.ReplaceAll(src, []byte(`}}`), []byte(`\}\}`))
[...]
str = bytes.ReplaceAll(str, []byte(`\{\{`), []byte(`{{`))
str = bytes.ReplaceAll(str, []byte(`\}\}`), []byte(`}}`))
```

pkg/deploy/automation_dummy_client.go
warber: yeah, although IMO there is absolutely no reason to put that in a separate file, that has in addition to it an awful name... but fine with it :man_shrugging:  :smile:

warber:
please use our PR template :)

pkg/config/v2/config.go
jskelin: You're right. removed

pkg/deploy/automation.go
jskelin: I'll change that in a separate PR.
I tried to move automation in the deploy package to a separate package like in the download package. But it isn't so easy.

internal/template/template_utils.go
```diff
@@ -116,3 +117,17 @@ func escapeCharactersForJson(rawString string) (string, error) {
...
+	src = bytes.ReplaceAll(src, []byte(`}}`), []byte(`\}\}`))
+	return src
+}
+
+// UnescapeJinjaTemplates replaces each occurrence of "\{\{" with \{{" and each occurrence of "\}\}" with "}}"
```
Laubi, line 128:
```suggestion
// UnescapeJinjaTemplates replaces each occurrence of "\{\{" with {{" and each occurrence of "\}\}" with "}}"
```

internal/template/template_utils_test.go
```diff
@@ -286,3 +286,34 @@ func Test_isListDefinition(t *testing.T) {
 		})
 	}
 }
+
+func TestEscapeJinjaTemplates(t *testing.T) {
```
Laubi, line 290: Please use table tests here. The test cases are more apparent, making the code easier to understand. 
https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

pkg/deploy/automation.go
Laubi: I also just saw the change and was wondering why. As you stated, please move it back to _mock in a followup PR, or even this one. Should be a simple renaming, and we would not need to touch the .gitignore

Laubi:
I wrote a quick fuzzer to see if the escape-unescape use-case works:
```
func FuzzEscapeJinjaTemplates(f *testing.F) {
	seeds := []string{"", "some", "seed", "{{", "}}", `\{\{`, "{{thisisanexample}}"}

	for i := range seeds {
		f.Add(seeds[i])
	}

	f.Fuzz(func(t *testing.T, given string) {
		result := string(UnescapeJinjaTemplates(EscapeJinjaTemplates([]byte(given))))
		assert.Equal(t, given, result)
	})
}
```

The code errors nearly immediately for the given string
```
           	Error:      	Not equal: 
            	            	expected: "\\{\\{"
            	            	actual  : "{{"
```

Afaik, we discussed that we want to escape the `\\{` case as well because we can never be sure what customers configure. They might just have it in a name, a field, wrongly configure it, or anything.

Laubi (changes requested):
See comment above, I really think we need to take care of `\{`

jskelin:
> See comment above, I really think we need to take care of `\{`

After live discussion and investigation, we found that the automation framework already handles this case. So it is not the issue anymore.

internal/template/template_utils_test.go
jskelin: In this case, table tests would not bring benefits.

Since this is more of a style issue and it deals with the test, I'll just leave it as it is.

Laubi (approved):

jskelin:
Due to change of requests, this PR is replaced with https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1007

PR #998, opened 2023-05-09
fix(rest): Re-introduce custom user-agent at the lowest REST lib layer

UnseenWizzard:
With introducing oAuth support, we have lost the dedicated user-agent from the oAuth client (never added) as well as API token authenticated connects (not sent/ overwritten somewhere after the Transport struct it was added to).

To ensure we don't forget to send this with any future client implementations, setting the User-Agent header is moved to the lowest possible layer in our API call methods, the final rest/request method that actually calls the http/client.

### DIFF
--- a/pkg/rest/request.go
+++ b/pkg/rest/request.go
@@ -19,8 +19,10 @@ package rest
 import (
 	"bytes"
 	"fmt"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/version"
 	"io"
 	"net/http"
+	"runtime"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/timeutils"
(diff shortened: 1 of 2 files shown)


pkg/rest/request.go
```diff
@@ -19,8 +19,10 @@ package rest
 import (
 	"bytes"
 	"fmt"
+	version2 "github.com/dynatrace/dynatrace-configuration-as-code/pkg/version"
```
Laubi, line 22: nitpick: Why are you using an alias?

Laubi (approved):

pkg/rest/request.go
UnseenWizzard: because I copied the header line from pkg/client/transport and didn't notice the import name that used :D 
version2 seems necessary there as well tho

UnseenWizzard: addressed

warber (approved):

PR #999, opened 2023-05-09
feat(delete): Introduce structured delete definition

UnseenWizzard:
As the 'externalID' by which Settings are identified by monaco is extended to include the project (full coordinate - project,type,id) - the simple string format is no longer enough to define delete entry.

A new format is added that defines a structured project, type and name or ID delete entry.
Loading is extended to be able to load both the full and the old simple format.

This PR still includes several PRs as it depends on the extended external ID generation to be fully functional.
As that part is also in progress, this is left as draft for early review - other than these TODOs it should be ready for review

Delete YAML format changes to a structured format. 
Old format is still usable, but for Settings will produce a warning log.

```yaml
delete: 
  - project: "project"
    type: alerting-profile
    name: "Star Trek Service"
  - project: "project"
    type: builtin:whatever
    id: super-awesome-config
  - "alerting-profile/Star Trek Service"
  - project: automation-project <--- future extension to include Automation, not part of this PR
    type: workflow 
    id: my-workflow
```

### DIFF
--- a/pkg/delete/delete.go
+++ b/pkg/delete/delete.go
@@ -27,9 +27,14 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/api"
 )
 
+// DeletePointer contains all data needed to identify an object to be deleted from a Dynatrace environment.
+// DeletePointer is similar but not fully equivalent to config.Coordinate as it may contain an Identifier that is either
+// a Name or a ConfigID - only in case of a ConfigID is it actually equivalent to a Coordinate
 type DeletePointer struct {
-	Project    string
-	Type       string
+	Project string
+	Type    string
+
+	//Identifier will either be the Name of a classic Config API object, or a configID for newer types like Settings
 	Identifier string
 }
 
@@ -95,7 +100,7 @@ func deleteSettingsObject(c dtclient.Client, entries []DeletePointer) []error {
 	for _, e := range entries {
 
 		if e.Project == "" {
-			log.Warn("Generating legacy externalID for deletion of %q - this will fail to identify newer Settings object. Consider defining a 'project' for this delete entry.", e)
+			log.Warn("Generating legacy externalID for deletion of %q - this will fail to identify a newer Settings object. Consider defining a 'project' for this delete entry.", e)
 		}
 		externalID, err := idutils.GenerateExternalID(e.asCoordinate())
 
--- a/pkg/delete/delete.go
+++ b/pkg/delete/delete.go
@@ -22,13 +22,23 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/coordinate"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/idutils"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/api"
 )
 
 type DeletePointer struct {
-	Type     string
-	ConfigId string
+	Project    string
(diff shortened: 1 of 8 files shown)


pkg/delete/delete.go
```diff
@@ -18,16 +18,17 @@ package delete
...
 	"fmt"
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/idutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/client/dtclient"
 
-	"github.com/dynatrace/dynatrace-configuration-as-code/internal/idutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/api"
 )
 
 type DeletePointer struct {
```
warber, line 28:
```suggestion
type DeletePointer coordinate.Coordinate
```
this is now just a coordinate righ?

pkg/delete/delete_loader.go
```diff
@@ -34,7 +36,14 @@ type loaderContext struct {
...
 type deleteFileDefinition struct {
-	DeleteEntries []string `yaml:"delete"`
+	DeleteEntries []interface{} `yaml:"delete"`
+}
+
+type deleteEntry struct {
+	Project    string `mapstructure:"project"`
+	Type       string `mapstructure:"type"`
+	ConfigId   string `mapstructure:"id"`
+	ConfigName string `mapstructure:"name"`
```
warber, line 46: good idea to utilize mapstructure for that :+1:

pkg/delete/delete.go
UnseenWizzard: You're right! It is. And used as a coordinate as well

UnseenWizzard: On further thought, it actually isn't! For a classic Config API it is still defined by the Name, not the ConfigID. 
So while it looks similar and for Settings and Workflows it is, it's not always a coordinate. 

That's actually also why I had renamed it's `Identifier` field from the previously used `ConfigID`.

With that I actually think it's better to keep the separation and the util method to turn it into a coordinate in the specific places in which it is one

pkg/delete/delete.go
```diff
@@ -83,7 +93,12 @@ func deleteSettingsObject(c dtclient.Client, entries []DeletePointer) []error {
 	errors := make([]error, 0)
 
 	for _, e := range entries {
-		externalID, err := idutils.GenerateExternalID(coordinate.Coordinate{Type: e.Type, ConfigId: e.ConfigId})
+
+		if e.Project == "" {
+			log.Warn("Generating legacy externalID for deletion of %q - this will fail to identify newer Settings object. Consider defining a 'project' for this delete entry.", e)
```
Laubi, line 98:
```suggestion
			log.Warn("Generating legacy externalID for deletion of %q - this will fail to identify newer Settings objects. Consider defining a 'project' for this delete entry.", e)
```

Laubi: I am not sure about this, but it reads funny. 

If we are talking about a single object, we should say 
```suggestion
			log.Warn("Generating legacy externalID for deletion of %q - this will fail to identify a newer Settings object. Consider defining a 'project' for this delete entry.", e)
```

pkg/delete/delete.go
Laubi: Could you please add this to the docs? 
And add docs altogether, please? :)

pkg/delete/delete.go
UnseenWizzard: Agreed, "a" is better

Laubi (approved):

PR #1001, opened 2023-05-10
feat: use full coordinate when generating external ids for settings 2.0 objects

warber:
This PR changes the way we generate external IDs for settings 2.0 object to use the "full Coordinate" (projectname, schemaID, ID) in order to support using the same configuration in multiple projects by multiple "users".

To stay backwards compatible, we still calclulate the "old" external id (without project name included) and try to update already existing settings objects accordingly.

### DIFF
--- a/cmd/monaco/integrationtest/assert.go
+++ b/cmd/monaco/integrationtest/assert.go
@@ -158,7 +158,12 @@ func assertConfig(t *testing.T, client dtclient.ConfigClient, theApi api.API, en
 }
 
 func assertSetting(t *testing.T, c dtclient.SettingsClient, typ config.SettingsType, environment manifest.EnvironmentDefinition, shouldBeAvailable bool, config config.Config) {
-	expectedExtId := idutils.GenerateExternalID(typ.SchemaId, config.Coordinate.ConfigId)
+	expectedExtId, err := idutils.GenerateExternalID(config.Coordinate)
+	if err != nil {
+		t.Errorf("Unable to generate external id: %v", err)
+		return
+	}
+
 	objects, err := c.ListSettings(typ.SchemaId, dtclient.ListSettingsOptions{DiscardValue: true, Filter: func(o dtclient.DownloadSettingsObject) bool { return o.ExternalId == expectedExtId }})
 	assert.NilError(t, err)
 
--- a/cmd/monaco/integrationtest/cleanup.go
+++ b/cmd/monaco/integrationtest/cleanup.go
@@ -82,8 +82,13 @@ func cleanupSettings(t *testing.T, fs afero.Fs, manifestPath string, loadedManif
 		}
 		for _, configs := range cfgsForEnv {
 			for _, cfg := range configs {
-				if typ, ok := cfg.Type.(*config.SettingsType); ok {
-					extID := idutils.GenerateExternalID(typ.SchemaId, cfg.Coordinate.ConfigId)
+				if typ, ok := cfg.Type.(config.SettingsType); ok {
+					extID, err := idutils.GenerateExternalID(cfg.Coordinate)
+					if err != nil {
+						t.Log(err)
+						continue
+					}
+
 					deleteSettingsObjects(t, typ.SchemaId, extID, c)
 				}
 			}
--- a/cmd/monaco/integrationtest/v2/settings_integration_test.go
+++ b/cmd/monaco/integrationtest/v2/settings_integration_test.go
@@ -20,13 +20,19 @@
 package v2
 
 import (
+	"github.com/dynatrace/dynatrace-configuration-as-code/cmd/monaco/dynatrace"
 	"github.com/dynatrace/dynatrace-configuration-as-code/cmd/monaco/integrationtest"
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/idutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/testutils"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/client/dtclient"
+	v2 "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/coordinate"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/project/v2/topologysort"
+	"github.com/stretchr/testify/assert"
 	"testing"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/cmd/monaco/runner"
 	"github.com/spf13/afero"
-	"gotest.tools/assert"
 )
 
 // tests all configs for a single environment
@@ -43,7 +49,7 @@ func TestIntegrationSettings(t *testing.T) {
 		cmd.SetArgs([]string{"deploy", "--verbose", manifest})
 		err := cmd.Execute()
 
-		assert.NilError(t, err)
+		assert.NoError(t, err)
 		integrationtest.AssertAllConfigsAvailability(t, fs, manifest, []string{}, specificEnvironment, true)
 
 		// This causes an Update of all Settings
@@ -51,7 +57,7 @@ func TestIntegrationSettings(t *testing.T) {
 		cmd.SetArgs([]string{"deploy", "--verbose", manifest})
 		err = cmd.Execute()
 
-		assert.NilError(t, err)
+		assert.NoError(t, err)
 		integrationtest.AssertAllConfigsAvailability(t, fs, manifest, []string{}, specificEnvironment, true)
 	})
 }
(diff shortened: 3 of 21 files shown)


internal/idutils/external_id.go
```diff
@@ -18,26 +18,34 @@ package idutils
...
 )
 
-// GenerateExternalID generates the externalID for settings 2.0 objects based on the schema, and ID.
-// The result of the function is pure.
-// Max length for the external ID is 500
-func GenerateExternalID(schema, id string) string {
+// GenerateExternalID returns a string that can be used as an external ID for a Settings 2.0 object
+func GenerateExternalID(coordinate coordinate.Coordinate) string {
+	if coordinate.Type == "" || coordinate.ConfigId == "" {
+		panic("Settings schema id and config id needs to be set to generate a valid external id for settings 2.0 objects")
```
UnseenWizzard, line 28: can this actually happen in some random runtime situations? If not, then why validate against it.
If there's a way that happens then I'd prefer returning an error.

internal/idutils/external_id.go
```diff
@@ -18,26 +18,34 @@ package idutils
...
 	const externalIDMaxLength = 500
 
-	formattedID := fmt.Sprintf(format, schema, id)
-	encodedID := base64.StdEncoding.EncodeToString([]byte(formattedID))
-
-	encodedIDMaxLength := externalIDMaxLength - len(prefix)
-	if len(encodedID) > encodedIDMaxLength {
-		encodedID = encodedID[encodedIDMaxLength:]
+	rawID := strings.Join(parts, "$")
+	if strings.HasSuffix(rawID, "$") {
```
UnseenWizzard, line 41: Wouldn't this only apply if the configID actually ends in $ ? And then we'd produce possibly wrong/overlapping external ID.

Join places the separator between elements AFAIK - so for no or one element it won't append a $

warber: true that

cmd/monaco/integrationtest/v2/multi_project_multi_user_test.go
```diff
@@ -0,0 +1,46 @@
...
+	RunIntegrationWithCleanup(t, multiProjectMultiUserFolder, multiProjectMultiUserManifest, "", "MultiProjectMultiUser", func(fs afero.Fs, _ TestContext) {
+
+		// This causes a POST for all configs:
+		cmd := runner.BuildCli(fs)
+		cmd.SetArgs([]string{"deploy", "--verbose", multiProjectMultiUserManifest})
+		err := cmd.Execute()
+
+		assert.NilError(t, err)
+
+		integrationtest.AssertAllConfigsAvailability(t, fs, multiProjectMultiUserManifest, []string{}, "", true)
```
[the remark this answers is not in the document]
warber: ewww

cmd/monaco/integrationtest/v2/settings_integration_test.go
```diff
@@ -68,5 +74,57 @@ func TestIntegrationValidationSettings(t *testing.T) {
...
+		OriginObjectId: configToDeploy.OriginObjectId,
+	})
+	assert.NoError(t, err)
+
+	settings, _ := c.ListSettings("builtin:anomaly-detection.metric-events", dtclient.ListSettingsOptions{
+		DiscardValue: true,
+	})
+
+	// assert that ther is only 1 settings object
+	assert.Len(t, settings, 1)
```
UnseenWizzard, line 129: This assert seems like it will be pretty flaky - if any other test has currently created a metric-event, or some cleanup failed, there might be more than one. 

You could pass a filter on the expected externalId to ListSettings and check that the legacy ID is not found, but the new one is

warber:
@UnseenWizzard I've addressed all of your comments :)

cmd/monaco/integrationtest/cleanup.go
```diff
@@ -82,8 +82,8 @@ func cleanupSettings(t *testing.T, fs afero.Fs, manifestPath string, loadedManif
 		}
 		for _, configs := range cfgsForEnv {
 			for _, cfg := range configs {
-				if typ, ok := cfg.Type.(*config.SettingsType); ok {
-					extID := idutils.GenerateExternalID(typ.SchemaId, cfg.Coordinate.ConfigId)
+				if typ, ok := cfg.Type.(config.SettingsType); ok {
+					extID, _ := idutils.GenerateExternalID(cfg.Coordinate)
```
UnseenWizzard, line 86: does it even make sense to continue with delete if the extId generation failed? Before we panicked. 
Maybe we should continue rather than even attempt the delete?

cmd/monaco/integrationtest/assert.go
```diff
@@ -158,7 +158,7 @@ func assertConfig(t *testing.T, client dtclient.ConfigClient, theApi api.API, en
 }
 
 func assertSetting(t *testing.T, c dtclient.SettingsClient, typ config.SettingsType, environment manifest.EnvironmentDefinition, shouldBeAvailable bool, config config.Config) {
-	expectedExtId := idutils.GenerateExternalID(typ.SchemaId, config.Coordinate.ConfigId)
+	expectedExtId, _ := idutils.GenerateExternalID(config.Coordinate)
```
UnseenWizzard, line 161: I think in the error case, the assert will not behave wrongly won't it?

If there's an error the id is empty string,  so the filter will include all Settings objects that don't  have an externalId attached. 

Would suggest to error out of the assert if generating the ID failed

pkg/delete/delete.go
```diff
@@ -82,7 +83,11 @@ func deleteSettingsObject(c dtclient.Client, entries []DeletePointer) []error {
 	errors := make([]error, 0)
 
 	for _, e := range entries {
-		externalID := idutils.GenerateExternalID(e.Type, e.ConfigId)
+		externalID, err := idutils.GenerateExternalID(coordinate.Coordinate{Type: e.Type, ConfigId: e.ConfigId})
+		if err != nil {
+			errors = append(errors, fmt.Errorf("unable to generate external id :%w", err))
```
Laubi, line 88: nitpicky, but customer-facing :colon placement :)
`"unable to generate external id: %w", err`

pkg/client/dtclient/client.go
```diff
@@ -392,36 +404,63 @@ func (d *DynatraceClient) upsertSettings(obj SettingsObject) (DynatraceEntity, e
...
 	}
 
-	externalId := idutils.GenerateExternalID(obj.SchemaId, obj.Id)
+	// generate legacy external ID without project name.
+	// and check if settings object with that external ID exists
+	// This exists for avoiding breaking changes when we enhanced external id generation with full coordinates (incl. project name)
+	// This can be removed in a later release of monaco
+	legacyExternalID, err := d.generateExternalID(coordinate.Coordinate{Type: obj.Coordinate.Type, ConfigId: obj.Coordinate.ConfigId})
+	if err != nil {
+		return DynatraceEntity{}, fmt.Errorf("unable to generate external id :%w", err)
```
Laubi, line 413: Same issue again with the colon placement

jskelin (approved):

pkg/client/dtclient/client.go
```diff
@@ -392,36 +404,63 @@ func (d *DynatraceClient) upsertSettings(obj SettingsObject) (DynatraceEntity, e
...
+		return DynatraceEntity{}, fmt.Errorf("unable to search Settings 2.0 object of schema %q with externalId %q", obj.SchemaId, legacyExternalID)
+	}
+
+	externalID, err := d.generateExternalID(obj.Coordinate)
+	if err != nil {
+		return DynatraceEntity{}, fmt.Errorf("unable to generate external id: %w", err)
+	}
+
+	if len(settingsWithExternalID) > 0 {
+		obj.OriginObjectId = settingsWithExternalID[0].ObjectId
```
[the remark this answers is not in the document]
Laubi: Since we filter the externalId in line 417, no, there is at most one object possible

internal/idutils/external_id.go
```diff
@@ -19,25 +19,31 @@ package idutils
...
-	formattedID := fmt.Sprintf(format, schema, id)
-	encodedID := base64.StdEncoding.EncodeToString([]byte(formattedID))
-
-	encodedIDMaxLength := externalIDMaxLength - len(prefix)
-	if len(encodedID) > encodedIDMaxLength {
-		encodedID = encodedID[encodedIDMaxLength:]
+	rawID := strings.Join(parts, "$")
+	encodedID := base64.StdEncoding.EncodeToString([]byte(rawID))
+	if len(encodedID) > externalIDMaxLength-len(prefix) {
+		encodedID = encodedID[len(encodedID)-(externalIDMaxLength-len(prefix)):]
```
Laubi, line 44: Is this correct? What should this do?

warber: what exactly do you mean? the `if` part ensures that we do not exceed the max length of an identifier

cmd/monaco/integrationtest/v2/multi_project_multi_user_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"testing"
+)
+
+var multiProjectMultiUserFolder = "test-resources/integration-multi-project-multi-user/"
+var multiProjectMultiUserManifest = multiProjectMultiUserFolder + "manifest.yaml"
+
+// TestIntegrationMultiProjectMultiUser tries to upload a project that contatins two projects with
+// the exact same settings 2.0 object and verifies that deploying such a monaco configuration results in
+// two different settings objects deployed on the environment
+func TestIntegrationMultiProjectMultiUser(t *testing.T) {
```
UnseenWizzard, line 40: nit: That name seems off for what this tests (also applies to its resources folder)
IMO something like "TestSettingsInDifferentProjectsGetDifferentExternalIDs" would tell me more

internal/idutils/external_id.go
UnseenWizzard, line 44: hmm... you changed how a >500 character externalID is cut off. Won't that mean that we can't detect long old externalIDs anymore? 

So changing this will impact/break the extID update in this case

UnseenWizzard (approved):

PR #1004, opened 2023-05-16
feat: Allow Automations Deletion

UnseenWizzard:
Allows deletion of automation resoruces.

This PR is done to "make it work"! There are a few small refactorings in dedicated commits, but there are rough edges in setup that I'd like to address in followup PRs and purposely ignored here.
Followup improvements are: 

- Refactor Automation Client setup! Unlike the "DynatraceClient" that's uniformly setup by a util shared between commands, each command currently sets up the AutomationClient in dedicated code. This should IMO move into the shared Client setup.
- Refactor UUID generation! Re-using the "FromName" with a coordinate string is strange, would introduce a from Coordinate
- Mapping from config.AutomationResource to the client's resource iota is clunky and done in several places. I'd like to introduce a single mapper that can be reused everywhere. 

**Suggested to review per commit**

Yes, Automation resources can now be deleted if the FF is active

### DIFF
--- a/cmd/monaco/delete/delete.go
+++ b/cmd/monaco/delete/delete.go
@@ -15,11 +15,17 @@
 package delete
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/cmd/monaco/dynatrace"
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/concurrency"
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/environment"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/errutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/client"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/client/automation"
+	config "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/delete"
 	"golang.org/x/exp/maps"
 	"path/filepath"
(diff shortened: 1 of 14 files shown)


pkg/delete/delete.go
```diff
@@ -46,22 +49,35 @@ func (d DeletePointer) asCoordinate() coordinate.Coordinate {
...
 }
 
+type ClientSet struct {
+	DTClient         dtclient.Client
+	AutomationClient *automationClient.Client
+}
+
 // Configs removes all given entriesToDelete from the Dynatrace environment the given client connects to
-func Configs(client dtclient.Client, apis api.APIs, entriesToDelete map[string][]DeletePointer) []error {
+func Configs(clients ClientSet, apis api.APIs, automationResources map[string]config.AutomationResource, entriesToDelete map[string][]DeletePointer) []error {
```
warber, line 58: I'd strongly recommend to change the name of that function :D

cmd/monaco/delete/delete.go
```diff
@@ -83,14 +89,35 @@ func deleteConfigs(environments []manifest.EnvironmentDefinition, apis api.APIs,
...
 func deleteConfigForEnvironment(env manifest.EnvironmentDefinition, apis api.APIs, entriesToDelete map[string][]delete.DeletePointer) []error {
 	dynatraceClient, err := dynatrace.CreateDTClient(env.URL.Value, env.Auth, false)
-
 	if err != nil {
 		return []error{
 			fmt.Errorf("It was not possible to create a client for env `%s` due to the following error: %w", env.Name, err),
 		}
 	}
 
+	var autClient *automation.Client
```
warber, line 98: Why not move clientset.go a layer up in to the client package and call `CreateClientSet()` ?`

pkg/delete/delete.go
```diff
@@ -46,22 +49,35 @@ func (d DeletePointer) asCoordinate() coordinate.Coordinate {
...
 
-		} else {
-			deleteErrs := deleteClassicConfig(client, theApi, entries, targetApi)
-			errs = append(errs, deleteErrs...)
+			if !featureflags.AutomationResources().Enabled() {
+				continue
+			}
+
+			if clients.AutomationClient == nil {
+				log.Warn("Skipped deletion of %d Automation configurations of type %q as API client was unavailable", len(entries), entryType)
```
Laubi, line 72: nit: We currently mix (even in this PR) messages with a trailing dot and without. I'd argue we should use them.
Also, some of the logging starts with a lowercase, also in this PR

pkg/delete/delete.go
UnseenWizzard: Agreed - forgot that in my refactoring notes. This one and the "per type" DeleteAll should IMO change

cmd/monaco/delete/delete.go
UnseenWizzard: See PR description - I wanted to keep "delete automations" separate from big refactoring. 
The clientSets in several places and the fact that each cmd sets up it's own automation client I'd change into a unified setup in the 'DynatraceClient' as a followup

pkg/config/v2/type_definition_test.go
```diff
@@ -219,7 +219,7 @@ settings:
 			name:  "wrong data type",
 			given: given{"0x12d4"},
 			expected: expected{
-				errorMessage: "'type' section is not filed with proper values",
+				errorMessage: "failed to parse 'type' section: '' expected a map, got 'int'",
```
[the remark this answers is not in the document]
UnseenWizzard: there is no type section in this error sample... The previous error message completely swallowed the actual parsing error message though, so IMO this is better, even though it produces an equally useless message if the data is way off - rather than just a typo or forgotten sub-field

pkg/delete/delete.go
UnseenWizzard: Addressed the logging for delete in extra commit. Also unified how statements end (`.`) and actions in progress (`...`).

Re casing: I found a single log line that started lowercase and that seems to have been a copy pasted error message. 
Errors start lowercase and don't have punctuation per Go convention.

pkg/delete/delete.go
```diff
@@ -18,7 +18,10 @@ package delete
 
 import (
 	"fmt"
+	"github.com/dynatrace/dynatrace-configuration-as-code/internal/featureflags"
+	automationClient "github.com/dynatrace/dynatrace-configuration-as-code/pkg/client/automation"
```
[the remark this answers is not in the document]
UnseenWizzard: copying from somewhere... addressed in added commit

PR #1003, opened 2023-05-16
feat: Download only automation resources

warber:
#### What this PR does / Why we need it:
Similar to settings and classic configs, thie PR introduces a `--only-automation` flag to specify that only automation API resources shall be downloaded by Monaco.

#### Does this PR introduce a user-facing change?

User can specify via `--only-automation` flag that only automation resources (no settings, no classic apis) schall be downloaded. Currently the feature flag `MONACO_FEAT_AUTOMATION_RESOURCES` must be set too.

PR #1007, opened 2023-05-17
escape jinja templates

jskelin:
##### escape workflow jinja template
Workflow definition can also accept Jinja templates. To avoid mixing of resolutions, workflow templates need to be escaped. With these changes, all jinja templates downloaded from DT instance like ```{{ .var }} ```  will be escaped with ```{{`{{`}} .var {{`}}`}} ```

##### add `--only-automation` flag
`--only-automation` flag behaves in a similar way as already existed `--only-apis` and `--only-setings`

This PR replaces https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/996

### DIFF
--- a/cmd/monaco/download/download_command.go
+++ b/cmd/monaco/download/download_command.go
@@ -84,6 +84,13 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
 	cmd.MarkFlagsMutuallyExclusive("api", "only-apis", "only-settings")
 	cmd.MarkFlagsMutuallyExclusive("only-apis", "only-settings")
 
+	if featureflags.AutomationResources().Enabled() {
+		cmd.Flags().BoolVar(&f.onlyAutomation, "only-automation", false, "Only download automation objects, skip another")
+		cmd.MarkFlagsMutuallyExclusive("only-apis", "only-settings", "only-automation")
+		cmd.MarkFlagsMutuallyExclusive("api", "only-automation")
+		cmd.MarkFlagsMutuallyExclusive("settings-schema", "only-automation")
+	}
+
 	if featureflags.Entities().Enabled() {
 		getDownloadEntitiesCommand(fs, command, cmd)
 	}
--- a/cmd/monaco/download/download_configs.go
+++ b/cmd/monaco/download/download_configs.go
@@ -37,6 +37,7 @@ type downloadCmdOptions struct {
 	specificSchemas         []string
 	onlyAPIs                bool
 	onlySettings            bool
+	onlyAutomation          bool
 }
 
 type auth struct {
@@ -119,6 +120,7 @@ func (d DefaultCommand) DownloadConfigsBasedOnManifest(fs afero.Fs, cmdOptions d
 		specificSchemas: cmdOptions.specificSchemas,
 		onlyAPIs:        cmdOptions.onlyAPIs,
 		onlySettings:    cmdOptions.onlySettings,
+		onlyAutomation:  cmdOptions.onlyAutomation,
 	}
 
 	downloaders, err := makeDownloaders(options)
@@ -148,6 +150,7 @@ func (d DefaultCommand) DownloadConfigs(fs afero.Fs, cmdOptions downloadCmdOptio
 		specificSchemas: cmdOptions.specificSchemas,
 		onlyAPIs:        cmdOptions.onlyAPIs,
 		onlySettings:    cmdOptions.onlySettings,
+		onlyAutomation:  cmdOptions.onlyAutomation,
 	}
 
 	downloaders, err := makeDownloaders(options)
@@ -164,6 +167,7 @@ type downloadConfigsOptions struct {
 	specificSchemas []string
 	onlyAPIs        bool
 	onlySettings    bool
+	onlyAutomation  bool
 }
 
 func doDownloadConfigs(fs afero.Fs, downloaders downloaders, opts downloadConfigsOptions) error {
@@ -205,7 +209,7 @@ func downloadConfigs(downloaders downloaders, opts downloadConfigsOptions) (proj
 		copyConfigs(configs, settingCfgs)
 	}
 
-	if shouldDownloadAutomationResources() {
+	if shouldDownloadAutomationResources(opts) {
 		automationCfgs, err := downloaders.Automation().Download(opts.projectName)
 		if err != nil {
 			return nil, err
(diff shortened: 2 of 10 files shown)


cmd/monaco/download/download_command.go
```diff
@@ -84,6 +84,13 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
 	cmd.MarkFlagsMutuallyExclusive("api", "only-apis", "only-settings")
 	cmd.MarkFlagsMutuallyExclusive("only-apis", "only-settings")
 
+	if featureflags.AutomationResources().Enabled() {
```
UnseenWizzard, line 87: Is the addition of only-automation needed for this PR, or was there some miscommunication with @warber, who also opened a PR for it? https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1003/files

I agree with placing this under the FF - could this change just be added to the other PR which also extends the tests, and be dropped from this changeset (if not needed for the changes here)?

pkg/download/automation/automation.go
```diff
@@ -78,6 +80,13 @@ func (d *Downloader) Download(projectName string, automationTypes ...config.Auto
 		for _, obj := range response.Results {
 
 			configId := obj.Id
+
+			if escaped, err := escapeJinjaTemplates(obj.Data); err != nil {
+				log.Warn("Failed to escape automation jinja for config %v (%s) - template needs manual adaptation: %v", configId, at.Resource, err)
```
UnseenWizzard, line 85: I'm not sure customers will be aware of what Jinja is, even though the Workflows docs mention it [0]

I'd suggest we use naming that fits the more general terms used in the docs

```suggestion
				log.Warn("Failed to escape automation templating expressions for config %v (%s) - template needs manual adaptation: %v", configId, at.Resource, err)
```


[0] https://www.dynatrace.com/support/help/shortlink/automations-reference

pkg/download/automation/automation.go
```diff
@@ -102,6 +111,12 @@ func (d *Downloader) Download(projectName string, automationTypes ...config.Auto
 	return configsPerType, nil
 }
 
+func escapeJinjaTemplates(src []byte) ([]byte, error) {
+	var prettyJSON bytes.Buffer
+	err := json.Indent(&prettyJSON, src, "", "\t")
```
UnseenWizzard, line 116: out of curiosity: why indent with tabs rather than spaces? Probably doesn't make any difference, but later on the JSON template will be re-indented with spaces.

pkg/download/automation/automation_test.go
```diff
@@ -28,31 +28,86 @@ import (
...
+				wfData, _ := os.ReadFile("./testdata/listWorkflows.json")
+				rw.Write(wfData)
+			case "/platform/automation/v1/business-calendars":
+				wfData, _ := os.ReadFile("./testdata/listBusinessCals.json")
+				rw.Write(wfData)
+			case "/platform/automation/v1/scheduling-rules":
+				wfData, _ := os.ReadFile("./testdata/listSchedulingRules.json")
+				rw.Write(wfData)
+			default:
+				t.Fatal("NO")
```
UnseenWizzard, line 44: nit: Not your change, but could we change this to a more useful message please? Just picturing a test failing with "NO" and trying to figure out what exactly when wrong 😅 

e.g.
```suggestion
				t.Fatal("Unexpected API call to %s", req.URL.Path)
```

pkg/download/automation/internal/jinja.go
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package internal
+
+import "bytes"
+
+// EscapeJinjaTemplates replaces each occurrence of "{{" with "\{\{" and each occurrence of "}}" with "\}\}"
```
UnseenWizzard, line 21: Comment is outdated. Also it would be great if it explains why the "pre-escaping" is needed :) 

Also is my assumption correct, that we don't actually need the `{{` -> `\{\{` escaping and just do it for completeness?
We'll only run into issues with the closing bracket escaping, as it would wrongly also escape the closing bracket of the previously added `{{'{{'}}`, right?

pkg/download/automation/internal/jinja_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestEscapeJinjaTemplates(t *testing.T) {
+	assert := assert.New(t)
+
+	assert.Equal([]byte("Hello, {{`{{`}}planet{{`}}`}}!"), EscapeJinjaTemplates([]byte(`Hello, {{planet}}!`)))
```
UnseenWizzard, line 27: nit: would personally prefer a table-driven test over a single one with many asserts. For me that makes finding which specific case has failed and isolating what's going wrong much easier. 
(Likely not an issue here as it's all the same 'context' - so I'd be fully fine if this stays as is)

cmd/monaco/download/download_command.go
jskelin: It was convenient for me to implement it to make manual testing easier.

I didn't see that @warber also did it.

pkg/download/automation/automation.go
jskelin: changed

pkg/download/automation/automation_test.go
jskelin: done

cmd/monaco/download/download_command.go
warber: closed #1003 in favor of this

cmd/monaco/download/download_configs.go
```diff
@@ -240,14 +244,16 @@ func copyConfigs(dest, src project.ConfigsPerType) {
...
 
 // shouldDownloadSettings returns true unless onlyAPIs or specificAPIs but no specificSchemas are defined
 func shouldDownloadSettings(opts downloadConfigsOptions) bool {
-	return !opts.onlyAPIs && (len(opts.specificAPIs) == 0 || len(opts.specificSchemas) > 0)
+	return !opts.onlyAutomation && !opts.onlyAPIs && (len(opts.specificAPIs) == 0 || len(opts.specificSchemas) > 0)
 }
 
-func shouldDownloadAutomationResources() bool {
-	return featureflags.AutomationResources().Enabled()
+func shouldDownloadAutomationResources(opts downloadConfigsOptions) bool {
```
UnseenWizzard, line 255: as this PR won out, could you please add some test coverage for the new logic? The other PR included some tests - maybe they can just be transferred over :)

pkg/download/automation/internal/jinja.go
jskelin: done

Laubi:
Hi, I created another test case that fails: `{}}`.
The reason is, that we replace it with this one:
```
{{{`}}`}}
^^ this is then the templating brackets
  ^ invalid character
```

pkg/download/automation/automation.go
UnseenWizzard: It might be more consistent if someone just looks for "Indent" via search. 
I don't think it has any performance impact if we'd indent in a way that does not actually need to be re-indented. 

Fine for me to leave as is

PR #1010, opened 2023-05-17
Refactor Client instantiation and unify usage between different commands/functions

UnseenWizzard:
refactor(client): Introduce a single ClientSet to remove instantiation logic out of specific cmds or pkgs

Previously commands and packages shared some logic on instantiation of the base DynatraceClient used to access
Classic Config and Settings APIs, but all implemented specicif setups of the new AutomationClient.

This takes the ClientSet concept that was introduced for the deploy package and moves it to the client/ package.
Additionally it unifies the client setup - in some places we added optional version checks, while in others we did
not. This also means that the concurrency limiter so far only applied to downloads is now on all clients - which might
be helpful/ensure we can't forget about it when adding deploy concurrency.

Wherever an API client is used ('function' packages) a dedicated ClientSet struct is introduced and prepared for a
possible followup to introduce dedicated interfaces per consumer.
For now all pkgs retain usage of the DynatraceClient 'fat interface' and only introduce a dedicated AutomationClient
interface for theit usecases.

Before the actual refactoring, an initial commit splits pkg/client into several sub-packages. 
This is done mostly to ensure some changes in the actual PR do not lead to cyclic dependencies from importing all of pkg/client (specifically: client/clientset.go -> client/dtclient/client -> client/[some auth and error methods])

Nope


warber (approved):

UnseenWizzard:
note: as before I'm neglecting the linter warning about code duplication in cmd/dynatrace/dynatrace.go
While two methods in it currently handle errors exactly the same, there's no reason to tightly couple them by making them share error handling code

PR #1012, opened 2023-05-19
fix: User defined preset dashboards are ignored

warber:
Preset Dashboards that are "pre-installed" and owned by Dynatrace are ignored during a download using Monaco, as these Dashboards cannot be modified by the user at all and re-deploying them will result in an error. 

However, we should not ignore preset dashboards defined by the user.
This PR changes the filter for persisting dashboards to skip dashboards only if they are presets AND owned by "Dynatrace"

Users are able to download dashboards marked as "preset" if they were created ba themselves

### DIFF
--- a/cmd/monaco/download/test-resources/integration-test-dashboard/dashboard/id-5.json
+++ b/cmd/monaco/download/test-resources/integration-test-dashboard/dashboard/id-5.json
@@ -1,0 +1,9 @@
+{
+    "id": "id-5",
+    "dashboardMetadata": {
+        "name": "Dashboard which is a preset by Dynatrace",
+        "owner": "Dynatrace",
+        "preset": true
+    },
+    "tiles": []
+}
--- a/pkg/download/classic/filter.go
+++ b/pkg/download/classic/filter.go
@@ -40,7 +40,7 @@ var apiContentFilters = map[string]contentFilter{
 			if json["dashboardMetadata"] != nil {
 				metadata := json["dashboardMetadata"].(map[string]interface{})
 
-				if metadata["preset"] != nil && metadata["preset"] == true {
+				if metadata["preset"] != nil && metadata["preset"] == true && metadata["owner"] == "Dynatrace" {
 					return false
 				}
 			}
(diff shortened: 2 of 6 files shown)


pkg/download/classic/filter.go
```diff
@@ -40,7 +40,7 @@ var apiContentFilters = map[string]contentFilter{
 			if json["dashboardMetadata"] != nil {
 				metadata := json["dashboardMetadata"].(map[string]interface{})
 
-				if metadata["preset"] != nil && metadata["preset"] == true {
+				if metadata["preset"] != nil && metadata["preset"] == true && metadata["owner"] == "Dynatrace" {
```
UnseenWizzard, line 43: So can non-Dynatrace presets be edited/modified?
Or are all presets 'create-only' rather than fully unmodifiable? If that's the case we should probably skip them on deploy if they exist on the environment, rather than ignore on download 🤔

warber: as far as I can see, user-defined preset dashboards are just as modifiable as non-preset dashboards

warber:
E2E test failing because of current issues in test setup wrt. automation resources

cmd/monaco/download/test-resources/integration-test-dashboard/dashboard/id-5.json
```diff
@@ -0,0 +1,9 @@
+{
+    "id": "id-4",
```
UnseenWizzard, line 2: Now that you fixed the other one, how about `id-5` ;)

pkg/download/classic/filter.go
UnseenWizzard: One more stupid question: are Dynatrace presets actually unmodifable on the API? Iirc this filter is quite old and might have been based on their immutability in the UI only

UnseenWizzard (approved):
Just nitpicks and questions. Changes lgtm!

pkg/download/classic/filter.go
UnseenWizzard: thx for checking

UnseenWizzard (approved):

PR #1018, opened 2023-05-23
feat(download): Skip unmodifiable settings objects 

UnseenWizzard:
Some Settings 2.0 objects can not be modified - e.g. Dynatrace default configurations.
When downloading these, the resulting project can not be deployed again successfully
as trying to change these unmodifable objects will fail.

As the Settings API now reports a modificationInfo object if requested, the Settings
client is extended to query and return this info.
Older Dynatrace versions may not return this field yet, thus it is handled as a pointer
to clearly indicate that it may not be present on a downloaded settings object.

The Settings downloader is extended to filter out unmodifiable Settings instead of
persisting them.

-

- Unmodifiable Settings won't be downloaded anymore
- Filtered out Settings are no longer shown as a warning (moved to Debug/verbose logs)


warber (dismissed):

pkg/client/dtclient/client.go
```diff
@@ -84,12 +84,21 @@ type ConfigClient interface {
...
+	Scope            string                    `json:"scope"`
+	Value            json.RawMessage           `json:"value"`
+	ModificationInfo *SettingsModificationInfo `json:"modificationInfo"`
+}
+
+type SettingsModificationInfo struct {
+	Deletable          bool          `json:"deletable"`
+	Modifiable         bool          `json:"modifiable"`
+	Movable            bool          `json:"movable"`
+	ModifiablePaths    []interface{} `json:"modifiablePaths"`
```
Laubi, line 100: I'd argue we don't need most of the fields here, even in near the future.

UnseenWizzard: Agreed, I wasn't sure about it either.

I've kept it in from the 'as-a-lib' perspective. 
I figure it has no real drawbacks to keep and doesn't hurt for the client to be aware of the full object.

PR #1020, opened 2023-05-23
fix(config): allow override of `OriginObjectId`

patrickpichler:
The override logic was missing logic required to allow overriding the `OriginObjectId`. The missing logic has been provided.

Fixes the override logic to also take the `OriginObjectId` into consideration. 


warber (approved):
lgtm

PR #1025, opened 2023-05-25
feat: Add structured logging capabilities

warber:
In order ease the processing of Monaco logs we need to support structured log output.
This PR introduces the popular zap logging library that will be used to output logs in JSON format.

The "old" log package was replaced by a newer package containing a dedicated `Logger` interface and two concrete implementations, "console" and "zap".
Per default, monaco still logs to the console in the unstructured text format as well as in a file in the `.logs` directory.

I also noticed that the timestamp for the log file name and the timestamp included in the download directory did not match and included a fix in this PR.

Further, time zones are now visible in the time stamps.

Last but not least I've moved the code for "traffic logging" from the log package into a dedicated "trafficlog" package, as these are not regular logs but more like a report of what http requests/responses have been made by monaco.

* Log output now contains a timestamp formatted with RFC3339 (which includes the time zone)
* User can switch to "structured" logging by setting `MONACO_LOG_FORMAT=json` (default = text) 
image
* User can switch to UTC-based timestamps by setting `MONACO_LOG_TIME=utc` (default = local)
image

These settings affects both logs written to the console as well as to the log file.


cmd/monaco/deploy/logging.go
```diff
@@ -27,10 +27,7 @@ func logProjectsInfo(projects []project.Project) {
 	for _, p := range projects {
 		log.Info("  - %s", p)
 	}
-
-	if log.DebugEnabled() {
```
warber, line 31: yes, I've deleted this check.
if you think this is absolutely needed I would probably need to add such a method to get to the log level to the logger interface.... also it feels not 100% correct to base this check on the logger configuration

internal/loggers/console/console.go
```diff
@@ -0,0 +1,48 @@
...
+type Logger struct {
+	consoleLogger *log.Logger
+}
+
+func (l Logger) Info(msg string, args ...interface{}) {
+	l.consoleLogger.Printf("INFO "+msg+"\n", args...)
+}
+
+func (l Logger) Error(msg string, args ...interface{}) {
+	l.consoleLogger.Printf("INFO "+msg+"\n", args...)
```
UnseenWizzard, line 35: console logger currently logs everything with "INFO" prefix

internal/loggers/loggers.go
```diff
@@ -0,0 +1,90 @@
...
+package loggers
+
+import (
+	"github.com/spf13/afero"
+	"io"
+	"strings"
+)
+
+// Logger defines the interface for a logging implementation
+type Logger interface {
```
UnseenWizzard, line 26: nit: We might want to add a compile time type assertion for zap and console logger? (var _ Logger = ...) 
Just to be safe for future extensions

internal/loggers/console/console.go
```diff
@@ -0,0 +1,48 @@
...
+
+package console
+
+import (
+	"log"
+)
+
+var Instance = Logger{consoleLogger: log.Default()}
+
+// Logger is a wrapper around the go default logger that just logs to the console
```
UnseenWizzard, line 25: I think it would be a good place to mention that this one's generally unused in production, as zap is also used for plain/non-JSON logging.

UnseenWizzard (commented):
lgtm overall, just some small comments

cmd/monaco/deploy/logging.go
UnseenWizzard: IIRC I was the one to add it, and I can generally live with that. 
But I added the guard as it's purely debug logging and the operation is somewhat heavy, especially for large config sets (nested loop O(n*m*k), n..projects, m...environments, k..configs).

Open to discussing the necessity/practice of debug logging guards, but in general, I think it's a good practice and might be worth re-adding to our logging interface. 
(btw: exposing the level/whether a level below INFO is active via a method is what typical Java logging frameworks do, that's probably why it does feel like the right way to expose the information me.)

warber: I will add it :)

PR #1029, opened 2023-05-30
feat: add pagination in automation client for download action

jskelin:
GET endpoints for automation for a bigger result set only returns a limited subset. To be able to get all records, the client needs to exploit the `offset` request parameter and repeat requests until all records from DT are gotten.

### DIFF
--- a/pkg/client/automation/client.go
+++ b/pkg/client/automation/client.go
@@ -27,8 +27,8 @@ import (
 
 // Response is a "general" Response type holding the ID and the response payload
 type Response struct {
-	// Id is the identifier that will be used when creating a new automation object
-	Id string `json:"id"`
+	// ID is the identifier that will be used when creating a new automation object
+	ID string `json:"id"`
 	// Data is the whole body of an automation object
 	Data []byte `json:"-"`
 }
@@ -39,7 +39,7 @@ func (r *Response) UnmarshalJSON(data []byte) error {
 	if err := json.Unmarshal(data, &rawMap); err != nil {
 		return err
 	}
-	if err := json.Unmarshal(rawMap["id"], &r.Id); err != nil {
+	if err := json.Unmarshal(rawMap["id"], &r.ID); err != nil {
 		return err
 	}
 	r.Data = data
@@ -166,7 +166,7 @@ func (a Client) upsert(resourceType ResourceType, id string, data []byte) (*Resp
 	if resp.IsSuccess() {
 		log.Debug("Updated object with ID %s", id)
 		return &Response{
-			Id:   id,
+			ID:   id,
 			Data: resp.Body,
 		}, nil
 	}
@@ -213,7 +213,7 @@ func (a Client) create(id string, data []byte, resourceType ResourceType) (*Resp
 	}
 
 	// check if id from response is indeed the same as desired
-	if e.Id != id {
+	if e.ID != id {
 		return nil, fmt.Errorf("returned object ID does not match with the ID used when creating the object")
 	}
 	log.Debug("Created object with ID %s", id)
--- a/pkg/client/automation/client.go
+++ b/pkg/client/automation/client.go
@@ -109,14 +109,14 @@ func WithClientRequestLimiter(limiter *concurrency.Limiter) func(client *Client)
 }
 
 // List returns all automation objects
-func (a Client) List(resourceType ResourceType) (result *ListResponse, err error) {
+func (a Client) List(resourceType ResourceType) (result *[]Response, err error) {
 	a.limiter.ExecuteBlocking(func() {
 		result, err = a.list(resourceType)
 	})
 	return
 }
 
-func (a Client) list(resourceType ResourceType) (*ListResponse, error) {
+func (a Client) list(resourceType ResourceType) (*[]Response, error) {
 	// try to get the list of resources
 	resp, err := rest.Get(a.client, a.url+a.resources[resourceType].Path)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


pkg/client/automation/client.go
```diff
@@ -109,36 +111,47 @@ func WithClientRequestLimiter(limiter *concurrency.Limiter) func(client *Client)
...
 	return
 }
 
-func (a Client) list(resourceType ResourceType) (*ListResponse, error) {
-	// try to get the list of resources
-	resp, err := rest.Get(a.client, a.url+a.resources[resourceType].Path)
-	if err != nil {
-		return nil, fmt.Errorf("unable to list automation resources: %w", err)
-	}
+func (a Client) list(resourceType ResourceType) (*[]Response, error) {
```
UnseenWizzard, line 121: does this still have to be a pointer? As a slice it can already be nil and the actual data of the underlying array is not copied if passed by value

pkg/client/automation/client.go
```diff
@@ -109,36 +111,47 @@ func WithClientRequestLimiter(limiter *concurrency.Limiter) func(client *Client)
...
+		}
+		retVal = append(retVal, result.Results...)
 	}
 
-	// unmarshal and return result
-	var result ListResponse
-	err = json.Unmarshal(resp.Body, &result)
-	if err != nil {
-		return nil, err
+	if len(retVal) != result.Count {
```
UnseenWizzard, line 151: We could technically get a case in which we have more return values than the last API result expects - if one was deleted between calls and the last `Count` ends up smaller than `retVal` + `results.Results`.

I think we should treat that case differently than getting fewer than expected. Possibly with just a warning log and returning the values normally. Wdyt?

pkg/client/automation/client.go
```diff
@@ -267,3 +280,18 @@ func rmIDField(data *[]byte) error {
 	}
 	return nil
 }
+
+func get(a Client, resourceType ResourceType, offset int) (rest.Response, error) {
+
+	params := url.Values{}
+	params.Add("offset", strconv.Itoa(offset))
+
+	u, _ := url.ParseRequestURI(a.url)
```
UnseenWizzard, line 289: please just return that error if happens - even if it's pretty unlikely to occur, we should IMO not omit errors if we could as well handle them

UnseenWizzard, line 289: Was there a reason to use `ParseRequestURI`? 
I'd just use `url.Parse` instead, seeing as this URL does not actually fulfill the assumption that the 'url was received in an HTTP request'

pkg/client/automation/client.go
```diff
@@ -267,3 +280,18 @@ func rmIDField(data *[]byte) error {
...
+	u.Path = a.resources[resourceType].Path
```
UnseenWizzard, line 290: I'm not 100% sure, but I think this could cause issues in some Managed or other proxy cases where the configured environment URL isn't just a host. 

If I'm not mistaken this would result in the created URL containing only the `host` part of the `client.url` value - possibly losing any path that already contains. 

I think just concatenating the URL and resource Path as was done before or via path.Join would be safer

pkg/client/automation/client.go
jskelin: done

pkg/client/automation/client.go
jskelin: done

pkg/client/automation/client.go
jskelin: Good observation.
changed to warn

pkg/client/automation/client.go
jskelin: done

pkg/client/automation/client.go
jskelin: The new logic has been implemented. It is now located in the `internal` package. Please take a look!

pkg/client/automation/client.go
```diff
@@ -109,36 +110,53 @@ func WithClientRequestLimiter(limiter *concurrency.Limiter) func(client *Client)
...
 
-	// handle http error
-	if !resp.IsSuccess() {
-		return nil, ResponseErr{
-			StatusCode: resp.StatusCode,
-			Message:    "Failed to list automation objects",
-			Data:       resp.Body,
+	for len(retVal) < result.Count {
+
+		u, err := internal.NextPageURL(a.url, a.resources[resourceType].Path, len(retVal))
```
UnseenWizzard, line 127: nit: I would give the sub-package a more speaking name, like `url`.

Or simply keep the `url.go` and `url_test.go` files in the `client/automation` package and un-export the method if it's only meant to be used here.

pkg/client/automation/client.go
```diff
@@ -109,36 +110,53 @@ func WithClientRequestLimiter(limiter *concurrency.Limiter) func(client *Client)
 }
 
 // List returns all automation objects
-func (a Client) List(resourceType ResourceType) (result *ListResponse, err error) {
+func (a Client) List(resourceType ResourceType) (result []Response, err error) {
```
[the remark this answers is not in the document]
jskelin: `ListResponse` is now unexported

pkg/client/automation/client.go
```diff
@@ -109,36 +110,53 @@ func WithClientRequestLimiter(limiter *concurrency.Limiter) func(client *Client)
...
+		retVal = append(retVal, result.Results...)
 	}
 
-	// unmarshal and return result
-	var result ListResponse
-	err = json.Unmarshal(resp.Body, &result)
-	if err != nil {
-		return nil, err
+	if len(retVal) != result.Count {
+		log.Warn("The number of expected (%d) and the number collected (%d) records from automation resource are different", result.Count, len(retVal))
```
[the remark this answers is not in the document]
jskelin: done

PR #1037, opened 2023-06-01
Ci/sign container images

Laubi:
To ensure the authenticity and integrity of our published container images, the images are now signed with a self-generated certificate.
This helps mitigate the risk of tampering and unauthorized modifications.

- Due to the nature of this change, details will be kept private.

Users can now use
`cosign verify --key cosign.pub <IMAGE-NAME>` to verify that we've signed the container.
The `cosign.pub` file will be attached to each GitHub release.


.ci/releasePipeline.groovy
```diff
@@ -136,15 +136,15 @@ pipeline {
...
+                                                    pushFileToGithub(rleaseName: release, source: "${release}/${release}", releaseId: releaseId)
+                                                    pushFileToGithub(rleaseName: release + ".sha256", source: "${release}/${release}.sha256", releaseId: releaseId)
                                                 } else {
                                                     echo "Skipping upload of binaries to GitHub"
                                                 }
                                                 break
                                             case "container":
                                                 // docker hub, only full releases and rc-releases
-                                                if (!isDevRelease) {
+                                                if (!isDevRelease) {//
```
UnseenWizzard, line 147:
```suggestion
                                                if (!isDevRelease) {
```

.ci/releasePipeline.groovy
```diff
@@ -279,19 +279,34 @@ void signWithSignService(Map args = [source: null, version: null, destDir: '.',
...
                         sh 'docker login --username $username --password $password $registry'
                         sh 'DOCKER_BUILDKIT=1 make docker-container OUTPUT=./build/docker/monaco CONTAINER_NAME=$registry/$repo/dynatrace-configuration-as-code VERSION=$version'
-                        sh 'docker push $registry/$repo/dynatrace-configuration-as-code:$version'
+
+                        pushAndSign(imageName: '$registry/$repo/dynatrace-configuration-as-code:$version')
+
                         if (args.tagLatest) {
                             sh 'docker tag $registry/$repo/dynatrace-configuration-as-code:$version $registry/$repo/dynatrace-configuration-as-code:latest'
-                            sh 'docker push $registry/$repo/dynatrace-configuration-as-code:latest'
+                            pushAndSign(imageName: '$registry/$repo/dynatrace-configuration-as-code:latest')
```
UnseenWizzard, line 319: Technically it should not be necessary to sign here? 
Unless I misunderstand what cosign signs. 

This is just a `tag` of the same image - if I understand it correctly, cosign will sign the image based on the sha disgest. The 'latest' tag should point to exactly the same sha, so it should already be signed.

Makefile
```diff
@@ -2,7 +2,7 @@ BINARY_NAME ?= monaco
 VERSION ?= 2.x
 RELEASES = $(BINARY_NAME)-windows-amd64.exe $(BINARY_NAME)-windows-386.exe $(BINARY_NAME)-linux-arm64 $(BINARY_NAME)-linux-amd64 $(BINARY_NAME)-linux-386 $(BINARY_NAME)-darwin-amd64 $(BINARY_NAME)-darwin-arm64
 
-.PHONY: lint format mocks build install clean test integration-test integration-test-v1 test-package default add-license-headers compile build-release $(RELEASES)
+.PHONY: lint format mocks build install clean test integration-test integration-test-v1 test-package default add-license-headers compile build-release $(RELEASES) docker-container
```
UnseenWizzard, line 5:
```suggestion
.PHONY: lint format mocks build install clean test integration-test integration-test-v1 test-package default add-license-headers compile build-release $(RELEASES) docker-container sign-image
```

UnseenWizzard (approved):
Looks good overall! 
I think the signing could be simplified/we sign the same thing twice at the moment - but I may have misunderstood how cosign works

.ci/releasePipeline.groovy
Laubi: image

You can see on DockerHub that the images, although they have the same 'content,' have different digests, so we need to sign them each time.

Interestingly, not every registry behaves the same. For docker, it's a different digest, while in our internal registry, it's the same.

UnseenWizzard: Huh, interesting...
Does 2.1.1 already point to the same image for the latest? I can't recall when exactly I fixed that from being a dedicated image to just a tag

UnseenWizzard: Ah.. nope `2.1.1` and `latest` on Dockerhub are still two different images: 

```sh
> docker run dynatrace/dynatrace-configuration-as-code:latest version
Unable to find image 'dynatrace/dynatrace-configuration-as-code:latest' locally
latest: Pulling from dynatrace/dynatrace-configuration-as-code
Digest: sha256:b01bcb0cdf6efcd938b699822f2a9ffddafb79c5a1ea9d99dc51ddd70cb7838c
[...]
monaco version latest
```

```sh
> docker run dynatrace/dynatrace-configuration-as-code:2.1.1 version
Unable to find image 'dynatrace/dynatrace-configuration-as-code:2.1.1' locally
2.1.1: Pulling from dynatrace/dynatrace-configuration-as-code
Digest: sha256:00e60cb1fe78fe0dbf89db03ddbde5ddbe181fa8c86e48d878b4cad21b2e8e7a
[...]
monaco version 2.1.1
```

Laubi: Ah, you are right. I reverted the line to push only. Since we sign it before it will be signed and pushed :)

.ci/releasePipeline.groovy
```diff
@@ -302,6 +327,12 @@ void createContainerAndPushToStorage(Map args = [version: null, tagLatest: false
     }
 }
 
+void pushAndSign(Map args = [imageName: null]) {
```
UnseenWizzard, line 330: nit: now that it's only used once this might just stay part of the overall createContainer method? It heavily depends on the other parts of that method having built the container

UnseenWizzard (approved):
one more nitpick ;)

UnseenWizzard (approved):

PR #1038, opened 2023-06-02
fix: conversion of environment variables refs does not consider compound cases

warber:
Fixes a bug while converting v1 config with multiple env var references to a v2 config, e.g.:
```yaml
- name: "{{ .Env.alertgroup }} - {{ .Env.application_name}}"
``` 

 produces only a single Environment type parameter for {{alertgroup}} and loses the {{application_name}} variable reference.

The fix contains the code to detect multiple environment variable references and converts them into a compound parameter, e.g.:
```yaml
  - name: "{{ .Env.GOPATH }} - {{ .Env.GOROOT }}"
```
gets converted to:
```yaml
    name:
      format: '{{ .__ENV_GOPATH__ }} - {{ .__ENV_GOROOT__ }}'
      references:
      - __ENV_GOPATH__ 
      - __ENV_GOROOT__ 
      type: compound
    parameters:
      __ENV_GOPATH__ :
        name: GOPATH
        type: environment
      __ENV_GOROOT__ :
        name: GOROOT
        type: environment
```

Multiple env var refs in v1 config yaml files are correctly transformed into compound parameters in the resulting v2 config

### DIFF
--- a/pkg/converter/converter.go
+++ b/pkg/converter/converter.go
@@ -20,23 +20,23 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/regex"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/slices"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/api"
-	listParam "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/parameter/list"
-	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/converter/v1environment"
-	projectV1 "github.com/dynatrace/dynatrace-configuration-as-code/pkg/project/v1"
-	"regexp"
-	"strings"
-
 	configV2 "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/coordinate"
 	configErrors "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/errors"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/parameter"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/parameter/compound"
 	envParam "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/parameter/environment"
+	listParam "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/parameter/list"
 	refParam "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2/parameter/reference"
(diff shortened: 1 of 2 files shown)


UnseenWizzard (commented):
Does this work for environment variables in the config YAML as well? 
In v1 the config YAML is also templated - exclusively for environment variables. 
So the compound env var reference case can happen in the JSON as well as the YAML.

I'm not sure if that takes the same code path.

warber:
yeah I've tried (multiple) env vars in the JSON payload as well. The result is somewhat different of course as it is not converted into a compound variable but I think it is fine:

json:
image

v2 yaml:
image

pkg/converter/converter_test.go
```diff
@@ -54,6 +55,8 @@ func TestConvertParameters(t *testing.T) {
 	listParameterValue := `"GEOLOCATION-41","GEOLOCATION-42","GEOLOCATION-43"`
 	envParameterName := "url"
 	envParameterValue := " {{ .Env.SOME_ENV_VAR }} "
+	compundEnvParameterName := "nickname"
+	compondEnvParameterValue := " {{ .Env.TITLE }} {{ .Env.NICKNAME }} "
```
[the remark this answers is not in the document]
warber: it should not make a difference if n=2 or n=200 but i can certainly add another env var :)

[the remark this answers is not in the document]
UnseenWizzard: It does not matter until I go and change the converter in a month or two and accidentally break it for n > 2 😉

Thanks!

UnseenWizzard:
> yeah I've tried (multiple) env vars in the JSON payload as well. The result is somewhat different of course as it is not converted into a compound variable but I think it is fine:
> [...]

Perfect thanks! 
As discussed in person, I was thinking about this wrong - works perfectly then.

pkg/converter/converter.go
```diff
@@ -404,8 +404,35 @@ func convertParameters(context *configConvertContext, environment manifest.Envir
...
+						Project:  config.GetProject(),
+						Type:     config.GetType(),
+						ConfigId: config.GetId(),
+					},
+					Property: parameterName,
+				})
+			}
+
+			// convert to compound parameter
+			c, err := compound.New(newName, valueWithPlaceHolders, refs)
```
UnseenWizzard, line 430: For the case of a single variable that is not a compound, couldn't we still exit before adding the compound iff there's only a single param and the `valueWithPlaceholders` is exactly equal to `{{ .Env.[p.Name] }}`?

warber: yes, but why?

UnseenWizzard: I think creating a compound where none is needed at all can be confusing to users. 
And if it's an easy fix to not do this with a 3 line early exit if in the right place, I don't see why we should why we should do this only half-way and produce unnecessary complexity.

UnseenWizzard (approved):
Thanks for addressing all the comments 🎉

PR #1046, opened 2023-06-12
Fix nightly E2E tests & several small improvements to pagination handling

UnseenWizzard:
- fix a small mistake in the pagination E2E test assertion (804aff68f991fc43f7492ec79128ee61c05ce74b)
- fix when a warning is logged about mismatch between received paginated entries and expected total count

Review per commit. The actual fix is in the second commit, the rest are small improvements found along the way,

YES - no incorrect warning is logged if Settings filtering results in less objects being returned from the client's List operation.

### DIFF
--- a/pkg/rest/list.go
+++ b/pkg/rest/list.go
@@ -166,7 +166,7 @@ func validateRespErrors(isNextCall bool, err error, resp Response, urlPath strin
 		return false, err
 	}
 	isLastAvailablePage := false
-	if success(resp) {
+	if resp.IsSuccess() {
 		return false, nil
 
 	} else if isNextCall {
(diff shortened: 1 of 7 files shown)


pkg/rest/list.go
```diff
@@ -28,8 +28,16 @@ import (
 
 const emptyResponseRetryMax = 10
 
+// AddEntriesToResult is a function which should parse an API response body and append the returned entries to a result slice
+// Handling the parsing, any possible filtering and owning and filling the result list is left to the caller of ListPaginated,
+// as it might differ notably between client implementations.
+// The function MUST return the number of entires it has parsed from the received API payload body. This is used to validate
```
warber, line 34:
```suggestion
// The function MUST return the number of entries it has parsed from the received API payload body. This is used to validate
```

warber (dismissed):
lgtm

pkg/rest/list.go
UnseenWizzard: Thanks, correct on the commit that introduced it.

UnseenWizzard:
@warber I've addressed the typo (thx) via fixup and rebased on the changes from you Platform/Classic URL improvement. 
Please take another look/hit approve again :)

warber (approved):

PR #1047, opened 2023-06-12
Fix: give user hint about correct platform URL

warber:
Give the user a hint when (s)he is trying to target a platform enabled environment but gave classic environment URL to monaco

### DIFF
--- a/pkg/client/dtclient/client.go
+++ b/pkg/client/dtclient/client.go
@@ -316,7 +316,6 @@ func NewPlatformClient(dtURL string, token string, oauthCredentials auth.OauthCr
 
 	classicURL, err := metadata.GetDynatraceClassicURL(oauthClient, dtURL)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/client/dtclient/client.go
```diff
@@ -316,7 +316,6 @@ func NewPlatformClient(dtURL string, token string, oauthCredentials auth.OauthCr
 
 	classicURL, err := metadata.GetDynatraceClassicURL(oauthClient, dtURL)
 	if err != nil {
-		log.Error("Unable to determine Dynatrace classic environment URL: %v", err)
```
UnseenWizzard, line 319: I assume this is removed because the error is handled by logging further out as well? So handled twice?
Or is there some other reason to remove logging here? :)

warber: yeah it pops up twice

UnseenWizzard (approved):
lgtm 🚀

PR #1048, opened 2023-06-13
Feat/extract meids on download

UnseenWizzard:
Extracts monitored entity and uuids from downloaded JSONs into YAML parameters, to simplify manual (or automated) modification. 

This simplifies changing identifiers that we can't auto-replace with references to downloaded configurations, but which might still differ between target environments. (e.g. a HOST identifier if a given Service is running on different machines in different monitored environments).

- Note the somewhat ugly comment added to cmd/download to make it clear that id replacement must not be done before dependency resolution, as it removes IDs from the JSON also searched for by dep-resolution. 
If you have a better idea, please let me know.
- This goes one step above the scope of extracting monitored entity IDs, but UUIDs seemed like a reasonable addition. I assume extraction will likely be extended with further identifiers in the future.

Download contains more parameters, JSONs are more templated as meIDs and unknown UUIDs are now pulled into params.

### DIFF
--- a/pkg/download/id_extraction/id_extraction.go
+++ b/pkg/download/id_extraction/id_extraction.go
@@ -41,8 +41,7 @@ func ExtractIDsIntoYAML(configsPerType project.ConfigsPerType) project.ConfigsPe
 			ids = deduplicateIDs(ids)
 
 			for _, id := range ids {
-				idKey := strings.ReplaceAll(id, "-", "_") // golang template keys must not contain hyphens
-				paramID := fmt.Sprintf("__EXTRACTED_ID_%s__", idKey)
+				paramID := createParameterKey(id)
 				c.Parameters[paramID] = value.New(id)
 
 				newContent := strings.ReplaceAll(c.Template.Content(), id, fmt.Sprintf("{{ .%s }}", paramID))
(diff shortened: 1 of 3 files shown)


pkg/download/id_extraction/id_extraction.go
```diff
@@ -0,0 +1,62 @@
...
+func ExtractIDsIntoYAML(configsPerType project.ConfigsPerType) project.ConfigsPerType {
+	for _, cfgs := range configsPerType {
+		for _, c := range cfgs {
+			ids := meIDRegexPattern.FindAllString(c.Template.Content(), -1)
+			ids = append(ids, uuidRegexPattern.FindAllString(c.Template.Content(), -1)...)
+			ids = deduplicateIDs(ids)
+
+			for _, id := range ids {
+				idKey := strings.ReplaceAll(id, "-", "_") // golang template keys must not contain hyphens
+				paramID := fmt.Sprintf("__EXTRACTED_ID_%s__", idKey)
```
Laubi, line 45: Why the prefix? I get that we use the pattern for conversation, but for new properties I would use a simpler `key` only or `id_key`. 
The double underscores are IMO really not necessary.

Laubi: Maybe it would even make sense to differentiate between uuids and meids?
```
parameters:
  uuid_1_2_3_4: 1_2_3_4
  meid_host_1234: host_1234
```

i take it that the prefix was introduced for tools to simply detect the ids as 'monaco found' ids? Otherwise, i would have suggested to get rid of the prefix.

UnseenWizzard: You're right, I applied it based on the conversion convention. 
But unlike there, where it visually differentiates keys that existed before to generated ones, all are generated on download. 

I've simplified to to `extractedID_[...]`

Laubi: Works for me

UnseenWizzard: just pushed as refactor commit (too lazy to fixup again)

Laubi (approved):

PR #1054, opened 2023-06-20
feat: Introduce structured log metadata to Deployment logging (Non-Errors)

UnseenWizzard:
Introduced structured log metadata (env, coordinate) to non-error deployment logging.

Please review #1049 first!


pkg/download/entities/entities.go
```diff
@@ -68,7 +69,7 @@ func (d *Downloader) Download(specificEntitiesTypes []string, projectName string
 	log.Debug("Fetching specific entities types to download")
 
 	// get ALL entities types
-	entitiesTypes, err := d.client.ListEntitiesTypes()
+	entitiesTypes, err := d.client.ListEntitiesTypes(context.TODO()) //TODO: real context
```
Laubi, line 72: Not included in any followup PR

UnseenWizzard: Thanks! I'll address the context TODOs I've missed in https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1055 as the all related to download or delete usage!

warber:
lgtm :) @UnseenWizzard 

Two notes: 
* I see a lot of _direct_ `log.XXX` calls in packages that are potential candidates for living in a library scope(client,....). I think we need to get rid of this way of calling the logger there and make the logger injectable.
* Nit: some of the methods in the client interfaces have a context as the first argument, others do not. Maybe it is just worth adding context argument to the others as well (even if it is not used right now....) 

Both are things to look at later tho :)

warber (approved):

UnseenWizzard:
> Two notes:
> 
> -     I see a lot of direct log.XXX calls in packages that are potential candidates for living in a library scope(client,....). I think we need to get rid of this way of calling the logger there and make the logger injectable.
> -     Nit: some of the methods in the client interfaces have a context as the first argument, others do not. Maybe it is just worth adding context argument to the others as well (even if it is not used right now....)

Agree @warber, both make sense as a library improvement followup. 
Do you feel actually passing the logger would be reasonable? Else we pull the logger through all method interfaces...
It feels a bit off for the context as it's shared and not call-path/goroutine specific..

warber:
> Do you feel actually passing the logger would be reasonable? Else we pull the logger through all method interfaces...

I think it should be injectable via the constructor of the clients first, e.g., `automation.NewClient(url,c, WithLogger(...))` where possible and then either be used inside the struct directly or handed over to the functions that need to log something as an arg.
Also, if I think about it from a library perspective, it would be easier for the user if (s)he would not need to think about adding the right fields to the context but the library code sets up the logger with additional information inferred from the other arguments given to the called functionality. Things like that actually reinforce our first idea that maybe the library code shouldn't do any logging at all :man_shrugging: 

So i maybe we would need to reiterate a bit on the desired design and behavior when we start extracting out the library parts....

PR #1055, opened 2023-06-20
feat: Structured Logging: Download & Delete logs

UnseenWizzard:
Introduced structured log metadata (env, coordinate) to delete and download logs. 

Please review #1049 and #1054 first!


pkg/delete/delete.go
```diff
@@ -77,40 +78,40 @@ func Configs(clients ClientSet, apis api.APIs, automationResources map[string]co
...
 	}
 
 	return errs
 }
 
-func deleteClassicConfig(client dtclient.Client, theApi api.API, entries []DeletePointer, targetApi string) []error {
+func deleteClassicConfig(ctx context.Context, client dtclient.Client, theApi api.API, entries []DeletePointer, targetApi string) []error {
 	errors := make([]error, 0)
 
 	values, err := client.ListConfigs(context.TODO(), theApi) //TODO: real context
```
Laubi, line 99: Overlooked the todo

pkg/download/settings/settings.go
```diff
@@ -117,7 +118,7 @@ func (d *Downloader) download(schemas []string, projectName string) v2.ConfigsPe
 	for _, schema := range schemas {
 		go func(s string) {
 			defer wg.Done()
-			log.Debug("Downloading all settings for schema %s", s)
+			log.WithFields(loggers.F("type", s)).Debug("Downloading all settings for schema %s", s)
 			objects, err := d.client.ListSettings(context.TODO(), s, dtclient.ListSettingsOptions{}) //TODO: real context
```
Laubi, line 122: Overlooked

pkg/download/classic/downloader.go
```diff
@@ -265,13 +266,13 @@ func (d *Downloader) createTemplate(mappedJson map[string]interface{}, value dtc
...
-		log.Debug("\tFetching singleton-configuration '%v'", currentApi.ID)
+		log.WithFields(loggers.TypeF(currentApi.ID)).Debug("\tFetching singleton-configuration '%v'", currentApi.ID)
 
 		// singleton-config. We use the api-id as mock-id
 		singletonConfigToDownload := dtclient.Value{Id: currentApi.ID, Name: currentApi.ID}
 		return []dtclient.Value{singletonConfigToDownload}, nil
 	}
-	log.Debug("\tFetching all '%v' configs", currentApi.ID)
+	log.WithFields(loggers.TypeF(currentApi.ID)).Debug("\tFetching all '%v' configs", currentApi.ID)
 	return d.client.ListConfigs(context.TODO(), currentApi) //TODO: real context
```
Laubi, line 276: Overlooked?

pkg/delete/delete.go
Laubi, line 1: Also overlooked on l 137

pkg/delete/delete.go
```diff
@@ -317,16 +320,16 @@ func AllAutomations(ctx context.Context, c automationClient) []error {
...
 		if err != nil {
 			errs = append(errs, err)
 			continue
 		}
 
-		log.Info("Deleting %d Automations of type %s...", len(objects), resource)
+		log.FromCtx(ctx).WithFields(loggers.TypeF(string(resource))).Info("Deleting %d Automations of type %s...", len(objects), resource)
 		for _, o := range objects {
-			log.Debug("Deleting Automation object with id %q...", o.ID)
+			log.FromCtx(ctx).WithFields(loggers.TypeF(string(resource)), loggers.F("object", o)).Debug("Deleting Automation object with id %q...", o.ID)
```
Laubi, line 332: objectId maybe?

UnseenWizzard: Here and in other cases I've opted to include the full object as a field, rather than just the ID or parts of it :) 

The content of "object" is `o`, so the full object struct, rather than just the ID

UnseenWizzard: I'm not 100% sure how helpful that is, but it seemed like an interesting thing to include in the logs. 
I also think that for future simplified debug information sharing we should write structured logs, as they contain so much possibly helpful metadata now!

pkg/delete/delete.go
UnseenWizzard: fixed

UnseenWizzard:
@Laubi I've addressed the forgotten context TODOs! Thanks for catching that. 

Note that in the Download case I've simply dropped the comment and still pass a "context.TODO()", which will result in structured logs getting zero fields added. 
Download has neither of the fields available that Deploy passes via context, no Coordinate exists (yet) and the environment is only know by its URL, not name or group. 

For Delete the context only contains the environment, but is now correctly passed. 

I've also caught a forgotten use of context.TODO where an actual context is available and amended that in #1054

internal/log/field/fields.go
warber, line 1: nit: pls align the name of the file with the package name.
also, I would suggest to change the package name to "field**s**"

warber (approved):
lgtm just some nits

internal/log/field/fields.go
```diff
@@ -47,6 +47,12 @@ func Coordinate(coordinate coordinate.Coordinate) Field {
 		}}
 }
 
+// Type builds a Field containing information about a config type. This is used in cases where no full coordinate exists,
+// but only a config type is known - for example in download or deletion
+func Type(t string) Field {
```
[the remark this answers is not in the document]
UnseenWizzard: As they all add one `field` I find `field.Type(...)` or `field.Environment` reads pretty nice and clearly.
If they added several `fields` based on some input I'd agree that `fields.From...` seems better, but no current method adds more than one `field`.

(Discussing based on how it reads because nitpick)

internal/log/field/fields.go
UnseenWizzard: done, thanks. As per comment above, personally I think singular actually fits here

warber (approved):

PR #1057, opened 2023-06-21
Improve logging for deprecated classic endpoints

jskelin:
When the `download` is used in a basic way (download all configs), for deprecated classical endpoints `monaco`  print a message that suggests that some of the configurations won't be downloaded. 
With this change that is changed in a way that now is more detailed message is print.


UnseenWizzard:
@jskelin something seems pretty wrong still, the test asserting that unknown requested APIs result in an error crashes with a segfault. 

Looking at the code, the parameter that the test sets seems to have changed meaning (called `ignore` now) AND is actually unused.

I assume you've moved the filtering/checking to different place. 
If the test has been replaced already, please remove it, else please make sure there is a test that covers that a user requesting a "specific-api" that does not exist, cleanly returns an error. 

And please look into the unused parameter. Should it be removed? Or used? 

```
=== FAIL: pkg/download/classic TestDownloadSpecific_ReturnsErrorIfUnknownAPIsAreRequested (unknown)
2023/06/21 09:05:30 DEBUG APIs to download: 
 - API_ID_1
 - API_ID_2
2023/06/21 09:05:30 DEBUG Fetching configs to download
2023/06/21 09:05:30 DEBUG Started all downloads
2023/06/21 09:05:30 DEBUG 	Fetching all 'API_ID_2' configs
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x925759]

goroutine 89 [running]:
github.com/dynatrace/dynatrace-configuration-as-code/pkg/download/classic.(*Downloader).findConfigsToDownload(0xc00006d3e0, {{0x9d0046, 0x8}, {0x9d0dd6, 0xa}, {0x0, 0x0}, 0x0, 0x0, {0x0, ...}, ...})
	/home/runner/work/dynatrace-configuration-as-code/dynatrace-configuration-as-code/pkg/download/classic/downloader.go:219 +0x319
github.com/dynatrace/dynatrace-configuration-as-code/pkg/download/classic.(*Downloader).downloadAPIs.func1()
	/home/runner/work/dynatrace-configuration-as-code/dynatrace-configuration-as-code/pkg/download/classic/downloader.go:89 +0x145
created by github.com/dynatrace/dynatrace-configuration-as-code/pkg/download/classic.(*Downloader).downloadAPIs
	/home/runner/work/dynatrace-configuration-as-code/dynatrace-configuration-as-code/pkg/download/classic/downloader.go:87 +0x479
```

cmd/monaco/download/download_configs.go
```diff
@@ -171,6 +181,19 @@ type downloadConfigsOptions struct {
 	onlyAutomation  bool
 }
 
+func (opts downloadConfigsOptions) valid() bool {
+	retVal := true
+	knownEndpoints := api.NewAPIs()
+	for _, e := range opts.specificAPIs {
+		if !knownEndpoints.Contains(e) {
+			log.Warn("unknown (or unsupported) classic endpoint with name %q provided via \"--api\" flag. More about supported classic endpoints can be found in documentation", e)
```
UnseenWizzard, line 189: this should be logged as an Error as we handle it as an error and don't continue if it happens

UnseenWizzard:
general comment: Please check the log messages for typos and wording, I found a few typos in several log lines

cmd/monaco/download/download_configs.go
```diff
@@ -124,6 +125,11 @@ func (d DefaultCommand) DownloadConfigsBasedOnManifest(fs afero.Fs, cmdOptions d
 		onlyAutomation:  cmdOptions.onlyAutomation,
 	}
 
+	if errs := options.valid(); len(errs) != 0 {
+		err := printAndFormatErrors(errs, "input data are not valid")
```
UnseenWizzard, line 129:
```suggestion
		err := printAndFormatErrors(errs, "command options are not valid")
```

Input data sounds pretty vague

cmd/monaco/download/download_configs.go
```diff
@@ -132,11 +138,11 @@ func (d DefaultCommand) DownloadConfigsBasedOnManifest(fs afero.Fs, cmdOptions d
 }
 
 func (d DefaultCommand) DownloadConfigs(fs afero.Fs, cmdOptions downloadCmdOptions) error {
-	a, errors := cmdOptions.auth.mapToAuth()
-	errors = append(errors, validateParameters(cmdOptions.environmentURL, cmdOptions.projectName)...)
+	a, errs := cmdOptions.auth.mapToAuth()
+	errs = append(errs, validateParameters(cmdOptions.environmentURL, cmdOptions.projectName)...)
```
UnseenWizzard, line 142: nit: could we rename `validateParameters` to make it clear it's specific to direct download? 
My initial though was that it should be part of the options.validate() function now, but I think it does not actually apply to the Manifest based download, right?

cmd/monaco/download/download_configs.go
```diff
@@ -196,12 +219,12 @@ func doDownloadConfigs(fs afero.Fs, downloaders downloaders, opts downloadConfig
...
 
-	if shouldDownloadClassicConfigs(opts) {
-		classicAPIs := makeClassicAPIs(opts.specificAPIs)
-		classicCfgs, err := downloaders.Classic().Download(opts.projectName, classicAPIs...)
+	{
+		classicCfgs, err := downloaders.Classic().Download(opts.projectName)
 		if err != nil {
 			return nil, err
 		}
+		log.Info("downloaded %q configurations from classic endpoints", len(classicCfgs))
```
UnseenWizzard, line 227:
```suggestion
		log.Info("downloaded %q configurations from classic Config API endpoints", len(classicCfgs))
```

cmd/monaco/download/download_configs_test.go
```diff
@@ -316,34 +316,6 @@ func Test_shouldDownloadSettings(t *testing.T) {
 	}
 }
 
-func TestDownloadConfigsExitsEarlyForUnknownAPI(t *testing.T) {
```
UnseenWizzard, line 319: I might be missing it, but is there a new test covering the functionality now?
We still fail on unknown APIs and IMO it's worth having a test at this level - after all we require known APIs from this level down.

pkg/download/classic/downloader.go
```diff
@@ -49,83 +49,37 @@ type Downloader struct {
...
-		d.apisToDownload = apis
-	}
-}
-
-// NewDownloader creates a new Downloader
-func NewDownloader(client dtclient.Client, opts ...func(*Downloader)) *Downloader {
+// NewDownloader creates a new sound Downloader.
+func NewDownloader(client dtclient.Client, opts ...func(*Downloader) error) (*Downloader, error) {
 	c := &Downloader{
-		apisToDownload:    api.NewAPIs(),
```
UnseenWizzard, line 68: shouldn't the default still be `api.NewAPIs()`?
As `WithEndpoints` is optional, the current default initialized downloader has no APIs to download at all.

pkg/download/classic/downloader.go
```diff
@@ -49,83 +49,37 @@ type Downloader struct {
...
 }
 
-// WithAPIContentFilters sets the api content filters for the Downloader - these will be used in addition to base API filtering
-func WithAPIContentFilters(apiFilters map[string]contentFilter) func(*Downloader) {
-	return func(d *Downloader) {
-		d.apiContentFilters = apiFilters
-	}
-}
-
-func WithAPIs(apis api.APIs) func(*Downloader) {
```
UnseenWizzard, line 59: nit: Technically WithAPIs and the new WithEndpoints have the same meaning, they set d.apisToDownload.
Why the rename?

pkg/download/classic/downloader.go
```diff
@@ -49,83 +49,37 @@ type Downloader struct {
 	client dtclient.Client
 }
 
-// WithAPIContentFilters sets the api content filters for the Downloader - these will be used in addition to base API filtering
-func WithAPIContentFilters(apiFilters map[string]contentFilter) func(*Downloader) {
```
UnseenWizzard, line 53: Should this option really be removed?
Seemingly they're only used in tests right now as production code defaults, but on an interface level these options make sense. I'd keep it, rather than moving it into a non-exported test only method

cmd/monaco/download/downloaders_test.go
```diff
@@ -77,3 +77,94 @@ func TestGetDownloaderPanic(t *testing.T) {
...
+				given:      downloadConfigsOptions{onlySettings: true},
+				deprecated: false,
+			},
+			{
+				name:       "onlyAPIs",
+				given:      downloadConfigsOptions{onlyAPIs: true},
+				deprecated: false,
+			},
+			{
+				name:       "specificAPIs is marked with 'deprecatedBy'",
```
UnseenWizzard, line 140: nit
```suggestion
				name:       "specificAPI marked with 'deprecatedBy' is not filtered out",
```

I think the name could be improved to make the meaning of the test case clearer. 
Compare the test method names in downloader_test that you've removed. I find those make it clearer what assumption is actually covered by the testcase

cmd/monaco/download/downloaders.go
```diff
@@ -33,28 +36,92 @@ func makeDownloaders(options downloadConfigsOptions) (downloaders, error) {
...
+	var automationDownloader download.Downloader[config.AutomationType] = dlautomation.NoopAutomationDownloader{}
 	if clients.Automation() != nil {
 		automationDownloader = dlautomation.NewDownloader(clients.Automation())
 	}
-	var settingsDownloader download.Downloader[v2.SettingsType] = settings.NewDownloader(clients.Settings())
-	var classicDownloader download.Downloader[v2.ClassicApiType] = classic.NewDownloader(clients.Classic())
-	return downloaders{settingsDownloader, classicDownloader, automationDownloader}, nil
+	var settingsDownloader download.Downloader[config.SettingsType] = settings.NewDownloader(clients.Settings())
+	var classicDow download.Downloader[config.ClassicApiType]
+	if d, e := classicDownloader(clients.Classic(), options); e != nil {
```
UnseenWizzard, line 45: this logic is inverted, currently if no error happens, a nil downloader and nil error are returned (else case).

I've run into this while trying it locally after commenting out the non-compiling tests. 
Please ensure this logic is properly covered by tests (if it's covered once the existing tests are fixed great, else please add test cases for the downloader initialization) !

UnseenWizzard, line 45:
```suggestion
	if d, e := classicDownloader(clients.Classic(), options); e == nil {
```

this logic is inverted, currently if no error happens, a nil downloader and nil error are returned (else case).

I've run into this while trying it locally after commenting out the non-compiling tests. 
Please ensure this logic is properly covered by tests (if it's covered once the existing tests are fixed great, else please add test cases for the downloader initialization) !

cmd/monaco/download/downloaders.go
```diff
@@ -33,28 +36,92 @@ func makeDownloaders(options downloadConfigsOptions) (downloaders, error) {
...
-	return getDownloader[v2.ClassicApiType](d)
+func skipDownloadFilter(api api.API) bool {
+	if api.SkipDownload {
+		log.Info("API can not be downloaded and needs manual creation: '%v'.", api.ID)
+		return true
+	}
+	return false
+}
+
+func removeDeprecated(log ...func(api api.API)) api.Filter {
```
UnseenWizzard, line 83: The functional options seems a bit overkill for this usecase, especially as you only ever apply the first (line 87).

Couldn't you achieve the same effect with a simple `logWarning bool` parameter?

UnseenWizzard (changes requested):
Tried the functionality in general, which works nicely! 
Moving the filtering to clear single place is a nice change as well 👍 

However please double check the logic and tests/test-coverage! 
Without changes no classic config API downloader will ever be created, and I'm not sure if the creation logic is covered by tests as some are currently not functional.

pkg/download/classic/downloader.go
jskelin: This doesn't have an impact on our code. It is more how we would like to design input for the library. 
I'll make `api.NewAPIs()` to be default

pkg/download/classic/downloader.go
jskelin: As you said, for now, the only purpose is in testing. For that reason I moved it to the test file.

It is moved back!

cmd/monaco/download/downloaders.go
jskelin: Just to be more readable on the caller side: `api.NewAPIs().Filter(removeSkipDownload, removeDeprecated(withWarn()))`

As it is an implementation detail, I'll leave it like it is

cmd/monaco/download/download_configs_test.go
jskelin: added a new test with the name `Test_downloadConfigsOptions_valid`

cmd/monaco/download/download_configs.go
jskelin: This observation is in place. But as this part is out of scope, and I have no idea to propose a new name, I'll leave it as it is.

cmd/monaco/download/download_configs.go
```diff
@@ -202,14 +224,14 @@ func doDownloadConfigs(fs afero.Fs, downloaders downloaders, opts downloadConfig
...
+	{
 		log.Info("Downloading configuration APIs")
 
-		classicAPIs := makeClassicAPIs(opts.specificAPIs)
-		classicCfgs, err := downloaders.Classic().Download(opts.projectName, classicAPIs...)
+		classicCfgs, err := downloaders.Classic().Download(opts.projectName)
 		if err != nil {
 			return nil, err
 		}
+		log.Info("downloaded %q configurations from classic Config API endpoints", len(classicCfgs))
```
UnseenWizzard, line 234: nit: don't think that number needs to be in quotes
```suggestion
		log.Info("downloaded %s configurations from classic Config API endpoints", len(classicCfgs))
```

cmd/monaco/download/download_integration_test.go
```diff
@@ -1115,65 +1115,6 @@ func TestDownloadIntegrationDownloadsOnlyAPIsIfConfigured(t *testing.T) {
 	assert.Assert(t, !settingsDownloaded, "Expected no Settings to the downloaded, when onlyAPIs is set")
 }
 
-func TestDownloadIntegrationDownloadsOnlySettingsIfConfigured(t *testing.T) {
```
UnseenWizzard, line 1118: Why is this test removed? Judging by the `DownloadsOnlySettingsIfConfigured` name, this should not be impacted by your changes and is still a valid case to be tested.

Edit: I guess it should use `makeDownloaders` (or an exported method further up that does) to get the correct instantiation now. But that likely goes for all these integration tests. Could they call exported methods or at least production setup methods, rather than building their own downladers and calling the private doDownloadConfigs?

pkg/download/classic/downloader_test.go
```diff
@@ -297,15 +296,6 @@ func TestDownload_FilteringCanBeTurnedOffViaFeatureFlags(t *testing.T) {
 	assert.Len(t, configurations, 1)
 }
 
-func TestDownload_EmptyAPIMap_ResultsInError(t *testing.T) {
```
UnseenWizzard, line 300: As the behavior has changed and the classic downloader silently returns an empty list of configurations, we're not telling users anymore if there where really no configs on the environment, or filtering has somehow removed all APIs and nothing was downloaded in the first place. 

I think we should still be clear about that. 
Please either reintroduce this check and error return in Download or (if it's possible on that level) differentiate the cases in download_configs.go and log differently for each case. 

Related to this: Please change the signature or `Download` to how it actually functions. As it does not accept API to filter for anymore, please remove that function argument. Also depending on the changes you decide for this case, remove the error return. It does not make sense to have that in the signature if the method always returns a nil error

jskelin: Currently, during the process of evaluation of input data for every unknown classic API endpoint will be reported as an error to the user (and the process will be terminated). So the user will only be able to continue the process if he enters the correct name.
By the point where `downloaders.Classic().Download(opts.projectName)` is called, all input data is clean, and all irregularities should be reported to the user.
_Is it enough to log (info level) the number of endpoints from which config is going to be downloaded?_

Signature of Download function is: `func (d *Downloader) Download(projectName string, _ ...config.ClassicApiType) (project.ConfigsPerType, error)`
To implement `type Downloader[T v2.Type] interface` second argument needs to be present.

cmd/monaco/download/download_integration_test.go
UnseenWizzard: As discussed directly: fine by me

pkg/download/classic/downloader_test.go
UnseenWizzard: Thanks, missed the generic interface part.

As discussed directly: please add logging so a user can still get that nothing is being downloaded without debug logging - under that assumption that in normal operation/usage of the interface this can never happen, no error is needed for this case.

PR #1058, opened 2023-06-21
Fix/module imports v2

UnseenWizzard:
This ensures monaco's v2 module actually matches the Go package manager requirement that module versions with breaking changes must reflect this in their module path (convention appends /vN+1 after the initial v0.x or 1.x version).

This ensures that an upgrade to a breaking change version is a different import than the previous version. 

No change to end users. 
Developers importing packages as a library will be able to access version 2.x packages.


jskelin:
Since this was never released as a library, why go with `v2`?

To introduce module/library versioning is OK.

UnseenWizzard:
> Since this was never released as a library, why go with `v2`?
> 
> Introducing module/library versioning is OK.

Please see the first commit message for the reasoning @jskelin :) 

In short, because 1.x and 2.x are still technically imported on the same name (github.com/dynatrace-oss/dynatrace-monitoring-as-code due to Github's building forwarding of renamed repos), Go uses git repos as the basis of it's package management, and Go's import name versioning (different major versions must not be importable on the same path/name), it is currently completely impossible to import v2. 

This makes it impossible for the Terraform provider to import our Workflow API client

jskelin (dismissed):

warber (dismissed):

PR #1059, opened 2023-06-21
Show an error if download of automation-resources is forced, but no OAuth credentials are configured

Laubi:
Up to now no error message was given if a user wanted to download automation resources, but no credentials were configured.

Now we get logs like this: 
```
> MONACO_FEAT_AUTOMATION_RESOURCES=1 monaco download manifest.yaml -e XXX --only-automation -v
2023-06-21T14:51:58+02:00       debug   request log not activated
2023-06-21T14:51:58+02:00       debug   response log not activated
2023-06-21T14:51:58+02:00       debug   Loading manifest "manifest.yaml". Restrictions: groups=[], environments=["XXX "]
2023-06-21T14:51:58+02:00       debug   Concurrent Request Limit: 5, 'MONACO_CONCURRENT_REQUESTS' environment variable is NOT set, using default value
2023-06-21T14:51:58+02:00       info    Downloading from environment 'XXX ' into project 'project_XXX '
Error: can't download automation resources: no OAuth credentials configured
```

Additionally, we now state when we can't create any project if the set of downloaded configurations is empty:
```
> monaco download manifest.yaml -e XXX --settings-schema builtin:rum.web.app-detection
2023-06-21T14:55:07+02:00       info    Downloading from environment 'https://XXX' into project 'project_XXX'
2023-06-21T14:55:07+02:00       info    Downloading settings objects
2023-06-21T14:55:08+02:00       info    Downloaded 0 settings for schema builtin:rum.web.app-detection
2023-06-21T14:55:08+02:00       info    No configurations downloaded. No project will be created.
```

### DIFF
--- a/cmd/monaco/download/download_configs.go
+++ b/cmd/monaco/download/download_configs.go
@@ -183,6 +183,11 @@ func doDownloadConfigs(fs afero.Fs, downloaders downloaders, opts downloadConfig
 		return err
 	}
 
+	if len(downloadedConfigs) == 0 {
+		log.Info("No configurations downloaded. No project will be created.")
+		return nil
+	}
+
 	log.Info("Resolving dependencies between configurations")
 	downloadedConfigs = dependency_resolution.ResolveDependencies(downloadedConfigs)
 
--- a/cmd/monaco/download/download_configs.go
+++ b/cmd/monaco/download/download_configs.go
@@ -15,11 +15,13 @@
 package download
 
 import (
+	"errors"
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/cmd/monaco/dynatrace"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/internal/log"
 	v2 "github.com/dynatrace/dynatrace-configuration-as-code/pkg/config/v2"
+	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/download/automation"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/download/dependency_resolution"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/download/id_extraction"
 	"github.com/dynatrace/dynatrace-configuration-as-code/pkg/manifest"
@@ -202,6 +204,8 @@ func downloadConfigs(downloaders downloaders, opts downloadConfigsOptions) (proj
 	configs := make(project.ConfigsPerType)
 
 	if shouldDownloadClassicConfigs(opts) {
+		log.Info("Downloading configuration APIs")
+
 		classicAPIs := makeClassicAPIs(opts.specificAPIs)
 		classicCfgs, err := downloaders.Classic().Download(opts.projectName, classicAPIs...)
 		if err != nil {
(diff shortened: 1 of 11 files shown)


cmd/monaco/download/download_configs.go
```diff
@@ -220,11 +226,17 @@ func downloadConfigs(downloaders downloaders, opts downloadConfigsOptions) (proj
 	}
 
 	if shouldDownloadAutomationResources(opts) {
-		automationCfgs, err := downloaders.Automation().Download(opts.projectName)
-		if err != nil {
-			return nil, err
+		if _, ok := downloaders.Automation().(automation.NoopAutomationDownloader); !ok {
```
UnseenWizzard, line 229: As far as I can see you have the Auth struct available in `opts` (though I'm not quite sure why/how clean that is tbh...)

To me it would seem less brittle to check for the existence of oAuth in the Auth struct, rather than against a specific implementation of the Automation downloader.

`if opts.auth.OAuth != nil {`

UnseenWizzard (commented):
I don't like that we check against the specific NoOp downloader implementation.

pkg/client/clientset.go
```diff
@@ -53,41 +52,50 @@ func (s ClientSet) Entities() *dtclient.DynatraceClient {
...
-		)
-		autClient = automation.NewClient(
-			url,
-			clientAuth.NewOAuthClient(context.TODO(), oauthCredentials),
-			automation.WithClientRequestLimiter(concurrency.NewLimiter(concurrentRequestLimit)),
-		)
+	oauthCredentials := clientAuth.OauthCredentials{
+		ClientID:     auth.ClientID,
+		ClientSecret: auth.ClientSecret,
+		TokenURL:     auth.Token,
```
UnseenWizzard, line 84: The OauthCredentials.TokenURL is the URL to get the oauth token for the given client from, isn't it?

It should not be the auth.Token (that's the classic API token right) but it's own value.
See it's initialization in the old version on line 76

UnseenWizzard (changes requested):

pkg/client/clientset.go
```diff
@@ -53,41 +52,50 @@ func (s ClientSet) Entities() *dtclient.DynatraceClient {
...
+	}
 
+	return &ClientSet{
+		dtClient:  dtClient,
+		autClient: nil,
+	}, nil
+}
+
+type PlatformAuth struct {
+	ClientID, ClientSecret, Token, TokenURL string
```
UnseenWizzard, line 75: nit: For clarity of the interface it might make sense to clearly differentiate the Token and TokenURL, currently it looks like they relate to each other. 

Maybe a sub-struct called 

```
OAuth {
  ClientID, ClientSecret, TokenURL
} 
```

and calling the `Token` - `ClassicAPIToken`? 

Or just `ClassicAPIToken` and `OauthTokenURL` ?

UnseenWizzard (dismissed):

pkg/client/clientset.go
warber: I've changed it to just:
```go
type PlatformAuth struct {
	OauthClientID, OauthClientSecret, OauthTokenURL string
	Token                                           string
}
```
fine?

UnseenWizzard (approved):
lgtm thanks for addressing the comments 🚀

PR #1062, opened 2023-06-27
feat/allow configuring custom user agent

UnseenWizzard:
This change allows library users to configure a custom user-agent, rather than having each HTTP request send our default "Monitoring as Code" user-agent.


warber:
Did you try to use a custom `RoundTripper` like it's already done to provide the old-fashioned api token header, instead of passing it via context? Imo that would be again a good fit for that since it is something that should be always set.
something like:
```go

... 
func (u *UATransport) RoundTrip(r *http.Request) (*http.Response, error) {
	r.Header.Set("User-Agent", u.Agent)
	return u.inner.RoundTrip(r)
}
...
```
(i would keep the ctx param itself tho, as you already identified it as an reasonable improvement in general)

warber (commented):
lgtm , conflicts need to be resolved

Laubi (approved):

PR #1066, opened 2023-06-29
refactor: Ensure custom errors use uniform names and have struct tags for JSON logging

UnseenWizzard:

refactor: Ensure custom errors use uniform names and have struct tags for JSON logging

To ensure that structured JSON logs are easy to parse, fields of custom errors that have the same meaning now have the same name.

For a nicer looking generated JSON all fields are extended with struct tags so they start lowercase, rather than than the upper-case of exported struct fields.

NO - but this should be included in the first release that contains structured logging fields, else there will be a change.

### DIFF
--- a/pkg/manifest/manifest_loader.go
+++ b/pkg/manifest/manifest_loader.go
@@ -58,53 +58,59 @@ type projectLoaderContext struct {
 	manifestPath string
 }
 
-type manifestLoaderError struct {
-	ManifestPath string
-	Reason       string
+type ManifestLoaderError struct {
+	ManifestPath string `json:"manifestPath"`
+	Reason       string `json:"reason"`
 }
 
-func (e manifestLoaderError) Error() string {
+func (e ManifestLoaderError) Error() string {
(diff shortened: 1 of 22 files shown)


pkg/manifest/manifest_loader.go
```diff
@@ -58,53 +58,59 @@ type projectLoaderContext struct {
...
-	Group       string
-	Environment string
+type EnvironmentDetails struct {
+	Group       string `json:"group"`
+	Environment string `json:"environment"`
+}
+
+type EnvironmentLoaderError struct {
+	ManifestLoaderError
+	EnvironmentDetails EnvironmentDetails `json:"EnvironmentDetails"`
```
Laubi, line 84:
```suggestion
	EnvironmentDetails EnvironmentDetails `json:"environmentDetails"`
```

Laubi (dismissed):

pkg/manifest/manifest_loader.go
UnseenWizzard: Thanks! Fixed

jskelin (approved):

PR #1068, opened 2023-07-03
Refactor/replace some ff with flags

jskelin:
Direct usage of feature flags in `pkg` prevents `pkg` to be reusable as a library. This PR replaces some usages of `MONACO_FEAT_DOWNLOAD_FILTER` and  `MONACO_FEAT_DOWNLOAD_FILTER_CLASSIC_CONFIGS` with the flags

### DIFF
--- a/cmd/monaco/download/downloaders.go
+++ b/cmd/monaco/download/downloaders.go
@@ -18,6 +18,7 @@ package download
 
 import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/dynatrace"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/dtclient"
@@ -59,7 +60,7 @@ func makeDownloaders(options downloadConfigsOptions) (downloaders, error) {
 
 func classicDownloader(client dtclient.Client, opts downloadConfigsOptions) *classic.Downloader {
 	endpoints := prepareAPIs(opts)
-	return classic.NewDownloader(client, classic.WithAPIs(endpoints))
+	return classic.NewDownloader(client, classic.WithAPIs(endpoints), classic.WithFiltering(shouldApplyFilter()))
 }
 
 func prepareAPIs(opts downloadConfigsOptions) api.APIs {
@@ -80,13 +81,17 @@ func prepareAPIs(opts downloadConfigsOptions) api.APIs {
 }
 
 func removeSkipDownload(api api.API) bool {
-	if classic.ShouldApplyFilter() {
+	if shouldApplyFilter() {
 		if api.SkipDownload {
 			log.Info("API can not be downloaded and needs manual creation: '%v'.", api.ID)
 			return true
 		}
 	}
 	return false
+}
+
+func shouldApplyFilter() bool {
+	return featureflags.DownloadFilter().Enabled() && featureflags.DownloadFilterClassicConfigs().Enabled()
 }
 
 func removeDeprecated(log ...func(api api.API)) api.Filter {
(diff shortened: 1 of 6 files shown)


pkg/download/classic/downloader.go
```diff
@@ -42,9 +41,11 @@ type (
 	Downloader struct {
 		apisToDownload api.APIs
 
+		filter bool
```
UnseenWizzard, line 44: please ensure the `NewDownloader` constructor sets this to `true` by default, before applying any options. We also assume the feature flags are ON by default in this case.

cmd/monaco/download/downloaders.go
```diff
@@ -89,6 +90,10 @@ func removeSkipDownload(api api.API) bool {
 	return false
 }
 
+func ShouldApplyFilter() bool {
```
UnseenWizzard, line 93: I don't think this should be exported anymore now

pkg/download/classic/downloader_test.go
```diff
@@ -14,14 +14,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package classic
+package classic_test
 
 import (
 	"context"
 	"fmt"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/download"
```
UnseenWizzard, line 22: even for tests, I don't think a package should ever import from cmd, please remove that (think it's to access the feature flag convenience method, IMO that one's not needed for testing. Currently we assume they are ON so keeping the tests at default filtering or where needed just explicitly setting WithFiltering(true) or WithFiltering(false) should be enough. 

Testing that the flags control the option should be done on the cmd package integration test IMO.

UnseenWizzard (changes requested):

cmd/monaco/download/downloaders.go
jskelin: I agree. To avoid massive changes in tests, this is exported. Is it enough to write a comment?

UnseenWizzard: Do the tests ever set the environment variables though?
From what I saw, only `TestDownload_SkipConfigBeforeDownload` does actually set the env vars, all other uses could set the value they need explicitly or just default to true. 

For `TestDownload_SkipConfigBeforeDownload` IMO on the downloader level this should change to just verify the boolean option works as intended - defaulting to true and possible to turn off. 

That the env vars translate to the correct input for the downloader that is created in the cmd package should IMO be tested there, not by tying the logic of cmd and pkg together. 
I think the test could move to the cmd/ tests easily as it just seems to use public methods of the downloader anyway?

pkg/download/classic/downloader.go
warber: if that's the reasonable default, you could turn it around, and change it to `skipFilter bool`

pkg/download/classic/downloader.go
```diff
@@ -74,12 +76,18 @@ func WithAPIs(apis api.APIs) ConstOption {
 	}
 }
 
-func WithAPIContentFilters(apiFilters map[string]contentFilter) ConstOption {
+func WithAPIContentFilters(apiFilters map[string]ContentFilter) ConstOption {
```
warber, line 79: just wondering, what does the "Const" mean in "ConstOption" i don't get it :D

pkg/download/classic/downloader.go
```diff
@@ -74,12 +76,18 @@ func WithAPIs(apis api.APIs) ConstOption {
...
 	return func(d *Downloader) {
 		d.apiContentFilters = apiFilters
 	}
 }
 
+func WithFiltering(b bool) ConstOption {
```
warber, line 85: just wondering, what does the "Const" in "ConstOption" mean? I don't get it :D

jskelin: from `Construction`
The name is far from perfect. Just propose a better one...

warber: `Option`, if that does not work, `DownloaderOption` or no name at all but just the signature

jskelin: Typing is forced to have a compile-time error if the signature of `func NewDownloader(client dtclient.Client, opts ...ConstOption) *Downloader` is changed!

I'll rename it to `Option`

UnseenWizzard (approved):

PR #1069, opened 2023-07-03
feat: cache settings objects when during upsert operations

warber:

The current process of upserting the Settings 2.0 object is highly inefficient since it requires retrieving all settings objects to find those with a matching legacy external ID. This pull request proposes the implementation of a basic cache to eliminate the need for this extra step for every settings object being upserted.

deploying settings 2.0 objects is faster :rocket:


pkg/client/dtclient/client.go
```diff
@@ -432,25 +435,21 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
 	if err != nil {
 		return DynatraceEntity{}, fmt.Errorf("unable to generate external id: %w", err)
 	}
-	settingsWithExternalID, err := d.listSettings(ctx, obj.SchemaId, ListSettingsOptions{
-		Filter: func(object DownloadSettingsObject) bool {
-			return object.ExternalId == legacyExternalID
-		},
-	})
 
+	settingsWithExternalID, err := d.getSettingsFromCache(ctx, obj.SchemaId, func(object DownloadSettingsObject) bool { return object.ExternalId == legacyExternalID })
```
UnseenWizzard, line 439: My initial thought would have been to make caching a built-in feature of ListSettings and transparent to the callsites. If I just use the public ListSettings now or listSettings in another internal method I'll have to know about and use the FromCache method explicitly. If I use the public interface I have no way to profit from caching.
What made you go for explicitly using the cache?

pkg/client/dtclient/settings_cache.go
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dtclient
+
+type SettingsCache struct {
```
UnseenWizzard, line 19: could/should we make this a generic key-value cache? We'll need a schema and config API object cache soon as well

warber:
@UnseenWizzard the PR is in draft an not yet ready to review :)

pkg/client/dtclient/client.go
warber: @UnseenWizzard I see the cache as something that will eventually removed again once we do not need to list all settings in upsert anymore. I don't see any other strict/important/useful reason why we need a cache everywhere, but maybe you can help me there :)

pkg/client/dtclient/settings_cache.go
warber: @UnseenWizzard  it's coming :) For now just imagine to read `T` insead of `DownloadSettingsObject` :D

pkg/client/dtclient/client.go
UnseenWizzard: > once we do not need to list all settings in upsert anymore.
I think with the change to look for unique key properties, this will never go away anymore.

UnseenWizzard:
> @UnseenWizzard the PR is in draft an not yet ready to review :)

@warber if you open a draft of something I find interesting I will take a look at it :P 
Of course, feel free to ignore comments at the draft stage, especially for things that'll change anyway (but I might ask them again once it's ready).

warber:
@UnseenWizzard ready for review.

pkg/client/dtclient/client.go
warber: I've changed it to be used "built-in" in `listSettings()` :)

pkg/client/dtclient/cache.go
```diff
@@ -0,0 +1,65 @@
...
+package dtclient
+
+import "sync"
+
+type cache[T any] struct {
+	cachedItems map[string][]T
+	mutex       sync.RWMutex
+}
+
+func (s *cache[T]) hasCache(id string) bool {
```
Laubi, line 26: Minor: `isCached` or just `has`? Or even `hasCached` instead of `hasCache`

pkg/client/dtclient/cache.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (s *cache[T]) filter(id string, filter func(T) bool) []T {
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+
+	if filter == nil {
+		filter = func(object T) bool { return true }
+	}
+	result := make([]T, 0)
+	for _, i := range s.cachedItems[id] {
```
Laubi, line 52: This will fail if cachedItems has not been initialized yet

pkg/client/dtclient/cache.go
```diff
@@ -0,0 +1,65 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package dtclient
+
+import "sync"
+
+type cache[T any] struct {
+	cachedItems map[string][]T
```
Laubi, line 22: map-key is what? Schema ID? Please specify or add brief docs.

pkg/client/dtclient/cache.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	s.cachedItems[id] = settings
+}
+
+func (s *cache[T]) filter(id string, filter func(T) bool) []T {
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+
+	if filter == nil {
+		filter = func(object T) bool { return true }
```
Laubi, line 49: I'd argue we should just return the entries here instead of looping them later and adding them to a map.

pkg/client/dtclient/cache.go
warber: it's **a** key.
might be schema id, depending on the use case...

pkg/client/dtclient/cache.go
warber: haha yeah, of course :smile:

pkg/client/dtclient/cache.go
warber: it's a good point :) , but you are wrong - ranging over or reading from a nil map will not panic.

Laubi: True - my error. Even accessing a  key that does not exist returns nil, thus ranging does work.

Laubi (approved):

pkg/client/dtclient/cache.go
warber: In this case, yes. Generally, it returns the zero value of the map's value type. :nerd_face:

pkg/client/dtclient/cache.go
```diff
@@ -0,0 +1,65 @@
...
+package dtclient
+
+import "sync"
+
+type cache[T any] struct {
+	cachedItems map[string][]T
+	mutex       sync.RWMutex
+}
+
+func (s *cache[T]) isCached(id string) bool {
```
UnseenWizzard, line 26: might be a stupid question / the cache usage is super specialized on filter(, but: 

Why is there no`get(id string) (T, bool)` at all? 
That would allow me to just do the _, ok check on my own where needed - following the idioms of map

pkg/client/dtclient/client.go
```diff
@@ -432,25 +435,22 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
...
 		return DynatraceEntity{}, fmt.Errorf("unable to generate external id: %w", err)
 	}
+
 	settingsWithExternalID, err := d.listSettings(ctx, obj.SchemaId, ListSettingsOptions{
 		Filter: func(object DownloadSettingsObject) bool {
 			return object.ExternalId == legacyExternalID
 		},
 	})
 
-	if err != nil {
```
UnseenWizzard, line 441: why did you remove this error handling? Won't we continue now in case the GET /settings API call failed?
In the case of client error here, it's likely further calls will just error as well, but still at this point data is invalid, isn't it?

UnseenWizzard (dismissed):
Thanks for the fix. Good to merge as far as I'm concerned, if you decide against the `get` method nitpick :)

pkg/client/dtclient/cache.go
warber: i like the idea. will do

warber:
@UnseenWizzard ready for review (again) :)

jskelin (changes requested):

jskelin (changes requested):
Using generics for values, this pretends to be a general solution for caching. On the other hand, the `get` method with integrated filtration is highly profiled for one purpose. 
Either make this solution a general solution (separate package with well-defined API). Or keep it local, non-exported and simple as possible that implement only specific demands (without generic)

warber:
@jskelin 

> the get method with integrated filtration is highly profiled for one purpose.

not really, it's a generic `get` method with a generic `filter` that can be applied (or not, you can also just pass `nil`)

> Either make this solution a general solution (separate package with well-defined API). Or keep it local, non-exported and simple as possible that implement only specific demands (without generic)!

For now, we don't need a cache anywhere but in this package and there we need it for multiple data types. To me that looks like a perfect fit for keeping it local and generic.

pkg/client/dtclient/cache.go
```diff
@@ -0,0 +1,71 @@
...
+	for _, i := range s.cachedItems[id] {
+		if filter(i) {
+			result = append(result, i)
+		}
+	}
+	return result, true
+
+}
+
+func (s *cache[T]) invalidate(id string) {
```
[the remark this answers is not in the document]
warber: could be done. however, i'd argure that it's also prette clear what the invalidate method might do in the context of a cache.
I am also fine with `delete' tho.

pkg/client/dtclient/cache.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (s *cache[T]) isCached(id string) bool {
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+
+	_, ok := s.cachedItems[id]
+	return ok
+}
+
+func (s *cache[T]) set(id string, settings []T) {
```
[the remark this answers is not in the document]
warber: agree

warber:
@jskelin 
I've refactored a bit, I think you will love it :)

jskelin (approved):
looks great

UnseenWizzard (approved):
I like the new version! 🚀

PR #1070, opened 2023-07-05
Consider if upsert tries to update object with uniqueness constraints or creates new one

warber:
Some of the settings schemas have constraints that enforce the uniqueness of some properties. If a user tries to create new settings object that "violates" some of those constraints, upsert process fails with an error.

With this change, monaco now tries to find an object that has matching values. If the object exists, it is updated, otherwise a new one is created.


cmd/monaco/integrationtest/v2/test-resources/settings-unique-properties/project1/builtinmonitoring.slo/d955d9b7-9630-3be5-aca5-3715779a7282.json
```diff
@@ -0,0 +1,15 @@
+{
+  "enabled": true,
+  "name": "Just a Name",
+  "metricName": "justaname",
```
jskelin, line 4: From the failing message:
`Metric name field can not start with a number, contain special characters or uppercase letters.`

jskelin (dismissed):
The tests look good. 
For the rest I need somebody else to take a look....

cmd/monaco/integrationtest/v2/test-resources/settings-unique-properties/project1/builtinmonitoring.slo/d955d9b7-9630-3be5-aca5-3715779a7282.json
UnseenWizzard: looking at the other SLO config it does not look like the metric name is modified? 
cmd/monaco/integrationtest/v2/test-resources/settings-unique-properties/project2/builtinmonitoring.slo/d955d9b7-9630-3be5-aca5-3715779a7282.json

still has `"metricName": "justaname",` - but changes the `name`

jskelin: The second comment is for documentation. `metricName` value was the only problem.

jskelin (approved):

pkg/client/dtclient/client_settings.go
```diff
@@ -187,6 +187,12 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
 		}
 	}
 
+	if objectID, err := d.findObjectWithMatchingConstraints(ctx, obj); err != nil {
+		return DynatraceEntity{}, err
+	} else if objectID != nil {
+		obj.OriginObjectId = objectID.ObjectId
```
UnseenWizzard, line 193: having debug logging for this case would be helpful I think

```suggestion
		log.Debug("Updating existing object %q with matching unique key properties", objectID.ObjectId")     
		obj.OriginObjectId = objectID.ObjectId
```

pkg/client/dtclient/client_settings.go
```diff
@@ -247,6 +253,76 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
...
+		return false, fmt.Errorf("failed to unmarshal data for key %q: %w", key, err)
+	}
+	v1 := u[key]
+
+	if err := json.Unmarshal(c2, &u); err != nil {
+		return false, fmt.Errorf("failed to unmarshal data for key %q: %w", key, err)
+	}
+	v2 := u[key]
+
+	return v1 == v2, nil
```
UnseenWizzard, line 268: this check does not work if the unique value is an object instead of a value. 

So e.g. if the key is "A" and it's value is another JSON object `{"key":"x", "val":"y"}` this will panic.

I've added a test case and adapt this accordingly in a commit.
I'm not sure if that case is possible, but we should not rule it out and fail execution if it happens.

UnseenWizzard (changes requested):
currently we panic in case a unique key's value is not a plain value, but another JSON object (marshaled as a map), as maps can not be checked for equality with `==`.

I have added a test case locally and will add a commit to address it as well! 
Otherwise looks good to me so far :) 

Marking as needs work for now

jskelin (dismissed):

pkg/client/dtclient/settings_client.go
```diff
@@ -247,6 +257,76 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
...
+					err = errors.Join(err, fmt.Errorf("can't update or create new configuration %q couse the already existing objects with ID %q and %q interfearing", source.Coordinate, candidate.ObjectId, objects[j].ObjectId))
+				}
+			}
+		}
+	}
+
+	return candidate, err
+}
+
+func areObjectsEqualForTheRule(rule []string, source SettingsObject, object DownloadSettingsObject) (bool, error) {
```
Laubi, line 318: I think the code that handles this is hard for me to wrap my head around. You need to know how `isSameValueOfConstraint` works internally to understand that the error is handled correctly.

I would go with a less clever but more straightforward solution: Checking the error explicitly, and also not returning the variable `b`, as it is clever, but the constants true/false are clearer.

I would suggest the change as follows:
```
func areObjectsEqualForTheRule(rule []string, source SettingsObject, object DownloadSettingsObject) (bool, error) {
	var eq bool
	var err error
	for _, c := range rule {
		eq, err = isSameValueOfConstraint(c, object.Value, source.Content)
		if err != nil {
			return false, err
		}

		if !eq {
			return false, nil
		}
	}

	return true, err
}
```

pkg/client/dtclient/settings_client.go
```diff
@@ -187,6 +190,13 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
 		}
 	}
 
+	if match, err := d.findObjectWithMatchingConstraints(ctx, obj); err != nil {
+		return DynatraceEntity{}, err
+	} else if match != nil {
```
Laubi, line 195: minor: unnecessary else - we return anyway before

pkg/client/dtclient/settings_client.go
```diff
@@ -247,6 +257,76 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
...
+	return target, nil
+}
+
+func findObjectWithSameConstraints(schema SchemaConstraints, source SettingsObject, objects []DownloadSettingsObject) (*DownloadSettingsObject, error) {
+	var candidate *DownloadSettingsObject
+	var err error
+	for _, rule := range schema.UniqueProperties {
+		for j, o := range objects {
+			var b bool
+			b, err = areObjectsEqualForTheRule(rule, source, o)
```
Laubi, line 303: We overwrite the error in every iteration, we need to create a second one, and append it to the first.

pkg/client/dtclient/settings_client.go
UnseenWizzard: Why do we return anyway before?
Frankly I haven't double-checked if it's impossible at the moment, but generally I think it's fine to assume that `findObjectWithMatchingConstraints` might return a `nil` entity even if there was no error

UnseenWizzard: Actually just saw while debugging an issue my recent changes seem to have introduced: 
Errors are only returned for actual errors. If there's no match found, the method will return `nil, nil`, no error, but also no match. So this is if/else seems to do exactly what it should - unless I'm missing something

pkg/client/dtclient/settings_client.go
UnseenWizzard: I've refactored this a bit in the last two commits I've added - I think it's easier to read now. please take another look

UnseenWizzard (approved):
My comments were addressed. Approving to remove my change request

Laubi (approved):

PR #1075, opened 2023-07-06
feat: Get single automation resource by ID

warber:
Add `Get` method to fetch a single automation resource by ID:

```go
httpClient := auth.NewOAuthClient(context.TODO(), auth.OauthCredentials{
	ClientID:     "######",
	ClientSecret: "######",
	TokenURL:     "https://sso.dynatrace.com/sso/oauth2/token",
})

automationClient := automation.NewClient("https://####.apps.dynatrace.com", httpClient)
wf, err := automationClient.Get(context.TODO(), "f41f0468-74c7-43e0-a4c4-6770ab336235", automation.Workflows)
if err != nil {
		log.Fatal(err.Error())
}
...
```

### DIFF
--- a/pkg/client/automation/client.go
+++ b/pkg/client/automation/client.go
@@ -120,6 +120,37 @@ func WithCustomUserAgentString(userAgent string) func(client *Client) {
 	}
 }
 
+// Get returns a single automation object matching the given resource type and id or an error
+// if a resource with the given ID and resourceType wasn't found
+func (a Client) Get(ctx context.Context, resourceType ResourceType, id string) (result *Response, err error) {
+	a.limiter.ExecuteBlocking(func() {
+		result, err = a.get(ctx, id, resourceType)
+	})
+	return
+}
+
+func (a Client) get(ctx context.Context, id string, resourceType ResourceType) (*Response, error) {
+	requestURL := a.url + a.resources[resourceType].Path + "/" + id
+	u, err := url.Parse(requestURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse URL '%s': %w", requestURL, err)
+	}
+
+	resp, err := rest.Get(ctx, a.client, u.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to GET automation resource of type %v with id %s: %w", resourceType, id, err)
+	}
+
+	if !resp.IsSuccess() {
+		return nil, rest.NewRespErr(fmt.Sprintf("failed to get automation resource with ID %s (HTTP %d): %s", id, resp.StatusCode, string(resp.Body)), resp).WithRequestInfo(http.MethodGet, u.String())
+	}
+
+	return &Response{
+		ID:   id,
+		Data: resp.Body,
+	}, nil
+}
+
 // List returns all automation objects
 func (a Client) List(ctx context.Context, resourceType ResourceType) (result []Response, err error) {
 	a.limiter.ExecuteBlocking(func() {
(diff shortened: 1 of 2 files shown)


pkg/client/automation/client.go
```diff
@@ -120,6 +120,37 @@ func WithCustomUserAgentString(userAgent string) func(client *Client) {
 	}
 }
 
+// Get returns a single automation object matching the given resource type and id or an error
+// if a resource with the given ID and resourceType wasn't found
+func (a Client) Get(ctx context.Context, id string, resourceType ResourceType) (result *Response, err error) {
```
UnseenWizzard, line 125: nit: having the resourceType before the ID would make more sense to me, listing them in a hierarchical way, and matches the (API, id) params of the config client:
```suggestion
func (a Client) Get(ctx context.Context, resourceType ResourceType, id string) (result *Response, err error) {
```

pkg/client/automation/client.go
```diff
@@ -120,6 +120,37 @@ func WithCustomUserAgentString(userAgent string) func(client *Client) {
...
+	a.limiter.ExecuteBlocking(func() {
+		result, err = a.get(ctx, id, resourceType)
+	})
+	return
+}
+
+func (a Client) get(ctx context.Context, id string, resourceType ResourceType) (*Response, error) {
+	requestURL := a.url + a.resources[resourceType].Path + "/" + id
```
UnseenWizzard, line 133: please use path.Join, manual concatenation and adding/not adding slashes is always prone to break eventually

```suggestion
	requestURL := path.Join(a.url, a.resources[resourceType].Path, id)
```

UnseenWizzard: As we just need the URL string, maybe `url.JoinPath` could be enough and also replace the `url.Parse` call. 
Although it seems to only parse the basepath

UnseenWizzard (dismissed):
comments are nitpicks, could be merged

pkg/client/automation/client.go
warber: your suggestion does not play well with the host part of the URL, especially with the "http://" part and produces an invalid request url. I will jsut leave it like that

pkg/client/automation/client.go
warber: yeah why not :)

pkg/client/automation/client.go
Laubi: I think we need to find a proper solution here, aka passing a proper URL object around instead of a plain string.
I recently had an issue where the concatenation caused a double `//`, thus failing requests. Would not have happened with path.Join.
I agree, maybe let's not fix this issue here, but we should be aware that we should not pass plain strings around.

Laubi: Since you're parsing the URL on line #134, you _could_ fix this issue fairly simple :)

pkg/client/automation/client.go
```diff
@@ -120,6 +120,37 @@ func WithCustomUserAgentString(userAgent string) func(client *Client) {
...
+	u, err := url.Parse(requestURL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse URL '%s': %w", requestURL, err)
+	}
+
+	resp, err := rest.Get(ctx, a.client, u.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to GET automation resource of type %v with  id %s: %w", resourceType, id, err)
```
Laubi, line 141:
```suggestion
		return nil, fmt.Errorf("failed to GET automation resource of type %v with id %s: %w", resourceType, id, err)
```

Laubi (approved):

PR #1077, opened 2023-07-07
Research/graphs

UnseenWizzard:
This is a PoC of using [gonum](https://github.com/gonum/gonum)'s graph package to build dependency trees of configurations. 

The first commit restructures existing code a bit, to split the current toplogysort from shared interfaces.
If this PoC ends up as the basis of how we implement this, more refactoring will be needed - the spread of toposort between project and internal should be removed, and sort might make sense as a project independent package.

The second commit introduces the actual gonum graph libraries. For reviewing the PoC focus on this one and the tests (and 'non-test' integration test demo).

### DIFF
--- a/cmd/monaco/dependencygraph/command.go
+++ b/cmd/monaco/dependencygraph/command.go
@@ -1,0 +1,76 @@
+/*
+ * @license
+ * Copyright 2023 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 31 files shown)


Laubi (dismissed):

UnseenWizzard:
@Laubi I've placed the CLI command behind it's own feature flag as discussed (latest commit) and rebased the branch. Please take another look after lunch.

PR #1078, opened 2023-07-10
feat: support archive

warber:
This PR introduces a new flag called `--support-archive` which is a substitute for the `MONACO_[REQUEST|RESPONSE]_LOG` environment variable. When enabled, it will produce a zip archive containing three files: request logs, response logs and regular logs.

The changes in this PR are bigger than you would expect for the simple reason that I didn't want to continue putting more and more fixed code into the rest/client package. Looking forward we want to use stuff from there as a library. For this use case we certainly do not want the library to create files and archives based on environmentvariable that you can't even change.

The code respponsible of making requests is refactored into a "rest client" that offers the usual rest operations as methods (get, post, put,....) Code previously making direct calls to those methods is now making use of the rest client.

Note, that there are a couple of issues that still require refactoring which will be done in a follow-up PR:

* Refactor TODOs about global "SupportArchive" variable
* Make sure that constructor for classic and platform client is easy to use
* review thread safety of traffic logger
* use base url field in rest client
* use an interface for the trafficlogger
* use traffic logger conditionally also for non-download/deploy/... operations (e.g. GetDynatraceVersion,....)

Request and response logs are now stored alongside regular logs in the `.logs` directory.
Further, the name of these files changed to `<timestamp>-req-.log` and `<timestamp>-resp.log`. The regular log file name remains as `<timestamp>.log`

If the user passes the `--support-archive` flag, a `support-archive.zip` file is created containing these trhee files.
The zip file can be used by users to share debug information while troubleshooting


internal/zip/zip.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	defer zipFile.Close()
+
+	zipWriter := zip.NewWriter(zipFile)
+	defer zipWriter.Close()
+
+	for _, f := range files {
+		err = addFileToZip(fs, zipWriter, f, preservePath)
+		if err != nil {
+			log.Warn("Unable to add %s file to support archive: %w", f, err)
```
UnseenWizzard, line 40: I can understand continuing and adding all the files we can to the archive, but shouldn't we still return an error in the end if we failed to add some of the files?

pkg/rest/client.go
```diff
@@ -0,0 +1,237 @@
...
+	return resp, RespError{
+		StatusCode: resp.StatusCode,
+		Reason:     fmt.Sprintf("GET request %s failed after %d retries: (HTTP %d)!\n    Response was: %s", url, settings.MaxRetries, resp.StatusCode, resp.Body),
+		Body:       string(resp.Body),
+	}
+}
+
+func (c Client) Delete(ctx context.Context, url string) (Response, error) {
+	req, err := c.request(ctx, http.MethodDelete, url)
+	if err != nil {
```
UnseenWizzard, line 88: OCD-nitpick: please add a newline before the `if` like in all other methods of the client

pkg/client/automation/client.go
```diff
@@ -116,7 +116,7 @@ func WithClientRequestLimiter(limiter *concurrency.Limiter) func(client *Client)
 // If none is set, the default Monaco CLI specific user-agent is sent.
 func WithCustomUserAgentString(userAgent string) func(client *Client) {
 	return func(d *Client) {
-		d.client = &http.Client{Transport: useragent.NewCustomUserAgentTransport(d.client.Transport, userAgent)}
+		d.client = rest.NewRestClient(&http.Client{Transport: useragent.NewCustomUserAgentTransport(d.client.Client().Transport, userAgent)}, rest.CreateRateLimitStrategy())
```
UnseenWizzard: this will overwrite the rate-limiting strategy as well as the traffic logger that were previously set for the `d.client`

you currently don't export any of the fields, but I think for options like this one, we still only want to replace the http.Client inside the rest.Client (or create a new one that re-uses the ratelimit and logger that was in the client before the option applied)

cmd/monaco/download/download_command.go
```diff
@@ -268,7 +268,7 @@ func printUploadToSameEnvironmentWarning(env manifest.EnvironmentDefinition) {
 		httpClient = clientAuth.NewOAuthClient(context.TODO(), credentials)
 	}
 
-	serverVersion, err = versionClient.GetDynatraceVersion(context.TODO(), rest.NewRestClient(httpClient, rest.CreateRateLimitStrategy()), env.URL.Value)
+	serverVersion, err = versionClient.GetDynatraceVersion(context.TODO(), rest.NewRestClient(httpClient, nil, rest.CreateRateLimitStrategy()), env.URL.Value)
```
UnseenWizzard, line 271: it was possible to write traffic logs for any command before - with this change, we're losing the option to write logs for downloads I believe. We should not

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -45,7 +46,7 @@ func VerifyEnvironmentGeneration(envs manifest.Environments) bool {
 }
 
 func isClassicEnvironment(env manifest.EnvironmentDefinition) bool {
-	if _, err := version.GetDynatraceVersion(context.TODO(), rest.NewRestClient(auth.NewTokenAuthClient(env.Auth.Token.Value), rest.CreateRateLimitStrategy()), env.URL.Value); err != nil {
+	if _, err := version.GetDynatraceVersion(context.TODO(), rest.NewRestClient(auth.NewTokenAuthClient(env.Auth.Token.Value), nil, rest.CreateRateLimitStrategy()), env.URL.Value); err != nil {
```
UnseenWizzard, line 49: here as well, this means these initial API calls are missing from the traffic logs - smaller issue than the whole download but still

cmd/monaco/support/archive.go
```diff
@@ -19,18 +19,19 @@ package support
...
 
+var SupportArchive bool // TODO: this needs to be refactored :_)
+
 func Archive(fs afero.Fs) error {
 	timeAnchorStr := timeutils.TimeAnchor().Format("20060102-150405")
 	zipFileName := "support-archive-" + timeAnchorStr + ".zip"
 	files := []string{
-		path.Join(log.LogDirectory, timeAnchorStr) + "-" + trafficlogs.RequestLog.FileName(),
-		path.Join(log.LogDirectory, timeAnchorStr) + "-" + trafficlogs.ResponseLog.FileName(),
+		path.Join(log.LogDirectory, timeutils.TimeAnchor().Format("20060102-150405")+"-"+"req.log"),
```
UnseenWizzard, line 33: why not keep using the `timeAnchorStr` here?

UnseenWizzard, line 33: it might still make sense to expose the filenames/filename format as a constant from trafficlogs I think? if the name changes in the logger, this breaks

internal/trafficlogs/logger.go
```diff
@@ -0,0 +1,161 @@
...
+
+	return tl
+}
+
+func (l *FileBasedLogger) Log(req *http.Request, reqBody string, resp *http.Response, respBody string) error {
+	l.lock.Lock()
+	defer l.lock.Unlock()
+
+	requestId := ""
+	if l.isRequestLogEnabled() {
```
UnseenWizzard, line 57: these will currently always be true won't they? Nothing is exported, so the `NewFileBased` constructor is the only way to ever get a logger

cmd/monaco/download/download_command.go
UnseenWizzard: nevermind, read this wrong, this only loses some traffic as do the other commands -> followup as discussed

pkg/client/automation/client.go
```diff
@@ -116,7 +116,8 @@ func WithClientRequestLimiter(limiter *concurrency.Limiter) func(client *Client)
 // If none is set, the default Monaco CLI specific user-agent is sent.
 func WithCustomUserAgentString(userAgent string) func(client *Client) {
 	return func(d *Client) {
-		d.client = rest.NewRestClient(&http.Client{Transport: useragent.NewCustomUserAgentTransport(d.client.Client().Transport, userAgent)}, rest.CreateRateLimitStrategy())
+		//d.client = rest.NewRestClient(&http.Client{Transport: useragent.NewCustomUserAgentTransport(d.client.Client().Transport, userAgent)}, trafficlogs.NewFileBased(), rest.CreateRateLimitStrategy())
```
UnseenWizzard, line 119: nit: drop commented out part

pkg/client/dtclient/client.go
```diff
@@ -303,8 +299,10 @@ func WithAutoServerVersion() func(client *DynatraceClient) {
 // If none is set, the default Monaco CLI specific user-agent is sent.
 func WithCustomUserAgentString(userAgent string) func(client *DynatraceClient) {
 	return func(d *DynatraceClient) {
-		d.client = rest.NewRestClient(&http.Client{Transport: useragent.NewCustomUserAgentTransport(d.client.Client().Transport, userAgent)}, rest.CreateRateLimitStrategy())
-		d.clientClassic = rest.NewRestClient(&http.Client{Transport: useragent.NewCustomUserAgentTransport(d.clientClassic.Client().Transport, userAgent)}, rest.CreateRateLimitStrategy())
+		d.client.Client().Transport = useragent.NewCustomUserAgentTransport(d.client.Client().Transport, userAgent)
+		d.clientClassic.Client().Transport = useragent.NewCustomUserAgentTransport(d.clientClassic.Client().Transport, userAgent)
+		//d.client = rest.NewRestClient(&http.Client{Transport: useragent.NewCustomUserAgentTransport(d.client.Client().Transport, userAgent)}, trafficlogs.NewFileBased(), rest.CreateRateLimitStrategy())
```
UnseenWizzard, line 304: remove?

pkg/client/dtclient/client.go
```diff
@@ -352,20 +347,18 @@ func NewPlatformClient(dtURL string, token string, oauthCredentials auth.OauthCr
...
 
-	tokenClient := auth.NewTokenAuthClient(token)
-	restClient := rest.NewRestClient(tokenClient, rest.CreateRateLimitStrategy())
 	d := &DynatraceClient{
 		serverVersion:         version.Version{},
 		environmentURL:        dtURL,
 		environmentURLClassic: dtURL,
-		client:                restClient,
-		clientClassic:         restClient,
+		client:                client,
```
UnseenWizzard, line 360: how does the DynatraceClient behave when `client` and `environmentURL` are not set?
Any platform related call will not work anyway (and hopefully not happen) for a classic environment, so can't we just leave these uninitialized rather than setting wrong data?

internal/trafficlogs/logger.go
warber: yes, this is just a logical leftover of the previous code will remove that check

cmd/monaco/support/archive.go
warber: absolutely

warber: homer-hides.... of course

pkg/rest/client.go
warber: sure

pkg/client/dtclient/client.go
warber: I've changed it, it required some minor adoptions to the test code, let's see if it breaks something :D

UnseenWizzard (dismissed):

pkg/client/clientset.go
```diff
@@ -98,17 +109,33 @@ func CreatePlatformClientSet(url string, auth PlatformAuth, opts ClientOptions)
 		ClientSecret: auth.OauthClientSecret,
 		TokenURL:     auth.OauthTokenURL,
 	}
+
+	tokenClient := clientAuth.NewTokenAuthClient(auth.Token)
+	oauthClient := clientAuth.NewOAuthClient(context.TODO(), oauthCredentials)
+	classicURL, err := metadata.GetDynatraceClassicURL(context.TODO(), rest.NewRestClient(oauthClient, nil, rest.CreateRateLimitStrategy()), url) //this will send the default user-agent
```
UnseenWizzard, line 115: nothing for this PR, just something I realized while reading this: we could move the optional user-agent down to rest client initialization now that we explicitly pass it in to the higher-level-logic `clients`.

cmd/monaco/runner/runner.go
```diff
@@ -58,15 +58,21 @@ Examples:
...
+			if support.SupportArchive {
+				return support.Archive(fs)
+			}
+			return nil
+		},
 	}
 
 	// global flags
 	rootCmd.PersistentFlags().BoolVarP(&verbose, "verbose", "v", false, "Enable debug logging")
+	rootCmd.PersistentFlags().BoolVar(&support.SupportArchive, "support-archive", false, "Create support archive")
```
UnseenWizzard, line 75: A bit more documentation on what a support archive will contain and where it can be found would be helpful for a user just seeing the flag

warber: yeah totally forgot about the actual feature :D will do

UnseenWizzard (dismissed):
Lgtm! 

I think the help message for the command flag could be improved with a bit more information on what a support archive actually is, but as long as the documentation describes it the short version is ok as well

cmd/monaco/support/archive.go
```diff
@@ -0,0 +1,38 @@
...
+var SupportArchive bool
+
+func Archive(fs afero.Fs) error {
+	timeAnchorStr := timeutils.TimeAnchor().Format(trafficlogs.TrafficLogFilePrefixFormat)
+	zipFileName := "support-archive-" + timeAnchorStr + ".zip"
+	files := []string{
+		path.Join(log.LogDirectory, timeAnchorStr+"-"+"req.log"),
+		path.Join(log.LogDirectory, timeAnchorStr+"-"+"resp.log"),
+		path.Join(log.LogDirectory, timeAnchorStr) + ".log"}
+	return zip.Create(fs, zipFileName, files, false)
```
UnseenWizzard, line 37: Actually now that I think of it: It would be great if we just log where that the support archive was created and where to find it here

UnseenWizzard (approved):

PR #1084, opened 2023-07-13
Research/concurrent deploy

UnseenWizzard:
This is a very rough PoC of using the new graph-based sort returning weakly connected components (config sets that depend on each other) to deploy them concurrently.

This is not tested much and probably has some issues.
This just exists as a showcase of how this could work if implemented.


UnseenWizzard:
closing draft as actual implementation goes a slightly different route

PR #1085, opened 2023-07-13
Fix/rest error handling

UnseenWizzard:
Small fixes to ensure errors are not lost/overwritten and HTTP requests can be made concurrently.

These issues where found while testing #1084 and the same commits are currently present on that branch.

In some (unlikely to occur cases) errors are correctly returned and shown to the user

### DIFF
--- a/pkg/rest/request.go
+++ b/pkg/rest/request.go
@@ -123,7 +123,15 @@ func executeRequest(client *http.Client, request *http.Request) (Response, error
 			return Response{}, fmt.Errorf("HTTP request failed: %w", err)
 		}
 		defer func() {
-			err = resp.Body.Close()
+			if closeErr := resp.Body.Close(); closeErr != nil {
+				if err != nil {
+					// don't overwrite an actual error for a body close issue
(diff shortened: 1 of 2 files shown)


pkg/rest/request.go
```diff
@@ -123,7 +123,13 @@ func executeRequest(client *http.Client, request *http.Request) (Response, error
 			return Response{}, fmt.Errorf("HTTP request failed: %w", err)
 		}
 		defer func() {
-			err = resp.Body.Close()
+			closeErr := resp.Body.Close()
+			if err == nil {
```
Laubi, line 127:
```suggestion
			if err == nil && closeErr != nil {
```
That's what you mean, right?

UnseenWizzard: ah... that's what I get for deciding later to wrap the close error in a nicer message 😅 The simple if err == nil worked when the code inside it was just err = closeErr

UnseenWizzard: fixed - a bit differently than suggested

Laubi (approved):

PR #1093, opened 2023-07-19
feat: cache classic configs

warber:
At present, when Monaco deploys, it checks all configurations of a specific type to determine whether an object with a given name exists and needs updating. To enhance efficiency, a cache is introduced, similar to what has been done for settings objects and schema constraints. This cache helps optimize the process.

The code underwent slight modifications to enable the disabling of the caching mechanism. This adjustment is essential for stable end-to-end tests, as the cache might sometimes provide inaccurate results, especially when the API hasn't yet returned the most up-to-date values shortly after their creation.

Performance improvements when deploying classic api configs. :rocket:

### DIFF
--- a/cmd/monaco/integrationtest/integration_test_utils.go
+++ b/cmd/monaco/integrationtest/integration_test_utils.go
@@ -19,7 +19,7 @@
 package integrationtest
 
 import (
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/dynatrace"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/support"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/testutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client"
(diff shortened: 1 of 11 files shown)


cmd/monaco/integrationtest/integration_test_utils.go
```diff
@@ -33,9 +33,25 @@ import (
 )
 
 func CreateDynatraceClients(t *testing.T, environment manifest.EnvironmentDefinition) *client.ClientSet {
-	clients, err := dynatrace.CreateClientSet(environment.URL.Value, environment.Auth)
+	var clients *client.ClientSet
+	var err error
+	if environment.Auth.OAuth == nil {
+		clients, err = client.CreateClassicClientSet(environment.URL.Value, environment.Auth.Token.Value, client.ClientOptions{
+			SupportArchive:  support.SupportArchive,
+			CachingDisabled: true,
```
UnseenWizzard, line 41: Please mention somewhere (e.g. the commit message ;) ) WHY caching is disabled for E2E tests. This seems like a prime candidate for forgetting about stumbling over again in the future

UnseenWizzard, line 41: Please mention somewhere (e.g. the commit message ;) ) WHY caching is disabled for E2E tests. This seems like a prime candidate for forgetting about stumbling over again in the future

internal/cache/cache.go
```diff
@@ -18,15 +18,36 @@ package cache
...
 import "sync"
 
-type Cache[T any] struct {
+type Cache[T any] interface {
+	Get(key string) (T, bool)
+	Set(key string, entries T)
+	Delete(key string)
+}
+
+type DefaultCache[T any] struct {
```
UnseenWizzard, line 27: nit: maybe move the struct declaration down to line 46 to group with its methods?

pkg/client/dtclient/client.go
```diff
@@ -295,6 +298,18 @@ func WithAutoServerVersion() func(client *DynatraceClient) {
 	}
 }
 
+// WithCachingDisabled lets disable the client builtin caching mechanism for
```
UnseenWizzard, line 301:
```suggestion
// WithCachingDisabled allows disabling the client builtin caching mechanism for
```

pkg/client/dtclient/client.go
```diff
@@ -295,6 +298,18 @@ func WithAutoServerVersion() func(client *DynatraceClient) {
...
+// classic configs, schema constraints and settings objects
+func WithCachingDisabled(disabled bool) func(client *DynatraceClient) {
```
UnseenWizzard, line 303: As there's no way to re-enable the caches by setting the flag to `true`, and the name already states disabled I guess we could just remove the boolean?

UnseenWizzard (approved):
Other than lacking some documentation on why things are the way they are (e.g. in commit or comments), just nitpicks. 

LGTM functionality wise

pkg/client/dtclient/client.go
warber: as discussed offline, let's keep it to avoid `if`s at the caller's side

UnseenWizzard (approved):

PR #1094, opened 2023-07-24
feat: Allow exporting dependency graph as DOT file

UnseenWizzard:
Fully finish the previously experimental command by:
- adding tests
- not overwriting files if they exist alrady
- removing the featureflag

Dependency graphs can be exported as DOT files with a new CLI command

### DIFF
--- a/cmd/monaco/dependencygraph/dependencygraph.go
+++ b/cmd/monaco/dependencygraph/dependencygraph.go
@@ -15,10 +15,11 @@
 package dependencygraph
 
 import (
-	"errors"
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/errutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/mutlierror"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/timeutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
@@ -28,6 +29,23 @@ import (
 	"github.com/spf13/afero"
 	"path/filepath"
 )
+
+// ExportError is returned in case any error occurs while creating a dependency graph file
+type ExportError struct {
+	message string
+	// Reason is the underlying error that occurred
+	Reason error `json:"reason"`
+	// ManifestFile the export failed for
+	ManifestFile string `json:"manifestFile"`
+	// Environment the dependency graph failed to be exported for - omitted if the error is not specific to an environment
+	Environment string `json:"environment,omitempty"`
+	// Filepath of the file that failed to be created - omitted if the error is not related to a file
+	Filepath string `json:"filepath,omitempty"`
+}
+
+func (e ExportError) Error() string {
+	return fmt.Sprintf("%s: %v", e.message, e.Reason)
+}
 
 func writeGraphFiles(fs afero.Fs, manifestPath string, environmentNames []string, environmentGroups []string, outputFolder string) error {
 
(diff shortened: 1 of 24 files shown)


cmd/monaco/dependencygraph/dependencygraph.go
```diff
@@ -74,6 +76,18 @@ func writeGraphFiles(fs afero.Fs, manifestPath string, environmentNames []string
 			return fmt.Errorf("failed to encode dependency graph to DOT for environment %q: %w", e, err)
 		}
 		file := filepath.Join(folderPath, fmt.Sprintf("dependency_graph_%s.dot", e))
+
+		exists, err := afero.Exists(fs, file)
+		if err != nil {
+			return fmt.Errorf("failed to validate if output file %q already exists: %w", file, err)
+		}
+		if exists {
```
Laubi, line 84: Why did you decide to only append the date if the file already exists? Why not always append it?

cmd/monaco/dependencygraph/dependencygraph.go
```diff
@@ -74,6 +76,18 @@ func writeGraphFiles(fs afero.Fs, manifestPath string, environmentNames []string
...
+			time := timeutils.TimeAnchor().Format("20060102-150405")
+			newFile := filepath.Join(folderPath, fmt.Sprintf("dependency_graph_%s_%s.dot", e, time))
+			log.Debug("Output file %q already exists, creating %q instead", file, newFile)
```
Laubi, line 87: Should we log the files as structured logs?

cmd/monaco/dependencygraph/dependencygraph_test.go
Laubi, line 1: Would a test make sense where we test if we check if the edges are correctly within the dot file?

UnseenWizzard: Does make sense, but already exists for the graph package export method:
graph_test.go:137 `TestGraphExport`

I didn't want to replicate that into reading and asserting the file contents, but I can if you want

cmd/monaco/dependencygraph/dependencygraph.go
UnseenWizzard: done

UnseenWizzard: Here's a sample error log

```json
{
  "level": "error",
  "ts": "2023-07-24T16:02:43+02:00",
  "msg": "Failed to create dependency graph files: failed to load manifest \"cmd/monaco/dependencygraph/test-resources/manifest.yaml\": encountered multiple errors: [cmd/monaco/dependencygraph/test-resources/manifest.yaml:default:env1: failed to parse auth section: error parsing token: environment-variable \"TOKEN\" was not found, cmd/monaco/dependencygraph/test-resources/manifest.yaml:default:env2: failed to parse auth section: error parsing token: environment-variable \"TOKEN\" was not found]",
  "error": {
    "type": "dependencygraph.ExportError",
    "details": {
      "reason": {
        "errors": [
          {
            "manifestPath": "cmd/monaco/dependencygraph/test-resources/manifest.yaml",
            "reason": "failed to parse auth section: error parsing token: environment-variable \"TOKEN\" was not found",
            "environmentDetails": {
              "group": "default",
              "environment": "env1"
            }
          },
          {
            "manifestPath": "cmd/monaco/dependencygraph/test-resources/manifest.yaml",
            "reason": "failed to parse auth section: error parsing token: environment-variable \"TOKEN\" was not found",
            "environmentDetails": {
              "group": "default",
              "environment": "env2"
            }
          }
        ]
      },
      "manifestFile": "cmd/monaco/dependencygraph/test-resources/manifest.yaml"
    }
  },
  "manifestFile": "cmd/monaco/dependencygraph/test-resources/manifest.yaml",
  "outputFolder": ""
}
```

cmd/monaco/dependencygraph/dependencygraph_test.go
Laubi: Not a fan of the string-compare there as it could be flaky on library update, but it is tested - which is the most important thing

cmd/monaco/dependencygraph/dependencygraph.go
Laubi: We do append the timestamp always on the created download folder (except if we pass an argument setting the folder).
Since this command will probably be used only by a few customers, I think we can let it be as is.

Laubi (approved):

cmd/monaco/dependencygraph/dependencygraph.go
UnseenWizzard: On the folder yes, on the containing files with fixed names only if they exist.
Unlike the download folder, where the timestamp tells me "this is what I downloaded on {date}", there's no inherent value of having the timestamp in these files or the download's manifest.yaml. 

Thus I really think appending it only if needed makes more sense in this case, even if it means an extra if in the code

PR #1097, opened 2023-07-25
feat: Add command to generate deletefiles

UnseenWizzard:
* Introduce 'generate' command and group dependency graph generation under it
* Add command allowing to generate a delete file from a given manifest

Points up for discussion:
- should delete persistence structs be exposed, or a dedicated write (with a dedicated exported input datatype) be created, mirroring manifest and config writers? Currently I went the easy route and exported the structs
- should we resolve references that are not to an `id`. As this only impacts classic config API types, and only if their name happens to be a reference, I elected to just handle this as a warning and to be documented limitation.

* A new command allows generating delete.yaml files
* The not yet released dependency graph generation command will be available as `monaco generate graph` instead of `monaco graph`


Laubi (approved):

Laubi:
Most of my comments are nitpicks - I do not see obvious issues :D 
I would still implement the --projects flag, but that can also be done later

UnseenWizzard:
Comments addressed @Laubi :)

Laubi (approved):

PR #1100, opened 2023-07-26
fix(deploy): Ensure thread safety

UnseenWizzard:
Previously goroutines would write and read the same context, resulting in race conditions (and unwanted behaviour).

Additionally testing found additional data races in the dry-run's dummyclient & our logger setup


warber (dismissed):

UnseenWizzard:
:-? still ran into race conditions... @warber 

Commit amended to a new try just building the new context outside the routine and giving each it's own new context.WithValue

Laubi (approved):

PR #1102, opened 2023-07-27
feat: log coordinate and gid in human readable logs

warber:
To enhance the usefulness of the (human-readable) logs, this pull request introduces additional context to the log lines. Specifically, it enriches the logs with details about the coordinate and the graph ID (gid). The graph ID proves particularly valuable in linking logs that occur out of sequence when configurations are deployed concurrently. To conserve characters, the coordinate is logged using a shorthand syntax.

image

Zap allows implementing the `Encoder` interface which makes it possible to choose the exact format of each log line on your own. This was done to change the order of when context fields are printed. Per default, they are just appended as JSON to each log line. However, this made the logs look very clumsy and unclear and confusing to look at.

### DIFF
--- a/internal/loggers/zap/console.go
+++ b/internal/loggers/zap/console.go
@@ -1,0 +1,85 @@
+/*
+ * @license
+ * Copyright 2023 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


internal/loggers/zap/console.go
```diff
@@ -0,0 +1,85 @@
...
+			mu:  sync.RWMutex{},
+			moe: zapcore.NewMapObjectEncoder(),
+		},
+	}
+	return cloneEncoder
+}
+
+func (e fixedFieldsConsoleEncoder) EncodeEntry(entry zapcore.Entry, _ []zapcore.Field) (*buffer.Buffer, error) {
+	line := Get()
+	line.AppendString(fmt.Sprintf("%s ", entry.Time.Format(time.RFC3339)))
```
Laubi, line 61: Why append an empty space if you append a tab anyway?

warber: typo, i even adopted failing unit tests for this.... :picard-facepalm: :D

Laubi (approved):

PR #1107, opened 2023-07-31
fix: remove early return when errors occur during parallel deploy

warber:
Since, all of the go-routines are already started, and cannot be canceled caching the errors and return early does not make much sense.

### DIFF
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -199,9 +199,6 @@ func deployComponentsParallel(ctx context.Context, components []graph.SortedComp
 	// Collect errors from goroutines and append to the 'errs' slice.
 	for range components {
 		componentDeployErrs := <-errChan
-		if len(componentDeployErrs) > 0 && !opts.ContinueOnErr && !opts.DryRun {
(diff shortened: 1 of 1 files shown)


pkg/deploy/deploy.go
```diff
@@ -199,9 +199,6 @@ func deployComponentsParallel(ctx context.Context, components []graph.SortedComp
 	// Collect errors from goroutines and append to the 'errs' slice.
 	for range components {
 		componentDeployErrs := <-errChan
-		if len(componentDeployErrs) > 0 && !opts.ContinueOnErr && !opts.DryRun {
```
UnseenWizzard, line 202: Random thought: currently, we kinda "hard-cancel" any goroutine still running by the time the CLI execution finished after this return. Probably not what we want to happen anyway

UnseenWizzard (approved):

Laubi (approved):

Laubi:
Another random thought: We don't wait at all currently until all components are deployed, right? 
Shouldn't be there a Waitgroup 🤔

PR #1110, opened 2023-08-01
feat: deploy configs within a component in parallel

Laubi:

### DIFF
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -226,9 +226,12 @@ func (c *componentDeployer) deploy(ctx context.Context) []error {
 		roots := graph.Roots(c.graph)
 
 		for _, root := range roots {
-			go func(node graph2.Node) {
+			node := root.(graph.ConfigNode)
+			ctx := context.WithValue(ctx, log.CtxKeyCoord{}, node.Config.Coordinate)
+
+			go func(ctx context.Context, node graph.ConfigNode) {
 				errChan <- c.deployNode(ctx, node)
-			}(root)
+			}(ctx, node)
 		}
 
 		for range roots {
@@ -249,11 +252,8 @@ func (c *componentDeployer) deploy(ctx context.Context) []error {
 	return errs
 }
 
-func (c *componentDeployer) deployNode(ctx context.Context, n graph2.Node) error {
-	cnf := n.(graph.ConfigNode).Config
-
-	ctx = context.WithValue(ctx, log.CtxKeyCoord{}, cnf.Coordinate)
-	entity, err := deployFunc(ctx, cnf, c.clients, c.apis, &c.resolvedEntities)
+func (c *componentDeployer) deployNode(ctx context.Context, n graph.ConfigNode) error {
+	entity, err := deployFunc(ctx, n.Config, c.clients, c.apis, &c.resolvedEntities)
 
 	// lock changes we will make to shared variables. Writing them is trivial compared to any http request
 	c.lock.Lock()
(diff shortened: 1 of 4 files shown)


pkg/deploy/deploy.go
```diff
@@ -208,51 +210,117 @@ func deployComponentsParallel(ctx context.Context, components []graph.SortedComp
...
+
+type componentDeployer struct {
+	lock             sync.Mutex
+	graph            deployGraph
+	clients          ClientSet
+	resolvedEntities entityMap
+	apis             api.APIs
+}
+
+func (c *componentDeployer) deployComponent(ctx context.Context) []error {
```
UnseenWizzard, line 227: nit: why not just `deploy` if it's a function of the componentDeployer already? 

```suggestion
func (c *componentDeployer) deploy(ctx context.Context) []error {
```

pkg/deploy/deploy.go
```diff
@@ -208,51 +210,117 @@ func deployComponentsParallel(ctx context.Context, components []graph.SortedComp
 	return errs
 }
 
-func deployComponent(ctx context.Context, component graph.SortedComponent, clientSet ClientSet, apis api.APIs, opts DeployConfigsOptions) []error {
+type deployGraph interface {
```
UnseenWizzard, line 213: Maybe this is an interface we can actually fulfill with `graph.ConfigGraph`? And not pull the underlying graph library directly into deploy, which you IMO reasonably argued against before.

We could implement things like Roots, Remove, etc. on this interface - which would allow us things like replacing the gonum lib with another one, without modifying deploy.

I think we just need Roots, RemoveNode and Copy methods?

Laubi: simple.DirectedGraph is still in the deploy, and we'd need to alias `graph.Node`, `graph.Copy`, and `simple.NewDirectedGraph` as well. 
For now, I replaced the interface with the ConfigGraph as suggested and added the NodeRemover interface to the ConfigGraph.

pkg/deploy/deploy.go
```diff
@@ -208,51 +210,111 @@ func deployComponentsParallel(ctx context.Context, components []graph.SortedComp
...
-	entityMap := newEntityMap(apis) //entityMap is only used to when resolving parameter references, and configs that reference each other are in the same component; no global entity map is needed
+	errChan := make(chan error)
+	for c.graph.Nodes().Len() != 0 {
+		roots := graph.Roots(c.graph)
 
-	g := component.Graph
-	sortedNodes := component.SortedNodes
+		for _, root := range roots {
+			go func(node graph2.Node) {
+				errChan <- c.deployNode(ctx, node)
```
UnseenWizzard, line 230: I think you can run into the same race condition as the go routines one level up in DeployParallel did? 
You pass the same ctx pointer into the different `deployNode` methods, which modify it (line 255)

If you create a copy of the context already with the Coordinate value in it and pass that into the go func (compare line 194) you'd be safe from that. And you have the coordinate available here from the root ConfigNode

pkg/deploy/deploy.go
```diff
@@ -208,51 +210,111 @@ func deployComponentsParallel(ctx context.Context, components []graph.SortedComp
...
 
-			if !errors.Is(err, skipError) {
-				deploymentFailed = true
-				errs = append(errs, err)
-				if !opts.ContinueOnErr && !opts.DryRun {
-					return errs
-				}
-			}
+	parentCnf := parent.(graph.ConfigNode).Config
+	rootCnf := parent.(graph.ConfigNode).Config
```
UnseenWizzard, line 280: should this be on the `root` rather than parent?

pkg/deploy/deploy.go
```diff
@@ -208,51 +210,111 @@ func deployComponentsParallel(ctx context.Context, components []graph.SortedComp
...
+
+		c.removeChildren(ctx, child, root, failed)
+
+		c.graph.RemoveNode(child.ID())
 	}
-	return errs
+}
+func deployComponent(ctx context.Context, component graph.SortedComponent, clientSet ClientSet, apis api.APIs, _ DeployConfigsOptions) []error {
+	g := simple.NewDirectedGraph()
+	graph2.Copy(g, component.Graph)
```
UnseenWizzard, line 308: is this copy necessary?

Laubi: It's a copy to not change the component.
My reasoning is that if you pass the graph in, you might not be expecting it to be completely deleted. My approach is to make a copy so that for the outside the graph does not change at all. Inside, we do remove each and every node of the component.

pkg/deploy/deploy.go
Laubi: I can't see how it would, as the WithValue is implemented as follows:
```go
func WithValue(parent Context, key, val any) Context {
	if parent == nil {
		panic("cannot create context from nil parent")
	}
	if key == nil {
		panic("nil key")
	}
	if !reflectlite.TypeOf(key).Comparable() {
		panic("key is not comparable")
	}
	return &valueCtx{parent, key, val}
}
```
> You pass the same ctx pointer into the different deployNode methods, which modify it (line 255)
It should not matter if we call it before or after as the parent is never modified.
To be sure, I tested it: https://go.dev/play/p/woxG3YbTiNZ


However, I think it makes sense to move it outside, as it is IMO a bit cleaner, and we can use the ConfigNode directly, meaning less unnecessary casts
It helps us generally and is a bit clearer imo.

pkg/deploy/deploy.go
```diff
@@ -208,51 +210,111 @@ func deployComponentsParallel(ctx context.Context, components []graph.SortedComp
...
+		}
 	}
 
-	dontDeploy := map[int64]struct{}{} //look-up map marking Nodes (by ID) that should not be deployed, because something they depend on was skipped or failed
+	close(errChan)
+
+	return errs
+}
+
+func (c *componentDeployer) deployNode(ctx context.Context, n graph2.Node) error {
```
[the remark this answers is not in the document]
Laubi: Changed it

pkg/deploy/deploy.go
Laubi: Nice catch!

pkg/deploy/deploy.go
UnseenWizzard: You're right, passing the ctx on, you're modifying a different context. 
The forgot the original issue had a different setup (and I forgot we had another issue in the logger further down)

UnseenWizzard (approved):

PR #1124, opened 2023-08-09
feature: Allow loading and writing bucket configs

Laubi:


internal/featureflags/featureflags.go
```diff
@@ -188,3 +188,10 @@ func DependencyGraphBasedDeployParallel() FeatureFlag {
 		defaultEnabled: false,
 	}
 }
+
+func Buckets() FeatureFlag {
+	return FeatureFlag{
+		envName:        "MONACO_FEAT_BUCKETS",
```
jskelin, line 194: for automation the name is `MONACO_FEAT_AUTOMATION_RESOURCES`
Should we keep that pattern?

jskelin (dismissed):
I would say that tests that prevent loading/writing configuration like next would be good to have
```
configs:
- id: profile-id
  type: 
    api: bucket
```

internal/featureflags/featureflags.go
Laubi: Good question. I thought about it but then decided against it. 
In automation, we had multiple resources: workflows, business calendars, and scheduling rules.
Here we only have buckets, and thats why I went with only 'MONACO_FEAT_BUCKETS'.

pkg/config/config_loader.go
```diff
@@ -370,6 +370,11 @@ func getType(typeDef typeDefinition) (Type, error) {
 		}, nil
 
 	case typeDef.isClassic():
+
+		if typeDef.Api == ApiTypeBucket {
```
UnseenWizzard, line 374: @Laubi when creating a followup for not allowing the `api: format` please link to this line and what refactorings would be needed to split loading a bucket definition from a shorthand API definition

pkg/config/config_writer.go
```diff
@@ -308,6 +308,14 @@ func extractConfigType(context *serializerContext, config Config) (typeDefinitio
 				Resource: t.Resource,
 			},
 		}, nil
+	case BucketType:
```
UnseenWizzard, line 311: Do we even want to write anything yet? IMO by the time we have bucket downloads the YAML format should be specific and not reuse the API definition. 

I'd just drop this case and let any writes fall into the default error case until we actually handle downloading and writing

UnseenWizzard (approved):
Okay for me as a basis for the feature flags bucket API create/update MVP. 
I'd actually strip it down a bit more, and not even implement the wrong writing though.

pkg/config/config_writer.go
Laubi: Dropped

pkg/config/config_loader.go
Laubi: Done

UnseenWizzard (approved):

PR #1127, opened 2023-08-11
fix(conversion): Don't fail when the same env var is referenced

Laubi:
When a Monaco v1 project used the same env variable both in the template and the config, we used to print an error that there is a collision.

This change compares now not only the name but also the payload of the variable. If the payload's the same, we drop one of the parameters. They're the same afterward and may be reused.

### DIFF
--- a/pkg/converter/converter_test.go
+++ b/pkg/converter/converter_test.go
@@ -581,7 +581,7 @@ func TestConvertConfigs(t *testing.T) {
 
 	properties := map[string]map[string]string{
 		configId: {
-			"name":                 configName,
+			"name":                 fmt.Sprintf("%s {{ .Env.%s }}", configName, envVariableName),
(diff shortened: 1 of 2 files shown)


pkg/converter/converter_test.go
```diff
@@ -581,7 +581,7 @@ func TestConvertConfigs(t *testing.T) {
 
 	properties := map[string]map[string]string{
 		configId: {
-			"name":                 configName,
+			"name":                 fmt.Sprintf("%s {{ .Env.%s }}", configName, envVariableName),
```
UnseenWizzard, line 584: nit: do we really want to keep stuffing additional cases into the same test? Might be nicer to add a dedicated test for this at least this very specific case.

UnseenWizzard (approved):

PR #1128, opened 2023-08-17
`deploy` functionality for Grail Bucket 

jskelin:
- feat: new client for bucket
- feat: create new bucket (without update)
- feat: upsert command for bucket (with update and create, without retry)
- feat: add concurrency to bucket upsert
- feat: extend platform exclusive configuration check with BucketType
- feat: add dummy client for Bucket
- test: add test cases for bucket.upsert
- doc: add documentation

Add cover for Grail Bucket. For now, only `deploy` is covered.

### DIFF
--- a/cmd/monaco/integrationtest/assert.go
+++ b/cmd/monaco/integrationtest/assert.go
@@ -26,9 +26,8 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/automation"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/bucket"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/dtclient"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/rest"
 	"github.com/stretchr/testify/assert"
-	"strings"
-
 	"testing"
 	"time"
 
@@ -252,22 +251,28 @@ func AssertBucket(t *testing.T, client bucket.Client, env manifest.EnvironmentDe
 	}
 
 	_, err := client.Get(context.TODO(), expectedId)
-	if err != nil && !strings.Contains(err.Error(), "404") {
-		// 404 is an allowed error to continue
+	exists := true
+
+	var respErr rest.RespError
+	if errors.As(err, &respErr) {
+		if respErr.StatusCode == 404 {
+			exists = false
+		} else {
+			assert.NoError(t, err)
+		}
+	} else if err != nil {
 		assert.NoError(t, err)
 	}
 
-	notFound := err != nil && strings.Contains(err.Error(), "404")
-
 	if cfg.Skip {
-		assert.Truef(t, notFound, "Skipped Automation Object should NOT be available but was. environment.Environment: '%s', failed for '%s'", env.Name, cfg.Coordinate)
+		assert.Falsef(t, exists, "Skipped Automation Object should NOT be available but was. environment.Environment: '%s', failed for '%s'", env.Name, cfg.Coordinate)
 		return
 	}
 
 	if available {
-		assert.Falsef(t, notFound, "Automation Object should be available, but wasn't. environment.Environment: '%s', failed for '%s'", env.Name, cfg.Coordinate)
+		assert.Truef(t, exists, "Automation Object should be available, but wasn't. environment.Environment: '%s', failed for '%s'", env.Name, cfg.Coordinate)
 	} else {
-		assert.Truef(t, notFound, "Automation Object should NOT be available, but was. environment.Environment: '%s', failed for '%s'", env.Name, cfg.Coordinate)
+		assert.Falsef(t, exists, "Automation Object should NOT be available, but was. environment.Environment: '%s', failed for '%s'", env.Name, cfg.Coordinate)
 	}
 }
 
--- a/cmd/monaco/integrationtest/assert.go
+++ b/cmd/monaco/integrationtest/assert.go
@@ -24,8 +24,10 @@ import (
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/automationutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/automation"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/bucket"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/dtclient"
 	"github.com/stretchr/testify/assert"
+	"strings"
 
 	"testing"
 	"time"
@@ -135,6 +137,12 @@ func AssertAllConfigsAvailability(t *testing.T, fs afero.Fs, manifestPath string
 						return
 					}
 					AssertAutomation(t, *clients.Automation(), env, available, typ.Resource, theConfig)
+				case config.BucketType:
+					if clients.Bucket() == nil {
+						t.Errorf("can not assert existience of Bucket config %q) because no BucketClient exists - was the test env not configured as Platform?", theConfig.Coordinate)
+						return
+					}
+					AssertBucket(t, *clients.Bucket(), env, available, theConfig)
 				default:
 					t.Errorf("Can not assert config of unknown type %q", theConfig.Coordinate.Type)
 				}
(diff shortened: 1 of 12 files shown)


UnseenWizzard:
@jskelin I think the first commit needs to be `docs` instead of `doc` to comply with the semantic commit rules. 
I've hit rebase in github before I saw that, so please make sure to fetch before rewriting the commit :)

cmd/monaco/integrationtest/v2/bucket_integration_test.go
```diff
@@ -0,0 +1,52 @@
...
+		cmd := runner.BuildCli(fs)
+		cmd.SetArgs([]string{"deploy", "--verbose", manifest})
+		err := cmd.Execute()
+		assert.NoError(t, err)
+
+		// This causes an Update of all automation objects
+		cmd = runner.BuildCli(fs)
+		cmd.SetArgs([]string{"deploy", "--verbose", manifest})
+		err = cmd.Execute()
+		assert.NoError(t, err)
```
UnseenWizzard, line 50: for a future PR/final implementation: we should extend `integrationtest/v2/assert.go` to allow checking for bucket API types and extend the test to include the assertion

pkg/client/bucket/bucket.go
```diff
@@ -0,0 +1,210 @@
...
+		return Response{}, fmt.Errorf("id must be non empty")
+	}
+	c.limiter.ExecuteBlocking(func() {
+		result, err = c.upsert(ctx, id, data)
+	})
+	return
+}
+
+// upsert first tries to create the bucket. If it does not work, we try to update it.
+// This has the advantage that we don't need to try to fetch the object first.
```
UnseenWizzard, line 73: This saves us 1 API call in the create case but potentially triggers unnecessary work on the Dynatrace environment. 
Generally fine with it, but I'm not sure if saving us one GET is worth the potential processing impact on Dynatrace. (We could check with GET /{id} in this case, right? we don't need to list all?)

Is the intention to keep this for the final version as well?

pkg/client/bucket/bucket.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return Response{}, err
+	}
+
+	r, err := c.client.Post(ctx, u, data)
+	if err != nil {
+		return Response{}, fmt.Errorf("unable to create object with ID %q: %w", id, err)
+	}
+	if !r.IsSuccess() {
+		return Response{}, rest.NewRespErr(fmt.Sprintf("failed to update object with ID %q (HTTP %d): %s", id, r.StatusCode, string(r.Body)), r).WithRequestInfo(http.MethodPut, u)
```
UnseenWizzard, line 109:
```suggestion
		return Response{}, rest.NewRespErr(fmt.Sprintf("failed to create object with ID %q (HTTP %d): %s", id, r.StatusCode, string(r.Body)), r).WithRequestInfo(http.MethodPut, u)
```

pkg/client/bucket/bucket.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	err = setBucketName(id, &data)
+	if err != nil {
+		return Response{}, err
+	}
+
+	r, err := c.client.Post(ctx, u, data)
+	if err != nil {
+		return Response{}, fmt.Errorf("unable to create object with ID %q: %w", id, err)
```
UnseenWizzard, line 106:
```suggestion
		return Response{}, fmt.Errorf("failed to create object with ID %q: %w", id, err)
```

IMO wording should be same for both client and HTTP statuscode errors

pkg/client/bucket/bucket.go
```diff
@@ -0,0 +1,210 @@
...
+	u.RawQuery = q.Encode()
+
+	var m map[string]any
+	err = json.Unmarshal(data, &m)
+	if err != nil {
+		return Response{}, fmt.Errorf("unable to unmarshal template: %w", err)
+	}
+	m["bucketName"] = b.BucketName
+	m["version"] = b.Version
+	m["status"] = b.Status
```
UnseenWizzard, line 143: is the status required to be set on update? could we just omit it?

pkg/client/bucket/bucket.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func (c Client) update(ctx context.Context, b Response, data []byte) (Response, error) {
+
+	u, err := url.Parse(c.url)
+	if err != nil {
+		return Response{}, fmt.Errorf("failed to parse url: %w", err)
+	}
+
+	u.Path, err = url.JoinPath(u.Path, endpoint, b.BucketName)
```
UnseenWizzard, line 127: nit: especially this use makes me wonder if we should call the `id` argument of other methods to explicitly be `bucketName` as well? for clarity

pkg/client/bucket/bucket.go
```diff
@@ -0,0 +1,210 @@
...
+	return nil
+}
+
+func (c Client) get(ctx context.Context, id string) (Response, error) {
+	u, err := url.JoinPath(c.url, endpoint, id)
+	if err != nil {
+		return Response{}, fmt.Errorf("faild to create sound url: %w", err)
+	}
+
+	retry := rest.RetrySetting{
```
UnseenWizzard, line 181: What lead to adding retries to the GET but none to POST and PUT calls for this API/client?

pkg/deploy/bucket.go
```diff
@@ -0,0 +1,53 @@
...
+
+	return &parameter.ResolvedEntity{
+		EntityName: id,
+		Coordinate: c.Coordinate,
+		Properties: properties,
+	}, nil
+}
+
+func coordinateID(c coordinate.Coordinate) string {
+	return fmt.Sprintf("%s_%s_%s", c.Project, c.Type, c.ConfigId)
```
UnseenWizzard, line 52: I assume the API does not allow us to just use the colon-separated format of coordinate.String() (`%s:%s:%s`)?

cmd/monaco/integrationtest/v2/bucket_integration_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/spf13/afero"
+)
+
+// tests all configs for a single environment
+func TestIntegrationBucket(t *testing.T) {
+
+	configFolder := "test-resources/integration-bucket/"
+	manifest := configFolder + "manifest.yaml"
+	specificEnvironment := ""
+	t.Setenv("MONACO_FEAT_BUCKETS", "1")
```
UnseenWizzard, line 36: I think the actual feature flag is still missing!

UnseenWizzard: Nevermind, I forgot that https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1124 already introduced the flag and loading bucket configs is behind the flag.

pkg/client/bucket/bucket.go
Laubi: With a proper list, we could save on requests (not data). However, with a simple Get, we ran into concurrency issues already in our very basic integration test: The object was created, but the GET did fail as the server said 'does not exist' as it has not yet 'recognized' the object; thus the update failed, as we expected to create an object. That's why we also have the retry in place. 
However, with the retry, the initial create took multiple requests, as the object did not exist in the first place (of cause, has not been created yet). With many objects, this really becomes an issue.

So actually, currently, without the list, we do safe on requests. Yes, the create could be more nice, but overall we do correctly create the object and don't need to execute multiple GET requests in order to do so.

pkg/client/bucket/bucket.go
Laubi: My issue is that that might be confusing. Buckets can have a name, the displayName. The bucketName is the ID. 
I would keep it as id, to clearly distinguish between {display,bucket}Name

pkg/client/bucket/bucket.go
Laubi: It's required. We can't update a bucket to something that it isn't, e.g. 'creating'

pkg/client/bucket/bucket.go
Laubi: As mentioned before, we had issues in our first approach in fetching buckets after the creation. We changed the code in a way that it's currently not necessary to have the retry, but keeping it would also not hurt too much.

pkg/client/bucket/bucket.go
UnseenWizzard: But we always use the `ID` as the JSON object's `bucketName` field. As the API has no ID but does have the bucketName as a dedicated field I find using that less confusing tbh

pkg/deploy/bucket.go
Laubi: Added a comment describing the reason. Simple reason: Colons are not allowed.

cmd/monaco/integrationtest/v2/bucket_integration_test.go
Laubi: Implemented

pkg/client/bucket/bucket.go
```diff
@@ -74,19 +75,19 @@ func (c Client) Upsert(ctx context.Context, id string, data []byte) (result Resp
...
+	log.WithCtxFields(ctx).WithFields(field.Error(err)).Debug("Failed to create new object with ID %q; trying to update existing: %s", id, err)
 
 	b, err := c.Get(ctx, id)
 	if err != nil {
 		return Response{}, fmt.Errorf("failed to get object with ID %q: %w", id, err)
 	}
 
 	r, err = c.update(ctx, b, data)
-	log.WithCtxFields(ctx).Debug("updated object with ID %q", id)
+	log.WithCtxFields(ctx).Debug("Update bucket with ID %q", id)
```
UnseenWizzard, line 90: Update**d**

cmd/monaco/integrationtest/assert.go
```diff
@@ -234,6 +242,35 @@ func AssertAutomation(t *testing.T, c automation.Client, env manifest.Environmen
...
+
+	var expectedId string
+	if cfg.OriginObjectId != "" {
+		expectedId = cfg.OriginObjectId
+	} else {
+		expectedId = deploy.BucketId(cfg.Coordinate)
+	}
+
+	_, err := client.Get(context.TODO(), expectedId)
+	if err != nil && !strings.Contains(err.Error(), "404") {
```
UnseenWizzard, line 255: This method returns a ResponseError doesn't it? 
Not 100% necessary for a testing util, but I think it'd be cleaner to errors.As it to a Response error and look at the actual status code for the checks.

UnseenWizzard (approved):
Thanks for addressing all my comments @Laubi 👍

PR #1133, opened 2023-08-22
refactor: Log errors correctly

Laubi:
Improves logging errors. 
We have one occurrence where we sometimes log a nil error, which results in the string `%w!(nil)`, and some other cases where the error struct is logged as `%w!(full error)`. Now the error string is logged, and the error field is set instead.


pkg/client/metadata/metadata.go
```diff
@@ -53,7 +54,11 @@ func GetDynatraceClassicURL(ctx context.Context, client *rest.Client, environmen
 
 	resp, err := client.Get(ctx, endpointURL)
 	if !resp.IsSuccess() || err != nil {
-		log.Debug("failed to query classic environment url from %q, falling back to deprecated endpoint %q: %v (HTTP %v)", ClassicEnvironmentDomainPath, DeprecatedClassicEnvDomainPath, err, resp.StatusCode)
+		if err == nil {
+			log.Debug("failed to query classic environment url from %q, falling back to deprecated endpoint %q (HTTP %v)", ClassicEnvironmentDomainPath, DeprecatedClassicEnvDomainPath, resp.StatusCode)
```
UnseenWizzard, line 58: I think it'd be helpful to also log the response body if this happens. The status code alone might not tell a user or us enough to resolve an issue with this

cmd/monaco/integrationtest/v2/test-resources/integration-bucket/project/bucket.json
```diff
@@ -1,6 +1,5 @@
 {
   "bucketName": "custom_logs",
```
UnseenWizzard, line 2: We should remove the bucketName from this one too. While deployment will overwrite it, people might look at our test configs as samples and having this field set can be misleading

UnseenWizzard (changes requested):
Log changes look good, but please split out the accidentally added bucket API e2e test!

cmd/monaco/integrationtest/v2/bucket_integration_test.go
```diff
@@ -41,16 +40,41 @@ func TestIntegrationBucket(t *testing.T) {
...
+		cmd.SetArgs([]string{"deploy", "--verbose", manifest, "-p", "project"})
 		err = cmd.Execute()
 		assert.NoError(t, err)
 
-		integrationtest.AssertAllConfigsAvailability(t, fs, manifest, []string{}, "", true)
+		integrationtest.AssertAllConfigsAvailability(t, fs, manifest, []string{"project"}, "", true)
+	})
+}
+
+func TestIntegrationComplexBucket(t *testing.T) {
```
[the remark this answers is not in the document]
Laubi: Yes, was by accident. Will fiddle it out again :D

cmd/monaco/integrationtest/v2/test-resources/integration-bucket/project/bucket.json
Laubi: I kept it in with the intent that it should work :(

UnseenWizzard (approved):

PR #1134, opened 2023-08-22
fix: Check classic config API configs for unique names in graph based deployment

jskelin:
With the new graph-based deploy functionality it is no more possible to check for name uniqueness during the deployment. It needs to be done before!
With this change, new method `ForEveryConfigDo` for the `project` structure is added. Utilizing that function, `name` argument for every settings configuration is checked.


cmd/monaco/deploy/deploy.go
```diff
@@ -111,6 +121,64 @@ func deployConfigs(fs afero.Fs, manifestPath string, environmentGroups []string,
...
+
+		if uniqueList[c.Environment] == nil {
+			uniqueList[c.Environment] = make(map[classicEndpoint][]config.Config)
+		}
+
+		for _, c2 := range uniqueList[c.Environment][a.Api] {
+			n1, err := getNameForConfig(c)
+			if err != nil {
+				errs = append(errs, err)
+				return
```
UnseenWizzard, line 147: if I'm not mistaken this method returns whenever an error occurs. Does it really need a slice of errors of can you just return a single error/nil

pkg/project/v2/project.go
```diff
@@ -62,10 +68,38 @@ func (p Project) HasDependencyOn(environment string, project Project) bool {
...
+			for _, c := range cs {
+				for _, f := range actions {
+					f(c)
+				}
+			}
+		}
+	}
+}
+
+func (p *Project) GetConfigurations(filter ...func(c config.Config) bool) []config.Config {
```
UnseenWizzard, line 91: I will be removing this method (it is unused anyway).
Getting all configurations in a project does generally not make sense, as they are duplicated per environment (with possible overwrites).

So for a given project of a single config, in a manifest defining three environments, this method would return a slice containing that config three times. 

Access to a projects configurations should always be for a specific environment, which can be done via the Env -> ConfigsPerType map.

pkg/project/v2/project.go
```diff
@@ -16,36 +16,44 @@ package v2
...
 
-// ConfigsPerTypePerEnvironments is a map of environment to api to configs
-type ConfigsPerTypePerEnvironments map[string]ConfigsPerType
+	// ConfigsPerEnvironment is a map of EnvironmentName to configs. This is a flattened version of ConfigsPerTypePerEnvironments
+	ConfigsPerEnvironment map[EnvironmentName][]config.Config
 
-// ConfigsPerEnvironment is a map of environment to configs
-type ConfigsPerEnvironment map[string][]config.Config
+	// DependenciesPerEnvironment is a map of EnvironmentName to project IDs
+	DependenciesPerEnvironment map[EnvironmentName][]string
```
warber, line 37: since you've used type aliases a lot, why not also use it here
```
DependenciesPerEnvironment map[EnvironmentName][]ProjectID
```

pkg/deploy/internal/classic/validation.go
```diff
@@ -0,0 +1,102 @@
...
+				return
+			}
+
+			if cmp.Equal(n1, n2) {
+				var nameDetails string
+				if s, ok := n1.(string); ok {
+					nameDetails = fmt.Sprintf(": %s", s)
+				}
+
+				errs = errs.Append(c.Environment, fmt.Errorf("duplicated config name found: configurations %q and %q define the same \"name\"%s", c.Coordinate, c2.Coordinate, nameDetails))
```
warber, line 70: errs = errs.Append(c.Environment, fmt.Errorf("duplicated config name found: configurations %q and %q define the same \"name\" %s", c.Coordinate, c2.Coordinate, nameDetails))

pkg/deploy/internal/resolve/resolve.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter/resolve"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/errors"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/internal/entitymap"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project/v2/sort"
+)
+
+func Properties(c *config.Config, entityMap *entitymap.EntityMap) (parameter.Properties, []error) {
```
warber, line 28: please add code doumentation

pkg/deploy/internal/resolve/resolve.go
```diff
@@ -0,0 +1,58 @@
...
+	errors = append(errors, errs...)
+
+	if len(errors) > 0 {
+		return nil, errors
+	}
+
+	return properties, nil
+}
+
+func ExtractConfigName(conf *config.Config, properties parameter.Properties) (string, error) {
```
warber, line 44: pls add code documentation

pkg/deploy/internal/resolve/resolve.go
warber, line 1: not sure if this deserves its own package.
maybe the code should just live in the "other" resolve package (v2/pkg/config/parameter/resolve)

UnseenWizzard: I think the Properties resolution could move there, but it's currently only used by deploy.
The entity name resolver is only used by deploy and currently uses deploy specific errors. 

I'm not sure it's worth refactoring and moving these out to the parameter package just for the sake of not having an internal package.

warber: it's more for the sake of avoiding questions like "what's the difference between the one and the other "resolve" package. 
but we can leave it like it is if you want :)

UnseenWizzard: After a closer look, every call to `config/parameter/resolve/ParameterValues` always first sorts the parameter. 

For deployment the `Properties` methods does it before calling `ParameterValues` and for all other uses it's called explicitly right before calling `ParameterValues`.

I will refactor `ParameterValues` to also sort the paramerts, rather than expecting a sorted list of them...
And then just leave an `extract.ConfigName()` in a `deploy/internal/extract` pkg for the deploy methods that use it.

pkg/deploy/internal/resolve/resolve.go
UnseenWizzard: function removed, config/parameter/resolve/ParameterValues documented instead

pkg/deploy/internal/classic/validation.go
UnseenWizzard: ammended fixing into the  `refactor(deploy): Log actual name value if available and match old logs more closely if duplicate name is found` commit

cmd/monaco/integrationtest/assert.go
```diff
@@ -119,12 +118,7 @@ func AssertAllConfigsAvailability(t *testing.T, fs afero.Fs, manifestPath string
 				continue
 			}
 
-			configParameters, errs := sort.Parameters(theConfig.Group, theConfig.Environment, theConfig.Coordinate, theConfig.Parameters)
-			testutils.FailTestOnAnyError(t, errs, "sorting of parameter values failed")
-
-			parameters = append(parameters, configParameters...)
```
UnseenWizzard, line 125: fun sidenote: this was pretty useless before.. we only ever happened the configParameters to it and directly used it to resolve parameter values...

warber (dismissed):

pkg/config/sort_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
```
UnseenWizzard, line 1: fixuped in this forgotten build tag

UnseenWizzard: @warber please re-approve :) 
vet was luckily failing here because this test imported something this also behind the unit build tag

UnseenWizzard:
note: golangcilint flags a pre-existing TODO to cleanup code once the sequential deploy is gone.
The line was moved in this PR. I will ignore the check and deactivate it temporarily to merge this.

warber (approved):

PR #1137, opened 2023-08-22
feat: Enable graph based parallel deploy by default

Laubi:

Enables parallel deploy by default.
Depends on #1134

### DIFF
--- a/pkg/deploy/deploy_test.go
+++ b/pkg/deploy/deploy_test.go
@@ -18,6 +18,7 @@ package deploy_test
 
 import (
 	"fmt"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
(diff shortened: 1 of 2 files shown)


pkg/deploy/deploy_test.go
```diff
@@ -747,7 +748,9 @@ func TestDeployConfigsWithDeploymentErrors(t *testing.T) {
 		deploy.EnvironmentInfo{Name: env}: deploy.DummyClientSet,
 	}
 
-	t.Run("deployment error - stop on error", func(t *testing.T) {
+	t.Run("[non-parallel] deployment error - stop on error", func(t *testing.T) {
```
UnseenWizzard, line 751: FYI @Laubi I've replaced your original "skip test if flag is on" version with one that explicitly setup the flags for the tests. 
I think that makes the desired behaviour clearer, and ensures all  behaviours are tested instead of basically always skipping the continue-test. 

It should also make it obvious which tests to remove when we drop the non parallel deployments eventually

warber (approved):

PR #1140, opened 2023-08-23
fix: Retry on bucket conflict

Laubi:
It can happen that while we update a bucket, the bucket's state changes. In that case, we want to retry again after some timeout.

### DIFF
--- a/pkg/client/bucket/bucket.go
+++ b/pkg/client/bucket/bucket.go
@@ -20,6 +20,7 @@ package bucket
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/concurrency"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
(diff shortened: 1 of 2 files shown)


pkg/client/bucket/bucket.go
```diff
@@ -85,17 +91,32 @@ func (c Client) upsert(ctx context.Context, bucketName string, data []byte) (Res
...
+		}
+
+		r, err = c.update(ctx, b, data)
+		if err == nil {
+			log.WithCtxFields(ctx).Debug("Updated bucket with bucketName %q", bucketName)
+			return r, nil
+		}
+
+		var respErr rest.RespError
+		if !(errors.As(err, &respErr) && respErr.StatusCode == http.StatusConflict) {
```
UnseenWizzard, line 112: nit: I wonder if this would be easier to read if it wasn't inverted but an "early exit" type sleep + continue if there's a Conflict HTTP status error. And the err == nil block was the normal path instead of a condition

UnseenWizzard (approved):

pkg/client/bucket/bucket.go
Laubi: I'd stay we can try it in the shared lib :)

PR #1141, opened 2023-08-23
refactor: Split up config package and simplify some code

Laubi:
Refactors the config writing/loading part.
Most notable points:
* Split up config -> config "core", reading, writing, persistence.
* Split up some files to multiple files in resulting packages 
* Config loader no longer loads multiple configs but only one -> moved to project loading
* Simplify loadconfigs


pkg/config/loader/config_loader.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/files"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/maps"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/slices"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
```
UnseenWizzard, line 21: nit: importing the parent package always feels a bit iffy to me, and like it will end with a cyclic dependency at some point

Laubi: I think, in this case, it's absolutely fine. If the config in any case depends on config-loading or config-writing, we did something absolutely wrong.
We could also move it to a completely different set of packages, away from the config. 

Similar case for the parameters

UnseenWizzard: > If **the config** in any case depends on config-loading or config-writing, we did something absolutely wrong.

That depends fully on what the "config" package actually is.
I'd argue that especially with sub-packages it can be understood as "all things config related", rather than "the config type". 

I agree that there is something very wrong if the config type ever depends on the loader.
But I can see how someone taking a casual look at this a few months from now or new to the code-base could end up adding something that causes a cycle/breaks current understanding. 

I've marked it as a nitpick because I agree it's fine in this case and can't think of a much better way. A config/type pkg makes the use of Config strange to read (type.Config)...
A fully separate config_loader package might actually not be too bad now that I think of it though?

pkg/writer/writer.go
```diff
@@ -74,7 +75,7 @@ func writeProjects(context *WriterContext, projectDefinitions manifest.ProjectDe
 
 		configs := collectAllConfigs(p)
 
-		errs := config.WriteConfigs(&config.WriterContext{
+		errs := writer.WriteConfigs(&writer.WriterContext{
```
UnseenWizzard, line 78: The new import paths seem pretty odd, especially in this case.
If we use readability of the public interface for design, I think a configwriter package or staying in config would make sense :/ `configwriter.Write()` / `config.Write()`.

Thinking about the loader, especially now loading a single configuration, `config.Load(...) config.Config, err` would be a very nice interface.

pkg/config/loader/paramters.go
```diff
@@ -0,0 +1,188 @@
+/*
```
UnseenWizzard, line 1: typo in the file name `paramters`->`parameters`

UnseenWizzard:
I really the changes splitting things into smaller chunks/files and leaving finding files to the project loader. 

I am however pretty unsure about the structure of new sub-packages. Overall I think it's better, but it could be even better - just not sure how. 
If we don't quickly agree on some great structure, I do think this can be merged as is, and future refactorings might point us to an even better package structure.

pkg/writer/writer.go
warber: IMO the package name was perfectly fine. Also, i don't see a problem with importing something that is also called writer into a package already named writer. Further, you can always alias your import to "configwriter" if needed. 

In general package names should not have underscores. (exception _test of course).

pkg/config/loader/config_loader.go
```diff
@@ -123,6 +103,30 @@ func LoadConfig(fs afero.Fs, context *LoaderContext, filePath string) ([]config.
...
+	definition := persistence.TopLevelDefinition{}
+	err = yaml.UnmarshalStrict(data, &definition)
+
+	if err != nil {
+		if strings.Contains(err.Error(), fmt.Sprintf("field config not found in type %s", persistence.GetTopLevelDefinitionYamlTypeName())) {
+			return nil, fmt.Errorf("config '%s' is not valid v2 configuration - you may be loading v1 configs, please 'convert' to v2:\n%w", filePath, err)
+		}
+
+		return nil, err
+	}
```
warber: nit(style): 
```
	var definition persistence.TopLevelDefinition
	if err := yaml.UnmarshalStrict(data, &definition); err != nil {
		if strings.Contains(err.Error(), fmt.Sprintf("field config not found in type %s", persistence.GetTopLevelDefinitionYamlTypeName())) {
			return nil, fmt.Errorf("config '%s' is not valid v2 configuration - you may be loading v1 configs, please 'convert' to v2:\n%w", filePath, err)
		}
		return nil, err
	}
```

pkg/config/loader/config_loader.go
```diff
@@ -127,32 +127,30 @@ func parseFile(fs afero.Fs, filePath string) ([]persistence.TopLevelConfigDefini
 	return definition.Configs, nil
 }
 
-// parseDefinition parses a single config entry
-func parseDefinition(
+// parseConfigEntry parses a single config entry
```
warber: nit: unnecessary comment

warber (approved):
only small nits

PR #1144, opened 2023-08-24
Fix/dont log platform warning for non platform delete

UnseenWizzard:
Small fixes to delete error and warning messages to help users (and support) troubleshoot issues.


pkg/delete/delete.go
```diff
@@ -71,7 +70,7 @@ func Configs(ctx context.Context, clients ClientSet, apis api.APIs, automationRe
 			deleteErrs := deleteClassicConfig(ctx, clients.Classic, targetApi, entries, entryType)
 			errs = append(errs, deleteErrs...)
 		} else if targetAutomation, isAutomation := automationResources[entryType]; isAutomation {
-			if clients.Automation == nil {
+			if reflect.ValueOf(clients.Automation).IsNil() {
```
UnseenWizzard, line 73: I personally dislike this, but there is no clean way to check the underlying object behind the interface for nil. The interface is never nil, it just happens to contain a nil pointer object.

The cleanest option would be to make the Automation client's zero/nil value useful and make it just do nothing. That would need us to change the client's method to pointer receivers, make them check for nil and then possible do nothing. 

That might be a good pattern to follow for the -core clients. But I think it's too much for this change, esp. as we want to move the automation client to -core soon. 
(Why I decided against the option to just produce an error earlier if there are 3rd gen types but no oauth, see the commit message. TL;DR: it seems more userfriendly.)

cmd/monaco/delete/delete.go
```diff
@@ -88,19 +88,29 @@ func deleteConfigForEnvironment(env manifest.EnvironmentDefinition, apis api.API
...
+		string(config.Workflow):         config.Workflow,
+		string(config.BusinessCalendar): config.BusinessCalendar,
+		string(config.SchedulingRule):   config.SchedulingRule,
+	}
+
+	platformTypes := maps.Keys(automationResources)
+	platformTypes = append(platformTypes, "bucket")
+
+	if env.Auth.OAuth == nil && containsPlatformTypes(entriesToDelete, platformTypes) {
+		log.WithCtxFields(ctx).Warn("Delete file contains Platform specific types, but no oAuth client is defined for environment - Dynatrace Platform configurations will not be deleted.")
```
Laubi, line 101:
```suggestion
		log.WithCtxFields(ctx).Warn("Delete file contains Dynatrace Platform specific types, but no oAuth client credentials are defined for environment %q. Dynatrace Platform configurations won't be deleted.", env.?Name)
```

pkg/delete/delete.go
```diff
@@ -151,7 +150,7 @@ func deleteSettingsObject(ctx context.Context, c dtclient.Client, entries []Dele
 			log.WithCtxFields(ctx).Debug("Deleting settings object %s/%s with objectId %s.", e.Type, e.Identifier, obj.ObjectId)
```
Laubi, line 150: Project is missing here (it's optional, but can be present). I'd suggest including it using the coordinate.

Additionally, in the error below, we should also include it.

pkg/delete/delete.go
```diff
@@ -107,7 +106,7 @@ func deleteClassicConfig(ctx context.Context, client dtclient.Client, theApi api
 	for _, v := range values {
 		log.WithCtxFields(ctx).WithFields(field.Type(theApi.ID), field.F("value", v)).Debug("Deleting %v (%v).", v, targetApi)
 		if err := client.DeleteConfigById(theApi, v.Id); err != nil {
-			errors = append(errors, err)
+			errors = append(errors, fmt.Errorf("could not delete %s configuration %q (%s): %w", theApi.ID, v.Name, v.Id, err))
```
Laubi, line 109: Very confusing message imo
```suggestion
			errors = append(errors, fmt.Errorf(`could not delete configuration "%s/%s" with ID %q: %w`, theApi.ID, v.Name, v.Id, err))
```

cmd/monaco/delete/delete.go
```diff
@@ -88,19 +88,29 @@ func deleteConfigForEnvironment(env manifest.EnvironmentDefinition, apis api.API
...
 	}
 
 	clientSet, err := dynatrace.CreateClientSet(env.URL.Value, env.Auth)
 	if err != nil {
 		return []error{
-			fmt.Errorf("It was not possible to create a client for env `%s` due to the following error: %w", env.Name, err),
+			fmt.Errorf("failed to create API client for enviroment %q due to the following error: %w", env.Name, err),
```
Laubi, line 107:
```suggestion
			fmt.Errorf("failed to create API client for environment %q due to the following error: %w", env.Name, err),
```

pkg/delete/delete.go
UnseenWizzard: Addressed but not exactly as you suggested. Logging the ID in brackets is consistent with many other logs in deployment and clients.

pkg/delete/delete.go
UnseenWizzard: Addressed by introducing a String() method on DeletePointer that will cleanly log in colon separated format with and without project being present

cmd/monaco/delete/delete.go
```diff
@@ -88,19 +88,29 @@ func deleteConfigForEnvironment(env manifest.EnvironmentDefinition, apis api.API
...
 		return []error{
-			fmt.Errorf("It was not possible to create a client for env `%s` due to the following error: %w", env.Name, err),
+			fmt.Errorf("failed to create API client for environment %q due to the following error: %w", env.Name, err),
 		}
 	}
 
-	log.WithCtxFields(ctx).Info("Deleting configs for environment `%s`...", env.Name)
+	log.WithCtxFields(ctx).Info("Deleting configs for environment %q...", env.Name)
 
+	//autClient := clientSet.Automation()
```
warber, line 113: remove

warber (approved):

warber (approved):

Issue #1129, opened 2023-08-18
Make --dry-run an actual dry-run


STgversluis:
**Describe the bug**
The term dry-run suggests that apart from actually applying the config, everything is tested. However, this is not the case.
After a lot of time spent troubleshooting together with Dynatrace support (case #172865), we came to the conclusion that the dry-run only runs a syntax check. It does not connect to the actual tenant, it does not check for conflicting settings, it does not verify the token… which means it leaves plenty of opportunity for disaster when someone makes a wrong change.

**How to reproduce**
Perform a dry-run known to result in an error, such as:
* set an incorrect token in the environment variable
* use a config that will overwrite or modify an existing setting

**Expected behavior**
* If an incorrect token is used, the dry-run should indicate this in a correct error message (in the log snippet below, the token is NOT an old format - I used a valid token and changed only 1 letter)
* Do not state incorrect or confusing info in the log (in the log snippet below, the remark about a Dynatrace Classic environment is confusing - what is a Dynatrace Classic environment?)
* If a configuration is overwritten, this should be logged and warned about in the dry-run

**Log output**

Output from dry-run:
2023-08-18T12:20:20+02:00	info	Validation finished without errors

Output from download, with same env var value for token:
2023-08-18T12:21:39+02:00	info	Loading manifest "manifest.yaml". Restrictions: groups=[], environments=["test"]
2023-08-18T12:21:39+02:00	warn	You used an old token format. Please consider switching to the new 1.205+ token format.
More information: https://www.dynatrace.com/support/help/dynatrace-api/basics/dynatrace-api-authentication
2023-08-18T12:21:39+02:00	error	Could not authorize against the environment with name "test" (https://dynatrace.europe.stater.corp/e/test) using token authorization: failed to query version of Dynatrace environment: (HTTP 401) Response was: {"error":{"code":401,"message":"Token Authentication failed"}}
2023-08-18T12:21:39+02:00	error	Please verify that this environment is a Dynatrace Classic environment.

**Environment (please complete the following information):**
* Dynatrace Managed 
* Monaco v2.6.0

**Additional context**
none.

UnseenWizzard:
Hi @STgversluis. 

You're correct that dry-run only validates configuration file syntax. 
This is documented: https://www.dynatrace.com/support/help/manage/configuration-as-code/reference/commands#deploy-flags

[...]

PR #1147, opened 2023-08-25
chore(http): Improve API token format warning

UnseenWizzard:
The previous log about using an old token format was misleading, as we don't actually check for an old format, but just if the new format is matched.

As the new format exists since October 2020, it is much more likely that a random string was provided as a token, than an actual old API token.

To reflect this, the log warning is changed to point the user towards checking their configuration first, but still mention to possibly update their tokens.

Old warning: 
```
2023-08-18T12:21:39+02:00 warn You used an old token format. Please consider switching to the new 1.205+ token format.
More information: https://www.dynatrace.com/support/help/dynatrace-api/basics/dynatrace-api-authentication
```

New warning: 
```
2023-08-25T10:37:33+02:00       warn    The supplied token does not match the expected format and may be invalid. If authentication fails, please check your manifest and environment variable configuration.
If you are using a token created before Dynatrace 1.205, please consider generating a new token: https://www.dynatrace.com/support/help/dynatrace-api/basics/dynatrace-api-authentication
```

relates to discussion on #1129

### DIFF
--- a/pkg/client/auth/http.go
+++ b/pkg/client/auth/http.go
@@ -39,7 +39,7 @@ type OauthCredentials struct {
 // NewTokenAuthClient creates a new HTTP client that supports token based authorization
 func NewTokenAuthClient(token string) *http.Client {
 	if !isNewDynatraceTokenFormat(token) {
-		log.Warn("You used an old token format. Please consider switching to the new 1.205+ token format.\nMore information: https://www.dynatrace.com/support/help/dynatrace-api/basics/dynatrace-api-authentication")
+		log.Warn("The supplied token does not match the expected format and may be invalid. If authentication fails, please check your manifest and environment variable configuration.\nIf you are using a token created before Dynatrace 1.205, please consider generating a new token: https://www.dynatrace.com/support/help/shortlink/api-authentication")
 	}
 	return &http.Client{Transport: NewTokenAuthTransport(nil, token)}
 }


pkg/client/auth/http.go
```diff
@@ -39,7 +39,7 @@ type OauthCredentials struct {
 // NewTokenAuthClient creates a new HTTP client that supports token based authorization
 func NewTokenAuthClient(token string) *http.Client {
 	if !isNewDynatraceTokenFormat(token) {
-		log.Warn("You used an old token format. Please consider switching to the new 1.205+ token format.\nMore information: https://www.dynatrace.com/support/help/dynatrace-api/basics/dynatrace-api-authentication")
+		log.Warn("The supplied token does not match the expected format and may be invalid. If authentication fails, please check your manifest and environment variable configuration.\nIf you are using a token created before Dynatrace 1.205, please consider generating a new token: https://www.dynatrace.com/support/help/dynatrace-api/basics/dynatrace-api-authentication")
```
Laubi, line 42: The link should probably be a short-link instead of the full one.

Laubi (dismissed):

pkg/client/auth/http.go
UnseenWizzard: Good point, I've updated it to the shortlink and rebased this pr. Please re-approve

Laubi (approved):

PR #1149, opened 2023-08-28
test: Assert that support-archives are created as expected

UnseenWizzard:
Assert that support-archives are actually being created as we expect by adding a small E2E test.

none

### DIFF
--- a/cmd/monaco/integrationtest/v2/supportarchive_e2e_test.go
+++ b/cmd/monaco/integrationtest/v2/supportarchive_e2e_test.go
@@ -1,0 +1,178 @@
+//go:build integration
+
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


cmd/monaco/integrationtest/v2/supportarchive_e2e_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/spf13/afero"
+)
+
+func TestSupportArchiveIsCreatedAsExpected(t *testing.T) {
+	configFolder := "test-resources/support-archive/"
+	manifest := configFolder + "manifest.yaml"
+	fixedTime := timeutils.TimeAnchor().Format(trafficlogs.TrafficLogFilePrefixFormat) // freeze time to ensure log files are created with expected names
+
+	RunIntegrationWithCleanup(t, configFolder, manifest, "valid_env", "SupportArchive", func(fs afero.Fs, _ TestContext) {
+		cleanupLogsDir(t)
```
warber, line 43: move to "RunIntegrationWithCleanup" ?

UnseenWizzard: it's not really an issue that any other test has, we write some unnecessary files into the container running the test, but no other test cares about them being there. 

If it was a general issue (and looked easy) I'd have refactored the traffic logger to accept a filesystem on creation, instead of always using the OS filesystem.

UnseenWizzard: also having this in `RunIntegrationWithCleanup` would not help for the error cases, as that util assumes that manifest can actually be loaded and it can connect to the environment. So it'd need to be called explicitly for these tests still.

warber (approved):

PR #1151, opened 2023-08-31
feat: use lib bucket client in CLI

warber:
Use Bucket client from core lib


internal/loggers/zap/console_test.go
```diff
@@ -38,7 +38,7 @@ func TestClone(t *testing.T) {
 	clone := encoder.Clone()
 	assert.IsType(t, encoder, clone, "Clone should return a fixedFieldsConsoleEncoder")
 	assert.Len(t, encoder.moe.Fields, 1)
-	assert.Len(t, clone.(fixedFieldsConsoleEncoder).moe.Fields, 0)
+	assert.Len(t, clone.(fixedFieldsConsoleEncoder).moe.Fields, 1)
```
UnseenWizzard, line 41: I like that we actually asserted that `Clone` does not actually clone the encoder with it's fields 😅

pkg/client/clientset.go
```diff
@@ -152,8 +153,15 @@ func CreatePlatformClientSet(url string, auth PlatformAuth, opts ClientOptions)
...
+	bucketClient, err := clients.Factory().
+		WithOAuthCredentials(clientcredentials.Config{
+			ClientID:     auth.OauthClientID,
+			ClientSecret: auth.OauthClientSecret,
+			TokenURL:     auth.OauthTokenURL,
+		}).
+		WithEnvironmentURL(url).
+		WithUserAgent(opts.getUserAgentString()).
+		BucketClient()
 	if err != nil {
```
UnseenWizzard, line 165: good catch by the linter above. 
The dtclient creation error from line 136 is never checked anymore.
Before it was checked by this err != nil, which now gets the error from the clients.Factory build.

UnseenWizzard (approved):

PR #1159, opened 2023-09-05
feat: implement traffic logging based on core lib clients (buckets)

warber:
Uses `HTTPListener` from configuraiton-as-code-core in order to do traffic logging for buckets API

### DIFF
--- a/internal/trafficlogs/logger.go
+++ b/internal/trafficlogs/logger.go
@@ -18,11 +18,13 @@ package trafficlogs
 
 import (
 	"fmt"
+	lib "github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/timeutils"
 	"github.com/google/uuid"
 	"github.com/spf13/afero"
+	"io"
 	"net/http"
 	"net/http/httputil"
 	"os"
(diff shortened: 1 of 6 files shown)


internal/trafficlogs/logger.go
```diff
@@ -51,24 +53,32 @@ func NewFileBased() *FileBasedLogger {
...
+	}
+	if resp, ok := record.IsResponse(); ok {
+		if err := l.logResponse(record.ID, resp, resp.Body); err != nil {
+			l.logError(record.ID, "response", err)
+		}
+	}
+}
+
+// Log takes request and response data and tries to write them into files created by this logger.
+// Note: this method is used by "old" clients that are not based on the rest client of the core lib.
```
UnseenWizzard, line 72: nit: technically used by just the old rest client, not several. You could even link to the method that calls it

internal/trafficlogs/logger.go
```diff
@@ -90,35 +106,70 @@ func (l *FileBasedLogger) logRequest(id string, request *http.Request, body stri
...
-	_, err = l.requestLogFile.WriteString(fmt.Sprintf("Request-ID: %s\n%s\n%s\n=========================\n", id, string(dump), body))
+
+	// write id
+	_, err = l.requestLogFile.WriteString(fmt.Sprintf("Request-ID: %s\n", id))
 	if err != nil {
 		return err
 	}
 
+	// write dump
+	if _, err := l.requestLogFile.WriteString(fmt.Sprintf("%s", string(dump))); err != nil {
```
UnseenWizzard, line 117: you've lost the newline after the dump I think

warber: for some reason there are alreade two newlines in the dump data itself :D

pkg/rest/client.go
```diff
@@ -186,7 +187,7 @@ func (c Client) executeRequest(request *http.Request) (Response, error) {
 		}
 
 		if c.trafficLogger != nil {
-			err := c.trafficLogger.Log(request, reqBody, resp, string(respBody))
+			err := c.trafficLogger.Log(request, io.NopCloser(strings.NewReader(reqBody)), resp, io.NopCloser(bytes.NewReader(respBody)))
```
UnseenWizzard, line 190: doesn't matter much, but couldn't you leave the interface of Log unchanged and just do this wrapping into a ReadCloser inside of that before calling the internal log methods?

warber: sure, I've thought about it but as you've already stated, it doesn't really matter :D 
But I will do it, yeah.

internal/trafficlogs/logger.go
```diff
@@ -51,24 +53,32 @@ func NewFileBased() *FileBasedLogger {
...
+	if req, ok := record.IsRequest(); ok {
+		if err := l.logRequest(record.ID, req, req.Body); err != nil {
+			l.logError(record.ID, "request", err)
+		}
+	}
+	if resp, ok := record.IsResponse(); ok {
+		if err := l.logResponse(record.ID, resp, resp.Body); err != nil {
+			l.logError(record.ID, "response", err)
+		}
+	}
```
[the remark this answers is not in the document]
warber: not sure if I like the idea tbh. Maybe these "report-like" files are not the correct place for errors...

UnseenWizzard (approved):

PR #1161, opened 2023-09-05
feat(delete): Delete configurations in parallel

UnseenWizzard:
⚠️ In a bit of local testing this doesn't seem to have much impact on deletion.

Similar to deploy a go routine is started for each config we delete.


warber:
@UnseenWizzard  can we close this?

PR #1163, opened 2023-09-05
fix(generate): Omit configs with empty names when generating deletefiles

UnseenWizzard:
Previously it was possible to generate invalid delete files if config had an empty string set as name but was a Config API type. As persisting delete entries uses the same object for all types it will omit empty names and IDs, and thus produce an invalid entry consisting only of the 'type' if a config had an empty 'name' defined.

### DIFF
--- a/cmd/monaco/generate/deletefile/deletefile.go
+++ b/cmd/monaco/generate/deletefile/deletefile.go
@@ -25,6 +25,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter/reference"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/persistence"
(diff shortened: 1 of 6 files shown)


warber (dismissed):

cmd/monaco/generate/deletefile/deletefile.go
```diff
@@ -144,9 +145,16 @@ func generateDeleteFileContent(environment string, projects []project.Project, a
 		for _, cfgs := range cfgsPerType {
 			for _, c := range cfgs {
 				if apis.Contains(c.Coordinate.Type) {
-					val, err := c.Parameters[config.NameParameter].ResolveValue(parameter.ResolveContext{ParameterName: config.NameParameter})
+					nameParam := c.Parameters[config.NameParameter]
+
+					if nameParam.GetType() == reference.ReferenceParameterType {
```
warber, line 150: pls add a short comment why we're leaving out that case?

UnseenWizzard: done.
On second thought I think we could actually improve the generator to resolve references that are not the IDs of other configs.
We'd just need to sort and build an entity map for each config. If I find the time I'll address this in a followup PR

warber (approved):

cmd/monaco/generate/deletefile/deletefile.go
warber: cool, or create a ticket otherwise :)

PR #1172, opened 2023-09-07
Feat/default mem limit

UnseenWizzard:
[feat: Introduce default soft memory limit](https://github.com/Dynatrace/dynatrace-configuration-as-code/commit/a34fca21e0d5db5d9c5b115a1f5bfc64821a4703)

It was found that with default settings, garbage collection
runs very seldomly during a deploymen, leading to memory consumption
to slowly build up and usually not go down before execution actuall
finishes.

This is likely due to the default GC setting of allocating up to 100%
of the current heap size within each next GC cycle.
With our memory pattern of allocating a large amount of
memory when loading all configurations at the start of a start of a
deployment, then slowly increasing memory with each deployed configuration
(rendered templates, logs, errors, etc.), this results in the GC mostly not
running after configs have been loaded.

As this in not a very reasonable memory behavior, and we can't expect all
users to set their own memory limits, a default limit of 2 GiB is introduced.

While hardcoding this isn't great, for now its a much simpler solution than
implementing OS specific ways to query the available memory and then still use
a hardcoded percentage.
Users that need finegrained control still have it by setting GOMEMLIMIT.

See message above for why I've decided not to query the system's available memory for now.
Some potential implementations for querying total memory for most operating systems:

- https://github.com/cloudfoundry/gosigar (bigger general system info API)
- https://pkg.go.dev/github.com/pbnjay/memory

I've tested a version using pbnjay/memory in #1173  - but I think it shows why a relative limit doesn't actually make sense.

Deployments and downloads might be slower than before, as GC will trigger more frequently.
It is less likely for containers running monaco to get OOM killed.

### DIFF
--- a/cmd/monaco/main.go
+++ b/cmd/monaco/main.go
@@ -15,12 +15,16 @@
 package main
 
 import (
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/memory"
 	"os"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/runner"
 )
 
 func main() {
+
+	memory.SetDefaultLimit()
+
 	statusCode := runner.Run()
 	os.Exit(statusCode)
 }
--- a/cmd/monaco/main.go
+++ b/cmd/monaco/main.go
@@ -15,16 +15,12 @@
 package main
 
 import (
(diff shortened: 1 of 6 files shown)


warber:
awesome - and a huge relief :D

warber (approved):

PR #1173, opened 2023-09-07
feat: Set default memory limit relative to available OS memory

UnseenWizzard:
This is a draft on how setting relative memory limits could work using github.com/pbnjay/memory. 

Note that running this on a machine with 32 GiB RAM shows that this simple relative limit does not make too much sense, as it would allow to use 24 GiB - which is probably not a reasonable assumption in a multi tenancy environment. 
In something like a container with 1-2GiB memory, setting a soft limit 25% below the hard limit of the container would make sense though. 

I think this level of intelligent setting of memory limits is better left to endusers who can configure what they need using GOMEMLIMIT.


UnseenWizzard:
Won't do - as described above, a lower default limit and docs on how to tune it to your usecase is preferrable

Issue #120, opened 2021-01-18
Update readme to indicate support for kubernetes credentials


robertjahn:
**Is your feature request related to a problem? Please describe.**
I saw that issue #3 was closed and I tested and it works, but the readme files table does not indicate this.  

**Describe the solution you'd like**
Update README section - https://github.com/dynatrace-oss/dynatrace-monitoring-as-code#supported-configuration-types-and-token-permissions

**Describe alternatives you've considered**
None

**Additional context**
None

kristofre:
azure credentials is also missing from the documentation. 
@UnseenWizzard can you please assign me to this one?

163a:
I added azure, kubernetes credentials to table while I was changing table for #187

with PR:
#236

If happy, this issue may be closed too.

didiladi:
Thanks for tackling this @163a. :rocket:

PR #1175, opened 2023-09-11
feat: delete buckets

warber:
Enable `monaco delete ...` to also recognize and delete bucket API type of configurations.

PR also contains a bit of code refactoring most notably cleaning up the delete function from
 ```go
func Delete(afero.Fs, string, string, []string, []string) error
``` 

to just

```go
func Delete(manifest.Environments, delete.DeleteEntries) error 
```

Loading the manifest and delete files happens outside of this function.

Obviously

### DIFF
--- a/cmd/monaco/delete/command.go
+++ b/cmd/monaco/delete/command.go
@@ -78,6 +78,7 @@ func GetDeleteCommand(fs afero.Fs) (deleteCmd *cobra.Command) {
 			if errs != nil {
 				return fmt.Errorf("encountered errors while parsing delete.yaml: %s", errs)
 			}
+
 			return Delete(manifest.Environments, entriesToDelete)
 		},
 		ValidArgsFunction: completion.DeleteCompletion,
--- a/cmd/monaco/delete/command.go
+++ b/cmd/monaco/delete/command.go
@@ -17,17 +17,21 @@
 package delete
 
 import (
+	"errors"
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/cmdutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/completion"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/errutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/files"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
+	"path/filepath"
 )
 
 func GetDeleteCommand(fs afero.Fs) (deleteCmd *cobra.Command) {
-
 	var environments, groups []string
 	var manifestName string
 	var deleteFile string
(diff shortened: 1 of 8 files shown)


pkg/delete/delete.go
```diff
@@ -104,7 +117,7 @@ func deleteClassicConfig(ctx context.Context, client dtclient.Client, theApi api
 	values, errs := filterValuesToDelete(ctx, entries, values, theApi.ID)
 	errors = append(errors, errs...)
 
-	log.WithCtxFields(ctx).WithFields(field.Type(theApi.ID)).Info("Deleting configs of type %s...", theApi.ID)
+	log.WithCtxFields(ctx).WithFields(field.Type(theApi.ID)).Info("Deleting %d config(s) of type %s...", len(entries), theApi.ID)
```
jskelin, line 120: I would go with `%q` so that the outprint of a type is more visible.

jskelin: log.WithCtxFields(ctx).WithFields(field.Type(theApi.ID)).Info("Deleting %d config(s) of type %q...", len(entries), theApi.ID)
```suggestion
	log.WithCtxFields(ctx).WithFields(field.Type(theApi.ID)).Info("Deleting %d config(s) of type %s...", len(entries), theApi.ID)
```

pkg/delete/delete.go
```diff
@@ -123,6 +136,10 @@ func deleteClassicConfig(ctx context.Context, client dtclient.Client, theApi api
 func deleteSettingsObject(ctx context.Context, c dtclient.Client, entries []DeletePointer) []error {
 	errors := make([]error, 0)
 
+	if len(entries) > 0 {
+		log.WithCtxFields(ctx).WithFields(field.Type(entries[0].Type)).Info("Deleting %d config(s) of type %s...", len(entries), entries[0].Type)
```
jskelin, line 140: same suggestion as for the line #120

jskelin (approved):

jskelin (dismissed):

jskelin (approved):

pkg/delete/delete.go
```diff
@@ -186,6 +204,22 @@ func deleteAutomations(c automationClient, automationResource config.AutomationR
 	return errors
 }
 
+func deleteBuckets(ctx context.Context, c bucketClient, entries []DeletePointer) []error {
```
UnseenWizzard, line 207: please add an AllBuckets variant and include it in the purge cmd as well - like the other AllX variants, list all objects and delete them.

cmd/monaco/delete/command.go
```diff
@@ -50,7 +54,31 @@ func GetDeleteCommand(fs afero.Fs) (deleteCmd *cobra.Command) {
 				return err
 			}
 
-			return Delete(fs, manifestName, deleteFile, environments, groups)
+			// Sanitize manifest file path to manifest yaml file
```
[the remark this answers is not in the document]
warber: i see a big distinction between
```
func Delete(afero.Fs, string, string, []string, []string) error
```
and
```
func Delete(manifest.Environments, delete.DeleteEntries) error 
``` 
tho

pkg/delete/delete.go
warber: will do

cmd/monaco/delete/command.go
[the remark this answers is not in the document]
UnseenWizzard: fair enough - but it still is a method that probably shouldn't be exported and is used in one place. Introducing types would probably achieve a similar nicer signature. 

Anyway it's a nitpick, I don't think there's a way or need to have these functions have perfectly split responsibilities or signatures :)

pkg/delete/delete.go
```diff
@@ -369,3 +372,55 @@ func AllAutomations(ctx context.Context, c automationClient) []error {
...
+	}
+
+	log.WithCtxFields(ctx).WithFields(field.Type("bucket")).Info("Deleting %d objects of type %q...", len(response.Objects), "bucket")
+	for _, obj := range response.Objects {
+		var bucketName struct {
+			BucketName string `json:"bucketName"`
+		}
+
+		// exclude builtin bucket names, they cannot be deleted anyway
+		if strings.HasPrefix(bucketName.BucketName, "dt_") || strings.HasPrefix(bucketName.BucketName, "default_") {
```
UnseenWizzard, line 427: that needs to happen after unmarshalling, doesn't it? 😅 `bucketName.BucketName` will still be zero initialized here

UnseenWizzard (approved):

PR #1178, opened 2023-09-12
fix: Settings matching based on unique properties

UnseenWizzard:
This addressed two issues in the settings_client's code finding existing objects to update based on unique properties.

The issues where both found when investigating memory consumption, which discovered that for some schemas we currently find hundreds of matches in a large test configurations. 
In this perfectly valid test project, many Settings define the same unique properties but apply to different scopes - a fact that the code didn't consider so far.

1. A refactoring aims to reduce heap memory consumption by removing the use of pointers and not including matching object IDs in error text if there are more than 5 matches. This ensures that matches don't escape to the heap, and that the currently stored errors don't get too big.
2. The logic is modified to only consider Settings that are within the same scope for finding matches.

A third commit is smuggled into this PR as I discovered some unnecessary logging while running E2E tests.

Yes - it fixes a bug which resulted in monaco not creating/updating valid Settings if their schema defined unique property constraints and a Setting with the same property existed in a different scope already.

### DIFF
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -298,6 +298,10 @@ func findObjectWithSameConstraints(schema SchemaConstraints, source SettingsObje
 
 	for _, uniqueKeys := range schema.UniqueProperties {
 		for j, o := range objects {
+			if o.Scope != source.Scope {
+				continue // settings in different Scopes can't be the same
+			}
+
 			matchFound, constraintMatches, err := doObjectsMatchBasedOnUniqueKeys(uniqueKeys, source, o)
 			if err != nil {
 				return match{}, false, err
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -193,17 +193,17 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
 		}
 	}
 
-	if match, err := d.findObjectWithMatchingConstraints(ctx, obj); err != nil {
+	if matchingObject, found, err := d.findObjectWithMatchingConstraints(ctx, obj); err != nil {
 		return DynatraceEntity{}, err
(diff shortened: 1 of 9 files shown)


pkg/client/dtclient/settings_client.go
```diff
@@ -271,61 +271,71 @@ type match struct {
...
 
-func findObjectWithSameConstraints(schema SchemaConstraints, source SettingsObject, objects []DownloadSettingsObject) (*match, error) {
+func findObjectWithSameConstraints(schema SchemaConstraints, source SettingsObject, objects []DownloadSettingsObject) (match, bool, error) {
 	candidates := make(map[int]constraintMatch)
 
 	for _, uniqueKeys := range schema.UniqueProperties {
 		for j, o := range objects {
-			match, constraintMatches, err := doObjectsMatchBasedOnUniqueKeys(uniqueKeys, source, o)
+			if o.Scope != source.Scope {
+				continue // settings in different Scopes can't be the same
```
Laubi, line 302: I am not sure about this check. If two settings have different scopes (for example, HOST-1 and HOST-2), the same uniqueness check could hit them. Or am I missing something obvious?

UnseenWizzard: Um... that is exactly the point of this check? Please read the commit message! (And/or my internal slack message from yesterday.)

Our code currently wrongly assumes that uniqueness constraints apply globally while the Settings framework applies them for a given scope. 
A great example of this - being the one we found it with - are the `builtin:host.process-groups.monitoring-state` settings that are also added as an E2E testcase. The process group in the setting must be unique - but is perfectly valid to define many Settings for the same PG in scope of different hosts.

Laubi: Well, it was also my wrong assumption that the checks are global. Well, learned something :)

Laubi (approved):

PR #1181, opened 2023-09-13
refactor(featureflags): Split featurelags defintions into files for permanent and temporary flags

UnseenWizzard:

As we have created quite a large amounts of feature flags but haven't removed any so far, the simple featureflags file is split into one for permanent flags and one for actually temporary flags.

For each temporary flag a comment is added denoting when it was introduced. This should allow better overview and actually help us make decisions to remove some of them.

I started this refactoring planning to add a new flag for writing error messages to a dedicated file, but then decided there's no reason to safeguard something as simple behind a flag. 

nope


warber (approved):
nice idea

PR #1182, opened 2023-09-13
fix(buckets): Ensure validating buckets works

UnseenWizzard:
Validation of bucket configs previously failed, because the DummyClient implemented for validation didn't set a status code.
Thus it's zero-initialized status code repsonse has treated as an error/non-success, and incorrectly flagged valid bucket configurations as invalid.


UnseenWizzard:
unrelated e2e test failed with a random 400 reply - retriggered

warber (approved):

PR #1184, opened 2023-09-14
chore: add deprecation notes to outdated files

warber:
* Adds deprecation notes and a pointer to updated information to and `api_coverage.md`
* delete outdated `RELEASE_NOTES.md`


api_coverage.md
```diff
@@ -1,3 +1,9 @@
+
+---
+**NOTE**: Information presented in this document is deprecated and **not updated** anymore. Please consult the official documentation to get
```
UnseenWizzard, line 3:
```suggestion
**NOTE**: This a working document, mainly focused on Configuration APIs being covered by Settings 2.0 support or not.
Information presented in this document does not represent the current coverage if Dynatrace APIs and is likely outdated anymore.
```

api_coverage.md
```diff
@@ -1,3 +1,9 @@
...
+information about [supported configuration types](https://www.dynatrace.com/support/help/manage/configuration-as-code/monaco/reference/supported-configuration#configuration-types-and-access-permissions).
```
UnseenWizzard, line 4:
```suggestion
 **Please consult the official documentation for information about [supported configuration types](https://www.dynatrace.com/support/help/manage/configuration-as-code/monaco/reference/supported-configuration#configuration-types-and-access-permissions)**.
```

UnseenWizzard: IMO the originial wordings is not quite correct, this isn't so much a deprecated document of API coverage (we always had the table in the docs, even for v1), but a working document of which Config APIs aren't covered directly or by Settings. 

While it's not been updated for a while, the section which config APIs are simply not covered at all is correct, and will likely become relevant again, if they are not covered by Settings 2.0 schemas soon.

warber: It's deprecated because there is another/a newer document to look at.. but ok

UnseenWizzard: No, that's what I mean - there is no newer document that provides the information of this one.
The documentation page (which is older) lists all types that are possible to configure.
This document is wrongly named, but it actually lists what **isn't supported** explicitly. 

If we wanted to provide a clean version of that information the https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/main/api_coverage.md#summarized-list-of-currently-unsupported-config-apis section would be the missing piece in the docs.

warber: oh, gotcha

UnseenWizzard (dismissed):

UnseenWizzard (approved):

PR #1183, opened 2023-09-13
feat: Error log file

UnseenWizzard:
#### What this PR does / Why we need it:
A new additional log file is written containing all errors. This gives users a simple way to access all errors without greping logs or an error summary.

#### Special notes for your reviewer:
Small refactorings to logging setup are included in the first few commits.

#### Does this PR introduce a user-facing change?
New .errors log file will be available to them.

PR #1185, opened 2023-09-14
feat/(fix): Do not collect large amounts of errors for final error report

UnseenWizzard:
Currently when using --continue-with-error, or by default for parallel deployment errors, monaco will collect all errors that occurred and will log them in a sorted report at the end of deployment.

In cases where many errors fail, that means we carry a memory overhead for presenting errors that have already been logged just to present the user with a sorted representation of issues.

These errors may carry additional information like full JSON templates that failed to render or HTTP request and response bodies.

This PR removes the collection of errors and ensures that no information is lost, by adding a few explicit logs.
It also attempts to present the information that is logged in a slightly clearer way.

- review per commit is recommended
- relates to https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1183 which produces a replacement "report" file excerpt of error messages.

Yes - monaco will no longer log an error summary after deployment errors. Instead if will log minimal information about the amounts of errors and require users to check the logs for error messages themselves if they want details.

### DIFF
--- a/cmd/monaco/deploy/deploy.go
+++ b/cmd/monaco/deploy/deploy.go
@@ -17,14 +17,17 @@ package deploy
 import (
 	"errors"
 	"fmt"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/deploy/internal/clientset"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/deploy/internal/logging"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/deploy/sequential"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/dynatrace"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/errutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/slices"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy"
 	deployErrors "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/errors"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/sequential"
 	"path/filepath"
 	"strings"
 
@@ -32,7 +35,6 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest"
 	project "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project/v2"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project/v2/sort"
 	"github.com/spf13/afero"
 )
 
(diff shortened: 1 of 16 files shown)


pkg/deploy/errors/error.go
```diff
@@ -87,3 +87,15 @@ func (e EnvironmentDeploymentErrors) Append(env string, err ...error) Environmen
...
+// DeploymentErrors is an error returned if any deployment errors occured. It carries a count of how many errors happened
+// during deployment, but no details on those errors. The specific errors that have happened during deployment are handled
+// by logging them, and never returned out of DeployConfigGraph.
+type DeploymentErrors struct {
+	// ErrorCount tells how many errors occurred during a deployment
+	ErrorCount int
+}
+
+func (d DeploymentErrors) Error() string {
+	return fmt.Sprintf("Deployment failed: %d errors occurred", d.ErrorCount)
```
UnseenWizzard: note: this is rephrased in a later commit

cmd/monaco/deploy/deploy.go
```diff
@@ -73,25 +73,17 @@ func deployConfigs(fs afero.Fs, manifestPath string, environmentGroups []string,
...
-			var environmentDeployErrs deployErrors.EnvironmentDeploymentErrors
-			if errors.As(deployErr, &environmentDeployErrs) {
-				for _, errs := range environmentDeployErrs {
-					// TODO error handling can change to remove the repetitive grouping to env errors - for now we just build a list to be grouped again by printErrorReport
-					deployErrs = append(deployErrs, errs...)
-				}
-			} else {
-				deployErrs = append(deployErrs, deployErr)
+		if err != nil {
+			var deployErrs deployErrors.DeploymentErrors
```
UnseenWizzard: note: this never works as DeployConfigGraph still always returns `EnvironmentDeploymentErrors`. This is addressed in a later commit, but improving `EnvironmentDeploymentErrors.Error()` and just returning it

cmd/monaco/deploy/internal/clientset/clientset.go
```diff
@@ -0,0 +1,58 @@
...
+		clients[deploy.EnvironmentInfo{
+			Name:  env.Name,
+			Group: env.Group,
+		}] = clientSet
+	}
+
+	return clients, nil
+}
+
+func NewClientSet(env manifest.EnvironmentDefinition, dryRun bool) (deploy.ClientSet, error) {
```
UnseenWizzard, line 42: self-nit: could be just `New`

cmd/monaco/deploy/internal/logging/logging.go
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package deploy
+package logging
```
UnseenWizzard, line 17: note: didn't want to call this `log` to not overlap with the actual `log` package. Would have made for nicer function names though...

pkg/deploy/errors/error.go
```diff
@@ -74,8 +74,16 @@ type EnvironmentDeploymentErrors map[string][]error
 
 func (e EnvironmentDeploymentErrors) Error() string {
 	b := strings.Builder{}
+	b.WriteString(fmt.Sprintf("Errors encountered for %d environment(s):", len(e)))
 	for env, errs := range e {
-		b.WriteString(fmt.Sprintf("%s deployment errors: %v", env, errs))
+		if len(errs) == 1 {
```
Laubi, line 79: nit: I think we can just skip the `len()==1`case. Might make some errors a bit nicer, IMO not worth the condition here

cmd/monaco/deploy/internal/clientset/clientset.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest"
+)
+
+func NewEnvironmentClients(environments manifest.Environments, dryRun bool) (deploy.EnvironmentClients, error) {
+	clients := make(deploy.EnvironmentClients, len(environments))
+	for _, env := range environments {
+		clientSet, err := NewClientSet(env, dryRun)
+		if err != nil {
+			return deploy.EnvironmentClients{}, err
```
Laubi, line 30: Why not return `nil, err` here?

Laubi (approved):

pkg/deploy/errors/error.go
UnseenWizzard: I don't think the condition makes this method overly complex does it?
As improving how errors are presented was part of what we set out to do here, I think presenting the error in a nicer is well worth it here. 

Note that this is generally the final error that is returned all the way to cobra and printed after execution.
I did only add the if in later, as it looked weird.

cmd/monaco/deploy/internal/clientset/clientset.go
UnseenWizzard: As I've just copied these functions out of the general deploy file, I didn't consider improving anything tbh. 
And even if, I'm not sure if a nil map over a zero-initialized one in this error case has any impact?

PR #1191, opened 2023-09-19
feat: Introduce Automation API client from -core library

UnseenWizzard:
Replace the monaco client implementation with the new client from the library.

There's also small improvements to E2E test utils and delete error handling that I noticed while getting the new client to work with monaco.
Review per commit recommended as always.

Tangential change: Bucket deletion will no longer report 404 responses as an error (and thus behave like other deletions that leniently accept a 404 as the desired state is reached anyway)


pkg/client/clientset.go
```diff
@@ -179,6 +171,11 @@ func CreatePlatformClientSet(url string, auth PlatformAuth, opts ClientOptions)
 		return nil, err
 	}
 
+	autClient, err := clientFactory.AutomationClient()
```
UnseenWizzard, line 174: setup is reused from the clientFactory built for the bucket client, which is nice and simple 👌

pkg/delete/delete.go
```diff
@@ -385,10 +395,13 @@ func AllAutomations(ctx context.Context, c automationClient) []error {
 		log.WithCtxFields(ctx).WithFields(field.Type(string(resource))).Info("Deleting %d objects of type %q...", len(objects), resource)
 		for _, o := range objects {
 			log.WithCtxFields(ctx).WithFields(field.Type(string(resource)), field.F("object", o)).Debug("Deleting Automation object with id %q...", o.ID)
-			err = c.Delete(t, o.ID)
+			resp, err := c.Delete(ctx, t, o.ID)
 			if err != nil {
 				errs = append(errs, err)
 			}
+			if err, isErr := resp.AsAPIError(); isErr {
```
UnseenWizzard, line 402: note to self: include 404 being fine here as well (line 206)

pkg/delete/delete.go
```diff
@@ -385,10 +395,13 @@ func AllAutomations(ctx context.Context, c automationClient) []error {
 		log.WithCtxFields(ctx).WithFields(field.Type(string(resource))).Info("Deleting %d objects of type %q...", len(objects), resource)
 		for _, o := range objects {
 			log.WithCtxFields(ctx).WithFields(field.Type(string(resource)), field.F("object", o)).Debug("Deleting Automation object with id %q...", o.ID)
-			err = c.Delete(t, o.ID)
+			resp, err := c.Delete(ctx, t, o.ID)
 			if err != nil {
 				errs = append(errs, err)
 			}
+			if err, isErr := resp.AsAPIError(); isErr && resp.StatusCode != http.StatusNotFound { // 404 means it's gone already anyway
```
warber, line 402: the error is added twice then, no?

warber: ah no, fortget it, thy are different of course

internal/automationutils/response_decoder.go
warber, line 1: maybe add tests for this?

warber (dismissed):

warber (approved):

PR #1192, opened 2023-09-20
fix: Add child-config as structured logging field

Laubi:
When printing the logs what configs are skipped, the child is not printed using structured logging. This is now fixed.

### DIFF
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -284,13 +284,17 @@ func (c *componentDeployer) removeChildren(ctx context.Context, parent, root gra
 		}
 		childCfg := child.Config
 
+		l := log.WithCtxFields(ctx).WithFields(
+			field.F("parent", parent.Config.Coordinate),
+			field.F("deploymentFailed", failed),
+			field.F("child", child.Config.Coordinate))
+
 		// after the first iteration
 		if parent != root {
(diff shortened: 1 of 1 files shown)


pkg/deploy/deploy.go
```diff
@@ -284,13 +284,17 @@ func (c *componentDeployer) removeChildren(ctx context.Context, parent, root gra
...
 
+		l := log.WithCtxFields(ctx).WithFields(
+			field.F("parent", parent.Config.Coordinate),
+			field.F("deploymentFailed", failed),
+			field.F("child", child.Config.Coordinate))
+
 		// after the first iteration
 		if parent != root {
-			log.WithCtxFields(ctx).WithFields(field.F("parent", parent.Config.Coordinate), field.F("deploymentFailed", failed), field.F("root", root.Config.Coordinate)).
+			l.WithFields(field.F("root", root.Config.Coordinate)).
```
UnseenWizzard, line 294: Isn't the `root` anyway always the coordinate the log is for?
At least per your internal message I understood that as the reason to keep the coordinate fixed to that of the config that's the source of things being skipped. 
So IMO it's not really necessary to add it another time

UnseenWizzard (approved):
Would personally drop the dedicated `root` coordinate field, but generally ok to merge.

PR #1196, opened 2023-09-22
fix: override retry settings in case a bucket is referenced in a settings obj to be deployed

warber:
We observed, that even if we wait for a new bucket definition to be in  "active" state, it turns out that settings objects referencing that bucket could still fail to be uploaded because the bucket could not be found.

In this PR a special handling for settings objects referencing buckets is added. If it is triggered, the retry settings for that upsert operation ar drastically increased.

### DIFF
--- a/pkg/rest/retry.go
+++ b/pkg/rest/retry.go
@@ -65,7 +65,7 @@ func SendWithRetry(ctx context.Context, sendWithBody SendRequestWithBody, object
 	if err != nil {
 		return Response{}, fmt.Errorf("HTTP send request %s failed after %d retries: %w", path, setting.MaxRetries, err)
 	}
-	return Response{}, NewRespErr(fmt.Sprintf("HTTP send request %s failed after %d retries: (HTTP %d)!\n    Response was: %s", path, setting.MaxRetries, resp.StatusCode, string(resp.Body)), resp)
+	return Response{}, NewRespErr(fmt.Sprintf("HTTP send request %s failed after %d retries: (HTTP %d)", path, setting.MaxRetries, resp.StatusCode), resp)
 }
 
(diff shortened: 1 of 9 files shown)


pkg/rest/retry.go
```diff
@@ -65,7 +65,7 @@ func SendWithRetry(ctx context.Context, sendWithBody SendRequestWithBody, object
 	if err != nil {
 		return Response{}, fmt.Errorf("HTTP send request %s failed after %d retries: %w", path, setting.MaxRetries, err)
 	}
-	return Response{}, NewRespErr(fmt.Sprintf("HTTP send request %s failed after %d retries: (HTTP %d)!\n    Response was: %s", path, setting.MaxRetries, resp.StatusCode, string(resp.Body)), resp)
+	return Response{}, NewRespErr(fmt.Sprintf("HTTP send request %s failed after %d retries: (HTTP %d)", path, setting.MaxRetries, resp.StatusCode), resp)
```
warber, line 68: this resulted in error messages printed twice in a row

pkg/rest/retry.go
```diff
@@ -65,7 +65,7 @@ func SendWithRetry(ctx context.Context, sendWithBody SendRequestWithBody, object
 	if err != nil {
 		return Response{}, fmt.Errorf("HTTP send request %s failed after %d retries: %w", path, setting.MaxRetries, err)
 	}
-	return Response{}, NewRespErr(fmt.Sprintf("HTTP send request %s failed after %d retries: (HTTP %d)!\n    Response was: %s", path, setting.MaxRetries, resp.StatusCode, string(resp.Body)), resp)
```
warber, line 68: this resulted in error messages printed twice in a row

Laubi (approved):

PR #1197, opened 2023-09-22
refactor(template): Introduce new File and InMemory Templates

UnseenWizzard:
The updated FileBasedTemplate loads file contents only when required, instead of at creation, reducing the long term in-use memory footprint
 of configurations.

The new InMemoryTemplate replaces the Download specific template, as well as some dedicated test implementations of templates. It currently carries an optional filepath, which is used for writing converted templates to the desired target file. This may be split off.

- exact memory impact during deployments still to be tested
- the general InMemoryTemplate now used for Download, Convert as well as testing, may not be the best way to go. 
  It might instead make sense to have dedicated template implementations in 

No visible change, template files are loaded from disk later.


pkg/config/config.go
```diff
@@ -149,8 +149,8 @@ func (c *Config) Render(properties map[string]interface{}) (string, error) {
 		return "", nil
 	}
 
-	templatePath := c.Template.Name()
-	if t, ok := c.Template.(template.FileBasedTemplate); ok {
+	templatePath := c.Template.Id()
+	if t, ok := c.Template.(*template.FileBasedTemplate); ok {
 		templatePath = t.FilePath()
 	}
```
warber, line 155: i guess you can delete these lines

UnseenWizzard: Upon closer inspection, no this is still needed. 

I didn't make Filepath() part of the interface and while both implementations currently have a Filepath() they behave differently. 
FileBasedTemplate always has a path and has a `Filepath() string` signature.
InMemoryTemplate _might_ have a defined path and has a `Filepath() *string` signature.

Usage is a bit odd at the moment... outside of tests, Render() will always get a FileBasedTemplate right now. 
The config writer might get InMemoryTemplates from download (no path), or convert (fixed path), and I think(!) will currently never get a FileBasedTemplate, but handles all in: pkg/persistence/config/writer/config_writer.go:611

I'm not sure of a better way to handle this at the moment. Pulling the optional Filepath *string method into the interface feels odd as well. 
If you have a good refactoring idea, let me know! Else maybe we leave it as is until someone has an idea to refactor?

pkg/config/template/filebased.go
```diff
@@ -0,0 +1,83 @@
...
+
+// FileBasedTemplate is a JSON Template stored in a file - when it's Template.Content is accessed, that file is read.
+// This is the usual type of Template monaco uses.
+type FileBasedTemplate struct {
+	fs   afero.Fs
+	path string
+}
+
+func (t *FileBasedTemplate) Id() string {
+	return t.path
```
jskelin, line 39: This is the total path of the template, but ID is only the file name. Right?

pkg/config/template/filebased.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (t *FileBasedTemplate) Content() (string, error) {
+	b, err := afero.ReadFile(t.fs, t.path)
```
jskelin, line 43: What do you think about ignoring an error?

Since the existence of the file is already checked in the constuction method, the error can be ignored. If it occurs, just write ERROR message and go to panic mode.

For more safety, performing a test read during creation is possible. And if it is necessary for performance, it can be done in parallel.

pkg/config/config.go
jskelin: With this changes `ID()` method for `FileBasedTemplate` returns path, but for `InMemory` real ID.  This addresses my complaint for different behavior for `ID()`

jskelin (commented):
having `ID()` method that returns differnet value for different implementation for me is problematic

pkg/config/template/filebased.go
UnseenWizzard: I consider it at the start but I would not.
While it generally won't happen, it can still happen that a file gets deleted or becomes inaccessible during deployment. Consider that even with parallel deployments, there might be an hour or two between a template being initially created and then actually read and rendered for an API call.

While unlikely, I don't think we should just ignore the error, so I've added it to the interface and handle it as an error everywhere.

pkg/config/config.go
UnseenWizzard: Technically neither the Id or Name make much sense here. 
The filepath is only ever included as a hint in errors that happend during rendering.
Additionally at least right now a non FileBasedTemplate will never be rendered, as that happens only during deploy. 

I've removed the unecessary and confusing initialization with the ID and just leave the path variable empyt in the unlikely case that a template without a path is being rendered.

pkg/config/template/filebased.go
UnseenWizzard: In both cases it's a cleaned version of the template's path - that is the combination of the folder the config YAML is in, plus the defined template path from the YAML, with any relative parts resolved. 

ID is really only used in logging and errors in a few places

pkg/config/template/template.go
```diff
@@ -0,0 +1,29 @@
...
+
+// Template is the main interface of a configuration payload that may contain template references (using Go Templates)
+// The main implementation is FileBasedTemplate, which loads its Content from a file on disk. An InMemoryTemplate exists
+// as well and is used in cases where Template data is not related to a file - e.g. during download or convert.
+type Template interface {
+	// ID of the template
+	Id() string
+
+	// Content returns the string content of the template, returns error if content is not accessible
+	Content() (string, error)
```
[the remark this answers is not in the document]
UnseenWizzard: Generally makes sense.
However Content is mostly read to be parsed into a Go template and Template.Parse accepts strings and has no functions taking byte slices.

pkg/config/template/inmemory.go
```diff
@@ -0,0 +1,58 @@
...
+package template
+
+var (
+	_ Template = (*InMemoryTemplate)(nil)
+)
+
+type InMemoryTemplate struct {
+	id      string
+	content string
+	path    *string
```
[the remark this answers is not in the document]
UnseenWizzard: added documentation comments, please take a look if they make sense to you

pkg/config/template/inmemory.go
```diff
@@ -0,0 +1,58 @@
...
+func (t *InMemoryTemplate) Content() (string, error) {
+	return t.content, nil
+}
+
+func (t *InMemoryTemplate) UpdateContent(newContent string) error {
+	t.content = newContent
+	return nil
+}
+
+func (t *InMemoryTemplate) FilePath() *string {
```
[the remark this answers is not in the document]
UnseenWizzard: added documentation comments, please take a look if they make sense to you

pkg/config/template/filebased.go
jskelin: It is not the best name, but we can leave it as it is. As this PR is not about improving the `Template` interface

jskelin (approved):

pkg/config/template/filebased.go
jskelin: As you said, it is rarely possible. In my opinion, the assumption that touching configuration is forbidden once the process starts is quite common. So panicking in that case is acceptable.

It is more of a discussion about possibilities. The current solution is acceptable to me.

UnseenWizzard: Let's discuss this in the team some time soon!

IMO panicking does not fit our tendency to handle problems and continue (e.g. if one deployment tree fails, we continue others). Would be worth discussing and defining that as a shared design philosophy and with that ban must uses of panic.
This is unlikely to happen, but it can - clean error handling is IMO a small cost to pay for not breaking a full deployment because some file is corrupt or was removed.

PR #1201, opened 2023-09-25
feat: status log field

warber:
This PR adds a field `deploymentStatus` to structured log lines related to configuration deployment, so it is e.g. easily possible to collect log lines that indicate that the deployment of certain configurations failed or were skipped etc.

It introduces the field `deploymentStatus` which is set whenever a log line related to a configuration coordinate is written during the execution of a deployment. It's set to one of the following values:

* deploying - to indicate that the log message was produced to tell that a deployment of a config is happening 
* deployed - to indicate that the log message was produced to tell that a deployment of a config was successful
* failed - to indicate that the log message was produced to tell that the deployment of a config failed
* skipped - to indicate that the log message was produced tell that the deployment of a config was skipped

A status for e.g. labeling log messages related to template loading was not yet introduced.
Also for downloading configuration, no status field is set.

### DIFF
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -202,7 +202,7 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
 			props = append(props, fmt.Sprintf("(%v = %v)", k, v))
 		}
 
-		log.WithCtxFields(ctx).Info("Updating existing object %q with matching unique properties: %v", matchingObject.object.ObjectId, strings.Join(props, ", "))
+		log.WithCtxFields(ctx).Debug("Updating existing object %q with matching unique properties: %v", matchingObject.object.ObjectId, strings.Join(props, ", "))
 		obj.OriginObjectId = matchingObject.object.ObjectId
 	}
 
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -285,15 +285,15 @@ func (c *componentDeployer) removeChildren(ctx context.Context, parent, root gra
 		l := log.WithCtxFields(ctx).WithFields(
 			field.F("parent", parent.Config.Coordinate),
 			field.F("deploymentFailed", failed),
-			field.F("child", child.Config.Coordinate))
+			field.F("child", child.Config.Coordinate),
+			field.StatusDeploymentSkipped())
 
 		// after the first iteration
 		if parent != root {
-			l.WithFields(field.F("root", root.Config.Coordinate), field.StatusDeploymentSkipped()).
+			l.WithFields(field.F("root", root.Config.Coordinate)).
 				Warn("Skipping deployment of %v, as it depends on %v which was not deployed after root dependency configuration %v %s", childCfg.Coordinate, parent.Config.Coordinate, root.Config.Coordinate, reason)
 		} else {
-			l.WithFields(field.StatusDeploymentSkipped()).
-				Warn("Skipping deployment of %v, as it depends on %v which %s", childCfg.Coordinate, parent.Config.Coordinate, reason)
+			l.Warn("Skipping deployment of %v, as it depends on %v which %s", childCfg.Coordinate, parent.Config.Coordinate, reason)
 		}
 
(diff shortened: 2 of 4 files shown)


pkg/deploy/deploy.go
```diff
@@ -221,11 +221,9 @@ func (c *componentDeployer) deploy(ctx context.Context) error {
 
 		for _, root := range roots {
 			node := root.(graph.ConfigNode)
-			ctx := context.WithValue(ctx, log.CtxKeyCoord{}, node.Config.Coordinate)
-
 			go func(ctx context.Context, node graph.ConfigNode) {
 				errChan <- c.deployNode(ctx, node)
-			}(ctx, node)
+			}(context.WithValue(ctx, log.CtxKeyCoord{}, node.Config.Coordinate), node)
```
Laubi, line 226: Wouldn't it be more readable if we kept the var on L224?

pkg/deploy/deploy.go
```diff
@@ -291,10 +289,11 @@ func (c *componentDeployer) removeChildren(ctx context.Context, parent, root gra
 
 		// after the first iteration
 		if parent != root {
-			l.WithFields(field.F("root", root.Config.Coordinate)).
+			l.WithFields(field.F("root", root.Config.Coordinate), field.StatusDeploymentSkipped()).
```
Laubi, line 292: Pls add the field to L288 :) 
The custom logger there already has a lot of common attributes between the two log lines :)

pkg/deploy/sequential/deploy_sequential.go
```diff
@@ -104,7 +104,7 @@ func DeployConfigs(clientSet deploy.ClientSet, apis api.APIs, sortedConfigs []co
 
 func deployConfig(ctx context.Context, clientSet deploy.ClientSet, apis api.APIs, em *entityMapWithNames, c *config.Config) (config.ResolvedEntity, []error) {
 	if c.Skip {
-		log.WithCtxFields(ctx).Info("Skipping deployment of config %s", c.Coordinate)
+		log.WithCtxFields(ctx).WithFields(field.StatusDeploymentSkipped()).Info("Skipping deployment %s")
```
Laubi, line 107:
```suggestion
		log.WithCtxFields(ctx).WithFields(field.StatusDeploymentSkipped()).Info("Skipping deployment")
```

Laubi (changes requested):

pkg/deploy/deploy.go
warber: yes, maybe. just did it because I don't like shadowed variables so much.  I can change it back if you want tho

Laubi (approved):

pkg/client/dtclient/settings_client.go
```diff
@@ -202,7 +202,7 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
 			props = append(props, fmt.Sprintf("(%v = %v)", k, v))
 		}
 
-		log.WithCtxFields(ctx).Info("Updating existing object %q with matching unique properties: %v", matchingObject.object.ObjectId, strings.Join(props, ", "))
+		log.WithCtxFields(ctx).Debug("Updating existing object %q with matching unique properties: %v", matchingObject.object.ObjectId, strings.Join(props, ", "))
```
UnseenWizzard, line 205: @Laubi didn't we discuss not just setting this to debug completely last week? :/ 
As said, I view this as the same type of INFO the logs on creating or updating configs by name or ID.
This tells me that an existing object was updated, rather than a new setting being created.

I agree to moving the full list of matching properties to debug, but as said last week, IMO this shouldn't fully vanish into Debug logs.

FYI @warber (not to worry for the bugfix release, but we should re-discuss this handling)

warber: > This tells me that an existing object was updated, rather than a new setting being created.

@UnseenWizzard IMO the way to go is to introduce a field of the general response struct the returned by the (new) api clients that can be used to check whether calling e.g. `Upsert` resulted in a modification or a creation. Based on that the deployStatus field can be set accordingly.

PR #1202, opened 2023-09-26
fix: Default TokenEndpoint not set when using `monaco downoad --url` 

warber:
When using `monaco download --url` the (default) token endpoint was not correctly set anymore.
This was introduced when we've started to use the clients from the core lib.
Unfortunately, the assumption that https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/406bafbbfeb87a04d91893ca3f38bb84ce623bc6/pkg/client/clientset.go#L160 
will contain the currently configured token endpoint OR the fallback to the default endpoint was wrong.

This PR contains a fix so that the fallback is already set when getting the auth credentials from the manifest.


pkg/manifest/manifest.go
```diff
@@ -38,10 +39,10 @@ type OAuth struct {
 	TokenEndpoint *URLDefinition
 }
 
-// GetTokenEndpointValue returns the defined token endpoint or an empty string if it's not set.
+// GetTokenEndpointValue returns the defined token endpoint or the default token endpoint if none is defined
 func (o OAuth) GetTokenEndpointValue() string {
-	if o.TokenEndpoint == nil {
-		return ""
+	if o.TokenEndpoint == nil || o.TokenEndpoint.Value == "" {
```
UnseenWizzard, line 44: nit: for this to be an empty string, a user would have to explicitly set it to "" or an env var that resolves to "" right? 
I wonder if we shouldn't print a warning for that case, before defaulting to the default URL. If you explicitly set this to empty, you probably made a mistake trying to set it to an actual value 🤔

UnseenWizzard (approved):
Changes look good for the bugfix :shipit: 
After getting the bugfix out, I think it'd still be worth thinking about adding at least one E2E test that runs without explicitly set token URL... With making our test configurations runnable against non-production, we kinda made them unrealistic - most users don't set an explicit token URL... would make sense for tests to cover that normal case IMO

jskelin (approved):

pkg/manifest/manifest.go
warber: it would be beneficial, however, it would be nicer to have that warning once and not whenever that method is called.

PR #1209, opened 2023-09-29
Reuse templatetools.JSONObject during the download process

jskelin:
…for the automation

Harmonize code for bucket and automation download. Because the functionality is similar, it is easier to have code also that looks similar.

### DIFF
--- a/pkg/download/bucket/bucket.go
+++ b/pkg/download/bucket/bucket.go
@@ -59,15 +59,14 @@ func (d *Downloader) Download(projectName string, _ ...config.BucketType) (v2.Co
 		return nil, nil
 	}
 
-	configs := d.convertAllObjects(projectName, response.Objects)
+	configs := convertAllObjects(projectName, response.Objects)
 	result["bucket"] = configs
 	return result, nil
 }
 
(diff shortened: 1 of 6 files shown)


Laubi:
This might be controversial, however, I'd rather not introduce the `templatetools.NewJSONObject(o)` to handle this case. 
In my personal opinion, I think it adds unnecessary overhead to the simple handling of the `map[k]v` structure. I personally have a harder time to read the object-oriented code than the existing one

pkg/download/bucket/bucket.go
```diff
@@ -77,14 +76,15 @@ func (d *Downloader) convertAllObjects(projectName string, objects [][]byte) []c
 
 		// exclude builtin bucket names
 		if tools.IsDefault(c.OriginObjectId) {
-			log.Debug("Skipping download of immutable default Bucket %s", c.OriginObjectId)
+			log.Debug("Skipping download of immutable default Bucket %s", c.OriginObjectId) //TODO: WithFields??
```
UnseenWizzard, line 79: IMO wouldn't hurt to include the ID as a field.

warber:
@jskelin Can we close this?

jskelin:
abandoned

PR #1212, opened 2023-10-02
feat: enable managing bucket definitions

warber:
Removes FF for bucket definitions, thus enabling support for buckets per default.

User is able to use monaco to configure grail buckets per default without setting using a feature flag


UnseenWizzard:
@warber I've just merged #1211 which sets the feature flag for all-configs and download/restore E2E tests - please rebase and search/remove usages of the flags one more time

UnseenWizzard (approved):
Changes look good to me. Several E2E tests appear to have failed with the unrelated API Error when saving app detection rules

jskelin (approved):

PR #1214, opened 2023-10-03
fix: unable to deploy app-detection-rules

warber:
Sometimes, the API returns 500 Internal Server Error e.g. when an application referenced by an application detection rule is not fully "ready" yet.

This PR adds another special treatment to handle that specific case.


UnseenWizzard (approved):
Yay... more special retry rules :/

Laubi (approved):

jskelin (approved):

PR #1219, opened 2023-10-05
chore: delete code responsible for performing sequential deployments

warber:
The "feature" of doing parallel deployments is now on per default for some time.
IMO this gives use enough confidence to delete the feature flag(s) and code paths for sequential deployments


pkg/deploy/deploy_test.go
```diff
@@ -748,38 +747,7 @@ func TestDeployConfigsWithDeploymentErrors(t *testing.T) {
...
-		envErrs := make(errors.EnvironmentDeploymentErrors)
-		assert.ErrorAs(t, err, &envErrs)
-		assert.Len(t, envErrs, 1)
-		assert.Len(t, envErrs[env], 1)
-		var depErr errors.DeploymentErrors
-		assert.ErrorAs(t, envErrs[env][0], &depErr)
-		assert.Equal(t, 2, depErr.ErrorCount, "Expected deployment to continue after the first error and count errors for both invalid configs")
-	})
-
 	t.Run("[parallel] deployment error - always continues on error", func(t *testing.T) {
```
UnseenWizzard, line 750: nit:
```suggestion
	t.Run("deployment error - always continues on error", func(t *testing.T) {
```

warber: thx, this means we can actually also delete the `ContinueOnError` Options field

UnseenWizzard: I think it's still used for environments? We deploy to envs in sequence don't we? And use cont-on-error there?

UnseenWizzard (approved):
Lgtm, parallel deployments have been out fully released since 2.8.0/3 weeks and longer partially and internally.

PR #1220, opened 2023-10-05
chore: use `MaskedString`for sensitive fields in AuthSecret

warber:
Introducing `MaskedString`type to use for fields that store potentially sensitive information.


UnseenWizzard (approved):
🚀

PR #1225, opened 2023-10-09
fix(generate): Prepend timestamp if generated delete.yaml already exists

UnseenWizzard:
As the delete command validates if it was passed a yaml file, by checking the file ending, the current handling of avoiding overwriting existing files by appending a filestamp, means that such generated deletefiles can't be used directly. To address this, the timestamp is simply pre-pended, which also makes sorting by filename easier.

### DIFF
--- a/cmd/monaco/generate/deletefile/deletefile_test.go
+++ b/cmd/monaco/generate/deletefile/deletefile_test.go
@@ -244,10 +244,36 @@ func TestDoesNotOverwriteExistingFiles(t *testing.T) {
 
 	t.Setenv("TOKEN", "some-value")
 
+	t.Run("default filename", func(t *testing.T) {
+		time := timeutils.TimeAnchor().Format("20060102-150405")
+		newFile := fmt.Sprintf("delete_%s.yaml", time)
+		testPreexistingFileIsNotOverwritten(t, "delete.yaml", newFile, false)
+	})
+
+	t.Run("custom filename", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cmd/monaco/generate/deletefile/deletefile_test.go
```diff
@@ -279,7 +279,7 @@ func TestDoesNotOverwriteExistingFiles(t *testing.T) {
 
 	// AND THEN new DOT file is created with timestamp appended
 	time := timeutils.TimeAnchor().Format("20060102-150405")
-	newFile := fmt.Sprintf("delete.yaml_%s", time)
+	newFile := fmt.Sprintf("%s_delete.yaml", time)
```
Laubi, line 282:
```suggestion
	newFile := fmt.Sprintf("delete_%s.yaml", time)
```
Wouldn't it be easier to log the timestamp after the delete? With that, autocompletion can already suggest what files are delete-files. 

We do it similarly with downloads (`download_{timestamp}`).

UnseenWizzard: As you can define your own filenames, it's not just "after the `delete`", it would mean splitting the string on the last dot and placing the timestamp between the actual name and extension.

Can be done, but not sure if it's that much better tbh. Prepended timestamps can also make sorting through the files a bit easier

UnseenWizzard: Addressed by inserting between filename and file-ending (before the last dot; if one exists) as discussed internally

PR #1227, opened 2023-10-09
feat: Skip "deleting" buckets

Laubi:
Up to now, if a bucket is in the "deleting" state, it is still downloaded. IMO this is wrong, thus this PR.

Log output looks like this after the change:
```
2023-10-09T15:58:36+02:00       debug   Skipping bucket: bucket "dt_query_executions" is immutable
2023-10-09T15:58:36+02:00       debug   Skipping bucket: bucket "default_logs" is immutable
2023-10-09T15:58:36+02:00       debug   Skipping bucket: bucket "default_bizevents" is immutable
2023-10-09T15:58:36+02:00       debug   Skipping bucket: bucket "default_events" is immutable
2023-10-09T15:58:36+02:00       debug   Skipping bucket: bucket "default_metrics" is immutable
2023-10-09T15:58:36+02:00       debug   Skipping bucket: bucket "default_spans" is immutable
2023-10-09T15:58:36+02:00       debug   Skipping bucket: bucket "default_security_events" is immutable
2023-10-09T15:58:36+02:00       debug   Skipping bucket: bucket "dt_system_events" is immutable
2023-10-09T15:58:36+02:00       debug   Skipping bucket: bucket "default_davis_events" is immutable
2023-10-09T15:58:36+02:00       debug   Skipping bucket: bucket "buckets_my-awesome-bucket" is deleting
```

Additionally, the 'updatable' state should not persist, AND it can be used to filter buckets that should not be downloaded.

### DIFF
--- a/pkg/download/bucket/bucket.go
+++ b/pkg/download/bucket/bucket.go
@@ -96,14 +96,6 @@ const (
 )
 
 func convertObject(o []byte, projectName string) (config.Config, error) {
-	c := config.Config{
-		Coordinate: coordinate.Coordinate{
-			Project: projectName,
-			Type:    "bucket",
-		},
-		Type: config.BucketType{},
-	}
-
 	r, err := templatetools.NewJSONObject(o)
 	if err != nil {
 		return config.Config{}, fmt.Errorf("failed to unmarshal bucket: %w", err)
@@ -114,26 +106,33 @@ func convertObject(o []byte, projectName string) (config.Config, error) {
 		return config.Config{}, fmt.Errorf("variable %q unreadable", bucketName)
 	}
 
-	c.Coordinate.ConfigId = id
-	c.OriginObjectId = id
-
 	// remove fields that will be set on deployment
 	r.Delete(bucketName)
 	r.Delete(status)
 	r.Delete(version)
 
 	// pull displayName into paramter if one exists
-	c.Parameters = map[string]parameter.Parameter{}
+	parameters := map[string]parameter.Parameter{}
 	p := r.Parameterize(displayName)
 	if p != nil {
-		c.Parameters[displayName] = p
+		parameters[displayName] = p
 	}
 
 	t, err := r.ToJSON()
 	if err != nil {
 		return config.Config{}, err
 	}
-	c.Template = template.NewInMemoryTemplate(id, string(jsonutils.MarshalIndent(t)))
+
+	c := config.Config{
+		Coordinate: coordinate.Coordinate{
+			Project:  projectName,
+			Type:     "bucket",
+			ConfigId: id,
+		},
+		OriginObjectId: id,
+		Type:           config.BucketType{},
+		Template:       template.NewInMemoryTemplate(id, string(jsonutils.MarshalIndent(t))),
+	}
 
 	return c, nil
 }
--- a/pkg/download/bucket/bucket.go
+++ b/pkg/download/bucket/bucket.go
@@ -18,6 +18,7 @@ package bucket
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/buckets"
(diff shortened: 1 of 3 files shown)


pkg/download/bucket/bucket.go
```diff
@@ -69,17 +77,16 @@ func (d *Downloader) convertAllObjects(projectName string, objects [][]byte) []c
 
 		c, err := convertObject(o, projectName)
 		if err != nil {
-			log.WithFields(field.Type("bucket"), field.Error(err)).
-				Error("Failed to decode API response objects for bucket resource: %v", err)
-			continue
-		}
+			if errors.As(err, &skipErr{}) {
```
Laubi, line 80: Not tooooooooo happy with that, but best I've come up with. We don't need the content of the error, just need to check if it's the correct type. 
`errors.Is` does not work with this

pkg/download/bucket/bucket.go
```diff
@@ -69,17 +77,16 @@ func (d *Downloader) convertAllObjects(projectName string, objects [][]byte) []c
...
+				log.WithFields(field.Type("bucket")).
+					Debug("Skipping bucket: %s", err.Error())
+			} else {
+				log.WithFields(field.Type("bucket"), field.Error(err)).
+					Error("Failed to decode API response objects for bucket resource: %v", err)
+			}
 
-		// exclude builtin bucket names
-		if tools.IsDefault(c.OriginObjectId) {
```
UnseenWizzard, line 78: IsDefault is used on deletion as well. Can we use the updatable property there as well and replace our name-based logic?

pkg/download/bucket/bucket.go
```diff
@@ -107,6 +114,16 @@ func convertObject(o []byte, projectName string) (config.Config, error) {
 		return config.Config{}, fmt.Errorf("variable %q unreadable", bucketName)
 	}
 
+	// skip unmodifiable buckets
+	if u, ok := r.Get(updatable).(bool); ok && !u {
```
UnseenWizzard, line 118: I find the error pretty weird too tbh. Wouldn't it be nicer to just partially deserialize the object (like we do for automation) and look at the updatable field - instead of looking and throwing an error on `convertObject` if it's un-modifiable? (just writing that makes it seem odd for 'convert' to fail if the object is unmodifiable)

pkg/download/bucket/bucket.go
Laubi: Could do. For now, I've reintroduced the IsDefault method, as older versions of Dynatrace may not set the 'new' field. Just to be sure.
For delete, we could also introduce the additional updatable check.

pkg/download/bucket/testdata/buckets.json
```diff
@@ -33,7 +36,18 @@
...
+        },
+        {
+            "bucketName": "my-bucket-name-that-shall-not-be-downloaded",
+            "table": "metrics",
+            "displayName": "Default metrics (15 months)",
+            "status": "deleting",
+            "retentionDays": 462,
+            "metricInterval": "PT1M",
+            "version": 1,
+            "updatable": true
```
UnseenWizzard, line 50: as you mentioned that older versions might not return that field, please don't forget to reintroduce some test configs that do not include it!

pkg/download/bucket/bucket.go
```diff
@@ -104,25 +105,38 @@ const (
...
-		return config.Config{}, fmt.Errorf("variable %q unreadable", bucketName)
+	// bucketName acts as the id, thus it must be set
+	if b.Name == "" {
+		return config.Config{}, fmt.Errorf("bucketName is not set")
 	}
 
 	// skip unmodifiable buckets
-	if u, ok := r.Get(updatable).(bool); ok && !u || buckettools.IsDefault(id) {
-		return config.Config{}, skipErr{fmt.Sprintf("bucket %q is immutable", id)}
+	if b.Updatable != nil && *b.Updatable == false || buckettools.IsDefault(b.Name) {
```
UnseenWizzard, line 127: If I got it correctly, the field is set not only if a bucket is a default object, but also if it's currently in a state where not changes are allowed, like 'deleting'?

I really think it would then be worth introducing in a general way, and also using if for the check in delete.
There's no reason to try and delete a bucket that's already being deleted.

pkg/download/bucket/bucket.go
Laubi: Yes, I agree, it's exactly my comment above. 
We have multiple checks, and moving them one level above also seems funny. Do you have a concrete idea on how we could solve this?

pkg/download/bucket/bucket.go
Laubi: No, updateable is only set if the bucket is a predefined object.

pkg/download/bucket/testdata/buckets.json
Laubi: Added

pkg/download/bucket/bucket.go
Laubi: It is; that's why I added the original comment to the PR. 
Moving the checks out also seems strange: We move it one level up with many continue statements or one level up and down into a new validation function - where we still need to differentiate between error and skip. Do you have a concrete solution for how we can do this nicely? I'd like to do it nicely :D

pkg/download/bucket/bucket.go
UnseenWizzard: Ah! I missed the code on line 132 😅

UnseenWizzard (approved):

PR #1235, opened 2023-10-13
feat: load and parse account management resources

warber:
This PR introduces the capability to read account management resources from files living inside of monaco project folders. The PR contains a function responsible for loading raw resources into a first in-memory representation as well as a function validating the references between users groups and policies.

### DIFF
--- a/pkg/persistence/account/load.go
+++ b/pkg/persistence/account/load.go
@@ -1,0 +1,84 @@
+/*
+ * @license
+ * Copyright 2023 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package account
+
+import (
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/files"
+	"github.com/mitchellh/mapstructure"
+	"github.com/spf13/afero"
+	"gopkg.in/yaml.v2"
+	"io"
+)
+
+// Load loads account management resources from YAML configuration files
+// located within the specified root directory path. It parses the YAML files, extracts policies,
+// groups, and users data, and organizes them into a AMResources struct, which is then returned.
+func Load(fs afero.Fs, rootPath string) (*AMResources, error) {
+	resources := &AMResources{
+		Policies: make([]Policy, 0),
+		Groups:   make([]Group, 0),
+		Users:    make([]User, 0),
+	}
+
+	yamlFilePaths, err := files.FindYamlFiles(fs, rootPath)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, yamlFilePath := range yamlFilePaths {
+		yamlFile, err := fs.Open(yamlFilePath)
+		if err != nil {
+			return nil, err
+		}
+		content, err := decode(yamlFile)
+		if err != nil {
+			return nil, err
+		}
+
+		var policies Policies
+		err = mapstructure.Decode(content, &policies)
+		if err != nil {
+			return nil, err
+		}
+		resources.Policies = append(resources.Policies, policies.Policies...)
+
+		var groups Groups
+		err = mapstructure.Decode(content, &groups)
+		if err != nil {
+			return nil, err
+		}
+		resources.Groups = append(resources.Groups, groups.Groups...)
+
+		var users Users
+		err = mapstructure.Decode(content, &users)
+		if err != nil {
+			return nil, err
+		}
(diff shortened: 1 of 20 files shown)


pkg/persistence/account/load.go
```diff
@@ -0,0 +1,84 @@
...
+			return nil, err
+		}
+		resources.Users = append(resources.Users, users.Users...)
+	}
+
+	return resources, nil
+}
+
+func decode(in io.ReadCloser) (map[string]any, error) {
+	defer in.Close()
```
UnseenWizzard, line 78: nit/philosophical: I know this is an internally used function, but IMO, closing the file/reader should be the responsibility of the function that opened it, not a function working on it. But I get why you're closing it right after decoding.

pkg/persistence/account/validate.go
```diff
@@ -0,0 +1,86 @@
...
+	if reference, isCustomRef := elem.(anyMap); isCustomRef {
+		idField, hasIdField := reference["id"]
+		if !hasIdField {
+			return fmt.Errorf("error validating account resources: %w", ErrIdFieldMissing)
+		}
+		idFieldStr, isIdFieldStr := idField.(string)
+		if !isIdFieldStr {
+			return fmt.Errorf("error validating account resources: %w", ErrIdFieldNoString)
+		}
+		policyRefExists := refCheckFn(idFieldStr)
```
UnseenWizzard, line 62:
```suggestion
		refExists := refCheckFn(idFieldStr)
```

I think "policyRef" is a left over from before you'v unified this?

pkg/persistence/account/types.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	Environment struct {
+		Name        string `mapstructure:"name"`
+		Permissions []any  `mapstructure:"permissions"`
+		Policies    []any  `mapstructure:"policies"`
+	}
+	ManagementZone struct {
+		Environment    string `mapstructure:"environment"`
+		ManagementZone string `mapstructure:"managementZone"`
+		Permissions    []any  `mapstructure:"permissions"`
```
UnseenWizzard, line 66: ManagementZones level can not have Policies?

pkg/persistence/account/validate.go
```diff
@@ -0,0 +1,86 @@
...
+
+import "fmt"
+
+// Validate checks the references in the provided AMResources instance to ensure
+// that all referenced groups and policies exist. It iterates through the users,
+// environment policies, and account policies, validating their references.
+func Validate(res *AMResources) error {
+	for _, user := range res.Users {
+		for _, groupRef := range user.Groups {
+			if err := refCheck(groupRef, res.GroupExists); err != nil {
```
UnseenWizzard, line 27: So for each user with group references, we iterate each group and check if the referenced ID is found.
And then we iterate the groups once again to iterate over all policies twice for each of them (once for env policies and once for account).

Couldn't we achieve the same with fewer iterations if we reversed the order of the checks?

1. Iterate res.Policies once, build set of IDs
2. Iterate res.Groups once, build set of IDs & check for each env and account policy if it's known/in the policy ID set
3. Iterate res.Users once, check for each group reference if it's in the group ID set

UnseenWizzard: Or alternatively if it makes sense store the objects in a id->object maps inside AMResources already, instead of slices.

With this we could already produce early errors for duplicate IDs, which I've just noticed is missing from validation at the moment?

UnseenWizzard (changes requested):
IMO validation should be improved.

pkg/persistence/account/validate.go
warber: i didn't care about performance. But sure we can do that :)

UnseenWizzard: 😱 You what?! I'm shocked @warber... shocked... 

Seriously though, I think it's really worth closely thinking about performance for this. 
If a user wants to configure account permissions as code, they're likely doing it at a larger scale. For a few users and default permissions, they probably won't start with config-as-code

warber: makes sense, didn't expect account resources on a larger scale, will update the pr

pkg/persistence/account/types.go
warber: afaik, nope

warber:
@UnseenWizzard I've improved the PR, pls take another look

pkg/persistence/account/load.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/mitchellh/mapstructure"
+	"github.com/spf13/afero"
+	"gopkg.in/yaml.v2"
+	"io"
+)
+
+// Load loads account management resources from YAML configuration files
+// located within the specified root directory path. It parses the YAML files, extracts policies,
+// groups, and users data, and organizes them into a AMResources struct, which is then returned.
+func Load(fs afero.Fs, rootPath string) (*AMResources, error) {
```
UnseenWizzard, line 31: I'd mark this as excluded from the gocognit check. 
I've quickly tried if it would actually become nicer to read if we split out the parts into `loadPolicies`, `loadGroups`, .. methods. But with error handling still needed there's not much of a difference at all, actually it becomes messier to read.

UnseenWizzard (approved):

pkg/persistence/account/load.go
warber: yes the "complexity" is absolutely manageable

PR #1238, opened 2023-10-17
Feat: notify users about new Monaco version

warber:
This PR enables the monaco CLI to perform a version check at the end of each run before exiting the CLI to notify the user about newer versions of monaco being available.

This behavior is turned ON by default and can be disabled by setting `MONACO_SKIP_VERSION_CHECK` if needed.

User is notified when a new version of monaco is available each time the CLI executes.

### DIFF
--- a/cmd/monaco/main.go
+++ b/cmd/monaco/main.go
@@ -15,12 +15,51 @@
 package main
 
 import (
-	"os"
-
+	"context"
+	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/runner"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/version"
+	monacoVersion "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/version"
+	"net/http"
+	"os"
 )
 
 func main() {
+	var versionNotification string
+	go setVersionNotificationStr(&versionNotification)
+
 	statusCode := runner.Run()
+	notifyUser(versionNotification)
 	os.Exit(statusCode)
+}
+
+func setVersionNotificationStr(msg *string) {
+	currentVersion, err := version.ParseVersion(monacoVersion.MonitoringAsCode)
+	if err != nil {
+		log.Debug("Could not perform version check: %s", err)
+		return
+	}
+
+	latestVersion, err := version.GetLatestVersion(context.TODO(), &http.Client{}, "https://api.github.com/repos/dynatrace/dynatrace-configuration-as-code/releases/latest")
+	if err != nil {
+		log.Debug("Could not perform version check: %s", err)
+		return
+	}
+
+	if currentVersion == version.UnknownVersion || latestVersion == version.UnknownVersion {
+		return
+	}
+
+	if latestVersion.GreaterThan(currentVersion) {
+		*msg = fmt.Sprintf("A newer version (%s) of Monaco is available! "+
+			"You can download the latest release from here: https://github.com/Dynatrace/dynatrace-configuration-as-code/releases/latest\n", latestVersion)
+	}
+}
+
+func notifyUser(msg string) {
+	if msg == "" {
+		return
+	}
+	fmt.Println(msg)
 }
--- a/cmd/monaco/main.go
+++ b/cmd/monaco/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/runner"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/version"
 	monacoVersion "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/version"
 	"net/http"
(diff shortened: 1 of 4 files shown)


internal/version/releases.go
```diff
@@ -0,0 +1,73 @@
...
+	"time"
+)
+
+type Releases []Release
+
+type Release struct {
+	TagName string `json:"tag_name"`
+}
+
+var DefaultReleaseProvider = getReleaseData(&http.Client{Timeout: time.Second * 3}, "https://api.github.com/repos/dynatrace/dynatrace-configuration-as-code/releases")
```
Laubi, line 33: Shouldn't it be enough to query the latest release?
`https://api.github.com/repos/dynatrace/dynatrace-configuration-as-code/releases/latest`

internal/version/releases.go
```diff
@@ -0,0 +1,73 @@
...
+
+func GetLatestVersion(releasesProvider func() (Releases, error)) (Version, error) {
+	releases, err := releasesProvider()
+	if err != nil {
+		return UnknownVersion, err
+	}
+	if len(releases) <= 0 {
+		return UnknownVersion, fmt.Errorf("no monaco Releases found")
```
Laubi, line 41:
```suggestion
		return UnknownVersion, fmt.Errorf("no monaco releases found")
```

internal/version/releases.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	tagName := releases[0].TagName // Assuming Releases are sorted in descending order
```
Laubi, line 44: Pls, don't assume. 
We have two possibilities: Either always query the last one (could be a backport in the future, currently this is not the case), or check properly for the correct latest version

Laubi (changes requested):

internal/version/releases.go
warber: I just took your proposal url in the ticket without further questioning it. 
In that case, yeah your aboslutely right, let's take that one.

internal/version/releases.go
```diff
@@ -0,0 +1,73 @@
...
+func getReleaseData(client *http.Client, url string) func() (Releases, error) {
+	return func() (Releases, error) {
+		resp, err := client.Get(url)
+		if err != nil {
+			return Releases{}, err
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode != http.StatusOK {
+			return Releases{}, fmt.Errorf("failed to fetch Release data. Status code: %d", resp.StatusCode)
```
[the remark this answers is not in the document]
warber: ide refactor/rename

cmd/monaco/cmdutils/checkversion.go
```diff
@@ -0,0 +1,42 @@
...
+	latestVersion, err := version.GetLatestVersion(version.DefaultReleaseProvider)
+	if err != nil {
+		log.Debug("Could not perform version check: %s", err)
+	}
+	currentVersion, err := version.ParseVersion(version2.MonitoringAsCode)
+	if err != nil {
+		log.Debug("Could not perform version check: %s", err)
+	}
+	if latestVersion.GreaterThan(currentVersion) {
+		fmt.Printf("A newer version (%s) of Monaco is available! "+
```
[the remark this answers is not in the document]
warber: that's what is currently the case, or do you mean the function should not take the burden of printing it to console?

internal/version/releases.go
```diff
@@ -0,0 +1,73 @@
...
+	if parsedVersion, err := ParseVersion(tagName); err != nil {
+		return UnknownVersion, err
+	} else {
+		return parsedVersion, nil
+	}
+}
+
+func getReleaseData(client *http.Client, url string) func() (Releases, error) {
+	return func() (Releases, error) {
+		resp, err := client.Get(url)
```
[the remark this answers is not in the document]
warber: technically not really needed in this case, but why not.

cmd/monaco/cmdutils/checkversion.go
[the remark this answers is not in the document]
Laubi: Uhh, I did overlook that the first time around. The current approach executes Monaco, then blocks until the new version is fetched, right?

A solution we discussed was that we start the check immediately after the CLI starts.
If the request is done successfully before the command is done, we print the version at the very end. 
If the request is not done (or not successful) after the command is done, cancel the request and do nothing.

cmd/monaco/main.go
```diff
@@ -15,12 +15,54 @@
...
+
 	statusCode := runner.Run()
+	notifyUser(versionNotification)
 	os.Exit(statusCode)
 }
+
+func setVersionNotificationStr(msg *string) {
+	currentVersion, err := version.ParseVersion(monacoVersion.MonitoringAsCode)
+	if err != nil {
+		log.Debug("Could not perform version check: %s", err)
```
Laubi, line 43: nit:
Intetion: Differentiate between the local check and the remote one. 
For us, the local one will always fail due to being a 'dev' version.

```suggestion
		log.WithFields(field.Error(err)).Debug("Can't parse current monaco version: %s", err)
```

cmd/monaco/main.go
```diff
@@ -15,12 +15,54 @@
...
+		*msg = fmt.Sprintf("A newer version (%s) of Monaco is available! "+
+			"You can download the latest release from here: https://github.com/Dynatrace/dynatrace-configuration-as-code/releases/latest\n", latestVersion)
+	}
+}
+
+func notifyUser(msg string) {
+	if msg == "" {
+		return
+	}
+	fmt.Println(msg)
```
Laubi, line 67: Is this Println on purpose or did you overlook it? :D 
```suggestion
	log.Info(msg)
```

cmd/monaco/main.go
```diff
@@ -15,12 +15,54 @@
...
+func setVersionNotificationStr(msg *string) {
+	currentVersion, err := version.ParseVersion(monacoVersion.MonitoringAsCode)
+	if err != nil {
+		log.Debug("Could not perform version check: %s", err)
+		return
+	}
+
+	latestVersion, err := version.GetLatestVersion(context.TODO(), &http.Client{}, "https://api.github.com/repos/dynatrace/dynatrace-configuration-as-code/releases/latest")
+	if err != nil {
+		log.Debug("Could not perform version check: %s", err)
```
Laubi, line 49: nit:
```suggestion
		log.WithFields(field.Error(err)).Debug("Could not perform version check: %s", err)
```

Laubi (approved):

Laubi (approved):

PR #1240, opened 2023-10-18
feat(manifest): Allow defining accounts

Laubi:
This PR allows defining accounts in the `manifest.yaml`.

To do that, a user must increase the manifest-version to `1.1`.

Example manifest:
```yaml
manifestVersion: 1.1                                # manifestVersion needs to be set to at least 1.1

accounts:                                           # New top level property
- name: other-account                               # name
  accountUUID: c3f50f90-a1e2-4e7b-aadb-f3dea28e2294 # account uuid
  apiUrl:                                           # allow overriding endpoint (optional)
    type: environment
    value: ENV_API_URL
  oAuth:                                            # credentials (only oAuth)
    clientId:
      name: ENV_CLIENT_ID
    clientSecret:
      name: ENV_CLIENT_SECRET
    tokenEndpoint:                                  # (optional)
      type: environment
      value: ENV_TOKEN_URL

```

### Does this PR introduce a user-facing change?
No, not really. Everything is still compatible if using monaco 1.0 manifests.

### Other changes included
`manifest_test.go` is now in the `manifest_test` package so we can do package tests.

### Future work
Some identified refactorings:
1) The manifest-version has 2 parts (major, minor), while the internal version used has three (major, minor, patch). Printing the version easily is not possible because of that, as we would suddenly print the three-part instead of the two parts on messages. A extra 'manifestversion' struct should be introduced.
2) The code is mostly inside one file (manifest_loader). It's quite difficult to navigate, and splitting it would make sense.

### DIFF
--- a/internal/featureflags/temporary.go
+++ b/internal/featureflags/temporary.go
@@ -75,3 +75,10 @@ func UpdateNonUniqueByNameIfSingleOneExists() FeatureFlag {
 		defaultEnabled: true,
 	}
 }
+
+func AccountManagement() FeatureFlag {
+	return FeatureFlag{
+		envName:        "MONACO_FEAT_ACCOUNT_MANAGEMENT",
+		defaultEnabled: false,
+	}
+}
--- a/pkg/manifest/account.go
+++ b/pkg/manifest/account.go
@@ -1,0 +1,97 @@
+/*
+ * @license
+ * Copyright 2023 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package manifest
+
+import (
+	"errors"
+	"fmt"
+	"github.com/google/uuid"
+)
+
+var (
+	errNameMissing   = errors.New("name is missing")
+	errAccUidMissing = errors.New("accountUUID is missing")
+)
+
+type invalidUUIDError struct {
+	uuid string
+	err  error
+}
+
+func (e invalidUUIDError) Error() string {
+	return fmt.Sprintf("invalid uuid %q: %s", e.uuid, e.err)
+}
+
+func (e invalidUUIDError) Unwrap() error {
+	return e.err
+}
+
+func convertSingleAccount(c *LoaderContext, a account) (Account, error) {
+
+	if a.AccountUUID == "" {
+		return Account{}, errAccUidMissing
+	}
+
+	accountId, err := uuid.Parse(a.AccountUUID)
+	if err != nil {
+		return Account{}, invalidUUIDError{a.AccountUUID, err}
+	}
+
+	oAuthDef, err := parseOAuth(c, a.OAuth)
+	if err != nil {
+		return Account{}, fmt.Errorf("oAuth is invalid: %w", err)
+	}
+
+	var urlDef *URLDefinition
+	if a.ApiUrl != nil {
+		if u, err := parseURLDefinition(c, *a.ApiUrl); err != nil {
+			return Account{}, fmt.Errorf("apiUrl: %w", err)
+		} else {
+			urlDef = &u
+		}
+	}
+
+	acc := Account{
+		Name:        a.Name,
+		AccountUUID: accountId,
+		ApiUrl:      urlDef,
+		OAuth:       oAuthDef,
+	}
+
+	return acc, nil
+}
+
+// convertAccounts converts the persistence definition to the in-memory definition
+func convertAccounts(c *LoaderContext, accounts []account) (map[string]Account, error) {
+
+	result := make(map[string]Account, len(accounts))
+
+	for i, a := range accounts {
+		if a.Name == "" {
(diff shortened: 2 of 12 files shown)


pkg/manifest/manifest.go
```diff
@@ -107,10 +109,29 @@ func (e Environments) Names() []string {
...
+type Account struct {
+	// Name is the account-name that is used to resolve user-defined lookup names.
+	Name string
+
+	// AccountUUID is the Dynatrace-account UUID
+	AccountUUID uuid.UUID
+
+	// ApiUrl is the target URL of this account.
+	// It is used when the default account management url is not the target account management url.
+	ApiUrl *url2.URL
```
UnseenWizzard: ~~Why use the `net/url` type here instead of the `*URLDefinition` that would allow setting in as env var as well?~~

Addressed/changed to URLDefinition in a later commit 👍

pkg/manifest/manifest_loader.go
```diff
@@ -293,43 +291,31 @@ func readManifestYAML(context *LoaderContext) (manifest, error) {
...
+		return fmt.Errorf("`manifestVersion` %s is no longer supported. Min required version is %s, please update manifest", m.ManifestVersion, version.MinManifestVersion)
 	}
 
 	if v.GreaterThan(maxSupportedManifestVersion) {
-		return fmt.Errorf("`manifestVersion` %s is not supported by monaco %s. Max supported version is %s, please check manifest or update monaco", manifestVersion, version.MonitoringAsCode, version.ManifestVersion)
+		return fmt.Errorf("`manifestVersion` %s is not supported by monaco %s. Max supported version is %s, please check manifest or update monaco", m.ManifestVersion, version.MonitoringAsCode, version.ManifestVersion)
+	}
+
+	if len(m.Accounts) > 0 && v.SmallerThan(manifestAccountSupport) {
+		return fmt.Errorf("`accounts` are unknown prior to `manifestVersion: \"1.1\"`, please update `manifestVersion` to at least `1.1`")
```
UnseenWizzard, line 327: nit: Maybe "unsupported" instead of "unknown"? It's a bit strange to say we don't "know" about something we've clearly just parsed and could load just fine in a monaco version that prints the warning

pkg/manifest/manifest_loader.go
```diff
@@ -149,36 +150,102 @@ func LoadManifest(context *LoaderContext) (Manifest, []error) {
...
 	environmentDefinitions, manifestErrors := toEnvironments(context, manifestYAML.EnvironmentGroups)
-
 	if manifestErrors != nil {
 		errs = append(errs, manifestErrors...)
 	} else if len(environmentDefinitions) == 0 {
 		errs = append(errs, newManifestLoaderError(context.ManifestPath, "no environments defined in manifest"))
 	}
 
+	// accounts
+	accounts, accErrs := convertAccounts(context, manifestYAML.Accounts)
```
UnseenWizzard: nit: `toAccounts` or also rename `toEnvironments` to `convert` - it's odd that functions that do the same operation on different types have different naming patterns

pkg/manifest/account.go
```diff
@@ -0,0 +1,97 @@
...
+package manifest
+
+import (
+	"errors"
+	"fmt"
+	"github.com/google/uuid"
+)
+
+var (
+	errNameMissing   = errors.New("name is missing")
```
UnseenWizzard, line 26: why introduce unexported sentinel errors? Are they just used in tests?

internal/featureflags/temporary.go
```diff
@@ -75,3 +75,10 @@ func UpdateNonUniqueByNameIfSingleOneExists() FeatureFlag {
 		defaultEnabled: true,
 	}
 }
+
+func AccountManagement() FeatureFlag {
```
UnseenWizzard, line 79: Should we put the loading of accounts under the feature flag as well? 
I guess we're fine with always understanding a 1.1 manifest, as long as we safeguard the actual deployment, so might not be needed.

UnseenWizzard (approved):
Changes look good to me overall 👍 Just a few small comments/suggestions

pkg/manifest/account.go
Laubi: Yes. I wanted to not rely on the exact messages of errors but wanted to test them properly.

internal/featureflags/temporary.go
Laubi: That would've been my idea as well. I think loading does no harm.

pkg/manifest/manifest_loader.go
Laubi: Good point, changed it :)

pkg/manifest/manifest_loader.go
Laubi: Moved everything to *parse*
Its a difficult topic, as we mix two things in this package:
* Reading
* Writing

`convert` would not state from what to what
`to` is just a non-telling prefix, also does not state from what to what
`parse` is also not perfect, as we're not really parsing anything. However, since we do validation and conversion, I think it's the most fitting.

According to Bing Chat, a good name would be `transform`, but same problem as `convert` and `to`.

UnseenWizzard (approved):

pkg/manifest/manifest_loader.go
Laubi: I thought about that but then decided against it in this PR to keep it simple. The structure we'd need to introduce is the following:

```
manifest/
  .                    # in memory definitions
  loading             
  writing
  internal/persistence # persistence definitions
```

I can try it on a follow-up if you agree that it makes sense and is not too much overhead.

UnseenWizzard: I think the change/effort would be mostly shuffling files around? IMO worth at least trying to see if feels any more organised :) 

And yes, 100% meant as an optional followup, thus already approved (and merged as I see 👍 )

PR #1242, opened 2023-10-19
chore: Split up 'setup' task of Makefile

Laubi:
Currently, when executing any makefile task, we usually install all tools that we do use somewhere in the makefile. This is most inconvenient and slow. 

This change does split up the setup step and moves the install calls to the targets where they're actually used.

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,12 +8,12 @@ default: build
 
 setup:
 	@echo "Installing build tools..."
-	@go install github.com/google/addlicense@v1.1.1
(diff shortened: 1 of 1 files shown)


UnseenWizzard (approved):

UnseenWizzard:
One note to this: originally, setup was an independent target that you manually run once - to setup the build dependencies. It was added to the steps that actually need the installed tools for convenience at some point.

warber (approved):
lgtm

PR #1245, opened 2023-10-23
chore: Split e2e workflow into separate jobs

Laubi:
Currently, our e2e tests are performed sequentially, which has the conclusion that they're taking quite a whileto execute.

This PR makes four improvements:
1) Use jobs instead of steps to parallelize e2e runs.
2) Use a generic setup step to fetch the correct sha for checkout.
3) Move the ifs from the steps to the jobs. This has the consequence that we can't use env vars anymore, but it marks the jobs correctly as 'skipped'.
4) Rename collect-workflows to use the naming pattern `{workflow}--collect-results.yml` to show immediately for what workflow they're collecting test results.

### Future work
* Move windows tests in (?)


.github/workflows/end-to-end-test.yml
```diff
@@ -22,120 +22,285 @@ env:
...
+        URL_ENVIRONMENT_2: ${{ secrets.URL_ENVIRONMENT_2 }}
+        TOKEN_ENVIRONMENT_1: ${{ secrets.TOKEN_ENVIRONMENT_1 }}
+        TOKEN_ENVIRONMENT_2: ${{ secrets.TOKEN_ENVIRONMENT_2 }}
+        PLATFORM_URL_ENVIRONMENT_1: ${{ secrets.PLATFORM_URL_ENVIRONMENT_1 }}
+        PLATFORM_URL_ENVIRONMENT_2: ${{ secrets.PLATFORM_URL_ENVIRONMENT_2 }}
+        OAUTH_CLIENT_ID: ${{ secrets.OAUTH_CLIENT_ID }}
+        OAUTH_CLIENT_SECRET: ${{ secrets.OAUTH_CLIENT_SECRET }}
+        OAUTH_TOKEN_ENDPOINT: ${{ secrets.OAUTH_TOKEN_ENDPOINT }}
+
+    - name: 🧹 Cleanup
```
UnseenWizzard, line 95: ! cleanup probably shouldn't run parallel to other jobs! It will purge the environments - which we certainly don't want to happen while other tests run. 

This should be removed from all but the new "nightly job" or be it's own thing - and we should limit that job to run after the other's have finished

Laubi: Yes, this version was the first 'let's see if this even works' version, the new one cleans it up correctly.

UnseenWizzard (approved):

PR #1256, opened 2023-11-08
feat: add build for linux/amd container

jskelin:
#### Add build procedure to build multi-arch container

With these changes, the produced docker container is now multi-arch and contains `linux/amd64` and `linux/arm64` containers.

### DIFF
--- a/.ci/jenkins/tools/docker.groovy
+++ b/.ci/jenkins/tools/docker.groovy
@@ -1,0 +1,87 @@
+/*
+ * @license
+ * Copyright 2023 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * installDockerQemuEmulator enables to run/build another architectures
+ */
+void installQemuEmulator() {
+    sh "docker run --privileged --rm tonistiigi/binfmt --install all"
+}
+
+void createNewBuilder() {
+    try { //in case of a quick rerun of the process, a container already has this builder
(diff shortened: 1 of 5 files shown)


.ci/jenkins/tools/docker.groovy
```diff
@@ -0,0 +1,91 @@
...
+                sh(label: "sing container",
+                    script: 'COSIGN_PASSWORD=$cosign_password cosign sign --key env://cosign_key $registry/$repo/dynatrace-configuration-as-code:$version -y --recursive')
+            } finally {
+                sh(label: "sign out from docker registry", script: 'docker logout $registry')
+            }
+        }
+    }
+}
+
+void signContainer() {
```
UnseenWizzard, line 87: empty leftover/non-implemented function

Dockerfile
```diff
@@ -1,16 +1,21 @@
...
 
-ARG NAME=monaco
-ARG SOURCE=/build/${NAME}-linux-amd64
+FROM --platform=$BUILDPLATFORM golang:1.20-alpine AS build
+WORKDIR /src
+ARG TARGETOS
+ARG TARGETARCH
+ARG VERSION=2.x
+COPY . .
+RUN GOOS=$TARGETOS GOARCH=$TARGETARCH CGO_ENABLED=0  go build -a -tags netgo -ldflags "-X github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/version.MonitoringAsCode=${VERSION} -w -extldflags '-static'" ./cmd/monaco
```
UnseenWizzard, line 9: Is there any really good reason to move building the binary into the docker container?
I get making each artifact build standalone that makes perfect sense - but making it standalone inside the docker builds feels like it just adds more overhead than benefit.

Couldn't we still build the correct binary inside the Jenkins pod by building it as part of the docker make target, or the Jenkins function `buildContainer`? 
Copying the sources and building inside Docker just seems like overhead and is probably slower than building outside.
Go should be able to build any target architecture on any architecture, doesn't it?

UnseenWizzard: Discussed in person: As we use buildx to build for different architectures in a single command, this is the easiest way to build the monaco binary for the correct architecture. Makes perfect sense to do it this way with that!

UnseenWizzard (approved):
Looks good to me, but I'll admit I trust you on the details and just gave this a high-level review!

Container is tested and works well on Apple silicon arm64 👍 

If you didn't, please make sure there's a final test release so we see all versions of the binary are built and pushed to at least the internal storage as expected as well.

PR #1261, opened 2023-11-10
chore: add basic data model for am resources in pkg/account package

warber:
Added data model to be used by deploy and download AM resources functions.
Additionally, the load am resources function is now considering references, and storing values of type `Reference` when necessary.


UnseenWizzard (approved):
Lgtm - let's merge it and see how/if it evolves when we use the data model for deployments and downloads

PR #1262, opened 2023-11-10
feat: Allow writing account resources

UnseenWizzard:
Extend the persistence package with a function to write Account Resources to disk. 
Currently this method is never called outside of disks - guarding behind the feature flag should be done once/where it's called (as part of downloading Account Resources)

Several small refactoring commits. 3rd commit is the actual writer feature.

NO, no introduced function is called anywhere yet.


pkg/persistence/account/loader/load.go
```diff
@@ -28,11 +29,11 @@ import (
 // Load loads account management resources from YAML configuration files
 // located within the specified root directory path. It parses the YAML files, extracts policies,
 // groups, and users data, and organizes them into a AMResources struct, which is then returned.
-func Load(fs afero.Fs, rootPath string) (*AMResources, error) {
-	resources := &AMResources{
-		Policies: make(map[string]Policy),
-		Groups:   make(map[string]Group),
-		Users:    make(map[string]User),
+func Load(fs afero.Fs, rootPath string) (*account.AMResources, error) {
```
UnseenWizzard, line 32: note that CI flags this method's cyclomatic complexity once again after this commit moved it. ignore that check

pkg/persistence/account/integration_test.go
```diff
@@ -0,0 +1,131 @@
+/*
```
Laubi, line 1: `//go:build unit` missing.

pkg/persistence/account/loader/load_test.go
Laubi, line 1: `//go:build unit` missing.

pkg/persistence/account/loader/validate_test.go
Laubi, line 1: `//go:build unit` missing.

pkg/persistence/account/types.go
Laubi, line 1: `GitHub Actions / Verify` fails (unused), but the GitHub check is marked as no issue. 
@UnseenWizzard, any Idea why?

pkg/persistence/account/writer/writer.go
```diff
@@ -0,0 +1,169 @@
...
+	"path/filepath"
+)
+
+type Context struct {
+	Fs            afero.Fs
+	OutputFolder  string
+	ProjectFolder string
+}
+
+func WriteAccountResources(writerContext Context, resources account.Resources) error {
```
Laubi, line 35:
```suggestion
func Write(writerContext Context, resources account.Resources) error {
```

I'd go with the simple wording. The import will probably called `accountWriter`, then the call would be `accountWriter.Write()`. IIRC the other loads have the same structure as well.

pkg/persistence/account/writer/writer.go
```diff
@@ -0,0 +1,169 @@
...
+	"fmt"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/account"
+	persistence "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/persistence/account"
+	"github.com/spf13/afero"
+	"gopkg.in/yaml.v2"
+	"path/filepath"
+)
+
+type Context struct {
```
Laubi, line 29: Please add documentation to the public functions.
E.g., is OutputFolder already the abs, or the relative one? Depending on that information, L37 could be wrong (as the afero.Fs path is not used).

pkg/persistence/account/writer/writer.go
```diff
@@ -0,0 +1,169 @@
...
+	if err := createFolderIfNoneExists(writerContext.Fs, projectFolder); err != nil {
+		return err
+	}
+
+	var errOccurred bool
+	if len(resources.Policies) > 0 {
+		policies := toPersistencePolicies(resources.Policies)
+		if err := persistToFile(policies, writerContext.Fs, filepath.Join(projectFolder, "policies.yaml")); err != nil {
+			errOccurred = true
+			log.Error("failed to persist policies: %w", err)
```
Laubi, line 54:
```suggestion
			log.Error("Failed to persist policies: %w", err)
```

Logging should start with uppercase. Same with the loglines below

pkg/persistence/account/writer/writer.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	if errOccurred {
+		return fmt.Errorf("failed to persist some account resources to folder %q", projectFolder)
+	}
+	return nil
+}
+
+func toPersistencePolicies(policies map[string]account.Policy) persistence.Policies {
+	out := make([]persistence.Policy, 0)
```
Laubi, line 81: Personally, I would preallocate it :) 
Does not matter as writing it takes way longer in this context, just a personal preference.

pkg/persistence/account/writer/writer_test.go
```diff
@@ -0,0 +1,398 @@
+/*
```
Laubi, line 1: `//go:build unit` missing.

Laubi (changes requested):
LGTM on a first glance, minor changes requested :)

pkg/persistence/account/types.go
UnseenWizzard: What do you mean by Verify fails? I can see no output of it failing and all but it's golangcilint run pass locally as well. What am I missing? 

If it's specifically the golangcilint check we run via reviewdog, that just seems to get attached to the wrong check: https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1262/checks?check_run_id=18564194587
(attached to the semgrep Action, while it actually started from Verify)
I don't know why that is, it might just attach the check to the last action running...

pkg/persistence/account/writer/writer.go
UnseenWizzard: > E.g., is OutputFolder already the abs, or the relative one? Depending on that information, L37 could be wrong (as the afero.Fs path is not used).

100% agree on lacking documentation, but I don't fully understand this question here.

If OutputFolder is relative, it will become absolute, if it is absolute already a `filepath.Abs` does nothing to it.

But what do you mean by:

> as the afero.Fs path is not used

The Fs is used - what do you mean by the "Fs path"?

pkg/persistence/account/writer/writer.go
UnseenWizzard: Thanks! Left over fmt.Errorf text from refactoring an initial version that returned on the first error instead of logging & carrying on.

UnseenWizzard:
@Laubi I've addressed most of your comments (the log capitalization via fixup, all others as additional commits) - but have left two where I have open questions unresolved

pkg/persistence/account/writer/writer.go
UnseenWizzard: documentation added, but I'm still not sure I understand what concern you have about the "Fs path" - lmk internally if we should discuss this on a quick call

Laubi: >edit: It this afero Fs has a basepath overlay/is a BasePathFs it will append that before the absolute path and create something weird yes (but that's the point of a basepath FS). Is that what you meant?

Yes, that's exactly what I meant. The abs method uses the path of the executable, while the afero basepath is something different. Could lead to some errors depending on the callee code.

pkg/persistence/account/types.go
Laubi: image
This error

UnseenWizzard: Huh... interesting! No clue how that gets lost on the actual check status! I will dig into this a bit after this PR if that's fine by you.
And leave the unused type in place for now, as the account types are IMO still subject to change between everyone's changes/upcoming PRs.

pkg/persistence/account/writer/writer.go
UnseenWizzard: Ok, then I got what you mean. 

However: We directly use filepath.Abs in several places that deal with files already - and afero's basepath Fs in very few other cases (convert, manifest load and the new account E2E test). 
(46 vs 5 usage search hits)

I think this is well worth discussing - e.g. we might want to just build fitting basepath filesystems at the very top of commands and not care about relative or absolute paths in most code. 
But for this piece of code I would for now lean towards keeping the direct use of filepath.Abs like most other places loading files, even though it might cause issues for basepath filesystems. 

IMO we should

Laubi (approved):

PR #1266, opened 2023-11-15
feat: deploy account management resources

warber:


cmd/monaco/account/deploy.go
```diff
@@ -107,23 +114,39 @@ func deploy(fs afero.Fs, opts deployOpts) error {
...
 	if len(errs) > 0 {
 		errutils.PrintErrors(errs)
 		return errors.New("failed to load all account management resources")
 	}
 
-	err := deployer.Deploy(accs, resources, deployer.Options{DryRun: opts.dryRun})
-	return err
+	accountClients, err := createAccountClients(accs)
+	if err != nil {
+		return fmt.Errorf("failed to creae account clients: %w", err)
```
UnseenWizzard, line 126: "create"

cmd/monaco/account/deploy.go
```diff
@@ -107,23 +114,39 @@ func deploy(fs afero.Fs, opts deployOpts) error {
...
+	}
+
+	supportedPermissions, err := deployer.DefaultPermissionProvider()
```
UnseenWizzard, line 129: nit: it's a bit weird that the caller needs to know and explicitly call the default permissions. IMO might be nicer to do this as part of the deployer/client initalization. Assuming you're only overwriting them for testing anyway, maybe you can just initialize the client directly for tests?

But at a second glance I also realize you query them once, but build several clients.. In a real scenario I'd expect most customers to not really have more than one account, so maybe we could live with the inefficiency - or cache as part of a DefaultProvider that's not just a function

cmd/monaco/account/deploy.go
```diff
@@ -132,3 +155,30 @@ func loadAccountManagementResources(fs afero.Fs, projs manifest.ProjectDefinitio
...
+	for _, acc := range manifestAccounts {
+		oauthCreds := clientcredentials.Config{
+			ClientID:     acc.OAuth.ClientID.Value.Value(),
+			ClientSecret: acc.OAuth.ClientSecret.Value.Value(),
+			TokenURL:     acc.OAuth.GetTokenEndpointValue(),
+		}
+
+		var apiUrl string
+		if acc.ApiUrl == nil || acc.ApiUrl.Value == "" {
+			apiUrl = "https://api-dev.internal.dynatracelabs.com"
```
UnseenWizzard, line 171: `api.dynatrace.com` !

cmd/monaco/account/deploy.go
```diff
@@ -132,3 +155,30 @@ func loadAccountManagementResources(fs afero.Fs, projs manifest.ProjectDefinitio
...
+		} else {
+			apiUrl = acc.ApiUrl.Value
+		}
+		accClient, err := clients.Factory().WithOAuthCredentials(oauthCreds).AccountClient(apiUrl)
```
UnseenWizzard: Add the monaco user-agent string to this client please. (I forgot that for delete as well... as it's merged could you maybe create and reuse a helper method for creating the client?)

pkg/account/deployer/client.go
```diff
@@ -0,0 +1,268 @@
...
+		supportedPermissions: supportedPermissions,
+	}
+}
+
+func (d *accountManagementClient) getAccountInfo() AccountInfo {
+	return d.accountInfo
+}
+
+func (d *accountManagementClient) getGlobalPolicies(ctx context.Context) (map[string]remoteId, error) {
+	globalPolicies, _, err := d.client.PolicyManagementAPI.GetLevelPolicies(ctx, "global", "global").Execute()
```
UnseenWizzard, line 57: is there no need for the full error check that looks at the error as well as the response here?

pkg/account/deployer/client.go
```diff
@@ -0,0 +1,268 @@
...
+	_, resp, err = d.client.PolicyManagementAPI.UpdateLevelPolicy(ctx, policyLevel, policyLevelId, existingPolicies[0].GetUuid()).CreateOrUpdateLevelPolicyRequestDto(policy).Execute()
+	if err = d.handleClientResponseError(resp, err, "unable to update policy with name: "+policy.Name); err != nil {
+		return "", err
+	}
+	return existingPolicies[0].GetUuid(), nil
+}
+
+func (d *accountManagementClient) upsertGroup(ctx context.Context, groupId string, group Group) (remoteId, error) {
+	if groupId != "" {
+		log.Debug("Trying to update policy with origin object ID (UUID) %q", groupId)
```
UnseenWizzard, line 118: `group`, not `policy`. Also  on line 121

pkg/account/deployer/client.go
```diff
@@ -0,0 +1,268 @@
...
+		resp, err := d.client.GroupManagementAPI.EditGroup(ctx, d.accountInfo.AccountUUID, groupId).PutGroupDto(group).Execute()
+
+		if err = d.handleClientResponseError(resp, err, "unable to update policy with UUID: "+groupId); err != nil {
+			return "", err
+		}
+		return groupId, nil
+	}
+
+	result, resp, err := d.client.GroupManagementAPI.GetGroups(ctx, d.accountInfo.AccountUUID).Execute()
+	if err = d.handleClientResponseError(resp, err, "unable to get groups with name: "+group.Name); err != nil {
```
UnseenWizzard, line 128: `group` not `groups` ?

pkg/account/deployer/client.go
```diff
@@ -0,0 +1,268 @@
...
+			return "", err
+		}
+		if len(createdGroups) < 1 {
+			return "", fmt.Errorf("unable to get UUID of created group with name: %s", group.Name)
+		}
+		return createdGroups[0].GetUuid(), nil
+	}
+
+	if len(existingGroups) > 1 {
+		log.Warn("Updating multiple policies with name %s", group.Name)
```
UnseenWizzard, line 153: I think this is leftover wording from different logic? Like policies you're only updating the first. And I do think group names can't be duplicated right so this is also a "should never happen" case, right?

pkg/account/deployer/client.go
```diff
@@ -0,0 +1,268 @@
...
+
+	resp, err = d.client.GroupManagementAPI.EditGroup(ctx, d.accountInfo.AccountUUID, groupToUpdate.GetUuid()).PutGroupDto(group).Execute()
+	if err = d.handleClientResponseError(resp, err, "unable to update group with name: "+group.Name); err != nil {
+		return "", err
+	}
+	return groupToUpdate.GetUuid(), nil
+}
+
+func (d *accountManagementClient) upsertUser(ctx context.Context, userId string) (remoteId, error) {
+	users, resp, err := d.client.UserManagementAPI.GetUsers(ctx, d.accountInfo.AccountUUID).Execute()
```
UnseenWizzard, line 166: How does the API behave if we just try to create a user that already exist? I guess we can't differentiate an actual error from "already exist" and thus need to do the GET first?

As the email is the ID we could just query for the exact user in that case though, instead of getting all and iterating :)

pkg/account/deployer/client.go
```diff
@@ -0,0 +1,268 @@
...
+	return userId, nil
+}
+
+func (d *accountManagementClient) updatePermissions(ctx context.Context, groupId string, permissions []accountmanagement.PermissionsDto) error {
+	if groupId == "" {
+		return fmt.Errorf("group id must not be empty")
+	}
+
+	if len(permissions) == 0 {
+		return nil
```
UnseenWizzard, line 190: Shouldn't we treat that as an error as well? When would this be correctly called with no permissions?

pkg/account/deployer/client.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	return nil
+}
+
+func (d *accountManagementClient) updateAccountPolicyBindings(ctx context.Context, groupId string, policyIds []string) error {
+	if groupId == "" {
+		return fmt.Errorf("group id must not be empty")
+	}
+	if len(policyIds) == 0 {
```
UnseenWizzard, line 210: same as for `permissions` (and then again for line 230 Env policies and for groups) - is really not a usage error to call these with empty slices?

pkg/account/deployer/deployer.go
```diff
@@ -17,15 +17,288 @@
...
+	}
+
+	d.deployedMgmtZones, err = d.accountManagementClient.getManagementZones(context.TODO())
+	if err != nil {
+		return err
+	}
+
+	for _, res := range resources {
+		for _, policy := range res.Policies {
+			log.Info("Deploying policy %s to account %s...", policy.Name, d.accountManagementClient.getAccountInfo().Name)
```
UnseenWizzard, line 95: I agree with the cognitive complexity linter, IMO this would be easier to read if you extracted either `deployPolicies(res.Policies)` or `deployPolicy(policy)` (no messing with the ID/uuid map then) functions (same for all types of course).

pkg/account/deployer/deployer.go
```diff
@@ -17,15 +17,288 @@
...
+	if err != nil {
+		return err
+	}
+
+	d.deployedMgmtZones, err = d.accountManagementClient.getManagementZones(context.TODO())
+	if err != nil {
+		return err
+	}
+
+	for _, res := range resources {
```
UnseenWizzard, line 93: Looking at this usage: Does it actually have any benefit to have the project -> resources mapping here?
If I'm not missing anything you're just looping over the resources and never even use the project name for anything. 

If you'd build a single combined account.Resources object when loading the projects, this deployer could simply deploy all policies, groups and users it got. 🤔

Also in that case we could flag user errors, like defining the same group twice in different projects, which I think we're missing right now.

pkg/account/deployer/client.go
warber: don't have a strong opinion on that.. I just thought, if you give it no permission (which may happen or not) it does nothing. if you count that operation as an error or not is arguable. should i change it to err?

pkg/account/deployer/deployer.go
warber: yes, just thought calling am method and putting the returned id into a map isn't that of a complexity tbh. but i can extract it int methods . no problem

pkg/account/deployer/client.go
warber: yep

pkg/account/deployer/client.go
warber: agreed to keep it like it is

pkg/account/deployer/client.go
warber: it was done like that under the assumption that the user ID is a UUDI or so. :D 
but it's the email address, so you are right

warber: but on the other hand. the api returns a 400 BadRequest err code (instead of 409 Conflict) so I cannot really count on that.. also there is no possibility for querying exactly one user. :/
so i guess i will keep it like it is...

cmd/monaco/account/deploy.go
warber: I've already done that today in the morning :) i will use that also in your code

pkg/account/deployer/deployer.go
warber: lets think about that in another PR?

cmd/monaco/account/deploy.go
warber: Let possibly target this in another PR.

cmd/monaco/account/deploy.go
```diff
@@ -121,9 +121,13 @@ func deploy(fs afero.Fs, opts deployOpts) error {
 		return errors.New("failed to load all account management resources")
 	}
 
+	if opts.dryRun {
+		return nil
```
UnseenWizzard, line 125: I think it'd make sense to log something similar to the config dry-run. E.g. "Successfully validated account resources"

pkg/account/deployer/deployer.go
UnseenWizzard: fine by me 👍 

As discussed directly, it might also make sense to just modify the persistence/account/loader to target several folders instead of a single one, as it already does this level of validation and builds a single struct for everything it finds.

cmd/monaco/account/deploy.go
UnseenWizzard: this is still open I think?

pkg/account/deployer/client.go
UnseenWizzard: There is [/iam/v1/accounts/{accountUuid}/users/{email}](https://api.dynatrace.com/spec/#/User%20management/UsersController_getUserGroups) - which is described as "list groups of a user" but behaves as the resource path suggests and returns the user object or 404 if the user does not exist.

We could call that and simply check if it returned a success code or 404

UnseenWizzard (dismissed):
Looks good to me 🚀 
I've merged the loader changes and it looks like you'll have to manually rebase to main and drop my two older commits that you've got on this branch.

UnseenWizzard (dismissed):

UnseenWizzard (approved):

PR #1268, opened 2023-11-20
Chore: Hardening of Account Management Deployment

warber:
This PR fixes a couple of issues for account management deployment:
* fix transformation of persistence model to in memory model to also consider types of references and policy levels
* fix loading of resources accross multiple project to also consider duplicates
* download all available groups to handle references to builtin groups
* enhance error handling a bit
* fix account mangement resources in e2e tests
* introduced Ref type in in-memory model


pkg/account/deployer/client.go
```diff
@@ -266,7 +280,7 @@ func (d *accountManagementClient) updateGroupBindings(ctx context.Context, userI
 	if len(groupIds) == 0 {
 		return nil
 	}
-	resp, err := d.client.UserManagementAPI.AddUserToGroups(ctx, d.accountInfo.AccountUUID, userId).RequestBody(groupIds).Execute()
+	resp, err := d.client.UserManagementAPI.ReplaceUserGroups(ctx, d.accountInfo.AccountUUID, userId).RequestBody(groupIds).Execute()
```
warber, line 283: changed to ReplaceUserGroups in order to also cover the "removal" of groups

pkg/account/deployer/client.go
```diff
@@ -213,7 +227,7 @@ func (d *accountManagementClient) updatePermissions(ctx context.Context, groupId
 			return fmt.Errorf("unsupported permission %q. Must be one of: %v", p.PermissionName, d.supportedPermissions)
 		}
 	}
-	resp, err := d.client.PermissionManagementAPI.AddGroupPermissions(ctx, d.accountInfo.AccountUUID, groupId).PermissionsDto(permissions).Execute()
+	resp, err := d.client.PermissionManagementAPI.OverwriteGroupPermissions(ctx, d.accountInfo.AccountUUID, groupId).PermissionsDto(permissions).Execute()
```
warber, line 230: changed to `OverwriteGroupPermissions` to also cover the removal of permissions.

pkg/account/deployer/client.go
```diff
@@ -67,6 +67,20 @@ func (d *accountManagementClient) getGlobalPolicies(ctx context.Context) (map[st
...
 
+func (d *accountManagementClient) getAllGroups(ctx context.Context) (map[string]remoteId, error) {
+	groups, resp, err := d.client.GroupManagementAPI.GetGroups(ctx, d.accountInfo.AccountUUID).Execute()
+	defer closeResponseBody(resp)
+	if err = d.handleClientResponseError(resp, err, "unable get all groups for account "+d.accountInfo.AccountUUID); err != nil {
+		return nil, err
+	}
+	result := make(map[string]remoteId)
+	for _, glP := range groups.GetItems() {
+		result[glP.Name] = glP.GetUuid()
```
Laubi, line 78: Any reason to use the getter on one site but the property on the other? (glP.Name vs getUuid). Property access should be enough, right?

pkg/persistence/account/loader/load.go
```diff
@@ -207,6 +207,32 @@ func decode(in io.ReadCloser) (map[string]any, error) {
 }
 
 func transform(resources *persistence.Resources) *account.Resources {
```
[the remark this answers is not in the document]
warber: i think the benefit here is the possibility to return nil. whereas otherwise you would need to construct a account.Resources value prefferably with valid (non nil) fields.

pkg/account/deployer/client.go
warber: Uuid is of type `*string`

pkg/persistence/account/loader/load.go
```diff
@@ -207,6 +207,32 @@ func decode(in io.ReadCloser) (map[string]any, error) {
...
+		default:
+			panic("unable to convert persistence model")
+		}
+	}
+
+	transformRefs := func(in []any) []account.Ref {
+		var res []account.Ref
+		for _, el := range in {
+			switch v := el.(type) {
+			case persistence.Reference:
```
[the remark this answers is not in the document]
warber: agreed with @Laubi  to add this later (maybe)

Laubi (approved):

PR #1271, opened 2023-11-27
chore: Account Management e2e tests

warber:
This PR enhances the current e2e tests for account management.
It covers
* deploying a user, a group and a policy as well as bindings between them.
* asserting that resources were correctly deployed
* deleting the deployed user, group and policy
* asserting that resources were indeed deleted

### DIFF
--- a/cmd/monaco/account/delete.go
+++ b/cmd/monaco/account/delete.go
@@ -103,19 +103,15 @@ func loadManifest(fs afero.Fs, manifestName string) (manifest.Manifest, error) {
 	}
 	// Sanitize manifest file path to manifest yaml file
 	manifestName = filepath.Clean(manifestName)
-	absManifestFilePath, err := filepath.Abs(manifestName)
-	if err != nil {
-		return manifest.Manifest{}, err
-	}
 
 	// Try to load the manifest file
 	m, errs := manifestloader.Load(&manifestloader.Context{
 		Fs:           fs,
-		ManifestPath: absManifestFilePath,
+		ManifestPath: manifestName,
 	})
 	if len(errs) > 0 {
 		errutils.PrintErrors(errs)
-		return manifest.Manifest{}, fmt.Errorf("error while loading manifest (%s)", absManifestFilePath)
+		return manifest.Manifest{}, fmt.Errorf("error while loading manifest (%s)", manifestName)
 	}
 	return m, nil
 }
--- a/cmd/monaco/account/deploy.go
+++ b/cmd/monaco/account/deploy.go
@@ -122,7 +122,7 @@ func deploy(fs afero.Fs, opts deployOpts) error {
 		return nil
 	}
 
-	accountClients, err := createAccountClients(accs)
+	accountClients, err := CreateAccountClients(accs)
 	if err != nil {
 		return fmt.Errorf("failed to create account clients: %w", err)
 	}
@@ -142,7 +142,7 @@ func deploy(fs afero.Fs, opts deployOpts) error {
 	return nil
 }
 
-func createAccountClients(manifestAccounts map[string]manifest.Account) (map[deployer.AccountInfo]*accounts.Client, error) {
+func CreateAccountClients(manifestAccounts map[string]manifest.Account) (map[deployer.AccountInfo]*accounts.Client, error) {
 	accClients := make(map[deployer.AccountInfo]*accounts.Client, len(manifestAccounts))
 	for _, acc := range manifestAccounts {
 		oauthCreds := clientcredentials.Config{
--- a/cmd/monaco/account/deploy.go
+++ b/cmd/monaco/account/deploy.go
@@ -19,21 +19,17 @@ package account
 import (
 	"errors"
 	"fmt"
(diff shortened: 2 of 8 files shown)


cmd/monaco/account/delete.go
```diff
@@ -103,19 +103,15 @@ func loadManifest(fs afero.Fs, manifestName string) (manifest.Manifest, error) {
 	}
 	// Sanitize manifest file path to manifest yaml file
 	manifestName = filepath.Clean(manifestName)
-	absManifestFilePath, err := filepath.Abs(manifestName)
```
warber, line 106: this just causes problems in test context with e.g. afero.
I've deleted it, as i think (?) it is not really necessary, please confirm

cmd/monaco/account/deploy.go
```diff
@@ -142,7 +142,7 @@ func deploy(fs afero.Fs, opts deployOpts) error {
 	return nil
 }
 
-func createAccountClients(manifestAccounts map[string]manifest.Account) (map[deployer.AccountInfo]*accounts.Client, error) {
+func CreateAccountClients(manifestAccounts map[string]manifest.Account) (map[deployer.AccountInfo]*accounts.Client, error) {
```
warber, line 145: this func might need another home, for now i've just left it where it is (cmd package)

pkg/account/delete/loader.go
```diff
@@ -64,19 +64,15 @@ func LoadResourcesToDelete(fs afero.Fs, deleteFile string) (Resources, error) {
 }
 
 func readDeleteFile(context *loaderContext) (FileDefinition, error) {
-	targetFile, err := filepath.Abs(context.deleteFile)
```
warber, line 67: this just causes problems in test context with e.g. afero.
I've deleted it, as i think (?) it is not really necessary, please confirm

cmd/monaco/account/deploy.go
UnseenWizzard: Weren't there some similar utils in a `cmd/monaco/dynatrace` package at some point? Might fit there

cmd/monaco/integrationtest/account/users_test.go
```diff
@@ -19,30 +19,235 @@
...
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/account/deployer"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
+	"net/http"
+	"slices"
+	"strings"
 	"testing"
 )
 
 func TestUsers(t *testing.T) {
```
UnseenWizzard, line 36: Can we rename this to something like "all_resources" on file and test level? (Random name based on all_configs)
This is currently the only test for all account resources, and not a test for configuring users.

cmd/monaco/integrationtest/account/users_test.go
```diff
@@ -19,30 +19,235 @@
...
+	assertElementInSlice(t, permissions.Permissions, func(el accountmanagement.PermissionsDto) bool {
+		permissionFound := el.PermissionName == permissionName
+		scopeTypeEqual := el.ScopeType == scopeType
+		scopeEqual := el.Scope == scope
+		return permissionFound && scopeTypeEqual && scopeEqual
+	})
+}
+
+func (a AccountResourceChecker) suffix(in string) string {
+	return strings.ReplaceAll(in, "$SUFFIX$", a.TestSuffix)
```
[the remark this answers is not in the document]
warber: not all resources require the suffix, e.g. default ones dont
but yeah, reusing the logic for the suffix is always better ... will do

UnseenWizzard (commented):
Basically all my comments can be viewed as nitpicks. Some might make the test-util/runner nicer to keep using for future tests. 

Please make sure you double check that using relative paths has no negative impacts on using monaco outside of the E2E tests and their setup of MemMapFs overlayed on OsFs - running account deploy & delete from the folder the files are in and from further up. Try some weird constructs like` monaco account delete -m some/path/to/manifest.yaml --file ../../delete/file/is/somewhere/else/delete.yaml` as well please

cmd/monaco/integrationtest/account/users_test.go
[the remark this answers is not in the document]
warber: don't know, but they can be referenced in bindings

cmd/monaco/integrationtest/account/runner.go
```diff
@@ -33,8 +37,9 @@ import (
 )
 
 type options struct {
-	fs     afero.Fs
-	suffix string
+	fs             afero.Fs
+	suffix         string
+	accountClients map[deployer.AccountInfo]*accounts.Client
```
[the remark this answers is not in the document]
warber: yes, that's better you are right, will do

cmd/monaco/integrationtest/account/users_test.go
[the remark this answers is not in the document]
UnseenWizzard: IIRC the rewriting done by the util at the moment only touches the names of configurations that we create/update, not everything that's a name via replace-all like the config rewriting. 

But I'm happy with some reuse, doesn't have to be fully automatic

UnseenWizzard (approved):

PR #1272, opened 2023-11-27
feat(manifest): Allow defining account UUIDs as environment variables

UnseenWizzard:
In some cases it is desired to not use hardcoded account UUIDs so they are now
allowed to be loaded from environment variables.
For convenience, both a full shorthand of and ommitting
the type and defaulting to value is allowed, making all of th following
defintions valid:

```yaml
accountUUID: 1234
accountUUID:
  value: 1234
accountUUID:
  type: value
  value: 1234
accountUUID:
  type: environment
  value: ENV_VAR
```

First commit in PR unifies the dedicated "type" structs into a single one - I see no reason why this should not be reused.

No - fully backwards compatible as the current shorthand is retained for convenience. 
Also this is not yet released..


pkg/manifest/loader/account.go
```diff
@@ -69,14 +66,51 @@ func parseSingleAccount(c *Context, a persistence.Account) (manifest.Account, er
...
+		return u.Value, nil
+	}
+
+	if u.Type == persistence.TypeEnvironment {
+		val, found := os.LookupEnv(u.Value)
+		if !found {
+			return "", fmt.Errorf("environment variable %q could not be found", u.Value)
+		}
+		if val == "" {
+			return "", fmt.Errorf("environment variable %q is defined but has no value", u.Value)
```
UnseenWizzard, line 106: not sure if we might want to use the `invalidUUIDError` here as well - semantically it seems to make sense, but IMO how it currently prints might look weird with an empty string

pkg/manifest/loader/account_test.go
```diff
@@ -63,36 +70,125 @@ func TestValidAccounts(t *testing.T) {
...
+	}
 
-	assert.Equal(t, v, map[string]manifest.Account{
-		"name": {
-			Name:        "name",
-			AccountUUID: uuid.MustParse(acc.AccountUUID),
-			ApiUrl: &manifest.URLDefinition{
-				Type:  manifest.ValueURLType,
-				Value: "https://example.com",
+	t.Run("full account", func(t *testing.T) {
```
UnseenWizzard, line 93: note: I found it nicer to have these organized as individual tests. I still keep the "load all" as a dedicated one.

pkg/manifest/internal/persistence/manifest_persitence.go
```diff
@@ -83,9 +84,33 @@ type Manifest struct {
...
 }
 
+type AccountUUID struct {
+	Type  Type   `yaml:"type,omitempty" mapstructure:"type"`
+	Value string `yaml:"value" mapstructure:"value"`
+}
+
+// UnmarshalYAML Custom unmarshaler for AccountUUID able to parse simple shorthands (accountUUID: 1234) and full values.
+func (c *AccountUUID) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var data interface{}
```
Laubi, line 94: Huge nitpick and personal optiniom: `any` is nicer than `interface{}`

UnseenWizzard: I honestly have no feelings either way.. `any` is nice syntactic sugar, but if there's nothing else to change in this PR, I'd leave this as is.

warber (approved):

Laubi (dismissed):

Laubi (approved):

PR #1275, opened 2023-11-28
chore: simplified randomization of am resources in e2e tests

warber:
Instead of loading resources and cherry picking what fields to change/randomize in code, the `RunAccountTestCase` func now just loops over all `yaml` files it can find and replaces potential `%RAND%` placeholders with a randomized string.

Advantages: 
* more flexibility of what shall be randomized
*  simpler code (strings.ReplaceAll)
* no need for duplicate the am resource model for e2e tests

Disadvantage:
* easier to "forget" putting the placeholders into the yaml files


UnseenWizzard (approved):
looks reasonable 👍 
much simpler code, and I think account test resources will not grow as much as the general config ones. which makes it less likely that we forgot to add the suffix to things.

PR #1278, opened 2023-11-29
Download of account configuration

jskelin:
This is the preliminary implementation of the `account download` command. The main functionality is in place.

The logging of errors is badly implemented - it will be added separately.
There is a lack of tests.  For that  `internal/http` package needs to be separated with an interface from the core part of the code (separate PR)


cmd/monaco/account/download2.go
```diff
@@ -0,0 +1,146 @@
...
+		if err != nil {
+			log.Error("Configuration download for account %q failed! Cause: %s", acc, err)
+			failedDownloads = append(failedDownloads, acc)
+		}
+	}
+
+	if len(failedDownloads) > 0 {
+		var es []string
+		for _, t := range failedDownloads {
+			log.Debug("Failed to download account %q (UUID: %q)", t.Name, t.AccountUUID)
```
UnseenWizzard, line 107: If this is meant to be a summary I'd log them as Error. Collecting things for debug logging seems of limited use to me

cmd/monaco/account/download2.go
```diff
@@ -0,0 +1,146 @@
...
+	manifestloader "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest/loader"
+	presistance "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/persistence/account/writer"
+	"github.com/google/uuid"
+	"github.com/spf13/afero"
+	"os"
+)
+
+func downloadAll(fs afero.Fs, opts downloadOpts) error {
+	if opts.outputFolder == "" {
+		opts.outputFolder = "project/accounts" //TODO: make output folder unique (Where to prevent overwriting - write/load module or here?)
```
UnseenWizzard, line 39: IMO it might be easiest to check if the default folder exists and append a timestamp here, rather than pulling that down into the writer.

pkg/account/downloader/groups.go
```diff
@@ -0,0 +1,197 @@
...
+	levelID = string
+	group   struct {
+		group         *account.Group
+		dto           *accountmanagement.GetGroupDto
+		permissionDTO *accountmanagement.PermissionsGroupDto
+		bindings      map[levelID]*accountmanagement.LevelPolicyBindingDto
+	}
+)
+
+func (a *Account) Groups(policies Policies, tenants Environments) (Groups, error) {
```
UnseenWizzard, line 40: Are these exported methods a leftover or does my IDE wrongly not find them used?
It looks like "Account.DownloadConfiguration" is the only entry point that is being used?

pkg/account/downloader/internal/http/http_utils.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"net/http"
+)
+
+func closeResponseBody(resp *http.Response) {
+	_ = resp.Body.Close()
+}
+
+func handleClientResponseError(resp *http.Response, clientErr error) error {
+	return clientErr
```
UnseenWizzard, line 28: Currently, this won't treat HTTP errors as an error. I don't think that's how the downloader should behave? I can't see any checks for 4xx/5xx status errors anywhere else in the code. There's more than the 404 exclusion to the TODO below unless I'm mistaken

pkg/account/info.go
```diff
@@ -0,0 +1,29 @@
...
+
+package account
+
+type AccountInfo struct {
+	Name        string
+	AccountUUID string
+}
+
+func (a AccountInfo) String() string {
+	if len(a.Name) > 0 {
```
UnseenWizzard, line 25: nit: `a.Name == ""` feels like the more idiomatic way to check if a string is empty. 

also: should we maybe log it as "name (uuid)" if both are set, rather than just the name?

cmd/monaco/account/download2.go
```diff
@@ -0,0 +1,146 @@
...
+		Fs:            fs,
+		OutputFolder:  opts.outputFolder,
+		ProjectFolder: accInfo.String(),
+	}
+	err = presistance.Write(c, *resources)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(resources)
```
warber, line 134: pls delete

pkg/account/downloader/internal/http/tenant.go
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package http
+
+import (
+	"context"
+	accountmanagement "github.com/dynatrace/dynatrace-configuration-as-code-core/gen/account_management"
+)
+
+func (c *Client) GetTenants(ctx context.Context, account string) (*accountmanagement.EnvironmentResourceDto, error) {
```
warber, line 24: Let's just name this `GetEnvironments`The term "tenant" is nowhere used in monaco... usually it's just "environment" also the generated client code speaks about "environments"

warber:
general question, shouldn't the download appear in a folder named "download_<timestamp>", similar to the regular config download?

cmd/monaco/account/download2.go
warber, line 1: download2.go?

cmd/monaco/account/download2.go
```diff
@@ -0,0 +1,146 @@
...
+			return errors.New("error while loading manifest")
+		}
+
+		if len(opts.accountName) > 0 {
+			for _, a := range opts.accountName {
+				if n, ok := m.Accounts[a]; !ok {
+					return fmt.Errorf("unknown enviroment %q", n.Name)
+				}
+			}
+			for _, a := range opts.accountName {
```
warber, line 59: why loop twice?

pkg/account/downloader/environments.go
```diff
@@ -0,0 +1,55 @@
...
+		name     string
+		originID string
+	}
+)
+
+func (e environment) String() string {
+	return e.id
+}
+
+func (e Environments) asList() []string {
```
warber, line 38: this is just used in a debug message write?
maybe what you want is a String() method for Environments type?

cmd/monaco/account/download.go
```diff
@@ -49,12 +48,12 @@ func downloadCommand(fs afero.Fs) *cobra.Command {
...
 		PreRun:            cmdutils.SilenceUsageCommand(),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return download(fs, opts)
+			return downloadAll(fs, opts)
 		},
 	}
 
 	cmd.Flags().StringVarP(&opts.manifestName, "manifest", "m", "manifest.yaml", "Name (and the path) to the manifest file. Defaults to 'manifest.yaml'")
-	cmd.Flags().StringVarP(&opts.accountName, "account", "a", "", "Account name defined in the manifest to download from")
+	cmd.Flags().StringSliceVarP(&opts.accountName, "account", "a", []string{}, "Account name defined in the manifest to download from")
```
warber, line 56: please adjust variable name and description accordingly....

cmd/monaco/account/download2.go
jskelin: It is just for UUID (which is only printed in debug-level messages). But we can live without this.

pkg/account/downloader/internal/http/http_utils.go
jskelin: For now, I am going to remove this part.
As I `download` is always getting the whole configuration, 404 is also an error in processing if some requested resources are absent.

There will be additional PR in which will this case be present and reused in delete/deploy

cmd/monaco/account/download2.go
jskelin: Just wanted to split the logic from the definition of the comment. Going to merge it.

jskelin: But not in this PR :) The file becomes overcrowded

pkg/account/downloader/account.go
```diff
@@ -0,0 +1,65 @@
...
+
+package downloader
+
+import (
+	"context"
+	accountmanagement "github.com/dynatrace/dynatrace-configuration-as-code-core/gen/account_management"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+)
+
+func (a *Account) Environments() (Environments, error) {
```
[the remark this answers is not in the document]
jskelin: All exported unused functions are removed

cmd/monaco/account/download.go
```diff
@@ -28,7 +27,7 @@ import (
 type downloadOpts struct {
 	auth
 	manifestName   string
-	accountName    string
+	accountName    []string
```
jskelin, line 30: rename to accountSlice

cmd/monaco/account/download2.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+func downloadAll(fs afero.Fs, opts *downloadOpts) error {
+	if opts.outputFolder == "" {
+		opts.outputFolder = "project/accounts"
+	}
+	if exists, err := afero.DirExists(fs, opts.outputFolder); err != nil {
+		return err
+	} else if exists {
+		opts.outputFolder = fmt.Sprintf("%s_%s", opts.outputFolder, time.Now().UTC().Format("20060102150405"))
```
jskelin, line 45: use timeutils.TimeAnchor() and log.LogFileTimestampPrefixFormat for formating

pkg/account/downloader/resource.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/accounts"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/account"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/account/downloader/internal/http"
+)
+
+type Account struct {
+	httpClient  *http.Client
+	accountInfo *account.AccountInfo
+}
```
jskelin, line 30: rename to `Downloader`!

warber (approved):

UnseenWizzard (approved):

PR #1279, opened 2023-11-29
fix: account management hardening pt.2

warber:
This PR fixes existing issues in current account management deployment. Primarily: 
* deleting all policy & permission bindings that were previously deployed

no, not released yet

### DIFF
--- a/pkg/account/deployer/client.go
+++ b/pkg/account/deployer/client.go
@@ -219,7 +219,7 @@ func (d *accountManagementClient) updatePermissions(ctx context.Context, groupId
 	}
 
 	if permissions == nil {
-		return nil
+		permissions = []accountmanagement.PermissionsDto{}
 	}
 
 	for _, p := range permissions {
@@ -241,7 +241,7 @@ func (d *accountManagementClient) updateAccountPolicyBindings(ctx context.Contex
 		return fmt.Errorf("group id must not be empty")
 	}
 	if policyIds == nil {
-		return nil
+		policyIds = []string{}
 	}
 	data := accountmanagement.PolicyUuidsDto{PolicyUuids: policyIds}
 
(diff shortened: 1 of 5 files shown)


pkg/account/deployer/client.go
```diff
@@ -240,7 +240,7 @@ func (d *accountManagementClient) updateAccountPolicyBindings(ctx context.Contex
 	if groupId == "" {
 		return fmt.Errorf("group id must not be empty")
 	}
-	if len(policyIds) == 0 {
+	if policyIds == nil {
```
warber, line 243: I was thinking of treating policyIds == nil the same as policIds == empty first, but then decided to just do nothing.... no strong feeling.. wdyt?

pkg/account/deployer/client.go
```diff
@@ -218,7 +218,7 @@ func (d *accountManagementClient) updatePermissions(ctx context.Context, groupId
 		return fmt.Errorf("group id must not be empty")
 	}
 
-	if len(permissions) == 0 {
```
warber, line 221: basically, this check prevented the flow from continuing, resulting in no updates happening when there are no references at all. 
same for other occurences.

pkg/account/deployer/client.go
UnseenWizzard: hmmm... can't think of an exact case where this would be called with an explicit nil list rather than an empty list. Did you think of one/thus handling them different actually make sense?
Gutfeeling: handling both as "remove bindings" would feel less prone to unintended behaviour in some edge cases we didn't think of yet

warber: yes, at the moment i fixed the code so that nil is never passed, but this could of course change in future when changing the code... will change to treat it like an empty list.

UnseenWizzard (approved):

PR #1280, opened 2023-11-29
fix: account resources persistence issues

UnseenWizzard:
Fixes the following issues: 

- **loader**: creating wrong defined environment policy passes without error
- **writer**:
  - for users without any group, group segment is present
  - write process panics for groups without policy

/

no, not released and just bug fixes

### DIFF
--- a/pkg/persistence/account/internal/types/types.go
+++ b/pkg/persistence/account/internal/types/types.go
@@ -42,7 +42,7 @@ type (
 		ID             string      `yaml:"id"`
 		Name           string      `yaml:"name"`
 		Level          PolicyLevel `yaml:"level"`
-		Description    string      `yaml:"description"`
+		Description    string      `yaml:"description,omitempty"`
 		Policy         string      `yaml:"policy"`
 		OriginObjectID string      `yaml:"originObjectId,omitempty"`
 	}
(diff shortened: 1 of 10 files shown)


pkg/persistence/account/internal/types/types.go
```diff
@@ -16,68 +16,101 @@
...
+
+const (
+	ReferenceType          = "reference"
+	PolicyLevelAccount     = "account"
+	PolicyLevelEnvironment = "environment"
+)
+
 type (
-	PolicyLevel = any // either PolicyLevelAccount or PolicyLevelEnvironment is allowed
-	PolicyRef   = any // either string or Reference is allowed
```
UnseenWizzard, line 21: note: I think we should reflect the change to explicit Level and Reference type(s) for the in-memory structs as well as a followup.

jskelin (approved):

jskelin (approved):

pkg/persistence/account/internal/types/types.go
warber: in memory types already have a reference type, if you mean that

warber:
lgtm, Nit: i think we can get rid of the two sentinel errors in errors.go as it is nowhere else used and/or needed. wdyt?

pkg/persistence/account/internal/types/types.go
UnseenWizzard: ah nice. Didn't look at them for a while

UnseenWizzard:
@warber the sentinel errors are used in a few test asserts.. generally agree we probably don't need them, but IMO they don't hurt to have?

PR #1283, opened 2023-11-30
fix: Fix settings update with overlapping ids

Laubi:

When updating a settings config, we try to update or onboard existing configs whenever the originObjectId is set.
However, there may be two objects, one with the expected externalId, and a different one with the expected originalObjectId. In this case, we update the setting with the externalId and print a proper warning.

### DIFF
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -219,7 +219,7 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
 		Filter: func(object DownloadSettingsObject) bool { return object.ExternalId == legacyExternalID },
 	})
 	if err != nil {
-		return DynatraceEntity{}, fmt.Errorf("unable to find Settings 2.0 object of schema %q with externalId %q: %w", obj.SchemaId, legacyExternalID, err)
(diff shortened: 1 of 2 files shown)


pkg/client/dtclient/settings_client.go
```diff
@@ -231,6 +231,32 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
...
+		var exIdSetting, ooIdSetting string
+		if settings[0].ExternalId == externalID {
+			exIdSetting = settings[0].ObjectId
+			ooIdSetting = settings[1].ObjectId
+		} else {
+			exIdSetting = settings[1].ObjectId
+			ooIdSetting = settings[0].ObjectId
+		}
+
+		log.WithCtxFields(ctx).Warn("Found two configs, one with the originObjectId (%q), and one with the externalId (%q). Updating the one with the externalId.", ooIdSetting, exIdSetting)
```
UnseenWizzard, line 256: I don't think it's fully clear to users what the "externalId" even is. 
I will rebase this PR and try to rephrase this directly,

UnseenWizzard (approved):
rebased and tried to improve the wording of the warning log a bit

PR #1289, opened 2023-12-06
fix: More correctly extract ids from configs

Laubi:
This PR fixes the first parts of the id extraction during download. There needs to be more improvements in the id_extraction, but to keep things simple, this PR introduces the first set of changes.

This is fixing the first set of issues. The id_extraction has similar issues, but another solution on how this can be done needs to be found.

The change has one minor caveat: It requires a character before and after the search string. An ID that is only the content can't be found. However, since we are always handling JSON contents which are  never (in our case) a plain number, we can dismiss this. We always have at least quotes around the object.

Yes: For large environments, different ids might be extracted due to being more correct;
* Some random strings like `HEY-ThisIsNotAnMeIDButIWasDetectedAsSuch` no longer are recognized as the ID `HEY-ThisIsNotAnMeIDB` but is skipped entirely.
* Ids that are substrings of other Ids are detected correctly:\
Given the two IDs `calc:metric` and calc:metric_new`. The config `"metric": "calc:metric_new" is no longer extracted as `"metric": "{{param}}_new"`, but as `"metric": "{param}"` with the correct ID.
* Ids that are complete substrings somewhere, are also no longer extracted from somewhere.\
Example the ID `1234` was found in this uuid: `c48504d9-085c-3e5d-9635-554fbcc12341` and extracted as `c48504d9-085c-3e5d-9635-554fbcc{{param}}1`. Instead, the ID should not be extracted, and if this not happens, the UUID is extracted correctly as its own parameter.
* See testcases for more examples. All of the test-cases are actual examples of what used to happen.

### DIFF
--- a/pkg/download/dependency_resolution/resolver/shared.go
+++ b/pkg/download/dependency_resolution/resolver/shared.go
@@ -23,6 +23,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter/reference"
 	valueParam "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter/value"
 	"regexp"
+	"strings"
 )
 
 func resolveScope(configToBeUpdated *config.Config, ids map[string]config.Config) {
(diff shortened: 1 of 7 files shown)


pkg/download/dependency_resolution/resolver/shared.go
```diff
@@ -61,3 +62,18 @@ var templatePattern = regexp.MustCompile(`[^a-zA-Z0-9_]+`)
 func sanitizeTemplateVar(templateVarName string) string {
 	return templatePattern.ReplaceAllString(templateVarName, "")
 }
+
+func replaceAll(content string, key string, s string) string {
+	// The prefix and suffix we search for are alphanumerical, as well as the "-", and "_".
+	// From investigating, this set seems at the moment like the most basic set to handle false positive substrings.
```
UnseenWizzard, line 68: nit
```suggestion
	// From investigating, this character set seems to be most most basic regex that still avoids false positive substring matches.
```

pkg/download/id_extraction/id_extraction.go
```diff
@@ -26,7 +26,7 @@ import (
 
 // meIDRegexPattern matching a Dynatrace Monitored Entity ID which consists of a type containing characters and
 // underscores, a dash separator '-' and a 16 char alphanumeric ID
```
UnseenWizzard, line 28:
```suggestion
// underscores, a dash separator '-' and 16 hex numbers
```

UnseenWizzard (dismissed):
LGTM - just small questions and suggestions for IMO clearer wording in the comments

pkg/download/id_extraction/id_extraction.go
```diff
@@ -26,7 +26,7 @@ import (
...
-var meIDRegexPattern = regexp.MustCompile(`[a-zA-Z_]+-[A-Za-z0-9]{16}`)
+var meIDRegexPattern = regexp.MustCompile(`[a-zA-Z_]+-[A-Fa-f0-9]{16}`)
```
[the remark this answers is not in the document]
Laubi: Yes, we could do that, but I argue against that. In that case, we would need to support any type that is possible to configure, which are all, and even if we only select the configurable, they would be a lot. I would just keep on using the open matching. With hex matching, we're already on a good track.

UnseenWizzard (approved):
:shipit:

PR #1292, opened 2023-12-07
fix: Dont always download buckets and config API types

UnseenWizzard:
1. drops the buckets feature flag, the feature is out without complaints for 2 months
2. fixes classic configs and buckets always being downloaded.

We should possibly discuss if always downloading classic config API types was desired.

YES
1. bucket feature can no longer be deactivated
2. config APIs and buckets are no longer always downloaded

### DIFF
--- a/cmd/monaco/download/download_configs.go
+++ b/cmd/monaco/download/download_configs.go
@@ -18,7 +18,6 @@ import (
 	"errors"
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/dynatrace"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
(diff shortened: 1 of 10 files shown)


Laubi (approved):

cmd/monaco/download/download_configs.go
```diff
@@ -233,7 +232,7 @@ func doDownloadConfigs(fs afero.Fs, downloaders downloaders, opts downloadConfig
 func downloadConfigs(downloaders downloaders, opts downloadConfigsOptions) (project.ConfigsPerType, error) {
 	configs := make(project.ConfigsPerType)
 
-	{
+	if shouldDownloadConfigs(opts) {
```
UnseenWizzard, line 235: double-checked this: this is not actually needed, classic config API download is triggered but the downloader is setup with zero "apis to download" if an "only-settings" is given. 
IMO that setup is pretty confusing and I'd keep the explicit exclusion here where all other types are excluded

PR #1294, opened 2023-12-11
feat(deploy): Print deprecation note for Settings

UnseenWizzard:
feat(deploy): Print deprecation note for Settings
Introduces warning log pattern and first entries for 'span-attribute' scopes


pkg/deploy/internal/setting/validation.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/entities"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter/value"
+)
+
+type Validator struct{}
+
+var deprecatedSchemas = map[string]string{
+	"builtin:span-attribute":       "this setting was replaced by 'builtin:attribute-allow-list' and 'builtin:attribute-masking'",
+	"builtin:span-event-attribute": "this setting was replaced by 'builtin:attribute-allow-list' and 'builtin:attribute-masking'",
```
Laubi, line 33: `builtin:resource-attribute` is also deprecated

pkg/deploy/internal/setting/validation.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	if msg, deprecated := deprecatedSchemas[s.SchemaId]; deprecated {
+		log.WithFields(field.Coordinate(c.Coordinate), field.Environment(c.Environment, c.Group)).Warn("Schema %q is deprecated - please update your configurations: %s", s.SchemaId, msg)
+	}
+
+	return validateScopeType(c)
+}
+
+func validateScopeType(c config.Config) error {
```
Laubi, line 51: I am not sure if we should split the scope validation like this. I understand the intention, but as you are already writing in the commit message, it's very limited. 

I would rather remove this and fail during a settings deploy if the value parameter can't be resolved correctly. This change would remove the duplicate effort and split it into two parts vs having the error later but only one scope validation step

Laubi (changes requested):

pkg/deploy/internal/setting/validation.go
UnseenWizzard: As is we do not report this configuration error on dry-run, but only when you deploy. That is a horrible user experience. IMO the limited check is the absolute minimum we should do for a dry-run - better would be to fully validate the actual content and follow references, but that is more effort than I wanted to include in this unrelated change

pkg/deploy/internal/setting/validation.go
UnseenWizzard: Thanks! Skimmed over that one

Laubi (approved):

pkg/deploy/internal/setting/validation.go
UnseenWizzard: I was wrong, it's checked in dry-runs as well already

Laubi (approved):

PR #1297, opened 2023-12-12
refactor: Cleanup config deploy function

warber:
This PR contains a cleanup of the deploy function(s) for configurations, in order to make it more readable.
It uses just functions in order to deploy the configuration, hence there is no `componentDeployer`anymore.
Further, the entitymap the shared resource between those functions. when the entity map is leaving the scope of the package its is passed as type "EntityLookup" that can just be queried.


pkg/deploy/deploy_test.go
```diff
@@ -34,342 +33,12 @@ import (
...
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/mock/gomock"
-	"golang.org/x/net/context"
 	"testing"
 )
 
 var dashboardApi = api.API{ID: "dashboard", URLPath: "dashboard", DeprecatedBy: "dashboard-v2"}
 var testApiMap = api.APIs{"dashboard": dashboardApi}
 
-func TestDeploy(t *testing.T) {
```
warber, line 44: These tests don't compile anymore. I've deleted them as they did not add much test coverage

jskelin (approved):

pkg/deploy/deploy_test.go
UnseenWizzard: Is the management zone numeric ID logic still covered by other tests? (part of the removed `TestDeploySetting`)

UnseenWizzard (approved):

PR #1298, opened 2023-12-12
feat: Create manifest.yaml on account download and align naming of created folders with config downloads

UnseenWizzard:
feat: Create manifest.yaml on account download and align naming of created folders with config downloads

not released

### DIFF
--- a/cmd/monaco/account/download.go
+++ b/cmd/monaco/account/download.go
@@ -39,13 +39,7 @@ import (
 
 func downloadAll(fs afero.Fs, opts *downloadOpts) error {
 	if opts.outputFolder == "" {
-		opts.outputFolder = "download_account"
-	}
-	if exists, err := afero.DirExists(fs, opts.outputFolder); err != nil {
-		return err
-	} else if exists {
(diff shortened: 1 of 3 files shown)


cmd/monaco/account/download.go
```diff
@@ -29,15 +29,17 @@ import (
...
 	"github.com/google/uuid"
 	"github.com/spf13/afero"
 	"os"
+	"path/filepath"
 )
 
 func downloadAll(fs afero.Fs, opts *downloadOpts) error {
 	if opts.outputFolder == "" {
-		opts.outputFolder = "project/accounts"
+		opts.outputFolder = "download_account"
```
warber, line 42: why not "download_account_<timestamp> like for "regular" downloads?

UnseenWizzard: It's`download_account_${timestamp}` in case `download_account` already exists. Honestly I thought that's the behavior of config downloads too!

UnseenWizzard: Changed to match the behaviour of config downloads

warber (approved):

PR #1300, opened 2023-12-13
chore:  Enhance info/debug logging of account deployment

warber:
image

This PR enhances current info and debug logging of account management resources deployment.


pkg/account/deployer/deployer.go
```diff
@@ -22,15 +22,28 @@ import (
...
 	idLookupFn func(id localId) remoteId
+
+	permissions []accountmanagement.PermissionsDto
 )
 
+func (p permissions) String() string {
+	sb := strings.Builder{}
+	sb.WriteString("[")
+	for _, e := range p {
+		fmt.Fprintf(&sb, "{%s %s %s}", e.PermissionName, e.ScopeType, e.Scope)
```
UnseenWizzard, line 41: Do you have a sample of how this prints? I wonder if just printing the three strings is clear enough or if we should add more details (or even print less to give users the main info they need)

pkg/account/deployer/deployer.go
```diff
@@ -185,12 +201,14 @@ func (d *AccountDeployer) updateGroupPolicyBindings(ctx context.Context, group a
...
 	}
-	remoteIds, err := d.getAccountPolicyRefs(group)
+	accPolicyUuids, err := d.getAccountPolicyRefs(group)
 	if err != nil {
 		return fmt.Errorf("failed to fetch policies for group %q: %w", group.Name, err)
 	}
 
-	if err = d.accountManagementClient.updateAccountPolicyBindings(ctx, remoteGroupId, remoteIds); err != nil {
+	log.Debug("Updating account level policy bindings for group with ID %q", remoteGroupId)
+	log.Debug("  Bind group %q --> %v", remoteGroupId, accPolicyUuids)
```
UnseenWizzard, line 210: nit: We've just started to get rid of indented log lines. If this is related to the log above, I'd rather we print it as one log entry (possibly with a newline and indent), rather than only "visually" making two log entries related by indenting the next line

UnseenWizzard (commented):
Some small thoughts on how to present the logs. 
A question this just brought up for me: are there structured logs we're still missing? 
Certainly logging errors as a field, but possibly at least the account ID could be reasonable to add? Maybe even as a visible field in normal logs, like the coordinates of configs?

pkg/account/deployer/deployer.go
warber:
```
[{permission-name account abcde-fghijk}{tenant-viewer tenant abd1392}]
```
so more or less like a `%v` on the regular struct but without CreatedAt and other fields.....
And I've removed e.Scope for not leaking account UUID (see updated PR)

pkg/account/deployer/client.go
```diff
@@ -181,7 +182,7 @@ func (d *accountManagementClient) upsertGroup(ctx context.Context, groupId strin
 	}
 
 	if len(existingGroups) > 1 { // shouldn't happen
-		log.Warn("Updating multiple policies with name %s. Updating group with UUID %q", group.Name, existingGroups[0].GetUuid(), group.Name)
+		logr.FromContextOrDiscard(ctx).V(1).Info("Updating multiple policies with name " + group.Name + ". Updating group with UUID " + existingGroups[0].GetUuid())
```
UnseenWizzard, line 185: You use V(1) for 'warn' here, and for 'debug' on other lines - please double check what the right verbosity levels are for logr calls

pkg/account/deployer/client.go
```diff
@@ -129,10 +130,10 @@ func (d *accountManagementClient) upsertPolicy(ctx context.Context, policyLevel
 	}
 
 	if len(existingPolicies) > 1 { // shouldn't happen
-		log.Warn("Found multiple policies with name %q. Updating policy with UUID %q", policy.Name, existingPolicies[0].GetUuid())
+		logr.FromContextOrDiscard(ctx).V(-1).Info("Found multiple policies with name " + policy.Name + ". Updating policy with UUID " + existingPolicies[0].GetUuid())
```
UnseenWizzard, line 133: as discussed directly: let's followup on this not actually logging as Warning but as Info, due to `logr` seemingly not working as `zapr` expects

UnseenWizzard (approved):

PR #1303, opened 2023-12-18
feat(account) Improve error and log messages

Laubi:
This PR improves the error & log messages by adding more context.

Additionally, it fixes a potential `nil` panic.

### DIFF
--- a/pkg/account/downloader/internal/http/groups.go
+++ b/pkg/account/downloader/internal/http/groups.go
@@ -18,7 +18,7 @@ package http
 
 import (
 	"context"
-	"errors"
+	"fmt"
 	accountmanagement "github.com/dynatrace/dynatrace-configuration-as-code-core/gen/account_management"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/account/downloader/internal/http/groups.go
```diff
@@ -30,7 +30,7 @@ func (c *Client) GetGroups(ctx context.Context, accUUID string) ([]accountmanage
 		return nil, err
 	}
 	if r != nil && int(r.Count) != len(r.Items) {
-		return nil, errors.New("the received data are inconsistent")
+		return nil, fmt.Errorf("the received data is inconsistent: count(%d) != items(%d)", int(r.Count), len(r.Items))
```
UnseenWizzard, line 33: I don't think "is inconsistent" tells a user anything about what's wrong or what to do. Can be log some concrete error here? I forsee all of these errors ending up at the dev team with a "what does this mean? what do I do?".

UnseenWizzard (approved):

PR #1306, opened 2023-12-19
feat(account): Don't fail on inconsistent API

Laubi:
If the API returns inconsistent data, we should not fail.
For a user, this check is meaningless as they can do nothing if it fails and not being able to use monaco in that case until it is fixed does not make sense.

### DIFF
--- a/pkg/account/downloader/internal/http/groups.go
+++ b/pkg/account/downloader/internal/http/groups.go
@@ -18,7 +18,6 @@ package http
 
 import (
 	"context"
-	"fmt"
 	accountmanagement "github.com/dynatrace/dynatrace-configuration-as-code-core/gen/account_management"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/account/downloader/internal/http/groups.go
```diff
@@ -29,9 +28,6 @@ func (c *Client) GetGroups(ctx context.Context, accUUID string) ([]accountmanage
 	if err != nil {
 		return nil, err
 	}
-	if r != nil && int(r.Count) != len(r.Items) {
```
UnseenWizzard, line 32: As discussed directly, makes sense to me for the reason you mentioned in the PR description. 
Neither a user nor we can actually do anything about that error case. 

@jskelin is there something @Laubi and I are missing though? Is this a case you actually ran into while developing the download?

PR #1307, opened 2023-12-19
feat(account): Don't create folder if no account data has been downloaded

Laubi:
If no data has been downloaded, we'd still create the download folder with only a manifest but nothing else. This is now no longer the case.

### DIFF
--- a/cmd/monaco/account/download.go
+++ b/cmd/monaco/account/download.go
@@ -64,7 +64,7 @@ func downloadAll(fs afero.Fs, opts *downloadOpts) error {
 
 	var failedDownloads []account.AccountInfo
 	for acc, accClient := range accountClients {
-		err := download(fs, opts, acc, accClient)
+		err := downloadAndPersist(fs, opts, acc, accClient)
 		if err != nil {
 			log.Error("Failed to download account resources for account %q: %s", acc, err)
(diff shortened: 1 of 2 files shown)


cmd/monaco/account/download.go
```diff
@@ -64,13 +64,20 @@ func downloadAll(fs afero.Fs, opts *downloadOpts) error {
...
+		err := downloadAndPersist(fs, opts, acc, accClient)
 		if err != nil {
 			log.Error("Failed to download account resources for account %q: %s", acc, err)
 			failedDownloads = append(failedDownloads, acc)
 		}
 	}
 
+	// all environments failed to download
+	if len(failedDownloads) == len(accountClients) {
+		log.Error("No resources were downloaded. No project will be created.")
```
UnseenWizzard, line 76: IMO this doesn't clearly enough state that downloading failed for all accounts, also isn't the returned error message logged? It looks like other places like the one on line 83 assume the error is handled/logged by the caller - if it is, then IMO no extra log is needed here, we don't log anything else before returning. 

Message suggestion for either case: 

`("failed to download any resources from accounts %q - not creating download folder", es)`

UnseenWizzard (approved):

PR #1309, opened 2023-12-20
feat(account): Accept manifest as optional flag instead of positional arg for all commands

UnseenWizzard:
Accept manifest as optional flag instead of positional arg for all commands.

This brings it in line with other commands and also reflects the usage change we want to introduce for config deploy in the near future. 

NO - not released yet


cmd/monaco/account/deploy.go
```diff
@@ -51,23 +51,20 @@ func deployCommand(fs afero.Fs) *cobra.Command {
 		Use:               "deploy <manifest.yaml> [flags]",
 		Short:             "Deploy account management resources",
 		Example:           "monaco account deploy manifest.yaml [--account <account-name-in-manifest>] [--project <project-defined-in-manifest>]",
-		Args:              cobra.ExactArgs(1),
```
Laubi, line 54: Should we use `cobra.MaximumNArgs` to still validate whether it has arguments set? could be a typo as well.

Laubi (approved):

PR #1311, opened 2023-12-21
unit test added for account management download

jskelin:
* introduced interface for http client
 * some code optimisation

To decouple code, a new interface for the client is introduced. The majority of units test is added. Coverage for the `group` download is unfinished and will be part of a separate PR


pkg/account/downloader/internal/http/client.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (c *Client) GetGroups(ctx context.Context, accUUID string) ([]accountmanagement.GetGroupDto, error) {
+	r, resp, err := c.GroupManagementAPI.GetGroups(ctx, accUUID).Execute()
+	defer closeResponseBody(resp)
+	if err = getErrorMessageFromResponse(resp, err); err != nil {
+		return nil, err
+	}
+	if r == nil {
+		return nil, errors.New("the received data are empty")
```
warber, line 83:
```suggestion
		return nil, errors.New("no group response data received")
```

warber (approved):

PR #1313, opened 2023-12-21
Enable featureflags.New only for testing

jskelin:
Since all FeatureFlags are listed inside `featureflags` package, and on-fly creation is not allowed New() function is useless except for testing. With this change, the `New()` function is moved to export_test.go file and is available only during the testing phase.


warber (approved):
lgtm,

just a thought, Since "New" is basically useless and just exists for being able to create a FeatureFlag value from another package, I think it would also be fine to just put the test code in the same package and just directly create the structs...

PR #1316, opened 2024-01-05
New test that covers groups with permissins

jskelin:
Since permissions are an important part of a configuration, this test case covers logic related to downloading it (the majority is inside the first commit).


Laubi:
Please update the PR description to give us a brief summary of what you have changed and why. For example, in our PR, you're renaming YAML fields and structs, way more than just adding a single test as the PR description describes :)

Laubi (approved):

Laubi (approved):

PR #1317, opened 2024-01-05
fix: handling of parameterized values that contain dt entity ids

warber:
Some configurations, e.g. synthetic monitors, support some special syntax to subistitute values, for example setting a header using the value like `"Bearer:  {SOME_ENTITY_ID | token}"` and things like that.
For quite some time, monaco inspects each payloads and searches for entity ids and replaces them with monaco references.
This process breaks go templating once we come across such values like above (e.g., In the example above something like {{{ externalIDs.SOME_ENTITY_ID }}} is created) 

Possible solution: as soon as we download configuration, parse the json tree and replace each opening curly brace ("{") with `{{'{'}}`. This would work, however this means that we would obfuscate a lot of configurations that naturally contain curly braces, e.g. workflow scripts. Also it is hard to force the go JSON marshaller to produce that kind of "unescaped/invalid" JSON.

Solution: It sounds ridiculous, but simply search for three subsequent "{" right befor rendering the payload for deployment and escape the first one, with {{"{"}} seems to work.

### DIFF
--- a/pkg/config/template/renderer.go
+++ b/pkg/config/template/renderer.go
@@ -17,6 +17,7 @@ package template
 import (
 	"bytes"
 	"fmt"
+	"strings"
 	templ "text/template" // nosemgrep: go.lang.security.audit.xss.import-text-template.import-text-template
 )
 
@@ -28,6 +29,10 @@ func Render(template Template, properties map[string]interface{}) (string, error
 		return "", fmt.Errorf("failure trying to render template %s: %w", template.ID(), err)
 	}
 
+	// special handling to fix the case that a payload was fetched that after the download and processing
+	// results in three subsequent {. This can happen e.g. if the payload allows to have content embraced between
+	// curly braces like {"somekey" : "some {VALUE}"}
+	content = strings.ReplaceAll(content, "{{{", "{{\"{\"}}{{")
 	parsedTemplate, err := ParseTemplate(template.ID(), content)
 
 	if err != nil {
--- a/pkg/config/template/renderer_test.go
+++ b/pkg/config/template/renderer_test.go
@@ -108,6 +108,42 @@ func TestRender(t *testing.T) {
 			false,
 		},
 		{
+			"renders simple template containing three subsequent {",
+			&InMemoryTemplate{
+				content: "{{{ .val }}}",
+			},
+			map[string]interface{}{"val": "the-key"},
+			`{the-key}`,
+			false,
+		},
+		{
+			"renders template containing three subsequent { ",
+			&InMemoryTemplate{
+				content: "{{{.val}} | bearer}",
+			},
+			map[string]interface{}{"val": "the-key"},
+			`{the-key | bearer}`,
+			false,
+		},
+		{
+			"renders simple template containing multiple three subsequent {",
+			&InMemoryTemplate{
+				content: "{{{ .val1 }}} {{{ .val2 }}}",
+			},
+			map[string]interface{}{"val1": "the-key1", "val2": "the-key2"},
+			`{the-key1} {the-key2}`,
+			false,
+		},
+		{
+			"renders simple template containing {",
+			&InMemoryTemplate{
+				content: "\\{\\{\\{ I just like curlies",
+			},
+			map[string]interface{}{"val1": "the-key1", "val2": "the-key2"},
+			`\{\{\{ I just like curlies`,
+			false,
+		},
+		{
 			"renders template #1",
 			&InMemoryTemplate{
 				content: templateString,


Laubi:
I do not have a better idea on how to solve this, but I would actually go with `{{ "{" }}` instead of `{{ curly }}`.
The reason is that we're already escaping workflows (jinja templates) like that, and I would rather reuse something that's already there. Not reuse from an internal perspective but from a customer/template perspective.

Laubi:
Please add some tests test that `{SOME_ENTITY_ID | something}` gets replaced correctly.

Laubi:
One more question: Does this change also change other existing configurations? E.g., if I download an environment before this change (with valid configs oc), will I get the same template after this change?

pkg/config/template/renderer_test.go
```diff
@@ -72,20 +64,16 @@ func TestParseTemplate(t *testing.T) {
...
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := ParseTemplate(tt.args.id, tt.args.content)
+			_, err := ParseTemplate(tt.args.id, tt.args.content)
 			if (err != nil) != tt.wantErr {
 				t.Errorf("ParseTemplate() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
-			if !reflect.DeepEqual(got, tt.want) {
```
UnseenWizzard, line 86: why did you remove asserting the actual template here? Also shouldn't this test move to template_test.go along with the function move?

UnseenWizzard:
> I do not have a better idea on how to solve this, but I would actually go with `{{ "{" }}` instead of `{{ curly }}`. The reason is that we're already escaping workflows (jinja templates) like that, and I would rather reuse something that's already there. Not reuse from an internal perspective but from a customer/template perspective.

I agree with @Laubi , IMO we should generally behave the same when templating out these and the workflow curlies. 
Adding a 'curly' function seems even harder to describe to users than the reason for "templating" curly brackets - basic value templating is a concept we expect users to know. Functions in templates are not something we so far expect a normal user to be aware off. 

re: invalid JSON - we already often produce invalid JSON with templating - e.g. templating integer values will result in a "string" for the template var which is not enclosed by quotes. I think it's fine to produce invalid JSON here as well.

warber:
@UnseenWizzard @Laubi If we don't care about the validity of the JSON template, we can certainly do it with the "regular" escaping strategy. And if are already in a situtation where we break JSON i think it's fine...
I will change it accordingly

pkg/config/template/renderer.go
```diff
@@ -28,6 +29,10 @@ func Render(template Template, properties map[string]interface{}) (string, error
 		return "", fmt.Errorf("failure trying to render template %s: %w", template.ID(), err)
 	}
 
+	// special handling to fix the case that a payload was fetched that after the download and processing
+	// results in three subsequent {. This can happen e.g. if the payload allows to have content embraced between
+	// curly braces like {"somekey" : "some {VALUE}"}
+	content = strings.Replace(content, "{{{", "{{\"{\"}}{{", -1)
```
Laubi, line 35:
```suggestion
	content = strings.ReplaceAll(content, "{{{", "{{\"{\"}}{{")
```

pkg/config/template/renderer_test.go
```diff
@@ -107,6 +107,15 @@ func TestRender(t *testing.T) {
 			`{ "key": the-key }`,
 			false,
 		},
+		{
+			"renders simple template containing three subsequent {",
```
Laubi, line 111: Please add a second test for two or more variables
```
content: "{{{ .val1 }}} {{{ .val2 }}}",
```

pkg/config/template/renderer_test.go
```diff
@@ -107,6 +107,24 @@ func TestRender(t *testing.T) {
...
+		},
+		{
+			"renders simple template containing multiple three subsequent {",
+			&InMemoryTemplate{
+				content: "{{{ .val1 }}} {{{ .val2 }}}",
+			},
+			map[string]interface{}{"val1": "the-key1", "val2": "the-key2"},
+			`{the-key1} {the-key2}`,
+			false,
+		},
```
UnseenWizzard, line 127: could you please add a few more test cases? 

1. The full sample that found this issue `{{{.val}} | bearer}` -> `{the-key | bearer}`
2. Some non-template cases that are not broken by this change like `\\{\\{\\{ I just like curlies` -> `{{{ I just like curlies`

UnseenWizzard (approved):

PR #1322, opened 2024-01-09
Adopt new names for config.yaml (account management)

jskelin:
* environment -> environments
  * name -> environment (for environment definition)
  * managementZone -> managementZones

Rename a few of the keys in yaml file on the suggestion of the first test users.

### DIFF
--- a/pkg/persistence/account/internal/types/types.go
+++ b/pkg/persistence/account/internal/types/types.go
@@ -59,9 +59,9 @@ type (
 		// Account level permissions and policies that apply to users in this group
 		Account *Account `yaml:"account,omitempty" json:"account,omitempty" jsonschema:"description=Account level permissions and policies that apply to users in this group."`
 		// Environment level permissions and policies that apply to users in this group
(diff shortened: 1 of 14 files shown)


pkg/persistence/account/internal/types/types.go
```diff
@@ -59,17 +59,17 @@ type (
 		// Account level permissions and policies that apply to users in this group
 		Account *Account `yaml:"account,omitempty" json:"account,omitempty" jsonschema:"description=Account level permissions and policies that apply to users in this group."`
 		// Environment level permissions and policies that apply to users in this group
-		Environment []Environment `yaml:"environment,omitempty" json:"environment,omitempty" jsonschema:"description=Environment level permissions and policies that apply to users in this group."`
+		Environment []Environment `yaml:"environments,omitempty" json:"environments,omitempty" jsonschema:"description=Environment level permissions and policies that apply to users in this group."`
```
UnseenWizzard, line 62: nit: Would make sense to make the struct fields for Env and MZ plural too IMO, they are slices anyway.

UnseenWizzard (approved):

PR #1327, opened 2024-01-11
chore: limit requests for deploying account management resources

warber:
This PR enables the dispatcher to dynamically create workers on the fly instead of having a fixed-sized pool.
This way the dispatcher behaves like creating a worker (aka go routing) for each resource.

In the end i am still not totally sure whether spawning a go routine for each and every resource is so nice.
So for now i've set max workers to be equal to `MONACO_CONCURRENT_REQUESTS`

It has the nice effect that there are not potentially `n - defaultLimit`  started go routines waiting to passing the gate :P

### DIFF
--- a/pkg/account/deployer/deployer.go
+++ b/pkg/account/deployer/deployer.go
@@ -69,9 +69,10 @@ type client interface {
 }
 
 type AccountDeployer struct {
-	accClient client
-	idMap     idMap
-	logger    loggers.Logger
+	accClient  client
+	idMap      idMap
+	logger     loggers.Logger
+	maxWorkers int
 }
 
 func NewAccountDeployer(client client) *AccountDeployer {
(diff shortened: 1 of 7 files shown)


pkg/account/deployer/deployer.go
```diff
@@ -69,17 +69,28 @@ type client interface {
...
-	logger    loggers.Logger
+	accClient  client
+	idMap      idMap
+	logger     loggers.Logger
+	maxWorkers int
 }
 
-func NewAccountDeployer(client client) *AccountDeployer {
-	return &AccountDeployer{
+func WithMaxWorkers(size int) func(*AccountDeployer) {
```
UnseenWizzard, line 78: nit: the "maxWorkers" naming makes sense on the Dispatcher, which has a worker pool and documentation defining how it creates workers. But to me it makes no sense on the AccountDeployer. Nothing else in the interface of the deployer would tell me what its workers are or why I would define WithMaxWorkers. 

Maybe a "max concurrent deploys" wording similar to the "max concurrent requests" would be easier to understand and document in relation to the Deployer.

UnseenWizzard (approved):

UnseenWizzard (approved):

PR #1330, opened 2024-01-12
feat(account): Write YAML file content in stable order on download

UnseenWizzard:
feat(account): Write YAML file content in stable order on download

As we do for config YAMLs, the content of account YAML files created by a download is now sorted in a stable order, so that files can actually be diffed and are only different if data that was downloaded has actually changed.

I've run a limited benchmark of running the download command 5 times on our E2E test account.
For the limited benchmark, there's no significant difference in speed or extra allocations incurred from the added sorting.

Not yet released, and not significant anyway: 
Order of downloaded user.yaml, group.yaml and policy.yaml changes compared to previous downloads, but previous DLs produced different file orders each time anyway.

### DIFF
--- a/pkg/persistence/account/writer/writer.go
+++ b/pkg/persistence/account/writer/writer.go
@@ -23,8 +23,10 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/account"
 	persistence "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/persistence/account/internal/types"
 	"github.com/spf13/afero"
+	"golang.org/x/exp/slices"
 	"gopkg.in/yaml.v2"
 	"path/filepath"
+	"strings"
 )
 
 // Context for this account resource writer, defining the filesystem and paths to create resources at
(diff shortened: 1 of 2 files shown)


pkg/persistence/account/writer/writer.go
```diff
@@ -145,15 +156,27 @@ func toPersistenceGroups(groups map[string]account.Group) []persistence.Group {
...
 				Environment:    e.Environment,
 				ManagementZone: e.ManagementZone,
 				Permissions:    e.Permissions,
 			}
+			// sort permissions so that they are stable within a persisted file
+			slices.SortFunc(mzs[i].Permissions, caseInsensitiveLexicographicSmaller)
 		}
+		// sort mzs by env and name so that they are stable within a persisted file
+		slices.SortFunc(mzs, func(a, b persistence.ManagementZone) bool {
+			return strings.ToLower(a.Environment) <= strings.ToLower(b.Environment) && caseInsensitiveLexicographicSmaller(a.ManagementZone, b.ManagementZone)
```
warber, line 178: why not use the compare func for Environment as well?

UnseenWizzard: I want MZ for the same environment to be sorted by MZ under that environment. the compare func only does 'smaller' not equal. 
Maybe I'm thinking too complicated though?

Laubi (approved):

pkg/persistence/account/writer/writer.go
warber: i mean why doing it inline here, whereas for the other's compare you've created a function.

warber: anyway, it's a nit

UnseenWizzard: Ah.. I created a function because I did the same check in several places. In this case I only do it once, so other than documentation via naming I saw/see no reason for adding a function here.

PR #1333, opened 2024-01-15
feat: Manage network zones configuration with monaco

warber:
This PR adds network zone configuration, to known classic api's so that they are downloaded with monaco.
Default network zone (e.g. network zone with id == "default") will be excluded from download as modifications to this network zone is anyway rejected by the API.
During deployment, we need to make sure that always the PUT method is used for update as well as creation.

Adds support for network-zones to monaco.

### DIFF
--- a/cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/network-zone/test-zone.json
+++ b/cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/network-zone/test-zone.json
@@ -1,0 +1,9 @@
+{
+  "alternativeZones": [],
+  "description": "This is a test zone",
+  "fallbackMode": "ANY_ACTIVE_GATE",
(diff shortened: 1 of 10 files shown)


Laubi (dismissed):

cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/network-zone/test-zone.json
```diff
@@ -0,0 +1,9 @@
+{
+  "alternativeZones": [],
+  "description": "This is a test zone",
+  "fallbackMode": "ANY_ACTIVE_GATE",
+  "numOfConfiguredActiveGates": 0,
```
Laubi, line 5: Please add one test without these properties, just to be sure in the future :)

Laubi (approved):

PR #1334, opened 2024-01-15
chore: fail fast if something errors prior to running the test fn

warber:
This PR adjusts the assertions, so that the test is aborted as soon as an error occurs before executing the test case.

### DIFF
--- a/cmd/monaco/integrationtest/account/all_resources_integration_test.go
+++ b/cmd/monaco/integrationtest/account/all_resources_integration_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/persistence/account/writer"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"net/http"
 	"slices"
 	"testing"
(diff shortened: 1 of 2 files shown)


cmd/monaco/integrationtest/account/all_resources_integration_test.go
```diff
@@ -42,7 +43,7 @@ func TestDeployAndDelete_AllResources(t *testing.T) {
 	cliDeployMZones.SetArgs([]string{"deploy", "testdata/all-resources/manifest-mzones.yaml"})
 
 	err := cliDeployMZones.Execute()
-	assert.NoError(t, err)
+	require.NoError(t, err)
```
jskelin, line 46: Will this work for local execution during debugging (where the version is not set to 2.x)?

jskelin (approved):

cmd/monaco/integrationtest/account/all_resources_integration_test.go
warber: I don't understand how this relates to the version being set?

Laubi (approved):

cmd/monaco/integrationtest/account/all_resources_integration_test.go
jskelin: Sorry, wrongly reported...

warber: no problem :)

PR #1335, opened 2024-01-17
fix: reimplement numeric id extraction to handle overflows

warber:
This PR fixes an error in id extraction which is happening if the regular implementation for var integer reading would cause an overflow, resulting in Monaco not calculating a numeric ID.

### DIFF
--- a/internal/idutils/numeric_id.go
+++ b/internal/idutils/numeric_id.go
@@ -60,17 +60,41 @@ func getNumericID(u uuid.UUID) int {
 }
 
 // getLegacyNumericID implements the Dynatrace logic for transforming a "legacy" (variant RFC, version 4 (random)) UUID to a numeric ID
-// by taking specific bytes of the UUID and decoding them as a signed variable-length integer
+// by taking specific bytes of the UUID and decoding them as a signed variable-length integer. The implementation of the bit shifting algorithm
+// for variable length integer decoding allows overflows to happen and are handled gracefully on purpose, as this is expected to happen for such IDs on the server side.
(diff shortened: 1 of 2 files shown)


jskelin (approved):

jskelin (dismissed):
lgtm

internal/idutils/numeric_id.go
```diff
@@ -62,15 +62,38 @@ func getNumericID(u uuid.UUID) int {
...
 // by taking specific bytes of the UUID and decoding them as a signed variable-length integer
 func getLegacyNumericID(u uuid.UUID) (int, error) {
-
 	var b []byte
 	b = u[0:6]                 // fill byte 0-5 with the UUID's most significant bytes (big-endian)
 	b = append(b, u[12:16]...) // fill byte 6-9 with the last 4 bytes of the UUID/ending "integer" of the UUID's least significant LSB
+	return int(zigzagDecode(varInt(b))), nil
+}
+
+func varInt(buf []byte) int64 {
```
UnseenWizzard, line 71: change looks good to me and - and at a glance the failing E2E tests don't seem to be directly related to management zone IDs calculations?

What I'd like to see either in the commit message or as a comment is the reason for the change "handle overflows" makes sense now, but not in 3-6months when someone else might again look at this and think "wait a minute, there's a Varint library implementation!". Noting down that we purposely replicate Dynatrace's bitshift algorithm exactly with it's 'feature' of silently overflowing to ensure we calculate the same IDs would be helpful IMO.

warber: added a few words

jskelin (approved):

PR #1338, opened 2024-01-18
feat: Extend API definitions that allow for complex API definitions

Laubi:
This PR adds the first version of parsing complex API definitions so that we can later extend it to reference existing objects.

Currently, the following would be possible:

```yaml
configs:
- id: my-profile
  type:
    api:
      name: profile
      scope: […] 
  config:
    …
```

Note that the scope is currently still unused and needs to be used in the future when implementing the APIs.

## Future Work

This PR is the first version to build the other stuff upon. There is still some stuff to be done.

### Refactorings
The type loading is a mess of yaml unmarshalling, mapstructure, multiple decoding steps, and code that is distributed throughout the config loading. 
The Unmarshalling step needs to be executed 5! times alone to make it work easily, which is obviously not what we want to do.
This should be refactored to load the type at the start and verify it, and then not do it again.

### Writing
The scope needs to be written as well if we download, this needs to be implemented as well.

### Actually implementing the APIs.
This PR only lays the foundation for users to define it.

### Other
It might be beneficial if we only let users define a scope IF they use an API that requires it. This could solve some confusion in usage maybe.

### DIFF
--- a/cmd/monaco/convert/convert_test.go
+++ b/cmd/monaco/convert/convert_test.go
@@ -135,7 +135,7 @@ profile:
 	assert.NoError(t, err)
 	assert.Equal(t, `{ "name":  "{{ .name }}" }`, string(payloadContent))
 
-	cfgs, errs := loader.LoadConfig(testFs, &loader.LoaderContext{
+	cfgs, errs := loader.LoadConfigFile(testFs, &loader.LoaderContext{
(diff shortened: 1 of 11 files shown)


warber (dismissed):
lgtm, pls run e2e tests tho

warber (approved):

PR #1340, opened 2024-01-19
fix: handle network zone deployment error

warber:
This, PR adds a special handling for network zone deployment errors that can occur when the feature for network zone is not (yet) enabled. 

In integration tests we need to make sure that we do not cleanup the specific setting responsible for enabling the network zone feature. Hence it needs to be ignored during cleanup/deletion

To do this, I've sneaked in an (already requested) feature to exclude or exclusively include specific configuration types when generating the delete file for configurations.

It fixes the bug. Additionally the user can add filters via flags to the delete file generation:

```
monaco generate deletefile ... --types builtin:alerting.profile  # generates delete file entries only for the given type(s)
monaco generate deletefile ... --exclude-types  # skips generating delete file entries for the given type(s)
```

### DIFF
--- a/cmd/monaco/generate/deletefile/command.go
+++ b/cmd/monaco/generate/deletefile/command.go
@@ -20,16 +20,24 @@ import (
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/cmdutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/completion"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/errutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/files"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
+	manifestloader "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest/loader"
+	project "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project/v2"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
+	"path/filepath"
 )
 
 func Command(fs afero.Fs) (cmd *cobra.Command) {
 
 	var fileName, outputFolder string
 	var projects, environments []string
+	var includeTypes, excludeTypes []string
 
 	cmd = &cobra.Command{
 		Use:               "deletefile <manifest.yaml>",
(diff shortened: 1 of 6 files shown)


cmd/monaco/generate/deletefile/command.go
```diff
@@ -47,14 +48,15 @@ func Command(fs afero.Fs) (cmd *cobra.Command) {
 				return err
 			}
 
-			return createDeleteFile(fs, manifestName, projects, environments, fileName, outputFolder)
+			return createDeleteFile(fs, manifestName, projects, environments, fileName, excludeTypes, outputFolder)
```
UnseenWizzard, line 51: Opinion: In terms of UX I find an "exclude" flag a bit weird personally. I get that there might a be need (like this one) to say "delete all but", but I would expect that the inverse "include/only" flag exists as well so I can say "remove all of this type configured via this project".

What was the existing request for this feature?

warber: i agree, the opposite would be more useful. Why not offer both? :)
Let me explain my reasoning. First i wanted to implement the filtering in the integration test cleanup only. However, it turns out that code cannot easily be reused for that purpose. Of course the "read-unmarshalToMap-modify-marshalFromMap-write" logic can be written manually, and I did that. However, there is the risk that this logic would silently fail if delete file structure slightly changes, and all of the sudden we have a failing test and someone needs to dig in why...

so i thought, why not introduce that flag, it might be not THE usecase everyone is searching for but hey it does no harm right?

Anyway, do you want me to revert that commit? :)

cmd/monaco/generate/deletefile/deletefile.go
```diff
@@ -162,13 +163,16 @@ func generateDeleteFileContent(projects []project.Project, specificEnvironments
...
 
-func generateDeleteEntries(projects []project.Project, apis api.APIs) []persistence.DeleteEntry {
+func generateDeleteEntries(projects []project.Project, apis api.APIs, excludeTypes []string) []persistence.DeleteEntry {
 	entries := make(map[persistence.DeleteEntry]struct{}) // set to ensure cfgs without environment overwrites are only added once
 
 	for _, p := range projects {
 		log.Info("Adding delete entries for project %q...", p.Id)
 
 		p.ForEveryConfigDo(func(c config.Config) {
+			if slices.Contains(excludeTypes, c.Coordinate.Type) {
```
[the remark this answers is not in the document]
warber: exactly

cmd/monaco/generate/deletefile/command.go
UnseenWizzard: as mentioned via DM - I'd suggest you discuss whether to keep, extend or revert the feature in the whole team next week. 

mentioned my alternative idea on how to ensure the setup for e2e tests in dm

pkg/client/dtclient/config_client.go
```diff
@@ -334,6 +336,9 @@ func isApplicationNotReadyYet(resp rest.Response, theApi api.API) bool {
 		isApplicationDetectionRuleException(resp, theApi) ||
 		strings.Contains(string(resp.Body), "Unknown application(s)")
 }
+func isNetworkZoneFeatureNotEnabledYet(resp rest.Response, theApi api.API) bool {
+	return strings.HasPrefix(theApi.ID, "network-zone") && (resp.Is4xxError())
```
UnseenWizzard, line 340: It makes it more prone to breaking, but should we maybe be more explicit than 4xx? E.g. the isApplicationNotReadyYet checks a specific response content, and thus is tied to the message but won't retry for any possibly unrelated user error.

cmd/monaco/generate/deletefile/command.go
```diff
@@ -47,14 +49,16 @@ func Command(fs afero.Fs) (cmd *cobra.Command) {
...
-			return createDeleteFile(fs, manifestName, projects, environments, fileName, outputFolder)
+			return createDeleteFile(fs, manifestName, projects, environments, fileName, includeTypes, excludeTypes, outputFolder)
 		},
 	}
 
 	cmd.Flags().StringVarP(&outputFolder, "output-folder", "o", "", "The folder the generated delete file should be written to. If not set, files will be created in the current directory.")
 	cmd.Flags().StringVarP(&fileName, "file", "", "delete.yaml", "The name of the generated delete file. If a file of this name already exists, a timestamp will be appended.")
 
 	cmd.Flags().StringSliceVarP(&projects, "project", "p", nil, "Projects to generate delete file entries for. If not defined, all projects in the manifest will be used.")
+	cmd.Flags().StringSliceVar(&excludeTypes, "exclude-types", nil, "Comma separated list of config types to be excluded from the generation process.")
```
Laubi, line 60:
```suggestion
	cmd.Flags().StringSliceVar(&excludeTypes, "exclude-types", nil, "Comma-separated list of config types to be excluded from the generation process.")
```

cmd/monaco/generate/deletefile/command.go
```diff
@@ -47,14 +49,16 @@ func Command(fs afero.Fs) (cmd *cobra.Command) {
...
+	cmd.Flags().StringSliceVar(&includeTypes, "types", nil, "Comma separated list of config types to be included in the generation process.")
```
Laubi, line 61:
```suggestion
	cmd.Flags().StringSliceVar(&includeTypes, "types", nil, "Comma-separated list of config types to be included in the generation process.")
```

cmd/monaco/generate/deletefile/deletefile.go
```diff
@@ -36,7 +36,7 @@ import (
 	"strings"
 )
 
-func createDeleteFile(fs afero.Fs, manifestPath string, projectNames, specificEnvironments []string, filename, outputFolder string) error {
+func createDeleteFile(fs afero.Fs, manifestPath string, projectNames, specificEnvironments []string, filename string, includeTypes []string, excludeTypes []string, outputFolder string) error {
```
Laubi, line 39: Should we create a struct here to pass the data through? We wire it through quite some methods.

Laubi (commented):
Please change the flag descriptions; the recfactoring comment is a nit.

Laubi (approved):

PR #1341, opened 2024-01-23
refactor: Type loading

Laubi:
## What this PR does:
This PR is a first step into refactoring more of the configuration loading, which is quite complicated at the current point.

This PR changes how we load (and write) the type information of a config. Instead of storing the possible information in the intermediate persisting layer, we parse and store the final config one already. This allows us to later validate and check more easily based on the type, not needing to check everything multiple times.

Additionally, this can be seen as the first step to abstract the type a bit more, although more work is still needed. 

Furthermore, the test coverage is increased by adding tests for branches that were not yet covered.

Stuff fixed mentioned in #1338:
✅ Writing complex API types is possible 
✅ Type loading is a bit less messy  

### Future work
- [ ] The scope for APIs should be verified iff the API is a subpath API.
- [ ] The config loading is still quite complicated. What can be done is unclear, but small steps should be taken in the right direction.
- [ ] Type loading could be made even easier using abstractions (type loader/writers/validator), but it's better than it was before.


pkg/persistence/config/internal/persistence/type_definition.go
```diff
@@ -71,154 +50,185 @@ type AutomationDefinition struct {
...
 
-func (c *TypeDefinition) GetApiType() (string, error) {
-	switch {
-	case c.IsSettings():
-		return c.Settings.Schema, nil
-	case c.IsClassic():
-		a, err := UnmarshalApiType(c.Api)
-		if err != nil {
-			return "", err
+func (c TypeDefinition) MarshalYAML() (interface{}, error) {
```
warber, line 194: does this need to be a non-pointer receiver?

pkg/persistence/config/internal/persistence/type_definition.go
```diff
@@ -71,154 +50,185 @@ type AutomationDefinition struct {
...
 
-	switch v := data.(type) {
-	case string:
-		if v == BucketType {
-			// string was a bucket
-			c.Bucket = v
-			return nil
+	// The TypeDefinition allows for the shorthand syntax of `api: my-api`.
+	// To catch that, let's try to unmarshal directly into a string. If it works, we know the shorthand is used.
+	// let's try to directly unmarshal into a string. If it works, we know that it's either a classic API, or Bucket API.
```
warber, line 56:
```suggestion
	// To catch that, let's try to unmarshal directly into a string. If it works, we know that the shorthand syntax is used.
```

pkg/persistence/config/internal/persistence/type_definition.go
Laubi: Yes, it is required

PR #1342, opened 2024-01-23
feat: Deploy mrum key user actions

jskelin:
This PR adds support for creating and updating mobile key user actions

### DIFF
--- a/pkg/deploy/internal/extract/configname.go
+++ b/pkg/deploy/internal/extract/configname.go
@@ -17,6 +17,7 @@
 package extract
 
 import (
+	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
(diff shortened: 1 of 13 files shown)


Laubi (approved):

pkg/deploy/internal/extract/configname.go
```diff
@@ -39,3 +40,24 @@ func ConfigName(conf *config.Config, properties parameter.Properties) (string, e
...
+		return "", fmt.Errorf("property '%s' not found, this is most likely a bug", config.ScopeParameter)
+	}
+
+	switch v := scope.(type) {
+	case string:
+		if v == "" {
+			return "", fmt.Errorf("resolved scope is empty")
+		}
+		return v, nil
+	case []any:
```
jskelin, line 56: why this separation is important? I think the default option is enough

PR #1353, opened 2024-02-02
Mrum/user action session properties

warber:
This PR enables "user-action-and-session-properties-mobile" to be downloaded and deployed.


pkg/api/api.go
```diff
@@ -44,9 +56,11 @@ type API struct {
 	TweakResponseFunc func(map[string]any)
 	// IsSubPathApi indicates whenever an API is a sub-path API.
 	SubPathAPI bool
-	// Parent is the parent API ID which is requred to download all possible values for sub-path apis.
-	// e.g. key-user-actions that are bound to applications
-	Parent string
+	// Parent is used for SubPath APIs to store information about the configuration type and ID of the related
+	// configuration once Resolved() is called.
+	Parent Config
```
warber, line 61: I don't like that "sort of dynamic data" inside this struct. However as a first solution i think it is ok.
A possible refactoring would be to provide a an own type for that data that is returned by `Resolve()`consumers of that data would not continue to work with the API directly but have their own data model for API related information

pkg/client/dtclient/config_client.go
```diff
@@ -275,20 +276,21 @@ func stripCreateOnlyPropertiesFromAppMobile(payload []byte) []byte {
...
-
 	if err != nil {
 		log.WithCtxFields(ctx).WithFields(field.Error(err)).Warn("Failed to send HTTP request: %v", err)
 	} else {
 
 		log.WithCtxFields(ctx).WithFields(field.F("statusCode", resp.StatusCode)).Warn("Failed to send HTTP request: (HTTP %d)!\n    Response was: %s", resp.StatusCode, string(resp.Body))
 	}
 
-	var setting rest.RetrySetting
+	if isCalculatedLogMonitoringAPIDisabled(resp) {
```
warber, line 289: since `callWithRetryOnKnownTimingIssues` is now also called from withing `getExistingValuesFromEndpoint` we run into that error for which it does not make sense to do any retry.

pkg/api/api.go
UnseenWizzard: That does feel extremely hacky.... Also it changes the concept of "Parent" of the API as in overarching API it's a sub-path of, to being both the parent type and config ID.
Also from what I can see you only ever care about the Type and Dynatrace Object ID - so writing it to a Config object and it's config ID is misusing the Config.
If you want to keep it as dynamic data in the API, I would not reuse the Config but have a dedicated struct limited to the data needed. 

I like the idea to extend Resolve() better

pkg/client/dtclient/config_client.go
```diff
@@ -477,7 +486,16 @@ func (d *DynatraceClient) getExistingValuesFromEndpoint(ctx context.Context, the
 
 	parsedUrl = addQueryParamsForNonStandardApis(theApi, parsedUrl)
 
-	resp, err := d.classicClient.Get(ctx, parsedUrl.String())
+	var resp rest.Response
+	// For any subpath API like e.g. Key user Actions, it can be that we need to do longer retries
+	// because the parent config (e.g. an application) is not ready yet
+	if theApi.SubPathAPI {
```
jskelin, line 492: As the comment suggests, I think it is more likely to be:
```
if theApi.Parent != nil
```

pkg/client/dtclient/entities.go
```diff
@@ -40,6 +40,16 @@ type KeyUserActionsWebResponse struct {
 	} `json:"keyUserActionList"`
 }
 
+type UserActionAndSessionPropertyEntry struct {
```
jskelin, line 43: I don't see the benefit of declaring these structures in a separate file when they are used only once. I suggest putting them closer to the usage, even using them as an unnamed type.

jskelin (approved):

jskelin (approved):

PR #1355, opened 2024-02-05
Enable to define the required feature flag for an API

jskelin:
There is a need to hide specific classic endpoints under development. For that reason api.API structure is expanded with a new field, and new filtering is added to prepareAPIs function

### DIFF
--- a/cmd/monaco/deploy/deploy.go
+++ b/cmd/monaco/deploy/deploy.go
@@ -114,7 +114,7 @@ func verifyEnvironmentGen(environments manifest.Environments, dryRun bool) bool 
 
 func loadProjects(fs afero.Fs, manifestPath string, man *manifest.Manifest) ([]project.Project, error) {
 	projects, errs := project.LoadProjects(fs, project.ProjectLoaderContext{
-		KnownApis:       api.NewAPIs().GetApiNameLookup(),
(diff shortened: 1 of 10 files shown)


warber:
lgtm, Only one concern with a general FF name lice "MONACO_EXPERIMENTAL" from my side is that if this gets used more often, it becomes unclear/hard see what **exactly** this FF is affecting. Imo we can be more specific like "MONACO_FEAT_KUA_API".

jskelin:
> lgtm, Only one concern with a general FF name lice "MONACO_EXPERIMENTAL" from my side is that if this gets used more often, it becomes unclear/hard see what **exactly** this FF is affecting. Imo we can be more specific like "MONACO_FEAT_KUA_API".

Regarding the actual Feature flag, `MONACO_EXPERIMENTAL` should be a general flag for a new feature so that `main` is always deployable. Because it is not yet defined what the scope of the new feature, it is questionable whether it should be just `key-user-actions-*` (e.g. `MONACO_FEAT_KUA_API`), or the whole `(M)RUM support` (e.g. `MONACO_FEAT_(M)RUM`)?

Either way, I don't have strong opinion on the name of the FF.

warber (approved):

warber (changes requested):
e2e tests

jskelin:
> e2e tests

e2e tests passes now

warber (approved):

PR #1357, opened 2024-02-05
feat: Build simple classic environment URL without API call

UnseenWizzard:
For most URLs simply replacing the .apps. part of the URL is enough to find the classic URL, which means we don't always need an authenticated API call to find the classic URL.

If a GET to the created URL does not reply with a 200 status code, the metadata endpoint will be called.

As there may be networking/DNS edge-cases where the replaced URL is valid (GET returns 200) but is not actually a Classic environment, a feature flag allows deactivation of the feature.

-

No noticeable change for users.

### DIFF
--- a/internal/featureflags/permanent.go
+++ b/internal/featureflags/permanent.go
@@ -89,6 +89,15 @@ func ExtractScopeAsParameter() FeatureFlag {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


warber (dismissed):
lgtm

PR #1360, opened 2024-02-07
Add support for `key-user-actions-web`

jskelin:
Add support for the classic API `key-mobile-actions-web` (delete is not supported)


warber (approved):

pkg/client/dtclient/config_client.go
```diff
@@ -393,11 +390,11 @@ func joinUrl(urlBase string, path string) string {
 	return trimmedUrl + "/" + url.PathEscape(trimmedPath)
 }
 
-func isLocationHeaderAvailable(resp rest.Response) (headerAvailable bool, headerArray []string) {
+func getLocationFromHeader(resp rest.Response) ([]string, bool) {
```
warber, line 393: nit: usually, bools are returned as last value in go as it was before.

warber (approved):

PR #1366, opened 2024-02-08
refactor: Get rid of entity id's in API struct

warber:
This PR gets rid of the need of storing the id's of dynatrace objects in the API struct.

Moreover the property `SubPathAPI` is gone and is derived exclusively by the presence of the Parent property via a Method `HasParent()` similarly to e.g. `IsStandardAPI()`

There is still room for further improvement, but i thought it would be good to make a first PR.


pkg/client/dtclient/config_client.go
```diff
@@ -35,7 +35,36 @@ import (
 	"strings"
 )
 
-func (d *DynatraceClient) upsertDynatraceObject(ctx context.Context, theApi api.API, objectName string, payload []byte) (DynatraceEntity, error) {
+// APIData holds information about the api of the config to be read or written on the server
+type APIData struct {
+	ID                           string
+	SingleConfiguration          bool
+	UrlPath                      string
+	SubPathAPI                   bool
```
warber, line 43: it would be really nice if at this point we don't need this information anymore, however, for now there a re a few checks for subpathAPI in the client 🤷

pkg/client/dtclient/config_client.go
```diff
@@ -35,7 +35,36 @@ import (
 	"strings"
 )
 
-func (d *DynatraceClient) upsertDynatraceObject(ctx context.Context, theApi api.API, objectName string, payload []byte) (DynatraceEntity, error) {
+// APIData holds information about the api of the config to be read or written on the server
+type APIData struct {
```
warber, line 39: I am open for better names ... just didn't want to name it API as well, to keep confusion at a minimum

UnseenWizzard, line 39: The way I had originally understood it, adding a different struct would allow to hold differing data to the api.API struct / allow removing that need to hold parent entity ID in the api.API struct. 

With the current change you've added the parent ID to the value "wrapper" specific to download (which looks like the perfect place).

This leaves this APIData struct as more or less a direct slightly slimmed down version of api.API - am I missing something, or do we maybe currently not need this struct after all?

pkg/api/api.go
```diff
@@ -74,13 +72,16 @@ func (a API) CreateURL(environmentURL string) string {
 	return environmentURL + a.URLPath
 }
 
+func (a API) IsSubPathAPI() bool {
```
jskelin, line 75: nit: I doubt about the name. Is it better to use HasParent() ??

jskelin (approved):

jskelin (approved):

jskelin (approved):

jskelin (approved):

PR #1369, opened 2024-02-14
`delete` feature for `key-user-actions-*` and `userActionAndSessionProperties`

jskelin:
Add `delete` functionality for next endpoints:
 * `userActionAndSessionProperties`
 * `key-user-actions-web`
 * `key-user-actions-mobile`

### DIFF
--- a/pkg/delete/internal/classic/delete.go
+++ b/pkg/delete/internal/classic/delete.go
@@ -21,106 +21,99 @@ import (
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/loggers"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/dtclient"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/pointer"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/rest"
 	"golang.org/x/net/context"
 	"net/http"
-	"strings"
 )
 
-type deleteValue struct {
-	pointer.DeletePointer
-	ID   string
-	Name string
-}
-
 // Delete removes the given pointer.DeletePointer entries from the environment the supplied client dtclient.Client connects to
-func Delete(ctx context.Context, client dtclient.Client, theApi api.API, entries []pointer.DeletePointer, targetApi string) error {
-	logger := log.WithCtxFields(ctx).WithFields(field.Type(theApi.ID))
-
-	deleteErrs := 0
+func Delete(ctx context.Context, client dtclient.Client, theAPI api.API, dps []pointer.DeletePointer) error {
+	client = newCachedDTClient(client)
 	var err error
-	var delValues []deleteValue
 
-	// if the api is *not* a subpath api, we can just list all configs that exist for a given api and then filter the items that need to be deleted
-	if !theApi.HasParent() {
-		var values []dtclient.Value
-		values, err = client.ListConfigs(ctx, theApi)
-		if err != nil {
-			logger.WithFields(field.Error(err)).Error("Failed to fetch existing configs of API type %q - skipping deletion: %v", theApi.ID, err)
-			return err
-		}
-
-		delValues, err = filterValuesToDelete(logger, entries, values, theApi.ID)
-
-	} else {
-		// for sub-path APIs, it is a bit more complex. we need to query all entries of each scope defined we can delete it.
-
-		// map all entries by scope, so we can later filter them by scope
-		scopedMapped := map[string][]pointer.DeletePointer{}
-		for _, entry := range entries {
-			scopedMapped[entry.Scope] = append(scopedMapped[entry.Scope], entry)
-		}
-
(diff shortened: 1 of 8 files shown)


cmd/monaco/integrationtest/account/deploy-download_test.go
```diff
@@ -94,6 +93,8 @@ func TestIdempotenceOfDeployment(t *testing.T) {
 		assert.Equal(t, deploy1st.Groups[g.ID], deploy2nd.Groups[g.ID])
 	}
 
-	err := monaco("account delete --manifest manifest.yaml --file delete.yaml").withFs(baseFs).run()
-	require.NoError(t, err)
+	t.Cleanup(func() {
+		_, err := test.Monacof("monaco account delete --manifest manifest.yaml --file delete.yaml").WithFs(baseFs).Run()
```
UnseenWizzard, line 97: nit: I find it confusing that you're sometimes using this prefixed with 'monaco' and other times without. The fact that the prefix is trimmed anyway is hidden deep in this helper method. I'd either consistently use it with 'monaco' or simply drop the trimming and make it clear from the signature/doc that it takes the command, args and flags that are input to the CLI

.editorconfig
```diff
@@ -11,6 +11,7 @@ insert_final_newline = true
 
 [*.go]
 indent_style = tab
+ij_go_move_all_imports_in_one_declaration = true
```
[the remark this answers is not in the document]
jskelin: moved to another PR

pkg/api/api.go
```diff
@@ -89,3 +89,7 @@ func (a API) Resolve(value string) API {
 	newA.URLPath = strings.ReplaceAll(a.URLPath, "{SCOPE}", value)
 	return newA
 }
+
+func (a API) String() string {
```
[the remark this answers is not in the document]
jskelin: moved to separate PR

cmd/monaco/integrationtest/config_rewrite.go
```diff
@@ -37,7 +37,10 @@ func GenerateTestSuffix(t *testing.T, generalSuffix string) string {
 	}
 
 	timestamp := time.Now().Format("20060102150405")
-	suffix := fmt.Sprintf("%s_%d_%s", timestamp, randomNumber, generalSuffix)
+	suffix := fmt.Sprintf("_%s_%d", timestamp, randomNumber)
+	if generalSuffix != "" {
```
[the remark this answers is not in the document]
jskelin: moved to separate PR

cmd/monaco/integrationtest/cleanup.go
```diff
@@ -45,6 +45,13 @@ func CleanupIntegrationTest(t *testing.T, fs afero.Fs, manifestPath string, envi
 		_, err := test.Monacof(
 			"monaco generate deletefile %s --file %s --exclude-types builtin:networkzones %s", absManifestPath, deleteFile, env,
 		).WithFs(fs).Run()
+		{
+			if df, err := filepath.Abs(deleteFile); err == nil {
+				if b, err := afero.ReadFile(fs, df); err == nil {
+					fmt.Println(string(b[:]))
```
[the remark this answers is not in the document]
jskelin: moved to separate PR

pkg/delete/internal/classic/delete.go
```diff
@@ -21,106 +21,140 @@ import (
...
 
-			var vals []deleteValue
-			vals, err = filterValuesToDelete(logger, scopeEntries, values, theApi.ID)
+// resolveIdentifier get the actual ID from DT and update entries with it
+func resolveIdentifier(ctx context.Context, client dtclient.Client, theAPI *api.API, e *entry) (*api.API, string, error) {
+	var err error
 
-			delValues = append(delValues, vals...)
+	if theAPI.Parent != nil {
+		if _, _, err := resolveIdentifier(ctx, client, theAPI.Parent, e.parent); err != nil && !is404(err) {
```
arthurpitman, line 112: considering that this is calling itself, what exactly does it do? why do we resolve if we dont care about the result?

pkg/delete/internal/classic/delete.go
```diff
@@ -21,106 +21,140 @@ import (
...
 
-	logger.Info("Deleting %d config(s) of type %q...", len(delValues), theApi.ID)
+	id, err := findUniqueID(knownValues, e.identifier)
+	if err != nil {
+		return nil, "", err
+	}
 
-	for _, v := range delValues {
-		vLog := logger.WithFields(field.Coordinate(v.AsCoordinate()), field.F("value", v))
+	e.resolvedID = id
```
arthurpitman, line 133: Do we need to mutate `e`?

pkg/delete/internal/classic/delete.go
arthurpitman: Ok, I see it is because `e` is being updated.

pkg/delete/internal/classic/delete.go
```diff
@@ -21,106 +21,140 @@ import (
...
+	return &a, id, nil
+}
 
-		a := theApi
-		if a.HasParent() {
-			a = a.Resolve(v.DeletePointer.Scope)
-		}
+func findUniqueID(knownValues []dtclient.Value, identifier string) (string, error) {
```
arthurpitman, line 137: How does deletion work? For a given API is it always by name or ID? Or is it really mixed?

pkg/delete/internal/classic/delete.go
arthurpitman: I think probably the complexity introduced by calling itself is not worth the saving in code and would simply be explicit about handling the two cases: with a parent and without.

warber (approved):

PR #1370, opened 2024-02-14
refactor: Drop manifest version back to 1.0

UnseenWizzard:
refactor: Drop manifest version back to 1.0
    
    When account resources where originally introduced, we expected
    the manifest extension to support them to be a breaking change.
    
    In their current version the addition of accounts to the manifest
    is not a breaking change, but they are just an optional field in
    the manifest. Older versions targeting config deployments will already
    produce an error we consider clear enough, mentioning that the accounts
    field can not be unmarshalled, and do not offer an account deployment
    command, so it is should be clear enough to users that they need a
    newer version to work with account resources.

Older versions produce the following error message when faced with an accounts field:
```sh
loading of manifest failed: test-resources/integration-all-configs/manifest.yaml: error during parsing the manifest: yaml: unmarshal errors:
          line 29: field accounts not found in type manifest.manifest
```

With the manifest version increase, they would produce:
```sh
loading of manifest failed: test-resources/integration-all-configs/manifest.yaml: invalid manifest definition: `manifestVersion` 1.1 is not supported by monaco 2.8.0. Max supported version is 1.0, please check manifest or update monaco
```

Looking at the error messages, the version increase still produces a nicer user-facing message...

Generally no. Any preview users having already activated the account management feature flag will have to drop their manifests back down to version 1.0 though - this should be communicated in release notes and internal/CA guild slack channels.

### DIFF
--- a/pkg/manifest/writer/manifest_writer_test.go
+++ b/pkg/manifest/writer/manifest_writer_test.go
@@ -623,47 +623,6 @@ environmentGroups:
 `,
 		},
 		{
-			"writes manifest 1.1 if account feature is active but account defined",
-			true,
-			manifest.Manifest{
-				Projects: manifest.ProjectDefinitionByProjectID{
-					"p1": {
-						Name: "p1",
-						Path: "projects/p1",
-					},
-				},
-				Environments: manifest.Environments{
-					"env1": {
-						Name: "env1",
-						URL: manifest.URLDefinition{
-							Value: "https://a.dynatrace.environment",
-						},
-						Group: "group1",
-						Auth: manifest.Auth{
-							Token: manifest.AuthSecret{
-								Name: "TOKEN_VAR",
-							},
-						},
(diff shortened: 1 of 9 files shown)


pkg/manifest/writer/manifest_writer_test.go
```diff
@@ -611,47 +611,6 @@ projects:
 - name: p1
   path: projects/p1
 environmentGroups:
-- name: group1
```
UnseenWizzard, line 614: note this part just happens to be the same in the following test-case so the diff looks weird. What's actually removed is the complete `"writes manifest 1.1 if account feature is active but account defined"` test case

Laubi (approved):

UnseenWizzard:
unrelated sub-path config E2E test failure, merging. Other tests looking good again after recent account mgmt fix & permission fix

PR #1371, opened 2024-02-14
Refactor: Create environment clients

warber:
For future development it's necessary to be able to create clients for a set of environments.
This functionality was already implemented in deploy package.
This PR moves it to a common place ("dynatrace" package) which hosts that kind of factory functions for clients to be called from the cmd package. 
Depending on the need, users can call these function from a central place.

need a set of clients for the APIs ? --> call `dynatrace.CreateClients(url, auth)`
need a set of clients for environments? --> call `dynatrace.CreateEnvironmentClients(environments)`
need a set of clients for accounts? --> call `dynatrace.CreateAccountClients(accounts)`


cmd/monaco/deploy/internal/clientset/clientset.go
```diff
@@ -1,58 +0,0 @@
...
-		clients[deploy.EnvironmentInfo{
-			Name:  env.Name,
-			Group: env.Group,
-		}] = clientSet
-	}
-
-	return clients, nil
-}
-
-func NewClientSet(env manifest.EnvironmentDefinition, dryRun bool) (deploy.ClientSet, error) {
```
UnseenWizzard, line 42: Am I missing something, or is the change to always use dynatrace.CreateClients breaking deploy dry-runs? They would previously get DummyClient instances instead of initialized clients - which I think you've just dropped now?

UnseenWizzard: I missed the if being moved further in to the deploy code

UnseenWizzard (approved):

PR #1372, opened 2024-02-19
Filter unsupported endpoints for delete command

jskelin:
When the new endpoint is under development, it is hidden behind the feature flag `MONACO_EXPERIMENTAL`. With this change, endpoints under development will also be unsupported for the `delete` command unless `MONACO_EXPERIMENTAL` ff is present.

The main change is in `pkg/delete/delete_loader.go`. Since `delete` is used in almost every integration test for the cleanup process, necessary changes are done in the rest of the code.

### DIFF
--- a/cmd/monaco/integrationtest/v2/all_configs_integration_test.go
+++ b/cmd/monaco/integrationtest/v2/all_configs_integration_test.go
@@ -19,52 +19,49 @@
 package v2
 
 import (
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/internal/test"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/testutils"
-	"testing"
-
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/runner"
 	"github.com/spf13/afero"
-	"gotest.tools/assert"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
 )
 
-// tests all configs for a single environment
-func TestIntegrationAllConfigsClassic(t *testing.T) {
-	specificEnvironment := "classic_env"
+func TestAllConfigs(t *testing.T) {
+	tests := []struct {
+		name, environment string
+	}{
+		{
+			name:        "tests all known configs against classic url",
+			environment: "classic_env",
+		},
+		{
+			name:        "tests all known configs against platform url",
+			environment: "platform_env",
+		},
+	}
 
-	runAllConfigsTest(t, specificEnvironment)
-}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			configFolder := "test-resources/integration-all-configs/"
+			manifest := configFolder + "manifest.yaml"
 
-func TestIntegrationAllConfigsPlatform(t *testing.T) {
-	specificEnvironment := "platform_env"
+			envVars := map[string]string{featureflags.Experimental().EnvName(): "true"}
 
-	runAllConfigsTest(t, specificEnvironment)
-}
-
-func runAllConfigsTest(t *testing.T, specificEnvironment string) {
-	configFolder := "test-resources/integration-all-configs/"
-	manifest := configFolder + "manifest.yaml"
-
-	envVars := map[string]string{featureflags.Experimental().EnvName(): "true"}
-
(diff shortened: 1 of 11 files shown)


cmd/monaco/integrationtest/v2/all_configs_integration_test.go
```diff
@@ -19,52 +19,49 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			configFolder := "test-resources/integration-all-configs/"
+			manifest := configFolder + "manifest.yaml"
+
+			envVars := map[string]string{featureflags.Experimental().EnvName(): "true"}
+
+			RunIntegrationWithCleanupGivenEnvs(t, configFolder, manifest, tc.environment, "AllConfigs", envVars, func(fs afero.Fs, _ TestContext) {
+				{
+					// This causes a POST for all configs:
+					_, err := test.Monacof("deploy %s --environment %s", manifest, tc.environment).WithFs(fs).Run()
```
UnseenWizzard, line 54: nit/thought: this change isn't really related to this commit, I think they should have gone in the refactor commit - having these additional changes makes it a bit harder to review the actual change.

pkg/delete/delete_loader.go
```diff
@@ -71,7 +71,7 @@ func LoadEntriesToDelete(fs afero.Fs, deleteFile string) (DeleteEntries, error)
 	context := &loaderContext{
 		fs:         fs,
 		deleteFile: filepath.Clean(deleteFile),
-		knownApis:  api.NewAPIs(),
+		knownApis:  api.NewAPIs().Filter(api.RemoveDisabled),
```
UnseenWizzard, line 74: This single line the 'real' fix, right? I'm okay adding refactoring changes to a PR, but as you already started, it would make it easier for me to review if things are nicely split into separate commits for refactoring and the change.

pkg/delete/delete_loader_test.go
```diff
@@ -173,6 +173,7 @@ func TestParseDeleteFileDefinitionsWithInvalidDefinition(t *testing.T) {
 }
 
 func TestLoadEntriesToDelete(t *testing.T) {
+	t.Setenv(featureflags.Experimental().EnvName(), "true")
```
UnseenWizzard, line 176: might have missed it, but could you please add covering the fact that delete filters out disabled APIs? So this can't accidentally break in the future. 
E.g. have a test that sets up a dedicated API map with an experimental one (so it's not impacted by setting current ones to non-experimental), then assert that there no delete entries loaded from a file that defines that API.

cmd/monaco/internal/test/cmd.go
```diff
@@ -0,0 +1,67 @@
...
+	"regexp"
+	"strings"
+)
+
+type Cmd struct {
+	cmd string
+	fs  afero.Fs
+}
+
+func Monacof(cmd string, args ...any) *Cmd {
```
Laubi, line 32: Please add documentation to all public symbols where it's not obvious what it does. 
Especially Monacof and Cmd are not clear; WithFs, Run, etc are clear

cmd/monaco/internal/test/cmd.go
Laubi, line 1: This file should still be in /cmd/monaco/integrationtest, as it's only used there, right?

cmd/monaco/internal/test/cmd.go
Laubi, line 32: `RunMonaco` imo would be better (?)

cmd/monaco/internal/test/cmd.go
```diff
@@ -0,0 +1,67 @@
...
+	cmd = fmt.Sprintf(cmd, args...)
+
+	cmd = regexp.MustCompile(`\s+`).ReplaceAllString(cmd, " ")
```
Laubi, line 35: Please add a brief documentation what and **why** you're doing this here.

To the what: I think replacing all spaces with one space is not intended, we should not have that. E.g. passing a file with multiple empty spaces would become invalid.

e.g. `deploy "man  ifest.yaml"` would become `deploy "man ifest.yaml"`, right?

cmd/monaco/internal/test/cmd.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (cmd *Cmd) WithFs(fs afero.Fs) *Cmd {
+	cmd.fs = fs
+	return cmd
+}
+
+func (cmd *Cmd) Run() (afero.Fs, error) {
+	fs := cmd.fs
+	if fs == nil {
+		fs = afero.NewCopyOnWriteFs(afero.NewOsFs(), afero.NewMemMapFs())
```
Laubi, line 50: We have a test util for that, and please, try to use the TempFs here instead

cmd/monaco/internal/test/cmd.go
warber: Then it would be `RunMonaco(...).Run()`

PR #1373, opened 2024-02-19
fix: Parameter resolving must not escape HTML character

Laubi:
## What this PR does / Why we need it:

When loading parameter types like the value-parameter, characters like <, >, and & were escaped to their unicode-representation, due to the json unmarshaler escaping those by default. This commit changes this behavior, and those characters are no longer escaped.

This fixes a 🐛 where duplicated configurations were created, as the local (escaped) name did not match the remote (unescaped) name. Now, the correct object is updated instead.

---

A real-world example is the string that is now also a test case:
`[8/5] Disk space available < 15% (/media/datastore)` is escaped to `[8/5] Disk space available \\u003c 15% (/media/datastore)` in the parameter. However, the remote object is still called `[8/5] Disk space available < 15% (/media/datastore)`.

Since the strings don't match, and thus, the name-matching fails, we create a new object, which is not intended.

### DIFF
--- a/internal/template/template_utils.go
+++ b/internal/template/template_utils.go
@@ -17,6 +17,7 @@
 package template
 
 import (
+	"bytes"
 	"encoding/json"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/regex"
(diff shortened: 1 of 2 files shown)


UnseenWizzard (approved):
Looks good, based on the existing tests and the function comment I'd think this fix won't have any unintended consequences :+1:

PR #1375, opened 2024-02-20
feat: consider scope when generating delete file

warber:
This PR adds the feature to support scoped configurations when generating the deletefile using `monaco generate deletefile...`. The "scope" property of each delete entry will be the "name" property of the referenced config.

There is potential for code reuse to follow references and get their properties, but i didn't extracted that functionality yet.


cmd/monaco/generate/deletefile/deletefile.go
```diff
@@ -238,8 +240,49 @@ func createConfigAPIEntry(c config.Config) (persistence.DeleteEntry, error) {
 		return persistence.DeleteEntry{}, fmt.Errorf("'name' parameter was empty")
 	}
 
+	var scopeValue string
+	if api.NewAPIs()[c.Coordinate.Type].HasParent() {
+		scopeParam, ok := c.Parameters[config.ScopeParameter]
```
warber, line 245: maybe i am overlooking something but i think it needs these 10+ steps to follow a reference and get a value of a parameter from it......

jskelin (dismissed):

jskelin (approved):

PR #1376, opened 2024-02-20
feat: Only load specified projects

arthurpitman:

This PR changes from a "load and then filter" approach for projects to a "load specified projects and then load dependencies" approach.

Changing approaches requires a significant change in testing, resulting in many of the test cases moving to the project loader (project_loader_test.go).

Yes!

Users will now be able to deploy select projects using a manifest etc that includes other projects that contain errors (for example, are incomplete). This is an advantage in terms of user experience. In addition, this PR might bring a small performance improvement as irrelevant projects are never loaded.


warber:
@arthurpitman Just saw the compile error in the `integrationtest`package. I know know it is annoying - you can catch these compilation locally by running `make compile` :)

pkg/project/v2/project_loader.go
```diff
@@ -84,13 +83,95 @@ func LoadProjects(fs afero.Fs, context ProjectLoaderContext) ([]Project, []error
 		return nil, []error{fmt.Errorf("no projects defined in manifest")}
 	}
 
-	log.Info("Loading %d projects...", len(context.Manifest.Projects))
+	if len(specificProjects) == 0 {
```
warber, line 86: nit: I think you can let "filterProjectDefinitionsByProjectNames()" just return the "unfiltered" projects if an empty slice is passed.
Then you don't specifically need to check for that case here.

pkg/project/v2/project_loader.go
```diff
@@ -84,13 +83,95 @@ func LoadProjects(fs afero.Fs, context ProjectLoaderContext) ([]Project, []error
...
+	log.Info("Loading %d projects...", len(specificProjectDefinitions))
+	projects, errors := loadProjectsFromProjectDefinitions(workingDirFs, context, specificProjectDefinitions, environments)
+	if errors != nil {
+		return nil, errors
+	}
 
-	for _, projectDefinition := range context.Manifest.Projects {
-		project, projectErrors := loadProject(workingDirFs, context, projectDefinition, environments)
+	additionalDepedencyProjectNames := getAdditionalDependencyProjectNames(projects, environments)
+	if len(additionalDepedencyProjectNames) > 0 {
```
Laubi, line 102: Is this only 1 pass?

What about dependencies that are layered 3 and deeper?
`A --dependsOn--> B --dependsOn--> C`

`Load A` should load `C` as well, but if I see it correctly, you do only one pass (?)

arthurpitman: Very good point, I will test this and see - but I think you're right!

pkg/project/v2/project_loader.go
warber: >  I feel that not adding special behavior avoids the need for any documentation as the function only does what the name implies.

the name of the function does not really imply what it does with empty slices.

warber: The program flow would also be easier too, as there is then really no need to check what function would be executed if specificProjects is empty.... But it's a nit, feel free to ignore :)

pkg/project/v2/project_loader.go
arthurpitman: Now fixed, together with some additional tests

warber (approved):

Laubi (approved):

PR #1379, opened 2024-02-21
feat: Mask Emails in Account management logs

warber:
This PR introduces the type `MaskedMail` in order to mask/hide sensitive/personal information like email addresses in logs.

image

The last [commit](https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1379/commits/4422948fb60c5339053cc0f79f5d8e61fbebadb7) uses the type `MaskedMail` directly inside the model (instead of string) wheras the [commit](https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1379/commits/ef483d75fa47402ec24c5540eb622b2ab1438d52) before casts the string type into `MaskedMail` on demand. 
You can decide which version you like more.
I'd use it as a field, because it's safer and you cannot forget it. Also it covers hidden leaks e.g. test assertions etc...

Email addresses in logs of account management are masked

### DIFF
--- a/internal/secret/email.go
+++ b/internal/secret/email.go
@@ -1,0 +1,61 @@
+/*
+ * @license
+ * Copyright 2024 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package secret
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// MaskedMail is a string that masks parts of its string representation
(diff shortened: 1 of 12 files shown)


internal/secret/email.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// MaskedMail is a string that masks parts of its string representation
+// if it's an email. If it's no valid email address the value is unchanged
+type MaskedMail string
```
jskelin, line 27: I would go with `Email`. Than on caller side it would look like `secret.Email`

Laubi (approved):

internal/secret/email_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func TestMaskEmail(t *testing.T) {
+	testCases := []struct {
+		email          string
+		expectedOutput string
+	}{
+		{"test@example.com", "te***@ex***.com"},
+		{"short@ex.com", "sh***@***.com"},
+		{"a@b.co.uk", "***@***.co.uk"},
+		{"invalid", "invalid"},
```
jskelin, line 33: As I mentioned in `email.go`, I would go with the garbage in - garbage out principle. If it isn't sound email, just return `****`.

jskelin (approved):

jskelin (approved):

jskelin (approved):

UnseenWizzard (approved):

PR #1380, opened 2024-02-22
fix: Duplicate name check must take environment into account

Laubi:
The duplicate name check marks all configs that are non-unique configs and where the name is declared as duplicate. This has the intention, that we don't update the same object multiple times in the same source config set.

However, a bug was present that if the same config was defined for multiple environments, that the duplicate name flag was set, since the config existed multiple times with the same name, **but for different environments**. This resulted in configs marked as duplicates, and thus not updating the remote object and creating a new one

This is now fixed by taking the target env into account during the check

### DIFF
--- a/pkg/project/v2/project_loader.go
+++ b/pkg/project/v2/project_loader.go
@@ -140,9 +140,10 @@ func loadProject(fs afero.Fs, context ProjectLoaderContext, projectDefinition ma
 		return Project{}, errors
 	}
 
-	// find and memoroize (non-unique-name) configurations with identical names and set a special parameter on them
+	// find and memorize (non-unique-name) configurations with identical names and set a special parameter on them
 	// to be able to identify them later
-	nonUniqueNameConfigCount := make(map[string]int)
+	// splitting is map[environment]map[name]count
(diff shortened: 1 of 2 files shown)


internal/testutils/test.go
```diff
@@ -32,3 +33,8 @@ func FailTestOnAnyError(t *testing.T, errors []error, errorMessage string) {
 	}
 	t.FailNow()
 }
+
+// Must verifies that the error passed is not nil or stops execution
+func Must(t *testing.T, err error) {
```
warber, line 38: why not use `require.NoError` directly?
this wrapper seems a bit unnecessary.

Laubi: 🤦

warber (approved):
lgtm, should we run e2e tests?
As a bonus, if you are motivated, can you replace the assertions in the modified file with testify? then we also don't have that ugly assert2 import alias anymore :)

Laubi:
@warber E2E tests make sense, started them now.

Ad refactoring: Yes, working on them rn, but I want to get this bugfix in immediately so that it's in the next release for sure

PR #1381, opened 2024-02-22
fix: timeout for automation client

warber:
This PR adds a timeout (of one minute, without any further reasoning why exactly one minute, please comment if you disagree with that value) to the context passed to the automation http client for deploy and download.

### DIFF
--- a/pkg/download/automation/download.go
+++ b/pkg/download/automation/download.go
@@ -35,6 +35,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/download/automation/internal"
 	v2 "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project/v2"
 	"golang.org/x/exp/maps"
+	"time"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/download/automation/download.go
```diff
@@ -59,7 +60,9 @@ func Download(client *client.Client, projectName string, automationTypes ...conf
 			lg.Warn("No resource mapping for automation type %s found", at.Resource)
 			continue
 		}
-		response, err := client.List(context.TODO(), resource)
+		ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+		defer cancel()
```
UnseenWizzard, line 64: As we're doing this in a loop, we're only canceling the old timeouts once all configuration types are downloaded. Not a huge issue with 3(?) types, but in general I think it'd be better to cancel after the API call - e.g. by wrapping the API call in a method that builds context, defers cancel and returns whatever response and error were

jskelin (approved):

UnseenWizzard (approved):

PR #1386, opened 2024-02-23
fix(client): Don't attempt to build simple classic URL if source URL is not a default platform URL

UnseenWizzard:
Currently if a user wrongly configures a classic URL with an oAuth client, a simple classic
URL will be built for it (not replacing anything from the valid classic URL) and we'll continue
with the assumption that the client is now configured for a platform and classic URL.
Once we attempt to make platform API calls against the configured URL, these calls will
result in hard to decypher errors.

The addition of creating a classic URL without making an API call, has thus removed our
check and error messages catching if the given URL is not for a platform environment at
all.

To be able to nicely test the mixture of URL pattern based logic and REST API calls, the metadata functions are refactored to accept a simple client interface instead of a plain rest.Client.
This allows to nicely mock interactions by just re-writing the Get method, while being able to use fake domains matching platform environment patterns instead of being limited to using the httptest.Server's IP address as URL input.

Yes - they get helpful error messages again when configuring an oAuth client for classic environment URLs, instead of getting difficult to understand errors once we try to hit invalid APIs thinking it's a platform env.


pkg/client/metadata/metadata.go
```diff
@@ -84,11 +88,16 @@ func GetDynatraceClassicURL(ctx context.Context, client *rest.Client, environmen
 	return jsonResp.GetURL(), nil
 }
 
-func findSimpleClassicURL(ctx context.Context, client *rest.Client, environmentURL string) (url string, ok bool) {
+func findSimpleClassicURL(ctx context.Context, client client, environmentURL string) (url string, ok bool) {
 	if !featureflags.BuildSimpleClassicURL().Enabled() {
```
warber, line 92: was not touched in this PR, but i would move the feature flag check one level up and only call that function when it is not enabled.

warber (approved):
lgtm, just one unrelated nit

pkg/client/metadata/metadata.go
UnseenWizzard: IIRC I had it outside of this function when I started the original change, then moved it in to make the calling function more readable. Honestly I'm not sure what's better or worse. The calling function does more, so outside you get a nested if (if FF on && able to build valid URL, exit) at the top of a longer function. 
As is the use of the flag is less visible which is also not great.

UnseenWizzard:
E2E tests once again failed with a 504 from the Account Management API..

PR #1392, opened 2024-02-27
chore: Add snyk security scan

warber:
This PR adds the snyk github action to scan for vulnerabilities when opening a PR.
This does not (yet) contain pushing the tag to snyk.io

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -1,0 +1,20 @@
+name: Snyk Security Scan
+
+on:
+  pull_request:
+    branches: [ main ]
+  merge_group: # run if triggered as part of a merge queue
+  push:
+    branches: [ main ]
+
+jobs:
+  security:
+    name: Code security scanning alerts
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


arthurpitman (dismissed):
LGTM

arthurpitman (approved):

Laubi (approved):

.github/workflows/snyk.yml
```diff
@@ -0,0 +1,20 @@
+name: Snyk Security Scan
+
+on:
+  pull_request:
+    branches: [ main ]
```
UnseenWizzard, line 5: how does snyk determine the version it is pushing the report for? Will these be reported for main or will we have a snyk entry for each branch (that we'll have to clean up later)

UnseenWizzard (approved):

.github/workflows/snyk.yml
UnseenWizzard: approved as I'm fine if we find this out after the action is merged, but we should take a look before we consider this fully done

PR #1395, opened 2024-02-28
feat: Download dashboard sharing settings

arthurpitman:
Adds support for downloading dashboard sharing settings

- As the feature is still in development, it must be enabled using the `MONACO_FEAT_DASHBOARD_SHARE_SETTINGS` feature flag.
- One and only one dashboard sharing configuration is available per dashboard.
- For security reasons `URLs` are removed from `publicAccess`

Yes, dashboard sharing settings configurations and templates will now be downloaded into `dashboard-share-settings`

### DIFF
--- a/internal/featureflags/temporary.go
+++ b/internal/featureflags/temporary.go
@@ -81,3 +81,12 @@ func GenerateJSONSchemas() FeatureFlag {
 		defaultEnabled: true,
 	}
 }
+
+// DashboardShareSettings toggles whether the dashboard share settings are downloaded and / or deployed.
+// Introduced: 2024-02-29; v2.12.0
+func DashboardShareSettings() FeatureFlag {
+	return FeatureFlag{
+		envName:        "MONACO_FEAT_DASHBOARD_SHARE_SETTINGS",
+		defaultEnabled: false,
+	}
+}
--- a/pkg/download/classic/download.go
+++ b/pkg/download/classic/download.go
@@ -175,15 +175,14 @@ func findConfigsToDownload(client dtclient.Client, apiToDownload api.API, filter
 	log.WithFields(field.Type(apiToDownload.ID)).Debug("\tFetching all '%v' configs", apiToDownload.ID)
 
 	if apiToDownload.HasParent() {
-		parentAPI := apisToDownload[apiToDownload.Parent]
 		var res values
-		parentAPIValues, err := client.ListConfigs(context.TODO(), parentAPI)
+		parentAPIValues, err := client.ListConfigs(context.TODO(), *apiToDownload.Parent)
 		if err != nil {
 			return values{}, err
 		}
 		for _, parentAPIValue := range parentAPIValues {
 
-			if skipDownload(parentAPI, parentAPIValue, filters) {
+			if skipDownload(*apiToDownload.Parent, parentAPIValue, filters) {
 				continue
 			}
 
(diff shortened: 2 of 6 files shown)


pkg/download/classic/download_test.go
```diff
@@ -364,3 +377,74 @@ func TestDownload_MalformedResponseFromAnAPI(t *testing.T) {
...
+			ShouldBeSkippedPreDownload: func(value dtclient.Value) bool { return false },
+		},
+	}
+
+	configurations, err := Download(c, "project", apiMap, contentFilters)
+	require.NoError(t, err)
+	require.Len(t, configurations, 2, "Expected two configurations")
+	require.Len(t, configurations["PARENT_API_ID"], 1)
+	require.Len(t, configurations["CHILD_API_ID"], 1)
+	assert.Equal(t, configurations["PARENT_API_ID"][0].Coordinate.ConfigId+configurations["PARENT_API_ID"][0].Coordinate.ConfigId, configurations["CHILD_API_ID"][0].Coordinate.ConfigId, "Single child config should have the same config ID as parent")
```
arthurpitman, line 449: This is the one thing I'm not really happy with: Currently the child config ID is the parent ID concatenated to itself, see https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/52f30ebbb945362f92ee89ecb9a7308678bc6291/pkg/download/classic/download.go#L312. Ideally the `value` would know it originated from a single configuration here.

UnseenWizzard (approved):
changes looks good & I don't have any great ideas on improving the template IDs rn..

Thanks for the nicely structured commits/changes!

pkg/download/classic/download_test.go
arthurpitman: Yes, that works for me 😄 . There may also be follow up chances to use `ID()` - particularly in logs - e.g. other places in `pkg/download/classic/download.go`

pkg/download/classic/download.go
```diff
@@ -162,10 +162,17 @@ type value struct {
 	parentConfigId string
 }
 
+func (v value) ID() string {
```
warber, line 165:
```suggestion
func (v value) id() string {
```

warber (dismissed):

internal/featureflags/temporary.go
```diff
@@ -81,3 +81,12 @@ func GenerateJSONSchemas() FeatureFlag {
 		defaultEnabled: true,
 	}
 }
+
+// DashboardShareSettings toggles whether the dashboard share settings are downloaded and / or deployed.
+// Introduced: 2024-02-29; v2.12.0
+func DashboardShareSettings() FeatureFlag {
```
jskelin, line 87: Instead of introducing new one, I would suggest to go with Experimental (from `permanent.go`)

warber: Did you discuss how to continue on that? I think it is also used already for diffent apis right?
my only concern is, that at some point you don't really know what you are enabling exactly with that general FF :D

jskelin (approved):

internal/featureflags/temporary.go
arthurpitman: That was my concern - `Experimental` doesn't say what it is enabling. I would stick with `MONACO_FEAT_DASHBOARD_SHARE_SETTINGS` for this feature as it is already there. Future ones we can decide in the refinement phase.

PR #1398, opened 2024-02-29
feat: Ignore dashboard-share-settings in context of deletion

warber:
A dashboard share setting cannot be deleted.
Hence they need to be ignored during the generation of the delete file as well as during deletion in case users manually put them inside the delete file.

### DIFF
--- a/pkg/api/endpoints.go
+++ b/pkg/api/endpoints.go
@@ -24,6 +24,7 @@ const (
 	ManagementZone                       = "management-zone"
 	Autotag                              = "auto-tag"
 	Dashboard                            = "dashboard"
+	DashboardShareSettings               = "dashboard-share-settings"
 	DashboardV2                          = "dashboard-v2"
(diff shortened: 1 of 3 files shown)


pkg/api/endpoints.go
```diff
@@ -56,6 +56,14 @@ var configEndpoints = []API{
...
 		NonUniqueName:                true,
 	},
+	{
+		ID:                           "dashboard-share-settings",
+		URLPath:                      "/api/config/v1/dashboards/{SCOPE}/shareSettings",
+		Parent:                       "dashboard",
+		PropertyNameOfGetAllResponse: "dashboards",
+		NonUniqueName:                true,
+		SingleConfiguration:          true,
+	},
```
warber, line 66: copied from https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1395

Laubi (approved):

PR #1399, opened 2024-02-29
Refactor/use testify

Laubi:


cmd/monaco/download/download_integration_test.go
```diff
@@ -125,6 +125,10 @@ func TestDownloadIntegrationSimple(t *testing.T) {
 			},
 		},
 	}, compareOptions...)
+
+	if diff != "" {
```
UnseenWizzard, line 129: nit: why not just assert.Empty(diff, "objects do not match")?

UnseenWizzard (approved):

PR #1400, opened 2024-02-29
chore: delete IntelliJ run configs

warber:


.run/shared/Mocks.run.xml
```diff
@@ -0,0 +1,8 @@
+<component name="ProjectRunConfigurationManager">
```
UnseenWizzard, line 1: did you include this on purpose?

e2e.env
```diff
@@ -0,0 +1,26 @@
+# This env file can be used to provide environment variables for running e2e tests
```
UnseenWizzard, line 1: Could you please: 
1. include this in gitignore, I'd like to safeguard against accidentally checking in a file filled with secrets.
2. mention this in CONTRIBUTING.md

And it's also helpful without an IDE, by just filling it out and sourcing it before running make integration-test, isn't it? Worth mentioning IMO

warber: ad 1) yes good point
ad 2) ok

ad bonus q) i think you would need to export them if you want to use them with the `source`command

.run/shared/Mocks.run.xml
warber: i think intellij did that

Laubi (approved):

PR #1401, opened 2024-03-01
chore: cleanup remaining plain api strings

warber:
I've just replaced all remaining occurrences of raw api type strings with their constant pendant.
Note that i deliberately left out any ocurrrences in `*_test.go` files as it is simply too much!


UnseenWizzard (approved):
lgtm - sonar simply reports that endpoints v1 is largely the same as the main endpoints definition. I think the check is not required to pass, so should be good to merge

PR #1406, opened 2024-03-05
print more details in log during test cleanup in integration test

jskelin:
To enable easier debugging of failed tests, more information is added to a log for the cleanup phase of the integration test.

### DIFF
--- a/cmd/monaco/integrationtest/account/utils_test.go
+++ b/cmd/monaco/integrationtest/account/utils_test.go
@@ -23,48 +23,9 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/files"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/require"
-	"regexp"
 	"strings"
 	"testing"
 )
-
-type monacoCmd struct {
-	cmd string
-	fs  afero.Fs
-}
-
-func monaco(cmd string) *monacoCmd {
-	cmd = regexp.MustCompile(`\s+`).ReplaceAllString(cmd, " ")
-	cmd = strings.Trim(cmd, " ")
-	cmd = strings.TrimPrefix(cmd, "monaco ")
-
-	return &monacoCmd{cmd: cmd}
-}
-func monacof(cmd string, args ...any) *monacoCmd {
-	return monaco(fmt.Sprintf(cmd, args...))
-}
-
-func (cmd *monacoCmd) withFs(fs afero.Fs) *monacoCmd {
-	cmd.fs = fs
-	return cmd
-}
-
-func (cmd *monacoCmd) run() error {
-	fs := cmd.fs
-	if fs == nil {
-		fs = afero.NewCopyOnWriteFs(afero.NewOsFs(), afero.NewMemMapFs())
-	}
-	fmt.Println(cmd)
-
-	cli := runner.BuildCli(fs)
-	cli.SetArgs(strings.Split(cmd.cmd, " "))
-
-	return cli.Execute()
-}
-
-func (cmd *monacoCmd) String() string {
-	return fmt.Sprintf("%s %s", "monaco", cmd.cmd)
-}
 
 func createMZone(t *testing.T) {
 	command := "deploy resources/mzones/manifest.yaml"
--- a/cmd/monaco/integrationtest/cleanup.go
+++ b/cmd/monaco/integrationtest/cleanup.go
@@ -20,62 +20,39 @@ package integrationtest
 
 import (
 	"fmt"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/runner"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/integrationtest/utils/monaco"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/timeutils"
-	"github.com/stretchr/testify/assert"
-	"path/filepath"
-	"strings"
-	"testing"
-
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/spf13/afero"
+	"github.com/stretchr/testify/require"
+	"path/filepath"
+	"testing"
 )
 
 // CleanupIntegrationTest deletes all configs that are defined in a test manifest. It uses the CLI runner, to call the
 // deletefile.Command to generate a delete file for the test manifest, and delete.GetDeleteCommand to remove configs using
 // the generated file.
-func CleanupIntegrationTest(t *testing.T, fs afero.Fs, manifestPath string, specificEnvironments []string, suffix string) {
-
-	var envArgs []string
-	if len(specificEnvironments) > 0 {
-		envArgs = []string{
-			"--environment",
-			strings.Join(specificEnvironments, ","),
-		}
+func CleanupIntegrationTest(t *testing.T, fs afero.Fs, manifestPath string, environment string, suffix string) {
+	var env string
+	if len(environment) > 0 {
+		env = fmt.Sprintf("--environment %s", environment)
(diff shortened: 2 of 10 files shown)


cmd/monaco/integrationtest/cleanup.go
```diff
@@ -20,62 +20,44 @@ package integrationtest
...
-	cmd = runner.BuildCli(fs)
-	args = append([]string{
-		"delete",
-		"--manifest", manifestPath,
-		"--file", deleteFile,
-	}, envArgs...)
-	cmd.SetArgs(args)
-	err = cmd.Execute()
+	require.NoError(t, err)
+	{
```
warber, line 44: why beginning a new block?

cmd/monaco/integrationtest/cleanup.go
```diff
@@ -20,62 +20,44 @@ package integrationtest
...
-	var envArgs []string
-	if len(specificEnvironments) > 0 {
-		envArgs = []string{
-			"--environment",
-			strings.Join(specificEnvironments, ","),
-		}
+func CleanupIntegrationTest(t *testing.T, fs afero.Fs, manifestPath string, environment string, suffix string) {
+	var env string
+	if len(environment) > 0 {
+		env = fmt.Sprintf("--environment %s", environment)
```
warber, line 37: this results in square brackets around the elements of `environment` passed to the ui, no?

cmd/monaco/integrationtest/cleanup.go
```diff
@@ -20,62 +20,44 @@ package integrationtest
...
+			}
+		}
+	}
 
-	if err != nil {
-		t.Log("Failed to cleanup all test configurations, manual/nightly cleanup needed.")
-	} else {
-		t.Log("Successfully cleaned up test configurations.")
+	{
+		_, err := test.Monacof("monaco delete --manifest %s --file %s %s", manifestPath, deleteFile, env).WithFs(fs).Run()
```
warber, line 57: why passing monaco as the first argument, and sometimes not?
Maybe we make it more strict, like always requiring the first word to be the command and use it always like this?
Otherwise it is unclear, and one would need to look down the code what actually happens to the monaco prefix.

UnseenWizzard: this - as commented on one of the other PRs that had these changes. I'd also just drop the monaco prefix entirely. It's currently only there for printing a "full" command including default CLI binary name, and code strips it out again anyway.

jskelin: With monaco it mimics the actual CLI command. Without it, it requires more internal knowledge of how things work. I found it helpful to have a `monaco` prefix when searching for a command.

My only argument for `monaco` prefix is that it looks nicer prefix :smiling_face_with_three_hearts: 

@arthurpitman , do you have some opinion on this
@UnseenWizzard , @warber Just make jour verdict, and I'll implement it.

Regarding showing/printing the command, it will always print monaco prefix. The idea is to print the exact command that is executed.

cmd/monaco/integrationtest/account/utils_test.go
```diff
@@ -23,49 +23,10 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/files"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/require"
-	"regexp"
 	"strings"
 	"testing"
 )
 
-type monacoCmd struct {
-	cmd string
```
arthurpitman, line 32: could we just use a slice of strings here rather than splitting them later?

cmd/monaco/integrationtest/account/utils_test.go
```diff
@@ -23,49 +23,10 @@ import (
...
-	cmd = strings.Trim(cmd, " ")
-	cmd = strings.TrimPrefix(cmd, "monaco ")
-
-	return &monacoCmd{cmd: cmd}
-}
-func monacof(cmd string, args ...any) *monacoCmd {
-	return monaco(fmt.Sprintf(cmd, args...))
-}
-
-func (cmd *monacoCmd) withFs(fs afero.Fs) *monacoCmd {
```
arthurpitman, line 47: Do we ever need a `monacoCmd` without a `fs`?

cmd/monaco/integrationtest/account/utils_test.go
```diff
@@ -23,49 +23,10 @@ import (
...
-
-func (cmd *monacoCmd) run() error {
-	fs := cmd.fs
-	if fs == nil {
-		fs = afero.NewCopyOnWriteFs(afero.NewOsFs(), afero.NewMemMapFs())
-	}
-	fmt.Println(cmd)
-
-	cli := runner.BuildCli(fs)
-	cli.SetArgs(strings.Split(cmd.cmd, " "))
```
arthurpitman, line 60: Could we just provide the arguments directly here? Splitting on a space feels dangerous

cmd/monaco/integrationtest/cleanup.go
warber: i misread it sorry, then it's fine oc

cmd/monaco/integrationtest/account/utils_test.go
jskelin: As this is nonproduction code and the input is static (the calling command will stay the same), I don't feel it is a threat.

cmd/monaco/integrationtest/account/utils_test.go
jskelin: In some cases, the output of one command is input to another. As the output of monaco is always on disk, it is necessary to chain the

cmd/monaco/integrationtest/account/utils_test.go
jskelin: It can be changed, it is internal implementation

cmd/monaco/integrationtest/cleanup.go
UnseenWizzard: I'm fine with having the `monaco` prefix - and even requiring it. I'm also fine if we drop it completely.
My only strong opinion is that I don't want it to sometimes have the prefix and sometimes not.

If I'd add a new test using this I'll look at what is there already, so having different ways it's used can be confusing.

arthurpitman: My preferred option would be with also a highly unique prefix so I can search for it even more easily that just with `monaco`. Personally, I rarely run things from the command line directly for testing but prefer launch configs set up in my IDE, but I do see the value in clearly seeing what command was run.

jskelin: `monaco` prefix is required than.

cmd/monaco/integrationtest/account/utils_test.go
jskelin: I change the implementation to a single step command. A separate command is available in case FS needs to be provided.

cmd/monaco/integrationtest/v2/config_restore_e2e_test.go
```diff
@@ -205,23 +206,21 @@ func TestDownloadWithSpecificAPIsAndSettings(t *testing.T) {
...
 
 	for _, tt := range tests {
-		RunIntegrationWithCleanup(t, configsFolder, configsFolderManifest, "", tt.name, func(fs afero.Fs, _ TestContext) {
-			t.Run(tt.name, func(t *testing.T) {
-
-				t.Log("Deploying configs")
-				cmd := runner.BuildCli(fs)
-				cmd.SetArgs([]string{"deploy", "-v", configsFolderManifest})
-				err := cmd.Execute()
+		tc := tt
```
warber, line 209: is this needed?

warber (dismissed):

arthurpitman (dismissed):

warber (approved):

PR #1407, opened 2024-03-05
feat: add api matcher (for use with gomock)

jskelin:
This matcher is for easier use with gomock.


arthurpitman (dismissed):

UnseenWizzard (approved):
thanks for moving it out of the public api package as discussed 👍

PR #1410, opened 2024-03-06
Fix: Proper handling of user actions and session properties

warber:
Fix for how user-actions-and-session-properties api is handled.
Most comments about the "why" are in the commits.

Added a simple integration test that for the moment asserts that expected configs are available.
This should be enhanced!

e2e test works but shows some error due to the fact that automatic cleanup (delete of scoped) configs is not working atm.

### DIFF
--- a/cmd/monaco/integrationtest/v2/scoped_configs_test.go
+++ b/cmd/monaco/integrationtest/v2/scoped_configs_test.go
@@ -1,0 +1,46 @@
+//go:build integration
+
+/**
+ * @license
+ * Copyright 2020 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package v2
+
+import (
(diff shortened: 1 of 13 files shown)


cmd/monaco/integrationtest/v2/scoped_configs_test.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/integrationtest"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/runner"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestDeployScopedConfigurations(t *testing.T) {
```
jskelin, line 30: is there a way to have a "unit" test with the mocked client? Not a must, but it would be nice to start with that practice

jskelin (approved):

jskelin (approved):

jskelin (approved):

pkg/deploy/internal/classic/validation.go
```diff
@@ -42,7 +42,7 @@ func (v *Validator) Validate(c config.Config) error {
 	}
 
 	a, ok := c.Type.(config.ClassicApiType)
-	if !ok || apis[a.Api].NonUniqueName {
+	if !ok || apis[a.Api].NonUniqueName || apis[a.Api].HasParent() {
```
arthurpitman, line 45: While this indicates that they _could_ have a different parent is it guaranteed? What if the scope is the same and the name is also the same?

warber: true, and good catch :) - i should have mentioned that we already have a ticket that's is targeting this particular issue

PR #1411, opened 2024-03-06
chore: Generate SBOM in JSON format

Churro:

To align with other SBOM initiatives, this PR changes the SBOM output format from XML to JSON and the used `args` to assert licenses, so that they are stored in the `licenses` section of the BOM. Since node.js 16 actions are [deprecated](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/), the integration is replaced by the go package that it wraps (also avoiding rate limit issues as occurred during the 2.11.0 release).

I've manually generated and validated the expected JSON output.

The SBOM is now published as a JSON document.


warber (approved):
lgtm

PR #1415, opened 2024-03-12
Fix: Correctly identify Key-User-Actions for web applciations

warber:
KUAs are identifies by name+domain+action type, and not by name only.
Since this information is present in the payload we need a way to compare their values to existing dynatrace objects.

I've introduced a custom function that can be registered on the API level to do this job.
This function can/should be re-used e.g. when dealing with deletion use case.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -63,6 +63,8 @@ type API struct {
 	Parent *API
 	// AppliedParentObjectID is the parent object ID for a SubPath API once it has been applied.
 	AppliedParentObjectID string
+	// CheckEqualFunc can be used to compare existing objects with current objects based on their payload
+	CheckEqualFunc func(map[string]any, map[string]any) bool
 	// RequireAllFF lists all feature flags that needs to be enabled in order to utilize this API
 	RequireAllFF []featureflags.FeatureFlag
 	// PropertyNameOfIdentifier defines the id field if it's not called 'ID'
--- a/pkg/api/endpoints.go
+++ b/pkg/api/endpoints.go
@@ -466,6 +466,11 @@ var configEndpoints = []API{
 		Parent:                       &applicationWebAPI,
 		RequireAllFF:                 []featureflags.FeatureFlag{featureflags.MRumProperties()},
 		TweakResponseFunc:            func(m map[string]any) { delete(m, "meIdentifier") },
+		CheckEqualFunc: func(existing map[string]any, current map[string]any) bool {
+			return existing["name"] == current["name"] &&
+				existing["actionType"] == current["actionType"] &&
+				existing["domain"] == current["domain"]
+		},
 	},
 	{
 		ID:           UserActionAndSessionPropertiesMobile,
--- a/pkg/client/dtclient/config_client.go
+++ b/pkg/client/dtclient/config_client.go
@@ -42,7 +42,7 @@ func (d *DynatraceClient) upsertDynatraceObject(ctx context.Context, theApi api.
 		// Single configuration APIs don't have an id which allows skipping this step
 		if !theApi.SingleConfiguration {
 			var err error
-			existingObjectID, err = d.getObjectIdIfAlreadyExists(ctx, theApi, theApi.CreateURL(d.environmentURLClassic), objectName)
+			existingObjectID, err = d.getObjectIdIfAlreadyExists(ctx, theApi, theApi.CreateURL(d.environmentURLClassic), objectName, payload)
 			if err != nil {
 				return DynatraceEntity{}, err
 			}
@@ -421,8 +421,8 @@ func getLocationFromHeader(resp rest.Response) ([]string, bool) {
 	return make([]string, 0), false
 }
 
-func (d *DynatraceClient) getObjectIdIfAlreadyExists(ctx context.Context, api api.API, url string, objectName string) (string, error) {
-	values, err := d.getExistingValuesFromEndpoint(ctx, api, url)
+func (d *DynatraceClient) getObjectIdIfAlreadyExists(ctx context.Context, theAPI api.API, url string, objectName string, payload []byte) (string, error) {
+	values, err := d.getExistingValuesFromEndpoint(ctx, theAPI, url)
 
 	if err != nil {
 		return "", err
@@ -432,11 +432,22 @@ func (d *DynatraceClient) getObjectIdIfAlreadyExists(ctx context.Context, api ap
 	var matchingObjectsFound = 0
 	for i := 0; i < len(values); i++ {
 		value := values[i]
-		if value.Name == objectName || escapeApiValueName(ctx, value) == objectName {
-			if matchingObjectsFound == 0 {
-				objectId = value.Id
+		if theAPI.CheckEqualFunc != nil {
+			var pl map[string]any
+			if err = json.Unmarshal(payload, &pl); err != nil {
+				return "", err
 			}
-			matchingObjectsFound++
+			if theAPI.CheckEqualFunc(value.CustomFields, pl) {
+				objectId = value.Id
+				matchingObjectsFound++
+			}
+		} else {
+			if value.Name == objectName || escapeApiValueName(ctx, value) == objectName {
+				if matchingObjectsFound == 0 {
+					objectId = value.Id
+				}
+				matchingObjectsFound++
+			}
 		}
 	}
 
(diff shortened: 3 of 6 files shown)


pkg/client/dtclient/config_client.go
```diff
@@ -414,11 +414,22 @@ func (d *DynatraceClient) getObjectIdIfAlreadyExists(ctx context.Context, api ap
...
+			var pl map[string]any
+			if err = json.Unmarshal(payload, &pl); err != nil {
+				return "", err
+			}
+			if theAPI.CheckEqualFunc(value.CustomFields, pl) {
 				objectId = value.Id
+				matchingObjectsFound++
+			}
+		} else {
+			if value.Name == objectName || escapeApiValueName(ctx, value) == objectName {
```
warber, line 427: this can/should be refactored in it's own CheckEqualFunc.
Maybe a follow up

pkg/client/dtclient/config_client.go
```diff
@@ -605,8 +618,9 @@ func unmarshalJson(ctx context.Context, theApi api.API, resp rest.Response) ([]V
 		}
 		for _, kua := range jsonResp.List {
 			values = append(values, Value{
-				Id:   kua.MeIdentifier,
-				Name: kua.Name,
+				Id:           kua.MeIdentifier,
+				Name:         kua.Name,
+				CustomFields: map[string]any{"name": kua.Name, "domain": kua.Domain, "actionType": kua.ActionType},
```
warber, line 623: i thought about populating CustomFields in general and always, however i think it causes too much memory overhead, compared to what we need it for

pkg/api/endpoints.go
```diff
@@ -466,6 +466,11 @@ var configEndpoints = []API{
 		Parent:                       &applicationWebAPI,
 		RequireAllFF:                 []featureflags.FeatureFlag{featureflags.MRumProperties()},
 		TweakResponseFunc:            func(m map[string]any) { delete(m, "meIdentifier") },
+		CheckEqualFunc: func(existing map[string]any, current map[string]any) bool {
+			return existing["name"] == current["name"] &&
+				existing["actionType"] == current["actionType"] &&
+				existing["domain"] == current["domain"]
```
arthurpitman, line 472: Probably could make `name`, `actionType`, and `domain` constants as we use them a few times

warber:
What I really don't like about my current solution is, that all of that is happening in the scope of the `doXByNAME` function... Maybe I can separate that better

pkg/client/dtclient/config_client.go
UnseenWizzard: Agree with this being unnecessary overhead in general - but we could consider if we can use it for any other special case APIs that we already have. e.g. dashboards have some extra fields that are just part of the API value struct now

pkg/api/api.go
```diff
@@ -61,7 +61,10 @@ type API struct {
 	TweakResponseFunc func(map[string]any)
 	// Parent is used for SubPath APIs to store information about the configuration type and ID of the related
 	// configuration once Resolved() is called.
-	Parent *API
+
+	// CheckEqualFunc can be used to compare existing objects with current objects based on their payload
+	CheckEqualFunc func(map[string]any, map[string]any) bool
+	Parent         *API
```
UnseenWizzard, line 67: you've accidentally removed this field from its comment - should stay on line 64

pkg/client/dtclient/config_client.go
```diff
@@ -414,11 +414,22 @@ func (d *DynatraceClient) getObjectIdIfAlreadyExists(ctx context.Context, api ap
...
 		value := values[i]
-		if value.Name == objectName || escapeApiValueName(ctx, value) == objectName {
-			if matchingObjectsFound == 0 {
+		if theAPI.CheckEqualFunc != nil {
+			var pl map[string]any
+			if err = json.Unmarshal(payload, &pl); err != nil {
+				return "", err
+			}
+			if theAPI.CheckEqualFunc(value.CustomFields, pl) {
 				objectId = value.Id
```
UnseenWizzard, line 423: In this case you overwrite the ID if more than one value matches, while so far we'd use the 1st ID (line 418 before, line 428 now) if more than one object match - as we seem to continue with just a warning log, I would not introduce different behavior for APIs with a check equal func and also only take the ID of the first matching object.

pkg/client/dtclient/config_client.go
```diff
@@ -485,7 +411,48 @@ func isUserSessionPropertiesMobile(a api.API) bool {
...
+		values, err := d.fetchExistingValues(ctx, theApi, theApi.CreateURL(d.environmentURLClassic))
+		if err != nil {
+			return "", err
+		}
+		return d.findUniqueByName(ctx, theApi, values, objectName)
+
+	}
+	// The network-zone API doesn't have a POST endpoint, hence, we need to treat it as an update operation
+	// per default
+	if theApi.ID == api.UserActionAndSessionPropertiesMobile {
```
UnseenWizzard: as these are now all returning directly, can you please structure this method so that they are early exit conditions? 

only if none of these ifs match - and the API is not a `SingleConfiguration` one (make early exit too), do we need to make any API calls to fetch existing values.

Suggestion: 

```go
if theApi.ID == api.UserActionAndSessionPropertiesMobile { return objectName }
if theApi.ID == api.NetworkZone { return objectName }
if theApi.ID == api.CalculatedMetricsLog { return objectName }
if theApi.SingleConfiguration { return "" }

values, err := d.fetchExistingValues(ctx, theApi, theApi.CreateURL(d.environmentURLClassic))
if err != nil {
	return "", err
}

if theApi.CheckEqualFunc != nil { return d.FindUnique } 

return d.FindUniqueByName
```

pkg/client/dtclient/config_client.go
```diff
@@ -42,6 +42,24 @@ func (d *DynatraceClient) upsertDynatraceObject(ctx context.Context, theApi api.
 			return DynatraceEntity{}, err
 		}
 
+		// The network-zone API doesn't have a POST endpoint, hence, we need to treat it as an update operation
```
UnseenWizzard, line 45: why did you move these back out after the fetching on IDs?

pkg/client/dtclient/config_client.go
```diff
@@ -627,24 +627,35 @@ func unmarshalJson(ctx context.Context, theApi api.API, resp rest.Response) ([]V
...
-				Domain       string `json:"domain"`
-				ActionType   string `json:"actionType"`
-			} `json:"keyUserActionList"`
-		}
+		var jsonResp map[string]any
 		err := json.Unmarshal(resp.Body, &jsonResp)
+
+		kuaL, ok := jsonResp[api.KeyUserActionsListField].([]any)
+		if !ok {
+			return nil, fmt.Errorf("unable to find field %s in key user alction list response", api.KeyUserActionsListField)
```
UnseenWizzard, line 635: `alction` typo in each `action`

pkg/client/dtclient/config_client.go
warber: because they are in the end deciding about whether an update or a create will be executed...
having that early return in another function hides that side effect entirely

pkg/client/dtclient/config_client.go
```diff
@@ -573,6 +540,62 @@ func (d *DynatraceClient) getExistingValuesFromEndpoint(ctx context.Context, the
...
+		}
+		if checkEqualFunc(value.CustomFields, pl) {
+			if matchingObjectsFound == 0 {
+				objectId = value.Id
+			}
+			matchingObjectsFound++
+		}
+	}
+
+	return objectId, nil
```
UnseenWizzard, line 596: you're now missing the warning and debug logs that the findUniqueByName has on lines 556-562. I think it's worth having that information in the logs for custom equals functions too

pkg/client/dtclient/config_client_test.go
```diff
@@ -408,14 +408,12 @@ func Test_getObjectIdIfAlreadyExists(t *testing.T) {
...
 
 			dtclient, _ := NewDynatraceClientForTesting(server.URL, server.Client(), nil)
-			got, err := dtclient.getObjectIdIfAlreadyExists(context.TODO(), testApi, server.URL, tt.givenObjectName, nil)
+
+			_, _, err := dtclient.configExistsByName(context.TODO(), testApi, tt.givenObjectName)
 			if (err != nil) != tt.wantErr {
 				t.Errorf("getObjectIdIfAlreadyExists() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
-			if got != tt.wantFoundId {
```
UnseenWizzard, line 416: why did you remove this check? If it is really not needed please remove the `wantFoundId` field from the testcases =

warber: accidentally

UnseenWizzard (approved):
thx for incorporating all comments

arthurpitman (approved):

PR #1419, opened 2024-03-14
Fix: Improve uniqueness checks for subpath APIs by taking scope into account

arthurpitman:
This PR adds stricter validation for sub-path APIs, only allowing the same name if the scope is different.

`key-user-action-web` configs are explicitly excluded as the uniqueness of these is determined by three fields in the payload

Yes, improved validation for classic configs for sub-path APIs: increases the chance of the user being presented with an error if the configs are not unique.

### DIFF
--- a/pkg/deploy/internal/classic/validation.go
+++ b/pkg/deploy/internal/classic/validation.go
@@ -42,7 +42,17 @@ func (v *Validator) Validate(c config.Config) error {
 	}
 
 	a, ok := c.Type.(config.ClassicApiType)
-	if !ok || apis[a.Api].NonUniqueName || apis[a.Api].HasParent() {
+	if !ok {
+		return nil
+	}
+
+	theAPI := apis[a.Api]
+	if theAPI.NonUniqueName {
+		return nil
+	}
+
+	// as the uniqueness of a key-user-action-web configuration is defined by its payload no validation can be performed
+	if a.Api == api.KeyUserActionsWeb {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/deploy/internal/classic/validation.go
```diff
@@ -51,6 +61,24 @@ func (v *Validator) Validate(c config.Config) error {
 	}
 
 	for _, c2 := range v.uniqueNames[c.Environment][a.Api] {
+
+		// if the configs have a scope and it is different then the configs are unique
+		if theAPI.HasParent() {
+			s1, err := config.GetScopeParameterForConfig(c)
+			if err != nil {
+				return err
```
UnseenWizzard, line 69: Do we actually need to treat not having a scope defined as an error here? Shouldn't loading already fail and ensure that a scope is set for APIs that need one? 

Then we could simplify this to accessing the scope parameters as `c.Parameters[ScopeParameter]` and comparing them? If one accidentally happens to be nil, then they are still different for the sake of validation

arthurpitman: No idea, I'm going to test this. But for the moment, I have put a marker in the error message so that I can see if it occurs

arthurpitman: OK, after some testing, it is possible to hit this error by having `scope: null` in the `config.yaml`

It is interesting that we get different behavior for `scope: ~`, `scope: {}` and `scope: null`.

warber (approved):

UnseenWizzard (approved):

warber (approved):

UnseenWizzard (approved):
:shipit:

PR #1420, opened 2024-03-14
Fix: for key-user-action-web that has the same name but a different domain (or actionType) create a correct template yaml file

jskelin:
For key-user-action-web that has the same name but a different domain (or actionType) create a correct template yaml file

### DIFF
--- a/pkg/download/classic/download.go
+++ b/pkg/download/classic/download.go
@@ -19,6 +19,7 @@ package classic
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
@@ -279,13 +280,29 @@ func downloadAndUnmarshalConfig(client dtclient.Client, theApi api.API, value va
 		return nil, err
 	}
 
-	if values, ok := data[theApi.PropertyNameOfGetAllResponse]; ok {
-		var res []map[string]any
-		err := mapstructure.Decode(values, &res)
-		return res, err
+	values, found := data[theApi.PropertyNameOfGetAllResponse]
+	if !found {
+		return []map[string]any{data}, nil
 	}
 
-	return []map[string]any{data}, nil
+	var res []map[string]any
+	err = mapstructure.Decode(values, &res)
+	if err != nil {
+		return []map[string]any{}, err
+	}
+	if theApi.ID == api.KeyUserActionsWeb { //clean unwanted configs
+		return filterResponses(res, value)
+	}
+	return res, nil
+}
+
+func filterResponses(res []map[string]any, value value) ([]map[string]any, error) {
+	for _, v := range res {
+		if v["meIdentifier"] == value.value.Id {
+			return []map[string]any{v}, nil
+		}
+	}
+	return nil, fmt.Errorf("unable to find %q configuration with ID %q", api.KeyUserActionsWeb, value.value.Id)
 }
 
 func createConfigForDownloadedJson(mappedJson map[string]interface{}, theApi api.API, value value, projectId string) (config.Config, error) {
--- a/pkg/download/classic/download_test.go
+++ b/pkg/download/classic/download_test.go
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package classic
+package classic_test
 
 import (
 	"context"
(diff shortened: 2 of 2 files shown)


pkg/download/classic/download.go
```diff
@@ -282,12 +282,25 @@ func downloadAndUnmarshalConfig(client dtclient.Client, theApi api.API, value va
 	if values, ok := data[theApi.PropertyNameOfGetAllResponse]; ok {
 		var res []map[string]any
 		err := mapstructure.Decode(values, &res)
+		if theApi.ID == api.KeyUserActionsWeb {
+			res = filterResponses(res, value)
```
arthurpitman, line 286: Is there a reason this step is performed regardless of whether an error occurred or not?

pkg/download/classic/download.go
```diff
@@ -282,12 +282,27 @@ func downloadAndUnmarshalConfig(client dtclient.Client, theApi api.API, value va
...
 		err := mapstructure.Decode(values, &res)
+		if err == nil {
+			if theApi.ID == api.KeyUserActionsWeb {
+				res = filterResponses(res, value)
+			}
+		}
 		return res, err
 	}
 
 	return []map[string]any{data}, nil
```
arthurpitman, line 293:
```suggestion
    values, found := data[theApi.PropertyNameOfGetAllResponse]
    if !found {
    	return []map[string]any{data}, nil
    }
    
	var res []map[string]any
	err := mapstructure.Decode(values, &res)
	if err != nil {
	 	return res, err // not sure if we need `res` here or not
	}
	
	if theApi.ID == api.KeyUserActionsWeb {
		res = filterResponses(res, value)
	}

	return res, nil
```

pkg/download/classic/download.go
```diff
@@ -282,12 +282,27 @@ func downloadAndUnmarshalConfig(client dtclient.Client, theApi api.API, value va
...
 }
 
+func filterResponses(res []map[string]any, value value) []map[string]any {
+	for _, v := range res {
+		if v["meIdentifier"] == value.value.Id {
+			res = append([]map[string]any{}, v)
+			break
```
arthurpitman, line 300:
```suggestion
			return []map[string]any{v}
```

pkg/download/classic/download_test.go
```diff
@@ -305,57 +184,127 @@ func TestDownload_FilteringCanBeTurnedOffViaFeatureFlags(t *testing.T) {
...
-	c.EXPECT().ListConfigs(gomock.Any(), gomock.Any()).DoAndReturn(func(ctx context.Context, a api.API) ([]dtclient.Value, error) {
-		if a.ID == "API_ID_1" {
-			return []dtclient.Value{{Id: "API_ID_1", Name: "API_NAME_1"}}, nil
-		} else if a.ID == "API_ID_2" {
-			return []dtclient.Value{}, nil
-		}
-		return nil, nil
-	}).Times(2)
-	c.EXPECT().ReadConfigById(gomock.Any(), gomock.Any()).Return([]byte("{}"), nil)
+func Test_generalCases(t *testing.T) {
```
arthurpitman, line 212: nit: I'm not convinced of the benefit these being a table test here

jskelin: I understand your problems, but let's go for now with this...

pkg/download/classic/download.go
```diff
@@ -279,13 +279,29 @@ func downloadAndUnmarshalConfig(client dtclient.Client, theApi api.API, value va
...
+		return []map[string]any{data}, nil
 	}
 
-	return []map[string]any{data}, nil
+	var res []map[string]any
+	err = mapstructure.Decode(values, &res)
+	if err != nil {
+		return []map[string]any{}, err
+	}
+	if theApi.ID == api.KeyUserActionsWeb {
```
warber, line 292: whenever we add special treatment with an if like that, please explain why with a short comment

pkg/download/classic/download.go
```diff
@@ -279,13 +279,29 @@ func downloadAndUnmarshalConfig(client dtclient.Client, theApi api.API, value va
...
+	return res, err
+}
+
+func filterResponses(res []map[string]any, value value) []map[string]any {
+	for _, v := range res {
+		if v["meIdentifier"] == value.value.Id {
+			return []map[string]any{v}
+		}
+	}
+	return res
```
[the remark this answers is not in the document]
jskelin: The default return should be nil!

pkg/download/classic/download.go
warber: i mean it is not clear why we need to inspect the payload and do all of that for that config type ;)

pkg/download/classic/download.go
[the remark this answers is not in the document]
warber: Since it was found a couple of lines above via "findConfigsToDownload" it must be found here as well.
I would say we should log it.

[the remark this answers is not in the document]
jskelin: It is a valid situation to return nil!
UC: application-web doesn't have any key-user-action. In that case, the return should be nil (or empty map)., not an error.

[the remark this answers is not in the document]
warber: so you can find a meID valud with findConfigsToDownload but are not able to find it via that filtering?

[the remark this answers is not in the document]
warber: oh ok, i see, since this is practically just one api call in before, we are not able to run into that situation.

warber (approved):

PR #1421, opened 2024-03-15
feat: generate delete entries for key user actions web

warber:


pkg/delete/persistence/delete_persistence.go
```diff
@@ -50,6 +54,22 @@ type DeleteEntry struct {
 
 	// Scope is the parent scope of a config. This field must be set if a classic config is used, and the classic config requires the scope to be set.
 	Scope string `yaml:"scope,omitempty" json:"scope,omitempty" mapstructure:"scope" jsonschema:"description=The scope of the config to be deleted - required for API configs that require a scope"`
+	// CustomValues holds special values that are not general enough to add as a field to a DeleteEntry but are still important for specific APIs
+	CustomValues map[string]string `yaml:",inline"`
+}
+
+func (d DeleteEntry) HashCode() string {
```
warber, line 61: I've added a map to the fields of `DeleteEntry` struct. (see line 58) Since maps are not Comparable `DeleteEntry` values cannot be used e.g. as keys in a `map` (which is the case on some places). So instead of the `DeleteEntry`value the string returned by `HashCode()`can be used as a map key instead.

cmd/monaco/generate/deletefile/deletefile.go
```diff
@@ -257,11 +261,42 @@ func createConfigAPIEntry(c config.Config, apis api.APIs, project project.Projec
 			return persistence.DeleteEntry{}, fmt.Errorf("resolved name parameter is no string")
 		}
 		scopeValue = nameOfRefCfgStr
+
+		if apis[c.Coordinate.Type].ID == api.KeyUserActionsWeb {
+			// while generating the delete file (which happens offline) we are not able
+			// to resolve that parameter, so we delete it
+			delete(c.Parameters, config.ScopeParameter)
+			props, resolvErr := c.ResolveParameterValues(entities.New())
```
UnseenWizzard, line 269: won't this also fail if the config has any other reference parameter defined?

cmd/monaco/generate/deletefile/deletefile.go
```diff
@@ -257,11 +261,42 @@ func createConfigAPIEntry(c config.Config, apis api.APIs, project project.Projec
...
+				return persistence.DeleteEntry{}, err
+			}
+
+			customValues["domain"] = jsonM["domain"]
+			customValues["actionType"] = jsonM["actionType"]
+		}
 	}
 
+	for _, s := range []string{"project", "type", "id", "name", "scope"} {
+		if _, ok = customValues[s]; ok {
```
UnseenWizzard, line 290: I don't get the point of this loop and the condition seems to not be covered by tests. 
You only ever write `domain` and `actionType` into the customValues map, but then check that it does not contain `project`, `type`, etc. keys - which I won't unless I'm missing where you've written them into the map. Did you mean to validate the content of the fields?

warber: If you reuse the customValues for a later use case (for which CustomValues was built) you are not allowed to use theses fields, as it would cause a crash of the program.

cmd/monaco/generate/deletefile/deletefile.go
warber: of course 🤦

cmd/monaco/generate/deletefile/deletefile.go
warber: maybe it would be better to error out here tho 🤔

cmd/monaco/generate/deletefile/deletefile.go
```diff
@@ -185,6 +189,20 @@ func skipping(ttype string, included, excluded map[string]struct{}) bool {
 	return false
 }
 
+func toMapKey(d persistence.DeleteEntry) string {
+	return fmt.Sprintf("%s/%s/%s/%s/%s/%s", d.Project, d.Type, d.ConfigId, d.ConfigName, d.Scope, sortedKeyValues(d.CustomValues, "/"))
```
arthurpitman, line 193: Would it matter if the same key was generated for different entries? Here it is certainly possible, even if unlikely.

cmd/monaco/generate/deletefile/deletefile.go
```diff
@@ -185,6 +189,28 @@ func skipping(ttype string, included, excluded map[string]struct{}) bool {
...
 
+func toMapKey(d persistence.DeleteEntry) string {
+	values := make(map[string]string)
+	values["project"] = d.Project
+	values["type"] = d.Type
+	values["configId"] = d.ConfigId
+	values["configName"] = d.ConfigName
+	values["scope"] = d.Scope
+	maps.Copy(values, d.CustomValues)
+	j, _ := json.Marshal(sortedKeyValues(values)) //nolint:errchkjson
```
UnseenWizzard, line 200: I know @arthurpitman suggested this, but: WHY?

We're now building a map, copying over the entries from CustomValues, sorting it and finally marshalling to JSON.
So just to get a unique string key to use for a map, we add the overhead of allocating a map and marshaling to JSON - where before we simply rendered the string with fmt.Sprintf.

I'm missing the benefit this refactoring provides tbh

cmd/monaco/generate/deletefile/deletefile.go
```diff
@@ -185,6 +189,28 @@ func skipping(ttype string, included, excluded map[string]struct{}) bool {
 	return false
 }
 
+func toMapKey(d persistence.DeleteEntry) string {
```
UnseenWizzard, line 192: nit:  a short doc comment for the method would be nice so it's clear what it's for/why we build string keys and can't use the DeleteEntries directly.

cmd/monaco/generate/deletefile/deletefile.go
warber: more specifically: 
```
func TestClash(t *testing.T) {
	type My struct {A, B string}
	my := My{A: "h/", B: "i"}
	my2 := My{A: "h", B: "/i"}
	fmt.Printf("%s/%s\n", my.A, my.B)   // Output: h//i
	fmt.Printf("%s/%s\n", my2.A, my2.B) // Output: h//i
}

```

UnseenWizzard: I'm missing how 
```
fmt.Sprintf("%s/%s/%s/%s/%s/%s", d.Project, d.Type, d.ConfigId, d.ConfigName, d.Scope, sortedKeyValues(d.CustomValues, "/"))
```

produced clashes that the map -> json string of the same object avoids?

UnseenWizzard: Ah, saw your sample after I sent my question.

I'd personally lean towards this type of clash being exceedingly unlikely  - it can only happen between config-name, scope, and the custom values (and not for the domain and actionType custom values we have so far, as only domain is freely definable and can't start with a /).
And we could still choose another even less likely separator (e.g. `{%$`). 

But as this is only in the scope of delete file generation which runs on its own and probably not often in a scale that makes the memory and runtime overhead noticeable, I can live with it in it's safe but less efficient form.

UnseenWizzard (approved):

PR #1423, opened 2024-03-18
Allow optionally writing coordinate as JSON in generated dependency graph DOT files

UnseenWizzard:


cmd/monaco/generate/dependencygraph/command.go
```diff
@@ -71,6 +72,8 @@ func Command(fs afero.Fs) (cmd *cobra.Command) {
 
 	cmd.Flags().StringVarP(&outputFolder, "output-folder", "o", "", "The folder generated dependency graph DOT files should be written to. If not set, files will be created in the current directory.")
 
+	cmd.Flags().BoolVar(&jsonIDs, "json-ids", false, "Set to generate a DOT file using a JSON representation of coordinates as node IDs, instead of the string representation. This can be useful when processing generated DOT files automatically.")
+
```
arthurpitman, line 76:
```suggestion
	cmd.Flags().BoolVar(&jsonIDs, "json-ids", false, "Set to generate a DOT file using a nodes IDs encoded as JSON instead of the default string representation. This can be useful when processing generated DOT files automatically.")

```

pkg/graph/graph.go
```diff
@@ -55,6 +56,9 @@ func (n ConfigNode) ID() int64 {
 
 // DOTID returns the node's identifier when printed to a DOT file. For readability of files this is the coordinate.Coordinate of the Config, instead of the node's ID integer.
 func (n ConfigNode) DOTID() string {
+	if n.DOTEncoding != "" {
+		return n.DOTEncoding
+	}
 	return n.Config.Coordinate.String()
```
arthurpitman, line 62: Just an observation: if using the "string" representation, the encoding is done late, here in this function, where as for JSON it is done much earlier, when building the graph. Is there a reason for this? All things being equal, would it be better to aways do the encoding at the same point in time?

UnseenWizzard: Personal preference would be to do it as late as possible. 
Based on your suggestion to allow the definition of the full encoding logic outside, rather than just flipping a bool option, I originally added an encoding function to the Node struct - but the realized that makes it non-comparable

cmd/monaco/generate/dependencygraph/test-resources/cycles/management-zone/zone.json
warber, line 1: midding terminating linebreak

cmd/monaco/generate/dependencygraph/test-resources/cycles/reports/report.json
warber, line 1: missing terminating linebreak

cmd/monaco/generate/dependencygraph/test-resources/cycles/management-zone/zone.json
UnseenWizzard: addressed for all test resources in this package as an added commit

UnseenWizzard:
@warber @arthurpitman addressed your comments & rebased

cmd/monaco/generate/dependencygraph/command.go
```diff
@@ -71,6 +72,8 @@ func Command(fs afero.Fs) (cmd *cobra.Command) {
 
 	cmd.Flags().StringVarP(&outputFolder, "output-folder", "o", "", "The folder generated dependency graph DOT files should be written to. If not set, files will be created in the current directory.")
 
+	cmd.Flags().BoolVar(&jsonIDs, "json-ids", false, "Set to generate a DOT file encoding each node's coordinate as JSON, instead of the default string representation. This can be useful when processing generated DOT files automatically.")
```
jskelin, line 75: I would go with a string value. For now we would have:
* string (default)
* json. 

I don't expect to have more options, but we keep the flexibility to easily add another one without breaking anything.

jskelin (approved):

jskelin (approved):

jskelin (dismissed):

cmd/monaco/generate/dependencygraph/command.go
UnseenWizzard: Good idea! I've added a commit changing this, please re-review/approve

cmd/monaco/generate/dependencygraph/dependencygraph.go
```diff
@@ -86,7 +87,20 @@ func writeGraphFiles(fs afero.Fs, manifestPath string, environmentNames []string
...
-	graphs := graph.New(projects, m.Environments.Names())
+	var opts []graph.NodeOption
+	if writeJSONIDs {
+		log.Debug("Encoding DOT Node IDs as JSON")
+		opts = append(opts, func(n *graph.ConfigNode) {
+			s, err := json.Marshal(n.Config.Coordinate)
+			if err == nil {
+				n.DOTEncoding = string(s)
+			} else {
+				n.DOTEncoding = fmt.Sprintf(`{ "encodingError": "%v" }`, err)
```
arthurpitman, line 98: Perhaps it is better to use `json.Marshal` to encode the encoding error?

UnseenWizzard: adapted as discussed internally - encoding will be an empty JSON object so we don't break any automation reading the graph and an error is logged.
In general we can still assume that this is a 'should never happen' case, as the coordinate is a simple struct of strings that should always be possible to marshal - and any substantial modification breaking marshalling, would also have impact on most of monaco.

arthurpitman (approved):

PR #1427, opened 2024-03-21
fix: for key-user-action-web that has the same name but a different d…

jskelin:
…omain (or actionType) delete correct configuration

To delete `key-user-action-web` delete.yaml is extended with `actionType` and `domain` fields. With this addition it is possible to uniquely specify configuration and delete it safely.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -17,8 +17,9 @@
 package api
 
 import (
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"strings"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 )
 
 const StandardApiPropertyNameOfGetAllResponse string = "values"
@@ -77,7 +78,7 @@ func (a API) CreateURL(environmentURL string) string {
 	return environmentURL + a.URLPath
 }
 
-// HasParent returns true if the API has a relation to another (parent) API.
+// HasParent returns true iff the API has a relation to another (parent) API.
 // This is typically the case for "Sub-path" APIs, e.g. Key User Actions for Mobile applications.
 // In this case "mobile-application" would be the parent API, which is also reflected in the URLs to be used to query
 // and create key user actions.
--- a/pkg/delete/delete_loader.go
+++ b/pkg/delete/delete_loader.go
@@ -19,13 +19,14 @@ package delete
 import (
 	"errors"
 	"fmt"
+	"path/filepath"
+	"strings"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/persistence"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/pointer"
 	"github.com/mitchellh/mapstructure"
-	"path/filepath"
-	"strings"
 
 	"github.com/spf13/afero"
 	"gopkg.in/yaml.v2"
@@ -158,7 +159,7 @@ func parseFullEntry(ctx *loaderContext, entry interface{}) (pointer.DeletePointe
 	if a, known := ctx.knownApis[parsed.Type]; known {
 		p, err := parseAPIEntry(parsed, a)
 		if err != nil {
-			return pointer.DeletePointer{}, fmt.Errorf("failed to parse entry for API %q: %w", a.ID, err)
+			return pointer.DeletePointer{}, fmt.Errorf("failed to parse entry for API '%s': %w", a.ID, err)
 		}
 		return p, nil
 	}
@@ -172,7 +173,27 @@ func parseAPIEntry(parsed persistence.DeleteEntry, a api.API) (pointer.DeletePoi
 	}
 
 	if parsed.ConfigId != "" {
-		log.Warn("Delete entry %q of API type defines config 'id' - only 'name' will be used.")
+		log.Warn("Delete entry '%s' of API type defines config 'id' - only 'name' will be used.")
+	}
+
+	if a.ID == api.KeyUserActionsWeb {
+		if parsed.Scope == "" {
+			return pointer.DeletePointer{}, fmt.Errorf("API of type '%s' requires a '%s', but none was defined", a, "scope")
+		}
+		if v := parsed.CustomValues["actionType"]; v == "" {
+			return pointer.DeletePointer{}, fmt.Errorf("API of type '%s' requires a '%s', but none was defined", a, "actionType")
+		}
+		if v := parsed.CustomValues["domain"]; v == "" {
+			return pointer.DeletePointer{}, fmt.Errorf("API of type '%s' requires a '%s', but none was defined", a, "domain")
+		}
+
+		return pointer.DeletePointer{
+			Type:       parsed.Type,
+			Identifier: parsed.ConfigName,
+			Scope:      parsed.Scope,
+			ActionType: parsed.CustomValues["actionType"],
+			Domain:     parsed.CustomValues["domain"],
+		}, nil
 	}
 
 	// The scope is required for sub-path APIs.
--- a/pkg/delete/delete_test.go
+++ b/pkg/delete/delete_test.go
@@ -21,6 +21,12 @@ package delete_test
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+	"testing"
+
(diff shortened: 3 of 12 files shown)


pkg/delete/delete_loader.go
```diff
@@ -175,6 +175,26 @@ func parseAPIEntry(parsed persistence.DeleteEntry, a api.API) (pointer.DeletePoi
 		log.Warn("Delete entry %q of API type defines config 'id' - only 'name' will be used.")
 	}
 
+	if a.ID == api.KeyUserActionsWeb {
+		if parsed.Scope == "" {
+			return pointer.DeletePointer{}, fmt.Errorf("API of type %q requires a %q scope, but non was defined", a, "scope")
```
arthurpitman, line 180:
```suggestion
			return pointer.DeletePointer{}, fmt.Errorf("API of type %q requires a %q, but none was defined", a, "scope")
```

pkg/delete/delete_loader.go
```diff
@@ -175,6 +175,26 @@ func parseAPIEntry(parsed persistence.DeleteEntry, a api.API) (pointer.DeletePoi
...
+		}
+		if v := parsed.CustomValues["actionType"]; v == "" {
+			return pointer.DeletePointer{}, fmt.Errorf("API of type %q requires a %q, but non was defined", a, "actionType")
```
arthurpitman, line 183:
```suggestion
			return pointer.DeletePointer{}, fmt.Errorf("API of type %q requires a %q, but none was defined", a, "actionType")
```

pkg/delete/delete_loader.go
```diff
@@ -175,6 +175,26 @@ func parseAPIEntry(parsed persistence.DeleteEntry, a api.API) (pointer.DeletePoi
...
+		}
+		if v := parsed.CustomValues["domain"]; v == "" {
+			return pointer.DeletePointer{}, fmt.Errorf("API of type %q requires a %q, but non was defined", a, "domain")
```
arthurpitman, line 186:
```suggestion
			return pointer.DeletePointer{}, fmt.Errorf("API of type %q requires a %q, but none was defined", a, "domain")
```

pkg/delete/delete_test.go
```diff
@@ -843,3 +837,27 @@ func TestConfigsWithParent(t *testing.T) {
 		})
 	}
 }
+
+func Test_KeyUserActionsWebUniqnes(t *testing.T) {
```
arthurpitman, line 841:
```suggestion
func Test_KeyUserActionsWebUniqueness(t *testing.T) {
```

pkg/delete/pointer/delete_pointer.go
```diff
@@ -33,6 +33,9 @@ type DeletePointer struct {
 
 	// Scope is the Entity ID / information necessary to delete the entity. This is required for sub-path entities.
 	Scope string
+
+	// ActionType and Doman plays its role only to delete api.KeyUserActionsWeb
```
arthurpitman, line 37:
```suggestion
	// ActionType and Domain are used when deleting key-user-actions-web entities
```

pkg/delete/internal/classic/delete.go
```diff
@@ -79,29 +89,33 @@ func is404(err error) bool {
...
+	id, err := findUniqueID(knownValues, identifier, theAPI.CheckEqualFunc)
 	if err != nil {
 		return "", err
 	}
 
 	return id, nil
 }
 
-func findUniqueID(knownValues []dtclient.Value, identifier string) (string, error) {
+func findUniqueID(knownValues []dtclient.Value, identifier identifier, uniqeFn func(map[string]any, map[string]any) bool) (string, error) {
```
arthurpitman, line 106: nit: maybe rename `uniqeFn`to `uniqueFn`

arthurpitman: Although, if this is really just the `api.CheckEqualFunc`, we should use that as a name, or come up with a better more general name and use it both places.

pkg/api/api.go
```diff
@@ -69,20 +69,25 @@ type API struct {
 	RequireAllFF []featureflags.FeatureFlag
 	// PropertyNameOfIdentifier defines the id field if it's not called 'ID'
 	PropertyNameOfIdentifier string
+	// child indicates that this API has child. It is field within NewAPIs function
+	haveChild bool
```
warber, line 73: If this is **only** necessary to make the cache work, i would strongly suggest to go with the nocache solution for deletion.

pkg/delete/pointer/delete_pointer.go
```diff
@@ -33,6 +33,9 @@ type DeletePointer struct {
 
 	// Scope is the Entity ID / information necessary to delete the entity. This is required for sub-path entities.
 	Scope string
+
+	// ActionType and Domain are used when deleting key-user-actions-web entities
+	ActionType, Domain string
```
warber, line 38: Why not CustomValues as above?

pkg/delete/internal/classic/client.go
```diff
@@ -18,34 +18,49 @@ package classic
...
+			c.mu.Unlock()
 		}
 	}
-	return client.listCache[api.URLPath], nil
+	return c.cache[a.URLPath], nil
+}
+
+func (c *cachedDTClient) DeleteConfigById(a api.API, id string) error {
+	// for endpoints that have a child, we
+	if a.ID == api.KeyUserActionsMobile || a.ID == api.KeyUserActionsWeb {
```
warber, line 60: is this needed?

pkg/delete/internal/classic/client.go
```diff
@@ -18,34 +18,49 @@ package classic
...
 		} else {
-			client.listCache[api.URLPath] = v
+			c.mu.Lock()
+			c.cache[a.URLPath] = v
+			c.mu.Unlock()
 		}
 	}
-	return client.listCache[api.URLPath], nil
+	return c.cache[a.URLPath], nil
+}
```
arthurpitman, line 56: Maybe the code could be simplified to:

```

func (c *cachedDTClient) ListConfigs(ctx context.Context, a api.API) ([]dtclient.Value, error) {
	if v, ok := c.cache[a.URLPath]; ok {
		return v, nil
	}

	v, err := c.Client.ListConfigs(ctx, a)
	if err != nil {
		return nil, err
	} 

	c.mu.Lock()
	c.cache[a.URLPath] = v
	c.mu.Unlock()
	return v, nil
}
```
**Question:**
But what are we trying to guard against with the locking here? I guess it is preventing concurrent writes, but it _would_ allow reads during writes, or? Could that be problematic?

pkg/delete/delete_loader.go
```diff
@@ -175,6 +175,26 @@ func parseAPIEntry(parsed persistence.DeleteEntry, a api.API) (pointer.DeletePoi
...
 
+	if a.ID == api.KeyUserActionsWeb {
+		if parsed.Scope == "" {
+			return pointer.DeletePointer{}, fmt.Errorf("API of type %q requires a %q, but none was defined", a, "scope")
+		}
+		if v := parsed.CustomValues["actionType"]; v == "" {
+			return pointer.DeletePointer{}, fmt.Errorf("API of type %q requires a %q, but none was defined", a, "actionType")
+		}
+		if v := parsed.CustomValues["domain"]; v == "" {
+			return pointer.DeletePointer{}, fmt.Errorf("API of type %q requires a %q, but none was defined", a, "domain")
```
arthurpitman, line 186: Here we have `%q`, do we also want `%#q`?

arthurpitman: OR do we want to revert to `%s` for this PR and consider this whole topic in a follow up?

pkg/delete/delete_test.go
```diff
@@ -843,3 +837,27 @@ func TestConfigsWithParent(t *testing.T) {
 		})
 	}
 }
+
+func Test_KeyUserActionsWebUniqueness(t *testing.T) {
```
arthurpitman, line 841: Should we add a few more test cases where we see specific KUA-web being deleted or _not_ deleted.

pkg/delete/internal/classic/client.go
jskelin: With pre dividing mechanism, it is not necessary to have that.

pkg/delete/internal/classic/client.go
jskelin: Here it forbids concurrent reading and writing. It could be a case that during the reading, another thread tries to write a new value. 

But we have `cache.DefaultCache` which is doing exactly this. So it is switched to that

pkg/delete/delete_loader.go
jskelin: for now let's use '%s'

warber (approved):

arthurpitman (approved):

PR #1428, opened 2024-03-21
fix: Apply externalId to settings objects with `builtin:oneagent.features` schema

arthurpitman:
This PR ensures `externalId`s are correctly applied to `builtin:oneagent.features` settings objects by removing code in ` upsertSettings(...)` that was preventing this.

This allows `builtin:oneagent.features` settings objects with scopes of monitored entities to be deleted as required.

To prevent `builtin:oneagent.features` settings objects with for example environment scope from being deleted, existing checks for `ModificationInfo.Deletable` are used.


warber (approved):
pls update PR description a bit

PR #1430, opened 2024-03-21
Adapt purge command to work with latest additions

warber:
I made two changes to

* skip trying to delete disabled api types
* skip trying to delete dashboard share settings

### DIFF
--- a/cmd/monaco/purge/purge.go
+++ b/cmd/monaco/purge/purge.go
@@ -41,7 +41,7 @@ func purge(fs afero.Fs, deploymentManifestPath string, environmentNames []string
 		return fmt.Errorf("error while finding absolute path for `%s`: %w", deploymentManifestPath, manifestErr)
 	}
 
-	apis := api.NewAPIs().Filter(api.RetainByName(apiNames), api.RemoveNonDeletable)
+	apis := api.NewAPIs().Filter(api.RetainByName(apiNames), api.RemoveDisabled, api.RemoveNonDeletable)
 
 	mani, manifestLoadError := manifestloader.Load(&manifestloader.Context{
 		Fs:           fs,
--- a/cmd/monaco/purge/purge.go
+++ b/cmd/monaco/purge/purge.go
@@ -41,7 +41,7 @@ func purge(fs afero.Fs, deploymentManifestPath string, environmentNames []string
 		return fmt.Errorf("error while finding absolute path for `%s`: %w", deploymentManifestPath, manifestErr)
 	}
 
-	apis := api.NewAPIs().Filter(api.RetainByName(apiNames))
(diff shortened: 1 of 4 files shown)


cmd/monaco/purge/purge.go
```diff
@@ -41,7 +41,7 @@ func purge(fs afero.Fs, deploymentManifestPath string, environmentNames []string
 		return fmt.Errorf("error while finding absolute path for `%s`: %w", deploymentManifestPath, manifestErr)
 	}
 
-	apis := api.NewAPIs().Filter(api.RetainByName(apiNames))
+	apis := api.NewAPIs().Filter(api.RetainByName(apiNames), api.RemoveDisabled, api.RemoveByName([]string{api.DashboardShareSettings}))
```
UnseenWizzard, line 44: long-form nit:
If dashboard share settings just can not be deleted, shouldn't we handle this in a more generalized way than introducing a special "RemoveByName" just for purge while the normal delete excludes them with an explicit if? 

I generally find the introduction of conflicting Remove and Retain by name functions problematic - this makes the Filter method potentially harder to use and understand.
And as user that can give purge a specific API to delete I get no indication why nothing happens if I ran the CLI to purge all share settings. 

Maybe a generalized 'Deletable' flag on the API struct (similar to Settings' ModificationInfo) would work better here

warber: didn't thought about adding a "Deletable" flag... 🤔  sounds interesting

warber: I would go with "NonDeletable" as the vast majority of configs are deletable, wdyt?

arthurpitman: This general idea would make looking at the APIs easier to understand. I'd go with a `Deletable` field though to avoid double negatives. We could have a helper function to use for creating the vast majority of APIs if that makes things easier.

warber: thx, I've updated the PR. but went with "NonDeletable" most of the time we just do not deal with this property and the double negation is tolerable, imo

arthurpitman (approved):

UnseenWizzard (approved):

pkg/api/endpoints.go
```diff
@@ -151,6 +151,7 @@ var configEndpoints = []API{
 		URLPath:             "/api/config/v1/dashboards/{SCOPE}/shareSettings",
 		Parent:              &dashboardAPI,
 		SingleConfiguration: true,
+		NonDeletable:        true,
```
UnseenWizzard, line 154: as a followup, could we introduce this in the regular delete please to replace the current if specifically checking for the dashboard-share-settings API.

PR #1432, opened 2024-03-22
fix: Avoid duplicate config downloads and template file overwrites

warber:
This PR contains couple of refactorings to cleanup logic for downloading configs.
The last two commits fixes downloading of scoped configuration like "key user actions" which were duplicated before which resulted in duplicate config writese to disk. Also possible template file clashes are fixed by the last commit.

### DIFF
--- a/pkg/download/classic/download.go
+++ b/pkg/download/classic/download.go
@@ -278,6 +278,12 @@ func download(client client.ConfigClient, theApi api.API, value value) ([]map[st
 		})
 	}
 
+	if theApi.PropertyNameOfIdentifier != "" {
+		return slices.DeleteFunc(res, func(m map[string]interface{}) bool {
+			return m[theApi.PropertyNameOfIdentifier].(string) != value.value.Id
+		}), nil
+	}
+
 	return res, nil
 }
 
--- a/pkg/download/classic/download.go
+++ b/pkg/download/classic/download.go
@@ -19,7 +19,6 @@ package classic
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
@@ -35,15 +34,11 @@ import (
 	projectv2 "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project/v2"
 	"github.com/mitchellh/mapstructure"
 	"golang.org/x/exp/maps"
+	"slices"
 	"strings"
 	"sync"
 	"time"
 )
-
-type downloadedConfig struct {
-	config.Config
-	value dtclient.Value
-}
 
 func Download(client client.ConfigClient, projectName string, apisToDownload api.APIs, filters ContentFilters) (projectv2.ConfigsPerType, error) {
 	log.Debug("APIs to download: \n - %v", strings.Join(maps.Keys(apisToDownload), "\n - "))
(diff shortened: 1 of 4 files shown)


jskelin (commented):
Better that it was.

jskelin (dismissed):

pkg/download/classic/download.go
```diff
@@ -261,7 +243,7 @@ func shouldFilter() bool {
 	return featureflags.DownloadFilter().Enabled() && featureflags.DownloadFilterClassicConfigs().Enabled()
 }
 
-func downloadAndUnmarshalConfig(client dtclient.Client, theApi api.API, value value) ([]map[string]interface{}, error) {
+func download(client dtclient.Client, theApi api.API, value value) (map[string]any, error) {
```
UnseenWizzard, line 246: You didn't introduce it, but the logic of this function would benefit from some explaining comments. That fact that it does a "read config by ID" call to list all configs of either a top-level API (id = "" so we just do a <api>/ call) or list all children of a parent API (<api>/<parentid>/<subpath>) confused me when trying to understand your changes

UnseenWizzard:
Overall I find it easier to follow to than before, but the mixture of "normal" config APIs types where this downloading will download and unmarshal single objects and things like kua-web where downloading will get and create several entries (and resolves parent API) makes it still a bit hard to follow. Some explanatory comments on the functions would be nice.

pkg/download/classic/download.go
```diff
@@ -56,92 +51,79 @@ func Download(client dtclient.Client, projectName string, apisToDownload api.API
...
 	wg := sync.WaitGroup{}
-	wg.Add(len(foundValues))
-	for _, v := range foundValues {
+	wg.Add(len(configsToDownload))
+	for _, v := range configsToDownload {
 		go func() {
 			defer wg.Done()
 
-			downloadedJsons, err := downloadAndUnmarshalConfig(client, api, v)
+			dlData, err := download(client, api, v)
```
UnseenWizzard, line 92: You didn't introduce this @warber, but: we capture the config to download `v` in the loop and go routine, rather than passing it as a variable to the go func. This seems not to have caused issues so far - but couldn't it?

warber: Isn't that only a problem for go <1.22 ?

warber:
> Overall I find it easier to follow to than before, but the mixture of "normal" config APIs types where this downloading will download and unmarshal single objects and things like kua-web where downloading will get and create several entries (and resolves parent API) makes it still a bit hard to follow. Some explanatory comments on the functions would be nice.

@UnseenWizzard yes i know 😿 if you have any further ideas, feel free to give it a try :)

pkg/persistence/config/writer/config_writer.go
```diff
@@ -790,3 +790,14 @@ func fmtDetailedConfigWriterError(context *serializerContext, format string, arg
 		Err:      fmt.Errorf(format, args...),
 	}
 }
+
+var fileNameClashes = make(map[string]int)
```
warber, line 794: I realized that passing this through each method as a pointer "only" for identifying possible clashes pollutes the code even more. So I've decided to use that package. private global here. tell me if you think that's acceptable :)

pkg/download/classic/download.go
```diff
@@ -281,32 +257,37 @@ func downloadAndUnmarshalConfig(client client.ConfigClient, theApi api.API, valu
...
-}
 
-func filterResponses(res []map[string]any, value value) ([]map[string]any, error) {
-	for _, v := range res {
-		if v["meIdentifier"] == value.value.Id {
-			return []map[string]any{v}, nil
-		}
+	if theApi.PropertyNameOfIdentifier != "" {
+		return slices.DeleteFunc(res, func(m map[string]interface{}) bool {
+			return m[theApi.PropertyNameOfIdentifier].(string) != value.value.Id
```
warber, line 283: we need a way to identify the matching object in the payload. To know its identifier (key) the `PropertyNameOfIdentifier` value is used

pkg/persistence/config/writer/config_writer.go
jskelin: nit: I would go with an array of strings ( `var []string`). 
And then in the next function, use `strings.Contains....` (https://pkg.go.dev/slices#Contains)

why: there is no need to keep track of the number of duplicates.

jskelin (changes requested):
Do we need to add a test for the new functionality (in case of duplicate filenames throw an error) ?

jskelin (approved):

pkg/persistence/config/writer/config_writer_test.go
```diff
@@ -1364,6 +1364,48 @@ func TestWriteConfigs(t *testing.T) {
...
+
+	fs := afero.NewMemMapFs()
+	errs := WriteConfigs(&WriterContext{
+		Fs:              fs,
+		OutputFolder:    "test",
+		ProjectFolder:   "project",
+		ParametersSerde: config.DefaultParameterParsers,
+	}, configs)
+
+	assert.Len(t, errs, 0)
```
jskelin, line 1399: nit: https://pkg.go.dev/github.com/stretchr/testify/assert#Empty

jskelin (approved):

PR #1433, opened 2024-03-22
test(e2e): Remove test case asserting references from Settings to Classic ManagementZones

UnseenWizzard:
test(e2e): Remove test case asserting references from Settings to Classic ManagementZones
In the past, Settings schemas used in this test, contained ManagementZone numeric IDs - which allowed downloading Classic Config MZs and resolving references from Settings objects to the Classic MZ. This is no longer possible as these Settings schemas contain Settings Object IDs for ManagementZones as of Dynatrace 1.287.

As this is only an issue if forcibly downloading Classic MZ objects together with Settings objects referencing that MZ - a normal download will skip the Classic Config MZ API in favour of the Settings schema, and resolve references correctly - no change on our side is deemed necessary and the testcase for this old behaviour is simply removed.

nope


warber (approved):
👌 failing e2e test is expected (acc test)

PR #1437, opened 2024-03-28
fix: deploy web key user actions sequentially

warber:
This makes sure that web key user actions can reliably deployed.
Since this API seems to work unreliably whenever we try to create multiple key user actions **in parallel**
this PR forces configs of this type to be deployed sequentially with a 2 second delay between each.

Note that we can only rely on configs to be in the same independent graph component when there are dependencies between them. Hence this approach does not work when hard coding the scope of a web key user actions to a specific web application. For this reason specifying a hard coded scope for web key user actions is not allowed and will result in an error printed to the user.

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -18,6 +18,7 @@ package api
 
 import (
 	"strings"
+	"time"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 )
@@ -72,6 +73,9 @@ type API struct {
 	PropertyNameOfIdentifier string
 	// NonDeletable indicates that configs of that type cannot be deleted
 	NonDeletable bool
+	// DeployWaitDuration defines the amount of time that shall elapse between deploying configs of this type.
+	// Note, that this only applies to configs within the same independent graph component
+	DeployWaitDuration time.Duration
 }
 
 func (a API) CreateURL(environmentURL string) string {
--- a/pkg/api/endpoints.go
+++ b/pkg/api/endpoints.go
@@ -16,7 +16,10 @@
 
 package api
 
-import "github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
+import (
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
+	"time"
+)
 
 const (
 	AlertingProfile                      = "alerting-profile"
@@ -472,6 +475,7 @@ var configEndpoints = []API{
 				existing["actionType"] == current["actionType"] &&
 				existing["domain"] == current["domain"]
 		},
+		DeployWaitDuration: 2 * time.Second,
 	},
 	{
 		ID:           UserActionAndSessionPropertiesMobile,
--- a/pkg/api/endpoints.go
+++ b/pkg/api/endpoints.go
@@ -17,6 +17,7 @@
 package api
 
 import (
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/environment"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"time"
 )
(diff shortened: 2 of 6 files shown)


pkg/deploy/deploy.go
```diff
@@ -164,9 +165,18 @@ func deployGraph(ctx context.Context, configGraph *simple.DirectedGraph, clients
 
 		for _, root := range roots {
 			node := root.(graph.ConfigNode)
-			go func(ctx context.Context, node graph.ConfigNode) {
-				errChan <- deployNode(ctx, node, configGraph, clients, resolvedEntities)
-			}(context.WithValue(ctx, log.CtxKeyCoord{}, node.Config.Coordinate), node)
+			switch deployStrategy := api.NewAPIs()[node.Config.Coordinate.Type].DeployStrategy.(type) {
+			case api.SeqDeployStrategy:
```
UnseenWizzard, line 169: I like the idea in general, but limiting it to just config APIs seems limited and means you need to check if it even is an API here (does the switch actually work for other settings or blow up accessing the DeployStratgy if the config is not an API type?)

warber: go returns the zero value for the map's value type if the key is not found, so it should not blow up.

UnseenWizzard: my main point was more that the DeployStrategy feels like a general concept, but is limited to API types and needs you to check the API map in two places now. Maybe we could set it on the config level?

UnseenWizzard, line 169: Upon closer inspection: Don't you currently make things sequential twice? 
When building the graph the fake dependencies are added, ensuring that configs aren't on the same level but in one chain. 

When deploying instead of firing off a go routine to deploy each node on the same level (the current roots), you now deploy inside the main routine of `deployGraph` and sleep in it.

I think the fake dependencies are always needed, else you won't get things into the same independent component. But then, wouldn't be enough to deploy them normally and just optional add the wait after deployment before returning on the error channel? Each 'root level' is synchronized via the error channel, so the wait duration should still effectively give us a delay before the next sequential config is deployed (as the depend on each other)

pkg/api/api.go
```diff
@@ -72,6 +73,9 @@ type API struct {
 	PropertyNameOfIdentifier string
 	// NonDeletable indicates that configs of that type cannot be deleted
 	NonDeletable bool
+	// DeployWaitDuration defines the amount of time that shall elapse between deploying configs of this type.
+	// Note, that this only applies to configs within the same independent graph component
+	DeployWaitDuration time.Duration
```
warber, line 78: better name wanted 🤷

pkg/project/v2/project_loader.go
```diff
@@ -189,6 +190,19 @@ func loadProject(fs afero.Fs, context ProjectLoaderContext, projectDefinition ma
 			errors = append(errors, newDuplicateConfigIdentifierError(c))
 		}
 	}
+
+	// The scope parameter of a key user actions web configuration needs to be a reference to another application-web config
+	// The reference parameter makes sure that rely on the fact that kua web configs are loaded/deployed within the same
+	// sub graph (independent component) later on as
+	for _, c := range configs {
```
warber, line 197: first i wanted to put that check into Typedefinition::Validate(), but turns out that this code is also used for deletion use case, where scope is just a string which is ok there

pkg/api/api.go
jskelin: Another field in api.go ...

jskelin (dismissed):

pkg/api/endpoints.go
```diff
@@ -472,6 +475,7 @@ var configEndpoints = []API{
 				existing["actionType"] == current["actionType"] &&
 				existing["domain"] == current["domain"]
 		},
+		DeployWaitDuration: 2 * time.Second,
```
arthurpitman, line 478: This delay works for 10 apps, each with 50 KUAWs: all were created as expected

arthurpitman (approved):

pkg/project/v2/project_loader.go
```diff
@@ -189,6 +190,19 @@ func loadProject(fs afero.Fs, context ProjectLoaderContext, projectDefinition ma
...
+		if c.Coordinate.Type == api.KeyUserActionsWeb {
+			if _, ok := c.Parameters[config.ScopeParameter].(*ref.ReferenceParameter); !ok {
+				errors = append(errors, fmt.Errorf("scope parameter of config of type '%s' with ID `%s` needs to be a reference "+
```
warber, line 200: pls change the backticks to single quotes (i think you also need to adjust the unit test then)

arthurpitman (approved):

PR #1438, opened 2024-03-28
Feat: Skip updates to SCIM and ALL_USERS account groups

arthurpitman:
This PR introduces a feature flag `MONACO_SKIP_READ_ONLY_ACCOUNT_GROUP_UPDATES` which, when enabled, causes groups owned by SCIM and ALL_USERS not be updated during deployment.

### DIFF
--- a/internal/featureflags/temporary.go
+++ b/internal/featureflags/temporary.go
@@ -117,3 +117,12 @@ func UserActionSessionPropertiesMobile() FeatureFlag {
 		defaultEnabled: true,
 	}
 }
+
+// SkipReadOnlyAccountGroupUpdates toggles whether updates to read-only account groups are skipped or not.
+// Introduced: 2024-03-29; v2.13.0
+func SkipReadOnlyAccountGroupUpdates() FeatureFlag {
+	return FeatureFlag{
+		envName:        "MONACO_FEAT_SKIP_READ_ONLY_ACCOUNT_GROUP_UPDATES",
+		defaultEnabled: false,
+	}
+}
--- a/internal/featureflags/temporary.go
+++ b/internal/featureflags/temporary.go
@@ -122,7 +122,7 @@ func UserActionSessionPropertiesMobile() FeatureFlag {
 // Introduced: 2024-03-29; v2.13.0
 func SkipReadOnlyAccountGroupUpdates() FeatureFlag {
 	return FeatureFlag{
-		envName:        "MONACO_FEAT_SKIP_READ_ONLY_ACCOUNT_GROUP_UPDATES",
+		envName:        "MONACO_SKIP_READ_ONLY_ACCOUNT_GROUP_UPDATES",
 		defaultEnabled: false,
 	}
 }
--- a/pkg/account/deployer/client.go
+++ b/pkg/account/deployer/client.go
@@ -145,54 +145,84 @@ func (d *accountManagementClient) upsertPolicy(ctx context.Context, policyLevel 
 func (d *accountManagementClient) upsertGroup(ctx context.Context, groupId string, group Group) (remoteId, error) {
 	if groupId != "" {
 		logr.FromContextOrDiscard(ctx).V(1).Info("Trying to update group with origin object ID (UUID) " + groupId)
-		resp, err := d.client.GroupManagementAPI.EditGroup(ctx, d.accountInfo.AccountUUID, groupId).PutGroupDto(group).Execute()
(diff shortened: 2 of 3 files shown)


pkg/account/deployer/client.go
```diff
@@ -145,54 +146,89 @@ func (d *accountManagementClient) upsertPolicy(ctx context.Context, policyLevel
...
+	if err != nil {
+		return "", err
+	}
+
+	if len(existingGroupsWithName) == 0 {
+		return d.createGroup(ctx, group)
+	}
+
+	if len(existingGroupsWithName) > 1 { // shouldn't happen
+		logr.FromContextOrDiscard(ctx).V(-1).Info("Updating multiple policies with name " + group.Name + ". Updating group with UUID " + existingGroupsWithName[0].GetUuid())
```
UnseenWizzard, line 167: not your change, but IIRC, negative verbosity is discarded/there's no Warning level in the logger combination we're using. (Bernd, did we every look into this any further?) So may as well log this on just Info

pkg/account/deployer/client.go
```diff
@@ -145,54 +146,89 @@ func (d *accountManagementClient) upsertPolicy(ctx context.Context, policyLevel
...
+	if len(createdGroups) < 1 {
+		return "", fmt.Errorf("unable to get UUID of created group with name: %s", group.Name)
+	}
+	return createdGroups[0].GetUuid(), nil
+}
 
-	groupToUpdate := existingGroups[0]
+func (d *accountManagementClient) updateExistingGroup(ctx context.Context, existingGroup accountmanagement.GetGroupDto, group Group) (remoteId, error) {
+	// Groups with owner "SCIM" or "ALL_USERS" cannot be modified and so updates should be skipped
+	if featureflags.SkipReadOnlyAccountGroupUpdates().Enabled() && ((existingGroup.Owner == "SCIM") || (existingGroup.Owner == "ALL_USERS")) {
```
UnseenWizzard, line 221: could we logically turn this around and only update `LOCAL` groups (may be mis-remebering the exact owner for groups from DT's permission management)?

arthurpitman: Good point, however I'm not sure whether restrictions also apply to groups owned by `SAML` or `DCS`, and so restricted this to not update those I was told cannot be updated.

pkg/account/deployer/client.go
arthurpitman: I'm too new to know, but is there a reason we log directly with `logr` here? Rather than use say `internal/log/log.go`?

UnseenWizzard: I can only guess, but I'd assume this happened with the idea that this client might move to the core library. 
(@warber @jskelin can one of you recall for sure why we use logr here?)

pkg/account/deployer/client.go
UnseenWizzard: Makes sense - I wasn't aware there's that many different types of owners.

pkg/account/deployer/client.go
warber: yes, the idea was to also move the "smart" AM client to the lib repository, which obviously never happened.

internal/featureflags/temporary.go
```diff
@@ -117,3 +117,12 @@ func UserActionSessionPropertiesMobile() FeatureFlag {
 		defaultEnabled: true,
 	}
 }
+
+// SkipReadOnlyAccountGroupUpdates toggles whether updates to read-only account groups are skipped or not.
+// Introduced: 2024-03-29; v2.13.0
+func SkipReadOnlyAccountGroupUpdates() FeatureFlag {
+	return FeatureFlag{
+		envName:        "MONACO_FEAT_SKIP_READ_ONLY_ACCOUNT_GROUP_UPDATES",
```
warber, line 125:
```suggestion
		envName:        "MONACO_SKIP_READ_ONLY_ACCOUNT_GROUP_UPDATES",
```
nit: as this is not really a "feature" but just changes how monaco behaves when it comes to these resources.

warber (approved):

pkg/account/deployer/client.go
UnseenWizzard: if we see any value/potential in it being reused, we should just move it IMO. If we don't, as @arthurpitman says, we might as welluse our logger directly and actually get warning level logs.

UnseenWizzard (approved):

warber (approved):

PR #1451, opened 2024-04-16
chore: Extend config to support dashboard and notebook documents

arthurpitman:
This PR extends configurations with support for dashboard and notebooks. This permits syntax like:

```
- id: MyDashboard
  config:
    name: My dashboard
    originObjectId: ext-ID-123 
    template: my-dashboard.json
    skip: false
  type:
    document:
      type: dashboard
```

To use this feature, enable the feature flag by setting `MONACO_FEAT_DOCUMENTS` to `true`

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -124,18 +124,14 @@ func (BucketType) ID() TypeId {
 	return BucketTypeId
 }
 
-type DocumentTypeType string
-
 const (
-	DashboardType DocumentTypeType = "dashboard"
-	NotebookType  DocumentTypeType = "notebook"
+	DashboardType DocumentType = "dashboard"
+	NotebookType  DocumentType = "notebook"
 )
 
-type DocumentType struct {
-	// Type identifies which Document type is used in the config.
-	// Currently, it can be a dashboard or a notebook.
-	Type DocumentTypeType
-}
+// DocumentType identifies which document type is used in the config.
+// Currently, it can be a dashboard or a notebook.
+type DocumentType string
 
 func (DocumentType) ID() TypeId {
 	return DocumentTypeId
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -66,6 +66,7 @@ const (
 	EntityTypeId     TypeId = "entity"
 	AutomationTypeId TypeId = "automation"
 	BucketTypeId     TypeId = "bucket"
+	DocumentTypeId   TypeId = "document"
 )
 
 type Type interface {
@@ -121,6 +122,23 @@ type BucketType struct{}
 
 func (BucketType) ID() TypeId {
 	return BucketTypeId
+}
+
+type DocumentTypeType string
+
+const (
+	DashboardType DocumentTypeType = "dashboard"
+	NotebookType  DocumentTypeType = "notebook"
+)
+
+type DocumentType struct {
+	// Type identifies which Document type is used in the config.
+	// Currently, it can be a dashboard or a notebook.
+	Type DocumentTypeType
+}
+
+func (DocumentType) ID() TypeId {
+	return DocumentTypeId
 }
 
 // Config struct defining a configuration which can be deployed.
--- a/pkg/persistence/config/internal/persistence/type_definition.go
+++ b/pkg/persistence/config/internal/persistence/type_definition.go
@@ -48,7 +48,7 @@ type AutomationDefinition struct {
 }
 
 type DocumentDefinition struct {
-	Type config.DocumentTypeType `yaml:"type" json:"type" jsonschema:"required,enum=dashboard,enum=notebook,description=This defines which document type this config is for."`
+	Type config.DocumentType `yaml:"type" json:"type" jsonschema:"required,enum=dashboard,enum=notebook,description=This defines which document type this config is for."`
 }
 
 // UnmarshalYAML Custom unmarshaler that knows how to handle TypeDefinition.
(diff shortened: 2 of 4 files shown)


pkg/config/config.go
```diff
@@ -123,6 +124,23 @@ func (BucketType) ID() TypeId {
 	return BucketTypeId
 }
 
+type DocumentTypeType string
```
arthurpitman, line 127: Any suggestions for a better name?

arthurpitman:
We will also have to address the Snyk check:
`the given struct should be annotated with the `mapstructure` tag (musttag)`

pkg/config/config.go
```diff
@@ -123,6 +124,23 @@ func (BucketType) ID() TypeId {
...
+
+const (
+	DashboardType DocumentTypeType = "dashboard"
+	NotebookType  DocumentTypeType = "notebook"
+)
+
+type DocumentType struct {
```
warber, line 134: This could just be:

`type DocumentType string `

and `DocumentTypeType` can be removed.

pkg/config/config.go
warber: see https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1451#discussion_r1567726029

pkg/persistence/config/writer/config_writer_test.go
```diff
@@ -1164,11 +1167,101 @@ func TestWriteConfigs(t *testing.T) {
 				"general/alerting-profile/a.json",
 			},
 		},
+		{
+			name: "Documents",
+			configs: []config.Config{
+				{
+					Template: template.NewInMemoryTemplateWithPath("project/dashboard/a.json", ""),
```
jskelin, line 1174: Is this the real path which should be on a disc?

pkg/persistence/config/writer/config_writer_test.go
```diff
@@ -1164,11 +1167,101 @@ func TestWriteConfigs(t *testing.T) {
...
+								Type: config.DocumentType{
+									Type: "notebook",
+								},
+							},
+						},
+					},
+				},
+			},
+			expectedTemplatePaths: []string{
+				"project/dashboard/a.json",
```
jskelin, line 1247: Are we going to put 2nd and 3rd gen dashboards together?

pkg/persistence/config/writer/config_writer_test.go
```diff
@@ -1164,11 +1167,101 @@ func TestWriteConfigs(t *testing.T) {
...
+				"project/notebook/a.json",
+			},
+			envVars: map[string]string{
+				featureflags.Documents().EnvName(): "true",
+			},
+		},
 	}
 
+	t.Setenv(featureflags.Documents().EnvName(), "true")
```
jskelin, line 1256: There is the filed `envVars map[string]string`! Why not set it there?

pkg/persistence/config/loader/config_loader_test.go
```diff
@@ -1323,6 +1324,117 @@ configs:
         wrong-field: var`,
 			wantErrorsContain: []string{"missing property"},
 		},
+		{
+			name: "Document dashboard config with FF on",
```
jskelin, line 1328: The name "... with FF on" suggests to me that it will also be a test "with FF off". I would argue (and no test with FF off exists) to remove that from the name.

pkg/persistence/config/internal/persistence/type_definition.go
```diff
@@ -227,7 +262,15 @@
 	case config.BucketType:
 		return BucketType, nil
 
-	default:
-		return nil, fmt.Errorf("unkown type: %T", c.Type)
+	case config.DocumentType:
```
jskelin, line 265: The language allows that, but to me, keeping "default" at the beginning or at the end helps to read the code easily.

jskelin (commented):
Our test files already have more than 1k lines. Is it time to start splitting it by domains?

pkg/persistence/config/writer/config_writer_test.go
arthurpitman: Correct, this line should be removed

pkg/persistence/config/internal/persistence/type_definition.go
```diff
@@ -227,7 +262,15 @@ func (c TypeDefinition) MarshalYAML() (interface{}, error) {
...
+		if featureflags.Documents().Enabled() {
+			return map[string]any{
+				"document": DocumentDefinition{
+					Type: t,
+				},
+			}, nil
+		}
 	}
+
+	return nil, fmt.Errorf("unkown type: %T", c.Type)
```
arthurpitman, line 275:
```suggestion
	return nil, fmt.Errorf("unknown type: %T", c.Type)
```

pkg/persistence/config/loader/config_loader_test.go
arthurpitman: A couple of tests down there is one with the feature flag off that expects an error to be generated

pkg/persistence/config/internal/persistence/type_definition.go
arthurpitman: Here it would require me to duplicate the return, which I'd prefer not to

arthurpitman:
> Our test files already have more than 1k lines. Is it time to start splitting it by domains?

I dont expect we'd touch these too often, so they are probably OK as they are. @warber what do you think?

warber (approved):

PR #1453, opened 2024-04-17
chore: adjusted error handling of automation client responses

warber:


cmd/monaco/integrationtest/assert.go
```diff
@@ -253,10 +253,8 @@ func AssertAutomation(t *testing.T, c client.AutomationClient, env manifest.Envi
 		expectedId = idutils.GenerateUUIDFromCoordinate(cfg.Coordinate)
 	}
 
-	resp, err := c.Get(context.TODO(), resourceType, expectedId)
-	assert.NoError(t, err)
-
-	exists := resp.IsSuccess()
+	_, err = c.Get(context.TODO(), resourceType, expectedId)
+	exists := err == nil
```
jskelin, line 257: Is this the same? What if the error is of the wrong type?

jskelin (approved):

PR #1455, opened 2024-04-18
fix: Delete automation resources in correct order

warber:
This PR adapts the Delete logic so that deletion of automation resources is happening in correct order, Workflows --> Scheduling Rules --> BusinessCalendars.

### DIFF
--- a/pkg/delete/delete.go
+++ b/pkg/delete/delete.go
@@ -46,37 +46,56 @@ type DeleteEntries = map[configurationType][]pointer.DeletePointer
 
 // Configs removes all given entriesToDelete from the Dynatrace environment the given client connects to
 func Configs(ctx context.Context, clients ClientSet, apis api.APIs, automationResources map[string]config.AutomationResource, entriesToDelete DeleteEntries) error {
-	deleteErrors := 0
-	for entryType, entries := range entriesToDelete {
-		if entryType == api.DashboardShareSettings {
-			log.Warn("Classic config of type %s cannot be deleted. Note, that they can be removed by deleting the associated dashboard.", api.DashboardShareSettings)
+	var deleteErrors int
+
+	// Delete automation resources (in the specified order
+	automationTypeOrder := []string{"workflow", "scheduling-rule", "business-calendar"}
+	for _, key := range automationTypeOrder {
+		entries := entriesToDelete[key]
+		if reflect.ValueOf(clients.Automation).IsNil() {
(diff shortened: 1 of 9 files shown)


pkg/delete/delete.go
```diff
@@ -46,37 +46,56 @@ type DeleteEntries = map[configurationType][]pointer.DeletePointer
...
-			err = bucket.Delete(ctx, clients.Buckets, entries)
 		} else { // assume it's a Settings Schema
 			err = setting.Delete(ctx, clients.Settings, entries)
 		}
-
 		if err != nil {
 			log.WithFields(field.Error(err)).Error("Error during deletion: %v", err)
 			deleteErrors += 1
 		}
+		delete(entriesToDelete, entryType)
```
UnseenWizzard, line 96: is there any need to delete the entries here? If we don't have to, why modify the thing we're looping over already

pkg/delete/delete.go
```diff
@@ -46,37 +46,56 @@ type DeleteEntries = map[configurationType][]pointer.DeletePointer
...
+		if err != nil {
+			log.WithFields(field.Error(err)).Error("Error during deletion: %v", err)
+			deleteErrors += 1
+		}
+		delete(entriesToDelete, key)
+	}
 
-		var err error
+	// Delete bucket resources
+	if reflect.ValueOf(clients.Buckets).IsNil() {
```
UnseenWizzard: nit: other than the reflection, and the fact that we can access it directly, is there a reason to handle bucket deletion separate to configs or settings? We could also just delete the bucket entries if no client is defined and leave the delete call in the main loop over delete entries

pkg/delete/delete.go
warber: not really, knowing that the map is empty at the end just has a calming effect :D

pkg/delete/delete.go
```diff
@@ -104,14 +123,14 @@ func All(ctx context.Context, clients ClientSet, apis api.APIs) error {
 		errs++
 	}
 
-	if reflect.ValueOf(clients.Automation).IsNil() {
+	if clients.Automation == nil {
 		log.Warn("Skipped deletion of Automation configurations as API client was unavailable.")
 	} else if err := automation.DeleteAll(ctx, clients.Automation); err != nil {
```
UnseenWizzard, line 128: does automation.Delete() adhere to the ordering when it deletes things?

UnseenWizzard (approved):
lgtm

UnseenWizzard (approved):

PR #1460, opened 2024-04-24
feat: Download dashboard and notebook documents

arthurpitman:
This PR adds support for downloading documents of type `dashboard` and `notebook`, placing these within in the project in folders of the same name.

To use this feature, enable the feature flag by setting `MONACO_FEAT_DOCUMENTS` to `true`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ bin/
 
 # Download files
 download_*
+!download_test.go
 
 # Generated JSON schema files
 schemas/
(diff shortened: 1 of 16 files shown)


arthurpitman:
> @arthurpitman looks great so far :) one thing i've noticed is that json data doesn't seem to be formatted or "prettified" before persisting, hence the whole data appears in a single line.

That seems to be way dashboards and notebooks save their data. Do you think it is something we should consider doing for the user?

warber:
> That seems to be way dashboards and notebooks save their data. Do you think it is something we should consider doing for the user?
good point, I thought we are doing that for other configs as well 🤔

warber (approved):

PR #1461, opened 2024-04-24
feat: Introduce `FileParameter` to allow reading parameter content from files

warber:
This PR introduces a new type of parameter, `FileParameter` which can be used to load content from a file and use it for templating in Monaco:

```yaml
configs:
- id: 2dab49b5-b58c-3f1c-b55f-342bee900db4
  config:
    name:
      type: value
      value: MyWorkflow
    parameters:
      comment: "// hey special comment"
      workflow:
        type: file
        path: "myWf.js"
        references:
          - comment
    template: 2dab49b5-b58c-3f1c-b55f-342bee900db4.json
    skip: false
    originObjectId: 2dab49b5-b58c-3f1c-b55f-342bee900db4
```

During deployment, Monaco will load the content of `myWf.js` into the parameter called `workflow`which can be referenced in the template using `{{ .workflow }}`
This particular example allows users to keep their workflow scripts in separate files instead of embedded inside the template, whicih improves usability a lot.
Note, that it is also possible to use templating inside the loaded file as well. In the example above this is demonstrated by another paramter `comment` which can be used inside the `myFs.js` file with `{{ .comment }}` 

Usually the `path` field of a `FileParamter` shall contain the filename. Monaco will search and load for that file located relative to the config folder. However it is also possible to break out of the config folder by using e.g. `../myFs.js`

### DIFF
--- a/cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/automation/config.yaml
+++ b/cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/automation/config.yaml
@@ -7,6 +7,12 @@ configs:
     name: '[CA] Notify Capability about new Jira Issues'
     template: jira-issue-workflow.json
     parameters:
+      comment: "// optional import of sdk modules"
+      jiraJs:
+        type: file
+        path: "jiraJs.txt"
+        references:
+          - comment
       actor:
         type: environment
         name: WORKFLOW_ACTOR
--- a/pkg/config/parameter/file/file.go
+++ b/pkg/config/parameter/file/file.go
@@ -21,6 +21,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/strings"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/template"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter"
+	tmpl "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/template"
 	"github.com/spf13/afero"
 	"path/filepath"
 )
@@ -33,9 +34,10 @@ var FileParameterSerde = parameter.ParameterSerDe{
 }
 
 type FileParameter struct {
-	WorkingDir string
-	Folder     string
-	Path       string
+	WorkingDir           string
+	Folder               string
+	Path                 string
+	referencedParameters []parameter.ParameterReference
 }
 
 func (f *FileParameter) GetType() string {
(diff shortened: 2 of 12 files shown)


pkg/config/parameter/file/file_test.go
```diff
@@ -0,0 +1,155 @@
...
+	})
+	assert.Nil(t, result)
+	assert.Error(t, err)
+
+}
+
+func TestResolveValue_FileNotFound(t *testing.T) {
+	param, _ := parseFileValueParameter(parameter.ParameterParserContext{
+		WorkingDir: "something",
+		Value:      map[string]any{"path": "something"},
```
arthurpitman, line 144: Little comment: if this is changed to a file the user can't access, the error message gets a bit more cryptic. On a Mac changing `WorkingDir` to `/` and `path` to `.file` produces: `failure trying to render template /.file: failed to read template content: open /.file: permission denied`, meaning it might be worth wrapping the second error in `(f *FileParameter) ResolveValue(...)` in `file.go`, such that the user understands that error happened inside a file parameter.

arthurpitman (dismissed):

pkg/config/parameter/file/file.go
```diff
@@ -33,34 +34,60 @@ var FileParameterSerde = parameter.ParameterSerDe{
...
 }
 
 func (f *FileParameter) GetType() string {
 	return FileParameterType
 }
 
 func (f *FileParameter) GetReferences() []parameter.ParameterReference {
-	// file parameters cannot have references
-	return []parameter.ParameterReference{}
+	if f.referencedParameters == nil {
```
UnseenWizzard, line 48: nit: is that check/initialized value return necessary? shouldn't a nil slice generally have the same effect as an empty initialized reference slice when used?

cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/automation/config.yaml
```diff
@@ -7,6 +7,12 @@ configs:
     name: '[CA] Notify Capability about new Jira Issues'
     template: jira-issue-workflow.json
     parameters:
+      comment: "// optional import of sdk modules"
+      jiraJs:
+        type: file
+        path: "jiraJs.txt"
```
UnseenWizzard, line 13: nit: why not.... `jira.js`? 😅

warber: haha yeah, intellij treats it as a javascript file and complains with parsing error because of our templating :P

pkg/config/parameter/file/file.go
```diff
@@ -0,0 +1,78 @@
...
+	return FileParameterType
+}
+
+func (f *FileParameter) GetReferences() []parameter.ParameterReference {
+	// file parameters cannot have references
+	return []parameter.ParameterReference{}
+}
+
+func (f *FileParameter) ResolveValue(context parameter.ResolveContext) (interface{}, error) {
+	content, err := afero.ReadFile(afero.NewOsFs(), filepath.Join(f.WorkingDir, f.Folder, f.Path))
```
UnseenWizzard: The direct use of the OS filesystem is the only part of this change that I'm unsure about. As nothing is actively writing this to files, it's not an issue that we can't inject the CopyOnWrite filesystem that we usually use in testing - but generally, it makes this parameter directly tied to the OS Fs rather the Fs that's actually being used by the config loader

It may be overkill - but I think we could hand in the same Fs that is used by the config-loader via the ResolveContext?

pkg/config/parameter/file/file.go
warber: depends on the usage, a nil slice is nil wheras the other one is not, but yes, for example `len` of both schould be 0.

pkg/config/parameter/file/file.go
warber: in a previous implementation i used a BasePath Fs. It required passing around that fs five times on the caller side just to support that special parameter loading. This and the fact that it is still 100% testable, and the fact that this type of parameter is -by definition- tied to the file system made me decide to do it like that.

warber: also, `ResolveContext` currently does not make any assumptions on "how" parameters are loaded, which is nice.

UnseenWizzard: I can live with it in general, especially because it's not an issue in production use (currently) 
But it still feels a like a potentially hard to debug case if something goes wrong. 

Thinking of an E2E test case: the copy-on-write FS is used and config files are rewritten. Now this parameter might read something else than expected by a test as the test would write to an overlay FS (in memory or temp files) but the parameter will always read from the OS Fs. 
The fact that this parameter uses than the rest of our loading code feels potentially brittle

UnseenWizzard (approved):
Approving as I can accept merging this as is, but my concerns stand.

pkg/config/parameter/file/file.go
UnseenWizzard: do other implemetations of GetReferences hold/make that promise that a zero lenght slice is returned instead of nil value?

pkg/config/parameter/file/file.go
warber: @UnseenWizzard I've pushed another [commit](https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1461/commits/994c55f95f4fb4051d3b11f3c4eb5976c07ee911) as an alternative implementation (reusing the afero filesystem used for config loading). :) please take a look and lets see if it is better :) also @arthurpitman  pls take a look 👋

warber: nice side effect, because of https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1461/files#diff-66ed2aae4cea3078ecd8fe5a748f3402a57747f3eb02981d07038496ade82ff5R133 we don't need to store working dir, nor folder into the parameter anymore 👌

arthurpitman (approved):

PR #1464, opened 2024-04-26
feat: introduce `FEAT_LOG_FILE` environment variable to deactivate log files

warber:
`MONACO_LOG_FILE` is treated to be set per default, hence Monaco will continue writing to the .logs directory. If it is set to false, it will not write log files. This is overruled as soon as the `--support-archive` flag is used. Then we obviously need/want to write logs to files and collect them.

### DIFF
--- a/internal/featureflags/permanent.go
+++ b/internal/featureflags/permanent.go
@@ -97,3 +97,11 @@ func BuildSimpleClassicURL() FeatureFlag {
 		defaultEnabled: true,
 	}
 }
+
+// LogToFile returns the feature flag to control whether log files shall be created or not
+func LogToFile() FeatureFlag {
+	return FeatureFlag{
+		envName:        "MONACO_LOG_FILE_ENABLED",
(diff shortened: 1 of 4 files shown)


internal/featureflags/permanent.go
```diff
@@ -97,3 +97,11 @@ func BuildSimpleClassicURL() FeatureFlag {
 		defaultEnabled: true,
 	}
 }
+
+// LogFile returns the feature flag to control whether log files shall be created or not
+func LogFile() FeatureFlag {
+	return FeatureFlag{
+		envName:        "MONACO_LOG_FILE",
```
arthurpitman, line 104:
```suggestion
		envName:        "MONACO_LOG_FILE_ENABLED",
```

arthurpitman: I think a name that implies a boolean rather than the file name would be more appropriate

jskelin (approved):

PR #1465, opened 2024-04-26
feat: ordered settings

warber:
This PR adds the feature of ordered settings. The settings type was extended to also contain the optional field "insertAfter" which is allowed to be a Reference Parameter:

```yaml
- id: 704aa921-9ff8-3f8e-bc6d-bedb63093ae9
  config:
    parameters:
    template: 704aa921-9ff8-3f8e-bc6d-bedb63093ae9.json
    skip: false
  type:
    settings:
      schema: builtin:container.monitoring-rule
      schemaVersion: 0.0.1
      scope: environment
      insertAfter:
        configId: c2314e1b-409c-3eaf-9efa-5dc593b14aff
        property: id
        type: reference
```

During download, Monaco detects whether a settings 2.0 object is ordered. If so, a reference parameter is created and the `insertAfter` field is populated. While deployment, this parameter is resolved and passed to the settings client.

User can use insertAfter property of settings 2.0 objects

### DIFF
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -54,6 +54,7 @@ type (
 
 	SchemaConstraints struct {
 		SchemaId         string
+		Ordered          bool
 		UniqueProperties [][]string
 	}
 
@@ -162,6 +163,8 @@ func (d *DynatraceClient) fetchSchemasConstraints(ctx context.Context, schemaID 
 			ret.UniqueProperties = append(ret.UniqueProperties, sc.UniqueProperties)
 		}
 	}
+	ret.Ordered = sd.Ordered
+
 	d.schemaConstraintsCache.Set(schemaID, ret)
 	return ret, nil
 }
@@ -257,7 +260,7 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
 		obj.OriginObjectId = ""
 	}
 
-	payload, err := buildPostRequestPayload(ctx, obj, externalID)
+	payload, err := buildPostRequestPayload(ctx, obj, externalID, options.InsertAfter)
 	if err != nil {
 		return DynatraceEntity{}, fmt.Errorf("failed to build settings object: %w", err)
 	}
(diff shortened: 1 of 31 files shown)


pkg/deploy/deploy.go
```diff
@@ -271,7 +271,11 @@ func deployConfig(ctx context.Context, c *config.Config, clients ClientSet, reso
 	var deployErr error
 	switch c.Type.(type) {
 	case config.SettingsType:
-		resolvedEntity, deployErr = setting.Deploy(ctx, clients.Settings, properties, renderedConfig, c)
+		var insertAfter string
+		if ia, ok := properties["insertAfter"]; ok {
```
UnseenWizzard, line 275: nit: I'd make this a fixed constant param name like name and scope are. IIRC defined in either the config or parameter package

pkg/deploy/internal/setting/setting.go
```diff
@@ -52,7 +52,7 @@ func Deploy(ctx context.Context, settingsClient client.SettingsClient, propertie
 		Content:        []byte(renderedConfig),
 		OriginObjectId: c.OriginObjectId,
 	}
-	upsertOptions := makeUpsertOptions(c)
+	upsertOptions := makeUpsertOptions(c, insertAfter)
```
UnseenWizzard, line 55: stupid functional question: how does the API behave if the config ID to insert after does not exist? Just ignore it I hope?

pkg/deploy/deploy.go
```diff
@@ -271,7 +271,11 @@ func deployConfig(ctx context.Context, c *config.Config, clients ClientSet, reso
...
+			insertAfter = ia.(string)
```
UnseenWizzard, line 276: with simply making this a fixed-name parameter, this might not actually be a string.

pkg/deploy/internal/setting/setting.go
UnseenWizzard: and how does it behave if it is not an object ID? With the specific parameter name and no extra validation that I can see so far, this might be any type of param.

warber: > Just ignore it I hope?
yes

pkg/client/dtclient/settings_client.go
```diff
@@ -257,7 +262,7 @@ func (d *DynatraceClient) upsertSettings(ctx context.Context, obj SettingsObject
 		obj.OriginObjectId = ""
 	}
 
-	payload, err := buildPostRequestPayload(ctx, obj, externalID)
+	payload, err := buildPostRequestPayload(ctx, obj, externalID, options.InsertAfter)
```
UnseenWizzard, line 265: Do we have any way to warn the user if they've used `insertAfter` for a Setting that is not actually ordered?
If it would mean a dedicated fetching of the schema just for that warning it might not be worth it - but I think when we check for matches by schema constraints we already grab the schema & the client caches the response, right? So it might be possible without notable overhead to read the schema and spit out a log

UnseenWizzard (dismissed):
Thanks for the changes to make it a special top-level parameter 👌
One final usability related comment to discuss, but generally looks good to me!

pkg/client/dtclient/settings_client.go
warber: Yes, it's easily possible to ask the cache and spit out a warning, however the API already returns a quite specific error that is logged in this case: ```{"code":400,"message":"insertAfter not supported for unordered schema builtin:rum.web.name"}```

warber: It can actually be treated as an error and would save us from retrying the operation a couple of times

UnseenWizzard (approved):

PR #1466, opened 2024-04-30
delete by objectId

jskelin:
Currently, delete file entries for automation resources and buckets contain `project`, `type` and `configId`, which are used to regenerate the ID of the object to delete during deletion. However, this doesn't work for objects not created by Monaco, as these have IDs that do not correspond.

Instead, to enable the deletion of pre-existing objects, delete file entries should support object ID (explicit field objectId ) as an alternative that, when available, is used instead of a generated ID.

When implementing the fix for automation and bucket resources we’ve agreed that we generally change the format of the delete file entries to be more expressive: Proposal for changes of the `delete.yaml` file:

```
delete:
- type: <ANY>
  objectId: 123453
- type: <SETTINGS>/<AUTOMATION>/bucket
  id: XYZ
  project: 42
- type: <CONFIG API>
  name: NAME
```

### DIFF
--- a/pkg/delete/internal/automation/delete.go
+++ b/pkg/delete/internal/automation/delete.go
@@ -19,6 +19,8 @@ package automation
 import (
 	"errors"
 	"fmt"
+	"net/http"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	automationAPI "github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/automation"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients/automation"
@@ -29,7 +31,6 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/pointer"
 	"golang.org/x/net/context"
-	"net/http"
 )
 
 type Client interface {
@@ -48,7 +49,10 @@ func Delete(ctx context.Context, c Client, automationResource config.AutomationR
 
 		logger := logger.WithFields(field.Coordinate(e.AsCoordinate()))
 
-		id := idutils.GenerateUUIDFromCoordinate(e.AsCoordinate())
+		id := e.OriginObjectId
+		if id == "" {
+			id = idutils.GenerateUUIDFromCoordinate(e.AsCoordinate())
+		}
 
 		logger.Debug("Deleting %v with id %q.", automationResource, id)
 
--- a/pkg/delete/internal/bucket/delete.go
+++ b/pkg/delete/internal/bucket/delete.go
@@ -20,6 +20,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients/buckets"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/buckettools"
@@ -28,7 +30,6 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/pointer"
 	"golang.org/x/net/context"
-	"net/http"
 )
 
 type Client interface {
@@ -46,7 +47,10 @@ func Delete(ctx context.Context, c Client, entries []pointer.DeletePointer) erro
 
 		logger := logger.WithFields(field.Coordinate(e.AsCoordinate()))
 
-		bucketName := idutils.GenerateBucketName(e.AsCoordinate())
+		bucketName := e.OriginObjectId
+		if e.OriginObjectId == "" {
+			bucketName = idutils.GenerateBucketName(e.AsCoordinate())
+		}
 
 		logger.Debug("Deleting bucket: %s.", e, bucketName)
 		_, err := c.Delete(ctx, bucketName)
--- a/pkg/delete/internal/classic/delete.go
+++ b/pkg/delete/internal/classic/delete.go
@@ -19,8 +19,9 @@ package classic
 import (
 	"errors"
 	"fmt"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client"
 	"net/http"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
@@ -37,7 +38,7 @@ func Delete(ctx context.Context, client client.ConfigClient, theAPI api.API, dps
 
 	for _, dp := range dps {
 		log := log.WithCtxFields(ctx).WithFields(field.Coordinate(dp.AsCoordinate()))
-		var id, parentID string
+		var parentID string
 		var e error
 		if theAPI.HasParent() {
 			parentID, e = resolveIdentifier(ctx, client, theAPI.Parent, toIdentifier(dp.Scope, "", ""))
@@ -52,14 +53,17 @@ func Delete(ctx context.Context, client client.ConfigClient, theAPI api.API, dps
 		}
 
 		a := theAPI.ApplyParentObjectID(parentID)
-		id, e = resolveIdentifier(ctx, client, &a, toIdentifier(dp.Identifier, dp.ActionType, dp.Domain))
-		if e != nil && !is404(e) {
-			log.WithFields(field.Error(e)).Error("unable to resolve config ID: %w")
-			err = errors.Join(err, e)
-			continue
-		} else if id == "" {
-			log.Debug("config doesn't exist - no need for action")
-			continue
+		id := dp.OriginObjectId
+		if id == "" {
+			id, e = resolveIdentifier(ctx, client, &a, toIdentifier(dp.Identifier, dp.ActionType, dp.Domain))
+			if e != nil && !is404(e) {
+				log.WithFields(field.Error(e)).Error("unable to resolve config ID: %w")
+				err = errors.Join(err, e)
+				continue
+			} else if id == "" {
+				log.Debug("config doesn't exist - no need for action")
+				continue
+			}
 		}
 
 		if e := client.DeleteConfigById(a, id); e != nil && !is404(e) {
--- a/pkg/delete/internal/setting/delete.go
+++ b/pkg/delete/internal/setting/delete.go
@@ -18,6 +18,7 @@ package setting
 
 import (
 	"fmt"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/idutils"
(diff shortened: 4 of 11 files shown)


pkg/delete/loader.go
```diff
@@ -137,99 +120,101 @@ func parseDeleteFileDefinition(ctx *loaderContext, definition persistence.FileDe
...
-	err := mapstructure.Decode(entry, &parsed)
-	if err != nil {
+	if err := mapstructure.Decode(entry, &parsed); err != nil {
 		return pointer.DeletePointer{}, err
 	}
 
-	if a, known := ctx.knownApis[parsed.Type]; known {
-		p, err := parseAPIEntry(parsed, a)
-		if err != nil {
+	if a, known := api.NewAPIs()[parsed.Type]; known {
```
arthurpitman, line 136: Do we need to check the presence of `parsed.Type` first? I think the field is required.

pkg/delete/persistence/delete_persistence.go
```diff
@@ -47,7 +47,8 @@ type DeleteEntry struct {
 	ConfigId string `yaml:"id,omitempty" json:"id,omitempty" mapstructure:"id" jsonschema:"description=The monaco ID of the config to be deleted - required for configs with generated IDs (e.g. Settings 2.0, Automations, Grail Buckets)."`
 	// ConfigName is the name of the config to be deleted - required for configs deleted by name (classic Config API types)
 	ConfigName string `yaml:"name,omitempty" json:"name,omitempty" mapstructure:"name" jsonschema:"description=The name of the config to be deleted - required for configs deleted by name (classic Config API types)."`
-
+	//OriginID is the dynatrace ID of the object
+	OriginID string `yaml:"objectId,omitempty" json:"objectId,omitempty" mapstructure:"objectId" jsonschema:"ID of the configuration in the Dynatrace. It can't combine with 'name'."`
```
arthurpitman, line 51:
```suggestion
	OriginID string `yaml:"objectId,omitempty" json:"objectId,omitempty" mapstructure:"objectId" jsonschema:"ID of the configuration in the Dynatrace. It can't be combined with 'name'."`
```

pkg/delete/loader.go
```diff
@@ -137,99 +120,101 @@ func parseDeleteFileDefinition(ctx *loaderContext, definition persistence.FileDe
...
-		log.Warn("Delete entry defines config 'name' - only 'id' will be used.")
+		return errors.New("'name' is not supported for this API")
 	}
-	return pointer.DeletePointer{
-		Project:    parsed.Project,
-		Type:       parsed.Type,
-		Identifier: parsed.ConfigId,
-	}, nil
+	if parsed.OriginID == "" && (parsed.ConfigId == "" && parsed.Project == "") {
+		return errors.New("TODO: finish me")
```
arthurpitman, line 201: Still need something here

pkg/delete/persistence/delete_persistence.go
arthurpitman, line 51: What about `configId` and `project`, they also cannot be combined with `objectId`?

arthurpitman: Should we rename `OriginID` here if it is called `objectId`?

pkg/delete/loader.go
```diff
@@ -137,99 +120,101 @@ func parseDeleteFileDefinition(ctx *loaderContext, definition persistence.FileDe
...
+		return errors.New("'id' is not supported for this API")
+	}
+	if parsed.Project != "" {
+		return errors.New("'project' is not supported for this API")
+	}
+	if parsed.ConfigName != "" && parsed.OriginID != "" {
+		return errors.New("'name' and 'objectId' can't be used together to define an entry")
+	}
+	if parsed.ConfigName == "" && parsed.OriginID == "" {
+		return errors.New("a 'name' or a 'originId' is required, but non was defined")
```
arthurpitman, line 174:
```suggestion
		return errors.New("a 'name' or a 'originId' is required, but none was defined")
```

pkg/delete/loader.go
```diff
@@ -137,99 +120,101 @@ func parseDeleteFileDefinition(ctx *loaderContext, definition persistence.FileDe
...
 	}
-
 	// The scope is required for sub-path APIs.
 	if a.HasParent() && parsed.Scope == "" {
-		return pointer.DeletePointer{}, errors.New("API requires a scope, but non was defined")
+		return errors.New("API requires a 'scope', but non was defined")
```
arthurpitman, line 178:
```suggestion
		return errors.New("API requires a 'scope', but none was defined")
```

pkg/delete/loader.go
```diff
@@ -137,99 +120,101 @@ func parseDeleteFileDefinition(ctx *loaderContext, definition persistence.FileDe
...
-
-	if str, ok := entry.(string); ok && err != nil {
-		return parseSimpleEntry(str)
+func convertToDeletePointer(entry any) (pointer.DeletePointer, error) {
+	if str, ok := entry.(string); ok {
+		return convertLegacy(str)
 	}
-
-	return ptr, err
+	return convert(entry)
```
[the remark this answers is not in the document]
jskelin: There's no case where a legacy and a "new" format could be switched!

pkg/delete/internal/automation/delete.go
```diff
@@ -48,7 +49,12 @@ func Delete(ctx context.Context, c Client, automationResource config.AutomationR
 
 		logger := logger.WithFields(field.Coordinate(e.AsCoordinate()))
 
-		id := idutils.GenerateUUIDFromCoordinate(e.AsCoordinate())
+		var id string
+		if e.OriginID != "" {
+			id = e.OriginID
+		} else {
+			id = idutils.GenerateUUIDFromCoordinate(e.AsCoordinate())
+		}
```
warber, line 57:
```suggestion
		id := e.OriginID
		if id == "" {
			id = idutils.GenerateUUIDFromCoordinate(e.AsCoordinate())
		}
```

pkg/delete/internal/bucket/delete.go
```diff
@@ -46,7 +47,12 @@ func Delete(ctx context.Context, c Client, entries []pointer.DeletePointer) erro
 
 		logger := logger.WithFields(field.Coordinate(e.AsCoordinate()))
 
-		bucketName := idutils.GenerateBucketName(e.AsCoordinate())
+		var bucketName string
+		if e.OriginID != "" {
+			bucketName = e.OriginID
+		} else {
+			bucketName = idutils.GenerateBucketName(e.AsCoordinate())
+		}
```
warber, line 55:
```suggestion
bucketName := e.OriginID
if bucketName == "" {
	bucketName = idutils.GenerateBucketName(e.AsCoordinate())
}

```

pkg/delete/internal/classic/delete.go
```diff
@@ -52,14 +52,18 @@ func Delete(ctx context.Context, client client.ConfigClient, theAPI api.API, dps
...
+			id, e = resolveIdentifier(ctx, client, &a, toIdentifier(dp.Identifier, dp.ActionType, dp.Domain))
+			if e != nil && !is404(e) {
+				log.WithFields(field.Error(e)).Error("unable to resolve config ID: %w")
+				err = errors.Join(err, e)
+				continue
+			} else if id == "" {
+				log.Debug("config doesn't exist - no need for action")
+				continue
+			}
 		}
```
warber, line 67:
```suggestion
		id = dp.OriginID
		if id == "" {
			id, e = resolveIdentifier(ctx, client, &a, toIdentifier(dp.Identifier, dp.ActionType, dp.Domain))
			if e != nil && !is404(e) {
				log.WithFields(field.Error(e)).Error("unable to resolve config ID: %w")
				err = errors.Join(err, e)
				continue
			}
			if id == "" {
				log.Debug("config doesn't exist - no need for action")
				continue
			}
		}
```

warber (commented):
Just some suggestions to avoid the use of `else` branches.

pkg/delete/pointer/delete_pointer.go
```diff
@@ -36,6 +37,9 @@ type DeletePointer struct {
 
 	// ActionType and Domain are used when deleting key-user-actions-web entities
 	ActionType, Domain string
+
+	//OriginID is DT ID of the configuration. Mutually exclusive with Identifier.
+	OriginID string
```
warber, line 42: nit: i think we call this "OriginObjectId" everywhere else.

warber (approved):

pkg/delete/pointer/delete_pointer.go
jskelin: I'll change this with the PR for the document delete

pkg/delete/internal/setting/delete.go
```diff
@@ -42,18 +43,21 @@ func Delete(ctx context.Context, c client.SettingsClient, entries []pointer.Dele
...
 
-		if e.Project == "" {
-			logger.Warn("Generating legacy externalID - this will fail to identify a newer Settings object. Consider defining a 'project' for this delete entry.")
+		var filterFn dtclient.ListSettingsFilter
+		if e.OriginID != "" { //delete by originID
+			filterFn = func(o dtclient.DownloadSettingsObject) bool { return o.ObjectId == e.OriginID }
+		} else {
+			externalID, err := idutils.GenerateExternalID(e.AsCoordinate())
+			if err != nil {
+				logger.Error("Unable to generate externalID, Setting will not be deleted: %v", err)
```
arthurpitman, line 52:
```suggestion
				logger.Error("unable to generate externalID, Setting will not be deleted: %v", err)
```

pkg/delete/loader.go
```diff
@@ -137,99 +120,104 @@ func parseDeleteFileDefinition(ctx *loaderContext, definition persistence.FileDe
...
+	}
+	if parsed.Project != "" {
+		return errors.New("'project' is not supported for this API")
+	}
+	if parsed.ConfigName != "" && parsed.ObjectId != "" {
+		return errors.New("'name' and 'objectId' can't be used together to define an entry")
+	}
+	if parsed.ConfigName == "" && parsed.ObjectId == "" {
+		return errors.New("a 'name' or a 'originId' is required, but none was defined")
 	}
```
arthurpitman, line 178: We need to update the name for `originId` or `objectId` in the error messages here

pkg/delete/pointer/delete_pointer.go
jskelin: done

arthurpitman:
nit: while we are changing things here, could we give these the same name:

https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/bf45cf6dfccce2b57ff5a04b0b8786b610f822c4/pkg/delete/loader.go#L155

pkg/delete/loader_test.go
```diff
@@ -29,325 +29,363 @@ import (
...
-  scope: ''
 `)
 
-	result, err := delete.LoadEntriesFromFile(createDeleteFile(t, fileContent))
+	result, err := delete.LoadEntriesFromFile(createDeleteFile(t, given))
 
 	var e delete.ParseErrors
 	assert.ErrorAs(t, err, &e)
-	assert.Equal(t, 5, len(e), "expected 5 errors")
+	assert.Equal(t, 3, len(e), "expected 4 errors")
```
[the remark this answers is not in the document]
jskelin: Excualy, I removed comments since they have no additional value or information.

arthurpitman (approved):

PR #1471, opened 2024-05-06
feat: Deploy dashboard and notebook documents

arthurpitman:


pkg/deploy/internal/document/document.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/entities"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter"
+	deployErrors "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/errors"
+	"github.com/go-logr/logr"
+)
+
+//go:generate mockgen -source=document.go -destination=document_mock.go -package=document documentClient
+type Client interface {
```
jskelin, line 37: I would suggest putting `Client` and `DummyClient` in different files to increase readability.

pkg/deploy/internal/document/document.go
```diff
@@ -0,0 +1,167 @@
...
+	if !ok {
+		return entities.ResolvedEntity{}, errors.New("missing name parameter")
+	}
+
+	documentType, err := getDocumentTypeFromConfigType(c.Type)
+	if err != nil {
+		return entities.ResolvedEntity{}, fmt.Errorf("cannot get document type for config: %w", err)
+	}
+
+	// strategy 1: if an origin id is available, try to update that document
```
jskelin, line 82: What do you think of putting this different strategy into different functions?

internal/idutils/external_id.go
```diff
@@ -50,3 +50,23 @@ func GenerateExternalID(c coordinate.Coordinate) (string, error) {
 }
 
 type ExternalIDGenerator func(coordinate.Coordinate) (string, error)
+
+// GenerateExternalIDForDocument generates an external ID for a document configuration. It is under 50 characters long and uses at most only "a-z", "A-Z", "0-9" and "-".
+func GenerateExternalIDForDocument(c coordinate.Coordinate) (string, error) {
```
jskelin, line 55: Could you rename `GenerateExternalID` so that name indicate it's just for Settings.

pkg/client/clientset.go
```diff
@@ -155,9 +155,8 @@ type BucketClient interface {
 type DocumentClient interface {
 	Get(ctx context.Context, id string) (documents.Response, error)
 	List(ctx context.Context, filter string) (documents.ListResponse, error)
-	Create(ctx context.Context, name string, data []byte, documentType documents.DocumentType) (documents.Response, error)
+	Create(ctx context.Context, name string, externalId string, data []byte, documentType documents.DocumentType) (documents.Response, error)
```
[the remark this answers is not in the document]
arthurpitman: This interface mirrors the shared client in `dynatrace/dynatrace-configuration-as-code`, so if this is something we want to do, I would do it in a follow up ticket

pkg/deploy/internal/document/document.go
arthurpitman: Could do, but in a follow up for all the deploy parts.

pkg/deploy/internal/document/document.go
arthurpitman: I'll give it a try to see if it makes it any clearer

pkg/deploy/internal/document/document.go
```diff
@@ -0,0 +1,174 @@
...
+
+	return entities.ResolvedEntity{
+		EntityName: documentName,
+		Coordinate: coordinate,
+		Properties: properties,
+	}
+}
+
+func getDocumentTypeFromConfigType(t config.Type) (documents.DocumentType, error) {
+	documentType, ok := t.(config.DocumentType)
```
jskelin, line 162: I'm not a fan (and a minority) of repetitive checks - the same check we had before the call.
But if we have it, I would put it as the very first check in the lifecycle of the `Deploy` function.

pkg/deploy/internal/document/document.go
jskelin: What do you think of the next order:
 * in case `ObjectOriginId` exists, use it (`id := c.ObjectOriginId`). Otherwise, try to find id via `externalId`
 * in case `id` is known, update the existing; otherwise, create a new entry.

With this approach, there will be only one place where the `client.Update` is called.

jskelin (dismissed):
There are two minor comments/suggestions.

arthurpitman:
From the spec, we need to see if we can update using the `originObjectId` before falling back to using the `externalId`, so two updates are necessary.

pkg/deploy/internal/document/document.go
arthurpitman: Two updates are indeed required as from the spec we must try with `originObjectId` and fall back to using `externalId` if the update doesn't work

pkg/deploy/internal/document/document.go
arthurpitman: `Deploy` itself doesn't offer much safety so I'd like to leave the check before using the value, but it does make sense to do it earlier

jskelin: nit: I suggest putting this check in the `Deploy` function as the first. To amplify visibility!

pkg/deploy/internal/document/document.go
jskelin: So, when the `originObjectID` is present, but there isn't a document, we try to find a document with the matching `externalId`?
What in the case where on the server exists both: a document with the same 'originObjectID` and another one that matches `externalId`?

arthurpitman: Correct, if `originObjectId` is present and there is a document with that ID then we use that. It has precedence over `externalId`.

pkg/deploy/internal/document/document.go
arthurpitman: Not sure I agree here, maybe we should then rework these types though!

jskelin (approved):

PR #1477, opened 2024-05-13
fix: Merge rather than overwrite downloaded configs of same type

arthurpitman:
With this PR, `download.copyConfigs(...)` merges the provided configs with those of the same type in the destination, rather than simply overwriting them.


UnseenWizzard (approved):
Technically we won't need this if we rename the new dashboard type, but great to have and avoid this error from happening again in the future.

PR #1479, opened 2024-05-13
chore: Use `dashboard-document` and `notebook-document` for document types

arthurpitman:

Currently the type `dashboard` is used for both platform and classic dashboard configs. This has two implications:
- It is not possible to have a platform dashboard and classic dashboard configuration with the same config ID
- A different value `type` must be chosen for platform dashboards  in delete files to ensure the entry is not ambiguous.

This PR removes the ambiguity by changing to `dashboard-document` for platform dashboards and `notebook-document` for notebooks.

Yes, with this PR, platform dashboard configs are specified as follows:

```
configs:
- id: my-dashboard
  config:
    name: My platform dashboard
    template: platform-dashboard.json
    skip: false
  type:
    document:
      type: dashboard-document
```

while for notebooks:

```
configs:
- id: my-notebook
  config:
    name: My demo notebook
    template: notebook.json
    skip: false
  type:
    document:
      type: notebook-document
```

#### Note
Documents are still behind `MONACO_FEAT_DOCUMENTS` feature flag

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -131,8 +131,8 @@ func (BucketType) ID() TypeId {
 }
 
 const (
-	DashboardType DocumentType = "dashboard"
-	NotebookType  DocumentType = "notebook"
+	DashboardType DocumentType = "dashboard-document"
+	NotebookType  DocumentType = "notebook-document"
 )
 
 // DocumentType identifies which document type is used in the config.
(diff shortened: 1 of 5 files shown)


UnseenWizzard (approved):
looks super strange in the sample... maybe our test users have some ideas as well.
Else my vote stands for `platform-dashboard`/`platform-notebook` - reads a bit better

UnseenWizzard:
> @arthurpitman  Let's use these for now and revisit the topic before completing the document support feature

Absolutely! That's why the PR is approved

PR #1489, opened 2024-05-24
fix(download): Log partial download failures as Warning

UnseenWizzard:
As monaco continues downloading and the command will not actuall fail if a single type or config can't be downloaded, logging these on Error level can be confusing and has caused customers to reach out to support.

Yes - should make it clearer that no error that would need an action is needed.

### DIFF
--- a/pkg/download/classic/download.go
+++ b/pkg/download/classic/download.go
@@ -100,7 +100,7 @@ func downloadConfigs(client client.ConfigClient, api api.API, projectName string
 	logger := log.WithFields(field.Type(api.ID))
 	foundValues, err := findConfigsToDownload(client, api, filters)
 	if err != nil {
-		logger.WithFields(field.Error(err)).Error("Failed to fetch configs of type '%v', skipping download of this type. Reason: %v", api.ID, err)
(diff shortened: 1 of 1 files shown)


arthurpitman (approved):

pkg/download/classic/download.go
```diff
@@ -100,7 +100,7 @@ func downloadConfigs(client client.ConfigClient, api api.API, projectName string
 	logger := log.WithFields(field.Type(api.ID))
 	foundValues, err := findConfigsToDownload(client, api, filters)
 	if err != nil {
-		logger.WithFields(field.Error(err)).Error("Failed to fetch configs of type '%v', skipping download of this type. Reason: %v", api.ID, err)
+		logger.WithFields(field.Error(err)).Warn("Failed to fetch configs of type '%v', skipping download of this type. Reason: %v", api.ID, err)
 		return results
```
arthurpitman, line 104:
```suggestion
		return []downloadedConfig{}
```

arthurpitman: Same nit for line 110

UnseenWizzard: true for this and line 110 there won't ever be anything in `results` - if fine by you, I'd elect to leave this as is for now though, as it's not functionally changing anything and is unrelated to the functional change of this PR.

PR #1490, opened 2024-05-27
fix: Close support archive at latest point in time

warber:
Users saw that when using `--support-archive` the very latest log messages are actually not contained in the support archive.
This PR fixes it by moving the code responsible for writing the support archive at the very end of program execution.

### DIFF
--- a/cmd/monaco/runner/runner.go
+++ b/cmd/monaco/runner/runner.go
@@ -35,7 +35,15 @@ import (
 )
 
 func Run() int {
-	rootCmd := BuildCli(afero.NewOsFs())
+	fs := afero.NewOsFs()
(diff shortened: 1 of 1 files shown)


arthurpitman (dismissed):

UnseenWizzard (changes requested):
Do I get it correctly, that the issue we're fixing with this, is that support-archives don't contain the logs from line 42 and 43?

If that's really necessary (isn't the final execution error just that errors occurred and to read the logs?), please implement this via `defer`. 
`cobra.OnFinalize` runs via defer and thus get's executed on panics - this change does not

UnseenWizzard (approved):

PR #1492, opened 2024-05-28
feat: Add support for private documents

arthurpitman:


cmd/monaco/generate/deletefile/test-resources/project/notebook-document/config.yaml
```diff
@@ -6,4 +6,4 @@ configs:
     skip: false
   type:
     document:
-      type: notebook-document
+      kind: notebook-document
```
warber, line 9: why is `kind` not just "notebook" or "dashboard" ?

arthurpitman: that change is coming in the next PR ;)

warber (approved):

pkg/client/clientset.go
```diff
@@ -156,8 +156,8 @@ type BucketClient interface {
 type DocumentClient interface {
 	Get(ctx context.Context, id string) (documents.Response, error)
 	List(ctx context.Context, filter string) (documents.ListResponse, error)
-	Create(ctx context.Context, name string, externalId string, data []byte, documentType documents.DocumentType) (documents.Response, error)
-	Update(ctx context.Context, id string, name string, data []byte, documentType documents.DocumentType) (documents.Response, error)
+	Create(ctx context.Context, name string, isPrivate bool, externalId string, data []byte, documentType documents.DocumentType) (documents.Response, error)
```
jskelin, line 159: I amplify my previous complaint about the increasing number of arguments.

Here are some thoughts:
 * https://matheus.ro/2018/01/29/clean-code-avoid-many-arguments-functions/
 * https://softwareengineering.stackexchange.com/a/145066/302528
 * https://dev.to/rweisleder/the-flag-parameter-anti-pattern-1j82

pkg/deploy/internal/document/document_mock_test.go
jskelin, line 1: This file shouldn't be in VCS!

pkg/deploy/internal/document/document.go
```diff
@@ -38,16 +38,16 @@ import (
 type Client interface {
```
jskelin, line 38: Does this interface need to be public?
Either make it private or (in the line above) create it in `document_mock.go` (without `_test` suffix

.gitignore
```diff
@@ -23,6 +23,7 @@ schemas/
 
 # Mock files
 **/*_mock.go
+**/*_mock_test.go
```
jskelin, line 26: See the comment on `document.go` file

pkg/deploy/internal/document/document.go
arthurpitman: Yes, it does as the type is used in the arguments of the `Deploy(...)` function, which itself is used from outside the package.

jskelin (approved):

PR #1493, opened 2024-05-28
feat: experimental event system support (POC)

warber:
This PR contains a POC that introduces a way handling application local events.
I've created a new package `events` that contains a very basic publish-subscribe like event system component that can be carried along with a go context and used wherever there is the need to emit an event.

On the other end it is possible to register subscribers for different types of events that are notified whenever a new event was sent. Note hat it is possible to aggregate events using a "correlation id". This is necessary because we need to be able to aggregate events that relate to each other. Subscribers shall receive only "internal" events and translate them to whatever they want to do with these events. In this PR there are two subscribers one for creating a "report" of deployed configs and one for creating an overall summary.

The feature can be tried out by enabling the feature flag `MONACO_FEAT_EVENTQUEUE` and performing a deployment.
As a result you should see two crated files one called `events_<timestamp>.jsonl` containing a line for each config Monaco processed during deployment. and another file called `summyry_<timestamp>.jsonl` containing a brief summary of the deployment containing information like the number of deployed configs, number of skipped configs, duration of exectution etc....

### DIFF
--- a/cmd/monaco/deploy/command.go
+++ b/cmd/monaco/deploy/command.go
@@ -20,8 +20,12 @@ import (
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/cmdutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/completion"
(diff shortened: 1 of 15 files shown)


warber:
Will be done in a new PR

PR #1494, opened 2024-05-29
rewrite `monaco.RunWithFs` to match `main.main`

jskelin:
The intention of the `monaco.RunWithFs` (and other variants) function, which is meant to be used in integration tests,  is to behave exactly as it should behave if the same command is executed via CLI. For that reason, some minor adaptation/ refactoring is done.

### DIFF
--- a/cmd/monaco/integrationtest/v2/all_configs_integration_test.go
+++ b/cmd/monaco/integrationtest/v2/all_configs_integration_test.go
@@ -21,9 +21,8 @@ package v2
 import (
 	"testing"
 
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/runner"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/integrationtest/utils/monaco"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/testutils"
 	"github.com/spf13/afero"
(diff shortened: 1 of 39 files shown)


cmd/monaco/integrationtest/v2/all_configs_integration_test.go
```diff
@@ -51,20 +50,12 @@ func runAllConfigsTest(t *testing.T, specificEnvironment string) {
...
 
 	RunIntegrationWithCleanupGivenEnvs(t, configFolder, manifest, specificEnvironment, "AllConfigs", envVars, func(fs afero.Fs, _ TestContext) {
-
 		// This causes a POST for all configs:
-
-		cmd := runner.BuildCli(fs)
-		cmd.SetArgs([]string{"deploy", "--verbose", manifest, "--environment", specificEnvironment})
-		err := cmd.Execute()
-
+		err := monaco.RunWithFsf(fs, "monaco deploy %s --environment=%s --verbose", manifest, specificEnvironment)
```
warber, line 54: Nit: maybe lets rename this function to avoid the question what a "Fsf" is :D
```
RunfWithFS(...)
```

jskelin: Can it be `RunWithFSf`?
There is some convention for this type of formatting that a function ends with `f`. Intellij recognizes this (check if enough arguments exist).

warber: ok :)

warber (approved):

PR #1500, opened 2024-06-05
feat: Split document config folders by document kind

arthurpitman:
This PR splits document configurations into different folders based on the document kind. That means:
- Dashboard document configurations are stored in `document-dashboard` within the project folder
- Notebook document configurations are stored in `document-notebook` within the project folder


pkg/persistence/config/writer/config_writer.go
```diff
@@ -149,7 +149,7 @@ func toTopLevelDefinitions(context *WriterContext, configs []config.Config) (map
 
 		apiCoord := apiCoordinate{
 			project: coord.Project,
-			api:     coord.Type,
+			api:     coord.extendedType,
```
UnseenWizzard, line 152: nit: maybe use the chae and rename this thing to type? and maybe the whole thing to something clearer that 'apiCoordinate' - can't think of anything though

warber (approved):

pkg/persistence/config/writer/config_writer.go
arthurpitman: I'm not sure here, so I think any further change should be done in a follow up ticket

PR #1501, opened 2024-06-10
Purge functionality for documents

jskelin:
This PR adds `purge` functionality for documents.

DT still doesn't support complete document deletion functionality. To be able to use the `delete` or `purge` command, the MONACO_FEAT_DELETE_DOCUMENTS flag needs to be set to true.

### DIFF
--- a/pkg/delete/delete.go
+++ b/pkg/delete/delete.go
@@ -87,7 +87,7 @@ func Configs(ctx context.Context, clients ClientSet, _ api.APIs, automationResou
 			}
 			err = bucket.Delete(ctx, clients.Buckets, entries)
 		} else if t == "document" {
-			if featureflags.Documents().Enabled() {
+			if featureflags.Documents().Enabled() && featureflags.DeleteDocuments().Enabled() {
 				if clients.Documents == nil {
 					log.WithCtxFields(ctx).WithFields(field.Type(t)).Warn("Skipped deletion of %d Document configuration(s) as API client was unavailable.", len(entries))
 					continue
@@ -141,6 +141,15 @@ func All(ctx context.Context, clients ClientSet, apis api.APIs) error {
 	} else if err := bucket.DeleteAll(ctx, clients.Buckets); err != nil {
 		log.Error("Failed to delete all Grail Bucket configurations: %v", err)
 		errs++
+	}
+
+	if featureflags.Documents().Enabled() && featureflags.DeleteDocuments().Enabled() {
+		if clients.Documents == nil {
+			log.Warn("Skipped deletion of Documents configurations as appropriate client was unavailable.")
+		} else if err := document.DeleteAll(ctx, clients.Documents); err != nil {
+			log.Error("Failed to delete all Document configurations")
+			errs++
+		}
 	}
 
 	if errs > 0 {
--- a/pkg/delete/delete.go
+++ b/pkg/delete/delete.go
@@ -147,7 +147,7 @@ func All(ctx context.Context, clients ClientSet, apis api.APIs) error {
 		if clients.Documents == nil {
 			log.Warn("Skipped deletion of Documents configurations as appropriate client was unavailable.")
 		} else if err := document.DeleteAll(ctx, clients.Documents); err != nil {
-			log.Error("Failed to delete all Document configurations")
+			log.Error("Failed to delete all Document configurations: %v", err)
 			errs++
 		}
 	}
--- a/pkg/delete/internal/classic/delete.go
+++ b/pkg/delete/internal/classic/delete.go
@@ -44,7 +44,7 @@ func Delete(ctx context.Context, client client.ConfigClient, dps []pointer.Delet
 		if theAPI.HasParent() {
 			parentID, e = resolveIdentifier(ctx, client, theAPI.Parent, toIdentifier(dp.Scope, "", ""))
 			if e != nil && !is404(e) {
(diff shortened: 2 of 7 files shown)


pkg/delete/delete.go
```diff
@@ -143,6 +143,15 @@ func All(ctx context.Context, clients ClientSet, apis api.APIs) error {
 		errs++
 	}
 
+	if featureflags.Documents().Enabled() && featureflags.DeleteDocuments().Enabled() {
+		if clients.Documents == nil {
+			log.Warn("Skipped deletion of Documents configurations as appropriate client was unavailable.")
+		} else if err := document.DeleteAll(ctx, clients.Documents); err != nil {
+			log.Error("Failed to delete all Document configurations")
```
arthurpitman, line 150: I think we should include the error in the log message here.

pkg/delete/internal/classic/delete.go
```diff
@@ -44,7 +44,7 @@ func Delete(ctx context.Context, client client.ConfigClient, dps []pointer.Delet
 		if theAPI.HasParent() {
 			parentID, e = resolveIdentifier(ctx, client, theAPI.Parent, toIdentifier(dp.Scope, "", ""))
 			if e != nil && !is404(e) {
-				log.WithFields(field.Error(e)).Error("unable to resolve config ID: %w")
+				log.WithFields(field.Error(e)).Error("unable to resolve config ID: %w", e)
```
arthurpitman, line 47: This will log the error twice

pkg/delete/internal/classic/delete.go
```diff
@@ -58,7 +58,7 @@ func Delete(ctx context.Context, client client.ConfigClient, dps []pointer.Delet
 		if id == "" {
 			id, e = resolveIdentifier(ctx, client, &a, toIdentifier(dp.Identifier, dp.ActionType, dp.Domain))
 			if e != nil && !is404(e) {
-				log.WithFields(field.Error(e)).Error("unable to resolve config ID: %w")
+				log.WithFields(field.Error(e)).Error("unable to resolve config ID: %w", e)
```
arthurpitman, line 61: As above

pkg/delete/internal/document/delete.go
```diff
@@ -40,7 +40,7 @@ func Delete(ctx context.Context, c client, dps []pointer.DeletePointer) error {
 	for _, dp := range dps {
 		err := deleteSingle(ctx, c, dp)
 		if err != nil {
-			log.WithCtxFields(ctx).WithFields(field.Type(dp.Type), field.Coordinate(dp.AsCoordinate())).Error("Failed to delete entry: %v", err)
+			log.WithCtxFields(ctx).WithFields(field.Type(dp.Type), field.Coordinate(dp.AsCoordinate())).Error("Failed to delete entry: %s", err)
```
arthurpitman, line 43: We should make the error logging consistent

pkg/delete/internal/document/delete.go
```diff
@@ -108,3 +107,19 @@ func isAPIErrorStatusNotFound(err error) bool {
...
+	listResponse, err := c.List(ctx, fmt.Sprintf("type='%s' or type='%s'", documents.Dashboard, documents.Notebook))
+	if err != nil {
+		return err
+	}
+
+	var retErr error
+	for _, x := range listResponse.Responses {
+		err := deleteSingle(ctx, c, pointer.DeletePointer{Type: x.Type, OriginObjectId: x.ID})
+		if err != nil {
+			retErr = errors.Join(retErr, err)
```
arthurpitman, line 121: I think it would be better to accumulate the errors first (for example in a slice) and then join them at the end outside of the loop

pkg/delete/internal/classic/delete.go
jskelin: for delete it is not the case. And I'm not going to expand the scope of this PR and fix it.

pkg/delete/internal/document/delete.go
jskelin: In this case it doesn't have any difference.

pkg/delete/internal/classic/delete.go
arthurpitman:
```suggestion
				log.WithFields(field.Error(e)).Error("unable to resolve config ID: %s", e.Error())
```

pkg/delete/internal/classic/delete.go
arthurpitman, line 61:
```suggestion
				log.WithFields(field.Error(e)).Error("unable to resolve config ID: %s", e.Error())
```

pkg/delete/delete.go
```diff
@@ -143,6 +143,15 @@ func All(ctx context.Context, clients ClientSet, apis api.APIs) error {
 		errs++
 	}
 
+	if featureflags.Documents().Enabled() && featureflags.DeleteDocuments().Enabled() {
+		if clients.Documents == nil {
+			log.Warn("Skipped deletion of Documents configurations as appropriate client was unavailable.")
+		} else if err := document.DeleteAll(ctx, clients.Documents); err != nil {
+			log.Error("Failed to delete all Document configurations: %v", err)
```
arthurpitman, line 150:
```suggestion
			log.Error("Failed to delete all Document configurations: %s", err.Error())
```

arthurpitman (approved):

PR #1512, opened 2024-06-26
OpenPipeline e2e test resources

warber:
This PR adds OpenPipeline test resources to the e2e tests

### DIFF
--- a/cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/openpipeline/events.json
+++ b/cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/openpipeline/events.json
@@ -1,0 +1,88 @@
+{
+    "customBasePath": "/platform/ingest/custom/events",
+    "editable": true,
+    "endpoints": [
+        {
+            "basePath": "/platform/ingest/v1/events",
+            "builtin": true,
+            "defaultBucket": null,
+            "displayName": "Default API",
+            "editable": false,
+            "enabled": true,
+            "processors": [],
+            "routing": {
+                "type": "dynamic"
+            },
+            "segment": ""
+        }
+    ],
+    "id": "events",
+    "pipelines": [
+        {
+            "builtin": true,
+            "dataExtraction": {
+                "editable": false,
+                "processors": []
+            },
+            "displayName": "events",
+            "editable": false,
+            "enabled": true,
+            "id": "default",
+            "metricExtraction": {
+                "editable": false,
+                "processors": []
+            },
+            "processing": {
+                "editable": false,
+                "processors": []
+            },
+            "securityContext": {
+                "editable": false,
+                "processors": []
+            },
+            "storage": {
+                "catchAllBucketName": "default_events",
+                "editable": false,
+                "processors": []
+            },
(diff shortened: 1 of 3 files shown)


cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/openpipeline/events.json
```diff
@@ -0,0 +1,60 @@
...
+      "enabled": true,
+      "processors": [],
+      "routing": {
+        "type": "dynamic"
+      },
+      "segment": ""
+    }
+  ],
+  "id": "events",
+  "pipelines": [
```
UnseenWizzard, line 20: Is this just the default configuration of this pipeline? If yes, could you please make this a custom configuration - e.g. by adding a pipeline, or such - so we can actually test applying a configuration and can't run into a situation where an optimized API just does nothing.

cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/openpipeline/events.json
```diff
@@ -0,0 +1,60 @@
...
+  ],
+  "routing": {
+    "catchAllPipeline": {
+      "editable": false,
+      "pipelineId": "default"
+    },
+    "editable": true,
+    "entries": []
+  },
+  "version": "1718793081420-294062f9e129495899197a5c9a0fb83a"
```
UnseenWizzard, line 59: q: can the version be stripped from the sample? IIRC it's ignored on create/update anyway and generated by the API?

cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/openpipeline/events.json
warber: I've added a second pipeline

cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/openpipeline/events.json
warber: yep

UnseenWizzard (approved):

cmd/monaco/integrationtest/v2/test-resources/integration-all-configs/project/openpipeline/events.json
```diff
@@ -0,0 +1,88 @@
...
+                "processors": []
+            },
+            "storage": {
+                "catchAllBucketName": "default_events",
+                "editable": false,
+                "processors": []
+            },
+            "type": "default"
+        },
+        {
```
arthurpitman, line 50: Adding a second pipeline to make this a non-default config is a good idea, but considering we can't delete (or cleanup) configurations, how do we track whether changes are really being accepted by the API?

warber: This would require a separate/dedicated e2e test which does the required assertions, we can think about adding one (in a follow up) :)

arthurpitman (approved):

PR #1514, opened 2024-06-26
feat: Make feature flag changes visible

UnseenWizzard:
Aim of this is to provide us and support with an overview of featureflags in support archives, and users with a warning if they've modified a feature flag. 

To be able to iterate over a list of FFs for logging them, the first commit proposes a move from FF definitions via functions to using a map. 
This makes access a bit more tedious, so it's only done for two to show the concept - if someone has a better idea, please propose it!

The changes in the second commit then produce: 

A warning if flags are modified at the start of our log: 

```sh
2024-07-12T13:01:34+02:00	info	Monaco version 2.x
2024-07-12T13:01:34+02:00	warn	Feature Flags modified - Dynatrace Support might not be able to assist you with issues.
2024-07-12T13:01:34+02:00	debug	Default soft memory limit set: 2.1 GB
```

And a file in support archive (and the .logs folder - if an archive is written) that lists FF state and defaults

```
Lines starting with '!' indicate that a flag has been modified from its default value.

Feature Flags:

 	MONACO_ENABLE_DANGEROUS_COMMANDS: false (default:false)
!	MONACO_FEAT_DOWNLOAD_FILTER: false (default:true)
 	MONACO_FEAT_DOWNLOAD_FILTER_CLASSIC_CONFIGS: true (default:true)
 	MONACO_FEAT_DOWNLOAD_FILTER_SETTINGS: true (default:true)
 	MONACO_FEAT_DOWNLOAD_FILTER_SETTINGS_UNMODIFIABLE: true (default:true)
 	MONACO_FEAT_EXTRACT_SCOPE_AS_PARAMETER: false (default:false)
 	MONACO_FEAT_FAST_DEPENDENCY_RESOLVER: false (default:false)
 	MONACO_FEAT_SIMPLE_CLASSIC_URL: true (default:true)
 	MONACO_FEAT_UPDATE_SINGLE_NON_UNIQUE_BY_NAME: true (default:true)
 	MONACO_FEAT_USE_MZ_NUMERIC_ID: true (default:true)
 	MONACO_FEAT_VERIFY_ENV_TYPE: true (default:true)
 	MONACO_LOG_FILE_ENABLED: true (default:true)
!	MONACO_SKIP_VERSION_CHECK: true (default:false)


Development and Experimental Flags:

 	MONACO_FEAT_DELETE_DOCUMENTS: false (default:false)
!	MONACO_FEAT_DOCUMENTS: true (default:false)
 	MONACO_FEAT_OPENPIPELINE: false (default:false)
 	MONACO_FEAT_PERSIST_SETTINGS_ORDER: false (default:false)
 	MONACO_SKIP_READ_ONLY_ACCOUNT_GROUP_UPDATES: false (default:false)

```

### DIFF
--- a/internal/featureflags/state.go
+++ b/internal/featureflags/state.go
@@ -25,19 +25,17 @@ import (
 
 // AnyModified returns true if any Permanent or Temporary feature flag value is different to its default.
 func AnyModified() bool {
-	for _, v := range Permanent {
-		enabled, def := v.Value()
-		if enabled != def {
-			return true
-		}
-	}
-	for _, v := range Temporary {
-		enabled, def := v.Value()
-		if enabled != def {
-			return true
-		}
-	}
+	return anyFeatureFlagModified(Permanent) || anyFeatureFlagModified(Temporary)
+}
 
+// anyFeatureFlagModified returns true if any feature flag value is different to its default.
+func anyFeatureFlagModified[K TemporaryFlag | PermanentFlag](featureFlags map[K]FeatureFlag) bool {
+	for _, v := range featureFlags {
+		enabled, def := v.Value()
+		if enabled != def {
+			return true
+		}
+	}
 	return false
 }
 
@@ -50,11 +48,21 @@ func StateInfo() string {
 		s.WriteString("Lines starting with '!' indicate that a flag has been modified from its default value.\n\n")
 	}
 
-	_, _ = fmt.Fprintf(&s, "Feature Flags:\n\n")
-	permanentKeys := maps.Keys(Permanent)
-	slices.Sort(permanentKeys)
-	for _, k := range permanentKeys {
-		v := Permanent[k]
+	s.WriteString("Feature Flags:\n\n")
+	s.WriteString(makeFeatureFlagTableString(Permanent))
+
+	s.WriteString("\n\nDevelopment and Experimental Flags:\n\n")
+	s.WriteString(makeFeatureFlagTableString(Temporary))
+
+	return s.String()
+}
+
+func makeFeatureFlagTableString[K TemporaryFlag | PermanentFlag](featureFlags map[K]FeatureFlag) string {
+	s := strings.Builder{}
+	keys := maps.Keys(featureFlags)
+	slices.Sort(keys)
+	for _, k := range keys {
+		v := featureFlags[k]
 		enabled, def := v.Value()
 		modifiedStr := " "
 		if enabled != def {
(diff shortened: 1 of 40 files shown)


arthurpitman:
I like the idea of this PR :) 
- Would it also make sense to include the state of the permanent feature flags in the support archive if we don't already?
- Would it be easier to use the environment variable as the key for the feature flag maps? 
- Should we pick more user-focused names for "temporary" and "permanent" feature flags, e.g. "stable feature flags" and "development feature flags"? Going back to the purpose of feature flags, my understanding is permanent ones are meant to aid the user in managing significant changes in behavior that they may or may not want to opt into, where as temporary ones are meant to aid us during development while not disturbing users.

internal/featureflags/state.go
```diff
@@ -0,0 +1,80 @@
...
+	return false
+}
+
+// StateInfo builds a string message describing the current and default values of all feature flags,
+// noting especially if any flag has been changed off its default value.
+func StateInfo() string {
+	s := strings.Builder{}
+
+	if AnyModified() {
+		s.WriteString("Some flags are modified off default!\nLines starting with '!' indicate modification\n\n")
```
arthurpitman, line 50:
```suggestion
		s.WriteString("Lines starting with '!' indicate that a flag has been modified from its default value\n\n")
```

internal/featureflags/state.go
```diff
@@ -0,0 +1,80 @@
...
+	slices.Sort(permanentKeys)
+	for _, k := range permanentKeys {
+		v := Permanent[k]
+		enabled, def := v.Value()
+		modifiedStr := " "
+		if enabled != def {
+			modifiedStr = "!"
+		}
+		_, _ = fmt.Fprintf(&s, "%v\t%v: %v (default:%v)\n", modifiedStr, v.envName, enabled, def)
+	}
```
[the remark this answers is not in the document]
UnseenWizzard: I didn't find a good way to extract this or the 'AnyModified' checks into generic functions, as the map keys are dedicated types rather than plain strings (to reduce the risk of mixing up keys/maps).
So I've opted for duplication rather.

If you have an idea that's not too messy, please suggest it :)

internal/featureflags/state.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return s.String()
+}
```
arthurpitman, line 80:
```suggestion
func AnyModified() bool {
	return anyFeatureFlagModified(Permanent) || anyFeatureFlagModified(Temporary)
}

// anyFeatureFlagModified returns true if any feature flag value is different to its default.
func anyFeatureFlagModified[K TemporaryFlag | PermanentFlag](featureFlags map[K]FeatureFlag) bool {
	for _, v := range featureFlags {
		enabled, def := v.Value()
		if enabled != def {
			return true
		}
	}
	return false
}

// StateInfo builds a string message describing the current and default values of all feature flags,
// noting especially if any flag has been changed off its default value.
func StateInfo() string {
	s := strings.Builder{}

	if AnyModified() {
		s.WriteString("Lines starting with '!' indicate that a flag has been modified from its default value.\n\n")
	}

	s.WriteString("Feature Flags:\n\n")
	s.WriteString(makeFeatureFlagTableString(Permanent))

	s.WriteString("\n\nDevelopment and Experimental Flags:\n\n")
	s.WriteString(makeFeatureFlagTableString(Temporary))

	return s.String()
}

func makeFeatureFlagTableString[K TemporaryFlag | PermanentFlag](featureFlags map[K]FeatureFlag) string {
	s := strings.Builder{}
	keys := maps.Keys(featureFlags)
	slices.Sort(keys)
	for _, k := range keys {
		v := featureFlags[k]
		enabled, def := v.Value()
		modifiedStr := " "
		if enabled != def {
			modifiedStr = "!"
		}
		_, _ = fmt.Fprintf(&s, "%v\t%v: %v (default:%v)\n", modifiedStr, v.envName, enabled, def)
	}
	return s.String()
}
```

arthurpitman: Fair point though, it doesn't cut so much code.

arthurpitman (approved):

PR #1515, opened 2024-06-27
feat: Add Fixed parameter for `network-zones`

### LINKED ISSUE #1356: Monaco 2.10.0 - incorrect order deploying network zones
During deployment Monaco tries to deploy management:network-zone before builtin:networkzones
it fails on first run with following message:
`Failed to send HTTP request: (HTTP 400)!     Response was: Not allowed because network zones are disabled`
it is successful on rerun

**How to reproduce**
- new environment with network zones disabled
- management:network-zone set to enable in /builtinnetworkzones/config.yaml+ json
- a network zone described network-zone/config.yaml + json

**Expected behavior**
monaco turns on network zones setting and sets them up

Monaco 2.10.0

warber:
When both, `builtin:networkzones` and `network-zone` apis are deployed, we need to make sure the settings 2.0 object is deployed first. Hence we introduce a fixed parameter as soon as we see both config appearing in a config set.

First commit fixes the problem, second commit reduces cognitive complexity of that code area in general.

fixes #1356

### DIFF
--- a/pkg/project/v2/project_loader.go
+++ b/pkg/project/v2/project_loader.go
@@ -182,28 +182,24 @@ func loadProject(fs afero.Fs, context ProjectLoaderContext, projectDefinition ma
 	log.Debug("Loading project `%s` (%s)...", projectDefinition.Name, projectDefinition.Path)
 
 	configs, errors := loadConfigsOfProject(fs, context, projectDefinition, environments)
+	errors = append(errors, findDuplicatedConfigIdentifiers(configs)...)
+	errors = append(errors, checkKeyUserActionScope(configs)...)
 
-	if d := findDuplicatedConfigIdentifiers(configs); d != nil {
-		for _, c := range d {
-			errors = append(errors, newDuplicateConfigIdentifierError(c))
-		}
-	}
-
-	for _, c := range configs {
-		// The scope parameter of a key user actions web configuration needs to be a reference to another application-web config
-		// The reference parameter makes sure that rely on the fact that kua web configs are loaded/deployed within the same
-		// sub graph (independent component) later on as
-		if c.Coordinate.Type == api.KeyUserActionsWeb {
-			if _, ok := c.Parameters[config.ScopeParameter].(*ref.ReferenceParameter); !ok {
-				errors = append(errors, fmt.Errorf("scope parameter of config of type '%s' with ID '%s' needs to be a reference "+
-					"parameter to another web-application config", api.KeyUserActionsWeb, c.Coordinate.ConfigId))
-			}
-		}
-	}
-	if errors != nil {
+	if len(errors) > 0 {
 		return Project{}, errors
 	}
 
+	insertFixedParameters(configs)
+
+	return Project{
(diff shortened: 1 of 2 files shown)


pkg/project/v2/project_loader.go
```diff
@@ -191,22 +189,46 @@ func loadProject(fs afero.Fs, context ProjectLoaderContext, projectDefinition ma
...
+	for _, c := range configs {
+		if c.Coordinate.Type == api.NetworkZone {
+			networkZones = append(networkZones, c)
+		}
+		if c.Coordinate.Type == "builtin:networkzones" {
+			networkZoneEnabled = c
+			networkZoneEnabledFound = true
+		}
+	}
+	// if we've found "networkzones" and "builtin:networkzones" settings 2.0 objects, then insert a fixed parameter.
```
UnseenWizzard, line 219: nit: this comment feels somewhat redundant, you have extensive description before the block, and that this loop adds the ref param is IMO clear from code.

pkg/project/v2/project_loader.go
```diff
@@ -191,22 +189,46 @@ func loadProject(fs afero.Fs, context ProjectLoaderContext, projectDefinition ma
...
 					"parameter to another web-application config", api.KeyUserActionsWeb, c.Coordinate.ConfigId))
 			}
 		}
 	}
-
 	if errors != nil {
 		return Project{}, errors
 	}
 
+	// Network zone API configs have a hard dependency on the "builtin:networkzones" setting 2.0 config
```
UnseenWizzard, line 203: Please describe WHY we actually need to do this, currently this extensive doc comment mainly tells me that there's some "dependency" and WHAT the code below does.

What I get from the code and variable naming, the setting is for (de)activating networkzones in general, so it needs to be configured for any zones (as their API calls would fail, if the setting is off I assume?)? 
If that's correct, the code actually tells me more about the WHY than 3 lines of comments ;)

pkg/project/v2/project_loader.go
```diff
@@ -182,28 +182,24 @@ func loadProject(fs afero.Fs, context ProjectLoaderContext, projectDefinition ma
...
+
+	return Project{
+		Id:           projectDefinition.Name,
+		GroupId:      projectDefinition.Group,
+		Configs:      toConfigMap(configs),
+		Dependencies: toDependenciesMap(projectDefinition.Name, configs),
+	}, nil
+}
+
+func insertFixedParameters(configs []config.Config) {
```
UnseenWizzard, line 202: I assume this is named with the intention, that it COULD do more things than just adding the one fixed reference for network zones in the future? 
IMO it can be changed to that if it ever happens - for now it'd be nicer to read with a clear name and by moving the documentation of WHY from comments inside the function to one on the function itself. 
Also a note that it modifies the configs it's given might not hurt

UnseenWizzard (approved):
Functionality looks good (if I understood correctly why we do this - see comments)
Comments are basically nitpicks on naming and documentation, and could be deferred if this fix needs to ship

UnseenWizzard (approved):

PR #1517, opened 2024-07-02
Avoid using whitespaces as IDs for Monaco coordinates

warber:
For some configuration types Monaco uses e.g. "names" as ID field in coordinates. These names can naturally contain whitespaces which are then part of monaco coordinates. While it is not a problem "per se" it introduces difficulties for tools that are processing coordinates, e.g. Upgrade Assistant. Further, it seems cleaner to not use whitespaces in identifiers. 

This, PR introduces logic to sanitize the values used as IDs inside coordinates after they are downloaded.

### DIFF
--- a/cmd/monaco/download/download_configs.go
+++ b/cmd/monaco/download/download_configs.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/download/openpipeline"
 	"os"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


warber:
Since e.g. key user actions with "MY KUA" and "MYKUA", will effectively result in the same sanitized ID, this is is not a good enough solution. Maybe a better approach would be to use something like base64 encoding for the IDs - with the side effect that IDs of downloaded configs wouldn't give the user a hint about the actual config items anymore.

warber:
Won't do at the moment. If this is still necessary, the idea would be to introduce a FF to let monaco generate uuids during download and use this as the id for the coordinate.

PR #1524, opened 2024-07-11
Remove old temporary feature flags

UnseenWizzard:
We have a large amount of FFs allowing to deactivate features we felt uncertain about, as well as development flags for features already released.

It also moves some flags I didn't feel safe removing to the 'permantent.go' file.

This was done to make reworking FFs so we can log them simpler by having fewer flags.

Yes and No - some features can no longer be turned off, via FFs that are not documented.


UnseenWizzard:
Test failures are Account Management E2E tests, which seem to be currently failing to clean up after themselves and thus fail with 400 errors from the API when trying to create more policies than possible.

arthurpitman (approved):

PR #1531, opened 2024-07-23
test: create initially public documents in e2e tests

warber:
based on https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1529 and https://github.com/Dynatrace/dynatrace-configuration-as-code-core/pull/118 this PR makes minor adjustments to the documents related e2e tests in order to test creating documents that are "initially public"

### DIFF
--- a/cmd/monaco/integrationtest/v2/documents_integration_test.go
+++ b/cmd/monaco/integrationtest/v2/documents_integration_test.go
@@ -46,11 +46,11 @@ func TestDocuments(t *testing.T) {
 
 	RunIntegrationWithCleanupGivenEnvs(t, configFolder, manifest, specificEnvironment, "Documents", envVars, func(fs afero.Fs, _ TestContext) {
 
-		// Create the buckets
+		// Create the documents
 		err := monaco.RunWithFsf(fs, "monaco deploy %s --project=project --verbose", manifest)
 		assert.NoError(t, err)
 
-		// Update the buckets
+		// Update the documents
 		err = monaco.RunWithFsf(fs, "monaco deploy %s --project=project --verbose", manifest)
 		assert.NoError(t, err)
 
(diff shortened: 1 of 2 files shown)


cmd/monaco/integrationtest/v2/documents_integration_test.go
```diff
@@ -92,7 +92,7 @@ func TestPrivateDocuments(t *testing.T) {
 		result, err := clientSet.Document().List(context.TODO(), fmt.Sprintf("name='my-notebook_%s'", testContext.suffix))
 		assert.NoError(t, err)
 		assert.Len(t, result.Responses, 1)
-		assert.True(t, result.Responses[0].IsPrivate)
+		assert.False(t, result.Responses[0].IsPrivate)
 
 		// change private field to false
```
arthurpitman, line 97:
```suggestion
		// change private field to true
```

arthurpitman: Also comment on line 90 needs updating, changing `true` to `false`

cmd/monaco/integrationtest/v2/test-resources/integration-documents/project/document-notebook/config.yaml
```diff
@@ -7,4 +7,4 @@ configs:
   type:
     document:
       kind: notebook
-      private: true
+      private: false
```
arthurpitman, line 10: Would we prefer to go with the default? I.e. omit `private`

cmd/monaco/integrationtest/v2/documents_integration_test.go
arthurpitman: And also line 112

arthurpitman (commented):
Would we prefer to keep the existing "initially private" test and add a new one rather than updating the existing one?

cmd/monaco/integrationtest/v2/test-resources/integration-documents/project/document-notebook/config.yaml
warber: also thought about it. Changing the value from "not present" to "private: true" in the test is then a bit more messy.
I think it is fine.

arthurpitman (approved):

PR #1542, opened 2024-07-31
feat: excluded skipped configs from being added to the dependency graph

warber:
If the feature flag `MONACO_FEAT_IGNORE_SKIPPED_CONFIGS` is enabled, configs having the property "skip" set to true will not only ignored during deployment but also excluded from the dependency graph. Hence, they are also validated prior to the deployment and referencing properties to these configs is not possible.


jskelin:
Please, C/P the integration tests but with this new FF enabled, just to be sure that everything works as expected.

pkg/graph/graph.go
```diff
@@ -221,6 +222,10 @@ func buildDependencyGraph(projects []project.Project, environment string, nodeOp
 
 	log.Debug("adding %d Config nodes to graph...", len(configs))
```
arthurpitman, line 223: Nit: Technically this log line might be wrong if the feature flag is enabled as then they wont be added.

arthurpitman (approved):

arthurpitman (approved):

PR #1543, opened 2024-08-02
Avoid creating circular dependencies between `network-zone` configs

warber:
Network zones can point to each other - if Monaco turn these pointers into actual parameters we end up creating circular dependencies which are not allowed in Monaco. Hence this PR skips extracting IDs out of the json payload and turning them into paramters for `network-zone` type of configs.

This PR apart from `dashboards` obviously only additionally covers `network-zone`s. Wheter this is enough or not depends a bit on how many other configurations can be configugurd in DT with "links" between each others. 
A possible improvement would be to detect such links during download, by detecting potential cirular dependencies that were caused by Monaco itself by injecting ID parameters.


jskelin (approved):

warber:
This change potentially introduces other errors where all of a sudden network zones cannot be migrated to other environments

PR #1547, opened 2024-08-05
chore: Update core library

arthurpitman:
This PR makes minimal changes to use the changes introduced into the dynatrace-configuration-as-code-core library with https://github.com/Dynatrace/dynatrace-configuration-as-code-core/pull/114


UnseenWizzard (dismissed):
looks all right (like you, just trusting the tests here though ;) )

PR #1557, opened 2024-08-14
fix: dependency circle between dashboards and dashboard-share-settings

warber:
Dashboards and Dashboard-share-settings have the same ids which causes clashes in code that collects configs by IDs, i.e. keys are overwritten, e.g. keys for dashboard-share-settings override keys for dashboards. 

Depending on the order of configs, this ultimately caused the code for excluding references between dashboards to be skipped.

I couldn't find out a strong requirement for using same IDs for this kind of parent-child configs.
So I've deleted the test for it. It would be great if somebody could verify that this is actually not a hard requirement.

### DIFF
--- a/pkg/download/classic/download.go
+++ b/pkg/download/classic/download.go
@@ -141,9 +141,6 @@ type value struct {
 }
 
 func (v value) id() string {
-	if v.value.Id == v.parentConfigId {
-		return v.value.Id
-	}
 	return v.value.Id + v.parentConfigId
 }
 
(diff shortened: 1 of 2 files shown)


jskelin (approved):

pkg/download/classic/download.go
```diff
@@ -141,9 +141,6 @@ type value struct {
 }
 
 func (v value) id() string {
-	if v.value.Id == v.parentConfigId {
```
UnseenWizzard, line 144: Did you test if dashboard-share-settings are correctly created after that change @warber?

I think this might have been added before the introduction of creating an API instance with the parent's ID pre-coded into the API path (`pkg/api/api.go:98`). We might have coded the ID into the URL via the config ID at the start and this ensures we have the right ID available? @arthurpitman does that maybe sound familiar as a reason to introduce this code?

warber: @UnseenWizzard yes, it works 🤷

UnseenWizzard (approved):

PR #1562, opened 2024-08-22
fix: avoid dots in extracted parameters

warber:
during download, we extract and convert IDs into parameters, we also do this for the "scope" parameter. Depending on the settings object the scope could contain dots (".") which are then used as a key of the reference parameters However, monaco fails to resolve these parameters as monaco assumes dots to be nested structures. Example:

```yaml
- id: 1
  config:
    parameters:
      extractedIDs:
        type: value
        value:
          id_metric_postgres.session_time.count: metric-postgres.session_time.count
    template: 1.json
    skip: false
  type:
    settings:
      schema: builtin:metric.metadata
      schemaVersion: "4.28"
      scope:
        property: extractedIDs.id_metric_postgres.session_time.count
        type: reference
```

monaco fails to resolve the parameter "extractedIDs.id_metric_postgres.session_time.count" as it assumes the following parameter structure: 

```yaml
parameters:
      extractedIDs:
        type: value
        value:
          id_metric_postgres:
            session_time:
              count: metric-postgres.session_time.count
```

This PR avoids running into this situation by simply making sure that parameter keys do not contain dots.

### DIFF
--- a/pkg/download/id_extraction/id_extraction.go
+++ b/pkg/download/id_extraction/id_extraction.go
@@ -101,6 +101,7 @@ func findAllIds(content string) []string {
 }
 
 func createParameterKey(id string) string {
-	idKey := strings.ReplaceAll(id, "-", "_") // golang template keys must not contain hyphens
+	idKey := strings.ReplaceAll(id, "-", "_")   // golang template keys must not contain hyphens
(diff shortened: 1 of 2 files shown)


UnseenWizzard (approved):

UnseenWizzard:
Ignoring flaky E2Es due to incomplete state of API clients on main

PR #1566, opened 2024-08-28
chore: removed v1 calculated metrics log api

warber:
This API seems to be not available anymore and is replaced by a settings 2.0 config.
Hence this API is marked as deprecated.


UnseenWizzard (changes requested):
Please do not not remove it, but mark it as deprecated by the setting!
On managed environments of older versions this API will still exist

PR #1574, opened 2024-09-11
fix: better max filename detection

warber:
This approach works also if due to deduplication of file names appending a suffix to the filename results in exceeding the limit again. Sanitizing files names and Sanitizing strings are separate now. Only when writing the template files, the file name deduplicate logic and filename length checks are done. In case there are still troubles with the default max length (=254) the user can override it using `MONACO_MAX_FILENAME_LEN` environment variable. Note that 2 places of the max fileName lenght are actually reserved for the duplication counter.


pkg/persistence/config/writer/config_writer.go
```diff
@@ -791,6 +792,36 @@ func fmtDetailedConfigWriterError(context *serializerContext, format string, arg
 	}
 }
 
+var maxFileNameLen = 254
```
jskelin, line 795: No need for this line. The default value will provide `GetEnvValueInt` in line 803.

jskelin (approved):
Generaly ok with some minor complaints.

pkg/persistence/config/writer/config_writer.go
warber: if its set using `GetEnvValueInt` then yes. Other than that I'd rather have it inititialized to a reasonable value if that's ok for you.

warber: Actually, this doesn't need to be a variable outside the function anymore at all.

jskelin (approved):

PR #1577, opened 2024-09-18
fix: only consider movable objects when injecting `insertAfter` parameters

warber:
When the `MONACO_FEAT_PERSIST_SETTINGS_ORDER` feature flag is turned on, Monaco is trying to inject `insertAfter` parameters in order to preserve the actual ordering of downloaded settings. However, it turned out that some of the settings objects are marked as not movable which ends up in an API error if `insertAfter`is used for these objects (even if the ordering does not change). This PR makes sure that `insertAfter`parameters are only created when the settings object is movable.

This makes turning on the feature flag safe(r) as Monaco will produce configuration sets that are deployable again.

### DIFF
--- a/pkg/download/settings/download_test.go
+++ b/pkg/download/settings/download_test.go
@@ -42,6 +42,7 @@ import (
 
 func TestDownloadAll(t *testing.T) {
 	uuid1 := idutils.GenerateUUIDFromString("oid1")
+	uuid2 := idutils.GenerateUUIDFromString("oid2")
 	uuid3 := idutils.GenerateUUIDFromString("oid3")
 
 	type mockValues struct {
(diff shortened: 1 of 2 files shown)


pkg/download/settings/download_test.go
```diff
@@ -334,6 +339,10 @@ func TestDownloadAll(t *testing.T) {
 							ObjectId:      "oid3",
 							Scope:         "tenant",
 							Value:         json.RawMessage("{}"),
+							ModificationInfo: &dtclient.SettingsModificationInfo{
+								Modifiable: true,
+								Movable:    true,
```
arthurpitman, line 344: I think we should also have an object that is _not_ movable as well here

arthurpitman (approved):

PR #1578, opened 2024-09-20
fix: include concurrent rate limit settings while creating new client…

jskelin:
… (update the core with bugfix)

Include fixed version of core library


go.mod
```diff
@@ -38,6 +38,6 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-go 1.22
+go 1.22.0
```
arthurpitman, line 41: I think this shouldn't be there?

arthurpitman (changes requested):

go.mod
jskelin: There is no need to prevent bugfix for this version of go.
Just learned today: https://go.dev/ref/mod#glos-patch-version

arthurpitman (approved):

PR #1583, opened 2024-09-26
Enhance HTTP traffic logging

warber:
This PR resolves some issues with logging HTTP traffic in Monaco:

* Avoid calling `file::sync`unnecessarily after writing each entry into the files
* Avoid using multiple instances of traffic logger
* Use Buffered Writer to write content into files
* Flush / Sync content once before adding files to the zip archive

### DIFF
--- a/cmd/monaco/runner/runner.go
+++ b/cmd/monaco/runner/runner.go
@@ -39,7 +39,9 @@ import (
 func RunCmd(fs afero.Fs, command *cobra.Command) error {
 	defer func() { // writing the support archive is a deferred function call in order to guarantee that a support archive is also written in case of a panic
 		if support.SupportArchive {
-			trafficlogs.GetInstance().Sync()
+			if err := trafficlogs.GetInstance().Sync(); err != nil {
+				log.WithFields(field.Error(err)).Error("Encountered error while syncing/flushing traffic log files: %s", err)
+			}
 			if err := support.Archive(fs); err != nil {
 				log.WithFields(field.Error(err)).Error("Encountered error creating support archive. Archive may be missing or incomplete: %s", err)
 			}
--- a/cmd/monaco/runner/runner.go
+++ b/cmd/monaco/runner/runner.go
@@ -15,6 +15,7 @@
 package runner
 
 import (
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/trafficlogs"
 	"io"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/account"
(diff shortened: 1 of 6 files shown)


internal/trafficlogs/logger.go
```diff
@@ -99,12 +106,28 @@ func (l *FileBasedLogger) Log(req *http.Request, reqBody string, resp *http.Resp
...
+		l.reqLogFile.Sync()    // nolint:errcheck
+		l.reqLogFile = nil
+	}
+
+	if l.respLogFile != nil {
+		l.respBufWriter.Flush() // nolint:errcheck
+		l.respLogFile.Sync()    // nolint:errcheck
+		l.respLogFile = nil
+	}
+}
```
arthurpitman, line 123: I think it would be good to collect and return the errors here

cmd/monaco/runner/runner.go
```diff
@@ -38,6 +39,7 @@ import (
 func RunCmd(fs afero.Fs, command *cobra.Command) error {
 	defer func() { // writing the support archive is a deferred function call in order to guarantee that a support archive is also written in case of a panic
 		if support.SupportArchive {
+			trafficlogs.GetInstance().Sync()
```
arthurpitman, line 42: Assuming this could return an error, we should log it too

arthurpitman (approved):

internal/trafficlogs/logger.go
```diff
@@ -164,13 +173,13 @@ func (l *FileBasedLogger) logResponse(id string, response *http.Response, body i
...
 	// write id
-	_, err = l.responseLogFile.WriteString(fmt.Sprintf("Request-ID: %s\n", id))
+	_, err = l.respBufWriter.WriteString(fmt.Sprintf("Request-ID: %s\n", id))
 	if err != nil {
 		return err
 	}
 
 	// write dump
-	if _, err = l.responseLogFile.WriteString(fmt.Sprintf("%s", string(dump))); err != nil {
+	if _, err = l.respBufWriter.WriteString(fmt.Sprintf("%s", string(dump))); err != nil {
```
arthurpitman, line 182: nit:
```suggestion
	if _, err = l.respBufWriter.WriteString(string(dump)); err != nil {
```

internal/trafficlogs/logger.go
```diff
@@ -121,13 +130,13 @@ func (l *FileBasedLogger) logRequest(id string, request *http.Request, body io.R
...
 	// write id
-	_, err = l.requestLogFile.WriteString(fmt.Sprintf("Request-ID: %s\n", id))
+	_, err = l.reqBufWriter.WriteString(fmt.Sprintf("Request-ID: %s\n", id))
 	if err != nil {
 		return err
 	}
 
 	// write dump
-	if _, err = l.requestLogFile.WriteString(fmt.Sprintf("%s", string(dump))); err != nil {
+	if _, err = l.reqBufWriter.WriteString(fmt.Sprintf("%s", string(dump))); err != nil {
```
arthurpitman, line 139: nit:
```suggestion
	if _, err = l.reqBufWriter.WriteString(string(dump)); err != nil {
```

arthurpitman (dismissed):

arthurpitman (approved):

PR #1584, opened 2024-09-26
chore: Fix typo in spelling of environment

arthurpitman:
This little PR fixes a typo in the spelling of environment

### DIFF
--- a/pkg/graph/graph.go
+++ b/pkg/graph/graph.go
@@ -74,7 +74,7 @@ type ConfigGraphPerEnvironment map[string]*simple.DirectedGraph
 func (graphs ConfigGraphPerEnvironment) EncodeToDOT(environment string) ([]byte, error) {
 	g, ok := graphs[environment]
 	if !ok {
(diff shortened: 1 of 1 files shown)


warber (approved):

jskelin:
I don't like forcing a function instead of `errors.New(...)` or `fmt.Errorf()`.
With this, it is easy to get the impression that the specialized error instead of basic is returned.

warber (dismissed):
maybe one possible alternative by extracting it into a method

```go
func (graphs ConfigGraphPerEnvironment) getGraph(environment string) (*simple.DirectedGraph, error) {
if g, ok := graphs[environment]; ok {
	return g, nil
}
return nil, fmt.Errorf("no dependency graph exists for envrionment %s", environment)
}
```

and use it like

```go
g, err := graphs.getGraph(environment)
if err != nil {
	return nil, err
}
...
```

jskelin (approved):

PR #1597, opened 2024-10-09
fix: Correctly compare extension versions

arthurpitman:
This PR fixes an issue where `1.9` was considered greater than `1.10` when comparing extension versions.
Rather than using a string compare, the existing `Version` struct from the `version` package  is used.
This was, however, generalized to accept `1`, i.e. just a major version, as a valid version, due to existing test cases.

### DIFF
--- a/internal/version/version_utils.go
+++ b/internal/version/version_utils.go
@@ -76,7 +76,7 @@ func (v Version) Invalid() bool {
 // according to the expected format: "MAJOR.MINOR.PATCH" or "MAJOR.MINOR" each component being a non-negative number
 func ParseVersion(versionString string) (Version, error) {
 	split := strings.Split(versionString, ".")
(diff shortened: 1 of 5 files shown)


internal/version/version_utils.go
```diff
@@ -76,17 +76,21 @@ func (v Version) Invalid() bool {
 // according to the expected format: "MAJOR.MINOR.PATCH" or "MAJOR.MINOR" each component being a non-negative number
 func ParseVersion(versionString string) (Version, error) {
 	split := strings.Split(versionString, ".")
-	if !(len(split) == 2 || len(split) == 3) {
+	if !(len(split) == 1 || len(split) == 2 || len(split) == 3) {
```
warber, line 79: nit: easier to read

```suggestion
	if len(split) < 1 || len(split) > 3 {
```

warber (approved):

PR #1598, opened 2024-10-09
chore: cache templates on project loading/validation

warber:
**What this PR does / Why we need it:**
During project loading, we verify the presence of template files before loading them from disk. Many projects reuse parameterized templates **a lot**, leading to repeated creation of the same template again and again, including validation to ensure that the file exists on disk which results in a massive amount of unnecessary syscalls for no reason. This PR introduces a cache to store already verified templates.

**Special notes for your reviewer:**
A next step would be to actually only load the template's content once when template::Content() is invoked. However, this is happening later and would probably change memory consumption significantly. This needs to be investeigated further.

Does this PR introduce a user-facing change?

### DIFF
--- a/cmd/monaco/convert/convert_test.go
+++ b/cmd/monaco/convert/convert_test.go
@@ -20,10 +20,12 @@ package convert
 
 import (
 	"fmt"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/cache"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/json"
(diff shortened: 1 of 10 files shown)


Laubi:
Afaik, we introduced reading templates when needed to reduce the memory footprint. We used to store the content of the JSON file directly in the config, and this had a huge memory impact. 

How would this PR change the memory usage when downloading and deploying an environment with tens of thousands of configurations? Bc even when we use a cache, when downloading and reuploading an environment, there won't be any duplicates of the JSON-files.

warber:
Realised that the changes in the PR raises unnecessary questions, given that it gives just a little performance gains in some situtations. So we've decided to close this.

PR #1614, opened 2024-10-24
fix(dependency_resolution): Dashboard-Share-Settings can no longer be referenced from dashboards.

Laubi:
Dashboard share settings are a bit special. One dashboard always has exactly 1 dashboard-share config connected to it. This share setting has the same ID as the dashboard itself. This is where the issues arise.

Dashboards can't depend on other dashboards, as we ignore those references. The references are usually only included in the markdown of the tiles, not the dashboards themselves. Since the dashboard share settings have the same ID as the dashboards, what means that instead of ignoring `dashboard->dashboard` references, we get unwanted `dashboard->dashboard-share-setting` references. What is more, the references are not within the same config, but some random dashboards can suddenly depend on the dashboard-share-settings.
Since dashboard-share settings do depend on their parent dashboard, it can happen that we suddenly introduce cyclic dependencies where there are actually none.

This fix ignores `dashboard-share-settings` completely as nothing can actually reference them. They just exist as the children of `dashboards`.

Changeset contains a small refactoring in extracting common code to `shared.go`

Yes, users will have a more stable experiencing when downloading/uploading dynatrace environments

### DIFF
--- a/pkg/download/dependency_resolution/resolver/ahocorasick_dep_resolver.go
+++ b/pkg/download/dependency_resolution/resolver/ahocorasick_dep_resolver.go
@@ -18,11 +18,13 @@ package resolver
 
 import (
 	"fmt"
+
 	goaho "github.com/anknown/ahocorasick"
+	"golang.org/x/exp/maps"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter/reference"
-	"golang.org/x/exp/maps"
 )
 
 // dependencyResolutionContext holds the important information for dependency resolution.
@@ -103,12 +105,8 @@ func findAndReplaceIDs(apiName string, configToBeUpdated config.Config, c depend
 			panic(fmt.Sprintf("No config found for given key %q", key))
 		}
 
-		if configToBeUpdated.Coordinate.Type == "dashboard" && conf.Coordinate.Type == "dashboard" {
-			continue // dashboards can not actually reference each other, but often contain a link to another inside a markdown tile
-		}
-
-		if conf.Coordinate == configToBeUpdated.Coordinate {
-			continue // skip self referencing configs
+		if !canReference(configToBeUpdated, conf) {
+			continue
 		}
 
 		log.Debug("\treference: '%v/%v' referencing '%v' in coordinate '%v' ", apiName, configToBeUpdated.Template.ID(), key, conf.Coordinate)
--- a/pkg/download/dependency_resolution/resolver/basic_dep_resolver.go
+++ b/pkg/download/dependency_resolution/resolver/basic_dep_resolver.go
@@ -17,11 +17,13 @@
 package resolver
 
 import (
+	"strings"
+
+	"golang.org/x/exp/maps"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter/reference"
-	"golang.org/x/exp/maps"
-	"strings"
 )
 
 type basicResolver struct {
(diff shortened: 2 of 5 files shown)


pkg/download/dependency_resolution/resolver/basic_dep_resolver.go
```diff
@@ -75,9 +77,9 @@ func basicFindAndReplaceIDs(apiName string, configToBeUpdated config.Config, con
...
 // cases there is no real valid reference even if the key is found in the content.
 func shouldReplaceReference(configToBeUpdated config.Config, configToUpdateFrom config.Config, contentToBeUpdated, keyToReplace string) bool {
-	if configToBeUpdated.Coordinate.Type == "dashboard" && configToUpdateFrom.Coordinate.Type == "dashboard" {
-		return false //dashboards can not actually reference each other, but often contain a link to another inside a markdown tile
+	if !canReference(configToBeUpdated, configToUpdateFrom) {
+		return false
 	}
 
-	return configToUpdateFrom.Template.ID() != configToBeUpdated.Template.ID() && strings.Contains(contentToBeUpdated, keyToReplace)
+	return strings.Contains(contentToBeUpdated, keyToReplace)
```
Laubi, line 84: To any reviewer: Does someone know why we compared the template-id here? It should be the coordinate like in the other resolver, right? 

@UnseenWizzard could you maybe also have a look?

pkg/download/dependency_resolution/resolver/ahocorasick_dep_resolver.go
```diff
@@ -103,12 +105,8 @@ func findAndReplaceIDs(apiName string, configToBeUpdated config.Config, c depend
 			panic(fmt.Sprintf("No config found for given key %q", key))
 		}
 
-		if configToBeUpdated.Coordinate.Type == "dashboard" && conf.Coordinate.Type == "dashboard" {
-			continue // dashboards can not actually reference each other, but often contain a link to another inside a markdown tile
-		}
-
-		if conf.Coordinate == configToBeUpdated.Coordinate {
-			continue // skip self referencing configs
+		if !canReference(configToBeUpdated, conf) {
```
arthurpitman, line 108: if it can not reference, would a debug log message be a good idea here?

pkg/download/dependency_resolution/resolver/basic_dep_resolver.go
arthurpitman: Is there a case where two configs are downloaded with the same template file? OR is this a crude way to ensure that references are not created to a config itself?

arthurpitman: Would it be valid to have both checks?
- A config should not reference itself, but
- Two different configs with the same template (can that occur?) cannot reference each other.

pkg/download/dependency_resolution/resolver/ahocorasick_dep_resolver.go
Laubi: While I agree that it could have some value, I don't necessarily think it would be a good idea. 
The basic resolver would print thousands on thousands of messages, as we're basically brute-forcing the dependencies. Each config will be compared with *all* configs, so 100 configs might produce 100 * 100 messages if they don't reference each other. 
For the faster algo, it will probably also provide a lot of spam, even though it might be improved.

pkg/download/dependency_resolution/resolver/basic_dep_resolver.go
Laubi: The issue I have is that the aho-algo does compare the coordinates; only the basic did not do that. Since all tests work fine without adoption, I think we're fine. I just want to double-check it.

Just triple-checked it, the Template-ID is usually initialized using the remove config-id of the config. Since this is part of the coordinate, it should be absolutely fine. We should have it consistent though.

pkg/download/dependency_resolution/resolver/shared.go
```diff
@@ -18,14 +18,17 @@ package resolver
...
 	valueParam "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter/value"
-	"regexp"
-	"strings"
 )
 
+// resolveScope updates the `scope` parameter of the config and converts it to a reference parameter iff the scope
+// is a known id of another downloaded config.
 func resolveScope(configToBeUpdated *config.Config, ids map[string]config.Config) {
 	if configToBeUpdated.Type.ID() != config.SettingsTypeId {
 		return
```
arthurpitman, line 34: Should `resolveScope` also continue for `dashboard-share-settings` configs?

arthurpitman (approved):

PR #1616, opened 2024-10-29
fix: Ensure environment is not accessed during a dry run

arthurpitman:
This PR ensures that no requests are sent to a Dynatrace environment during a dry run by switching to a no-op client early during deployment.

Furthermore, this PR tries to move towards the goal of only having one sort of "clientset" in Monaco, by removing those in from the delete and deploy packages. It also finishes splitting the old DTClient into a classic client and a settings client.

### DIFF
--- a/cmd/monaco/integrationtest/v2/dry-run_test.go
+++ b/cmd/monaco/integrationtest/v2/dry-run_test.go
@@ -46,6 +46,7 @@ func TestDryRun(t *testing.T) {
 		}))
 		defer server.Close()
 
+		// ensure all URLs used in the manifest point at the test server
 		t.Setenv("URL_ENVIRONMENT_1", server.URL)
 		t.Setenv("PLATFORM_URL_ENVIRONMENT_2", server.URL)
 		t.Setenv("OAUTH_TOKEN_ENDPOINT", server.URL)
--- a/cmd/monaco/integrationtest/v2/dry-run_test.go
+++ b/cmd/monaco/integrationtest/v2/dry-run_test.go
@@ -47,9 +47,7 @@ func TestDryRun(t *testing.T) {
 		defer server.Close()
 
 		// ensure all URLs used in the manifest point at the test server
-		t.Setenv("URL_ENVIRONMENT_1", server.URL)
-		t.Setenv("PLATFORM_URL_ENVIRONMENT_2", server.URL)
-		t.Setenv("OAUTH_TOKEN_ENDPOINT", server.URL)
+		setAllURLEnvironmentVariables(t, server.URL)
 
 		// This causes a POST for all configs:
 		err := monaco.RunWithFSf(fs, "monaco deploy %s --environment=%s --verbose --dry-run", manifest, specificEnvironment)
@@ -59,4 +57,12 @@ func TestDryRun(t *testing.T) {
 		err = monaco.RunWithFSf(fs, "monaco deploy %s --environment=%s --verbose --dry-run", manifest, specificEnvironment)
 		assert.NoError(t, err)
 	})
+}
+
+func setAllURLEnvironmentVariables(t *testing.T, url string) {
+	t.Setenv("URL_ENVIRONMENT_1", url)
+	t.Setenv("URL_ENVIRONMENT_2", url)
+	t.Setenv("PLATFORM_URL_ENVIRONMENT_1", url)
+	t.Setenv("PLATFORM_URL_ENVIRONMENT_2", url)
+	t.Setenv("OAUTH_TOKEN_ENDPOINT", url)
 }
--- a/cmd/monaco/integrationtest/v2/dry-run_test.go
+++ b/cmd/monaco/integrationtest/v2/dry-run_test.go
@@ -19,12 +19,15 @@
 package v2
 
 import (
+	"net/http"
+	"net/http/httptest"
 	"testing"
+
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/integrationtest/utils/monaco"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
-	"github.com/spf13/afero"
-	"github.com/stretchr/testify/assert"
 )
 
 func TestDryRun(t *testing.T) {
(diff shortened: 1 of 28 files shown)


pkg/delete/delete.go
```diff
@@ -38,91 +35,78 @@ import (
...
+	//  Dashboard share settings cannot be deleted
+	if _, ok := remainingEntriesToDelete[api.DashboardShareSettings]; ok {
+		log.Warn("Classic config of type %s cannot be deleted. Note, that they can be removed by deleting the associated dashboard.", api.DashboardShareSettings)
+		delete(remainingEntriesToDelete, api.DashboardShareSettings)
+
+	}
+
+	// Delete rest of config types
+	for t, entries := range remainingEntriesToDelete {
+		if err := deleteOtherConfigs(ctx, clients, t, entries); err != nil {
```
jskelin, line 56: Is `deleteConfig` sounds better than `deleteOthereConfigs`?
`deleteAutmoationConfig` is a special case, but this is standard behavior.

jskelin (approved):

jskelin (dismissed):

cmd/monaco/integrationtest/v2/dry-run_test.go
```diff
@@ -38,6 +41,16 @@ func TestDryRun(t *testing.T) {
 	}
 
 	RunIntegrationWithCleanupGivenEnvs(t, configFolder, manifest, specificEnvironment, "AllConfigs", envVars, func(fs afero.Fs, _ TestContext) {
+		server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+			t.Fatalf("unexpected HTTP request made during dry run: %s", req.RequestURI)
+		}))
+		defer server.Close()
+
+		// ensure all URLs used in the manifest point at the test server
+		t.Setenv("URL_ENVIRONMENT_1", server.URL)
```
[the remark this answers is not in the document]
tomazpu: It's just a nit, but maybe we could pack this even in a helper function of the tests.

pkg/client/clientset.go
```diff
@@ -185,8 +174,12 @@ var DefaultRetryOptions = corerest.RetryOptions{MaxRetries: 10, ShouldRetryFunc:
 // Each field may be nil, if the ClientSet is partially initialized - e.g. no autClient will be part of a ClientSet
 // created for a 'classic' Dynatrace environment, as Automations are a Platform feature
 type ClientSet struct {
-	// dtClient is the client capable of updating or creating settings and classic configs
-	DTClient DynatraceClient
+	// ClassicClient is the client capable of updating or creating classic configs
+	ClassicClient ConfigClient
+
```
tomazpu, line 179: There is no need for a new line, the same after `SettingsClient`

pkg/delete/delete.go
```diff
@@ -38,91 +35,79 @@ import (
...
 	automationTypeOrder := []config.AutomationResource{config.Workflow, config.SchedulingRule, config.BusinessCalendar}
 	for _, key := range automationTypeOrder {
-		entries := copiedDeleteEntries[string(key)]
-		if clients.Automation == (*coreAutomation.Client)(nil) {
+		entries := allEntries[string(key)]
+		delete(remainingDeleteEntries, string(key))
+		if autClient == nil {
 			log.WithCtxFields(ctx).WithFields(field.Type(key)).Warn("Skipped deletion of %d Automation configuration(s) of type %q as API client was unavailable.", len(entries), key)
-			delete(copiedDeleteEntries, string(key))
 			continue
```
tomazpu, line 77: I think this should be at least a return. Or we check for the null client in the Configs method.

Laubi: I had a closer look at the code, as I agreed with this comment. However, I think it's fine like this (for now). The reason is, that we report each config & component that it can't be deployed. A refactoring here that gives us the same log message information would be way more work and we should not do that right now.

cmd/monaco/integrationtest/v2/dry-run_test.go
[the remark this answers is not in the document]
arthurpitman: Good ideas, I have moved this into an unexported function, if we need it more often let's turn into a shared helper function.

pkg/delete/delete.go
```diff
@@ -38,91 +35,79 @@ import (
...
 // DeleteEntries is a map of configuration type to slice of delete pointers
 type DeleteEntries = map[configurationType][]pointer.DeletePointer
 
 // Configs removes all given entriesToDelete from the Dynatrace environment the given client connects to
-func Configs(ctx context.Context, clients ClientSet, _ api.APIs, automationResources map[string]config.AutomationResource, entriesToDelete DeleteEntries) error {
-	copiedDeleteEntries := make(DeleteEntries)
-	for k, v := range entriesToDelete {
-		copiedDeleteEntries[k] = v
+func Configs(ctx context.Context, clients client.ClientSet, entriesToDelete DeleteEntries) error {
+	remainingEntriesToDelete, errCount := deleteAutomationConfigs(ctx, clients.AutClient, entriesToDelete)
```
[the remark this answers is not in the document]
arthurpitman: We could but the automation configs need to be deleted in a certain order, so it seems simple to remove them as they are filtered out.

pkg/delete/delete.go
arthurpitman: Let's revisit this when we fix logs and hopefully simplify the whole deletion code.

cmd/monaco/purge/purge.go
```diff
@@ -73,33 +74,17 @@ func purgeConfigs(environments []manifest.EnvironmentDefinition, apis api.APIs)
 }
 
 func purgeForEnvironment(env manifest.EnvironmentDefinition, apis api.APIs) error {
+	ctx := context.WithValue(context.TODO(), log.CtxKeyEnv{}, log.CtxValEnv{Name: env.Name, Group: env.Group})
 
-	deleteClients, err := getClientSet(env)
+	clients, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{SupportArchive: support.SupportArchive})
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to create a client for env `%s` due to the following error: %w", env.Name, err)
```
arthurpitman, line 81:
```suggestion
		return fmt.Errorf("failed to create a client for env `%s`: %w", env.Name, err)
```

pkg/client/clientset.go
```diff
@@ -255,13 +248,13 @@ func validateURL(dtURL string) error {
 	return nil
 }
 
-func CreateClientSet(url string, auth manifest.Auth, opts ClientOptions) (*ClientSet, error) {
+func CreateClientSet(ctx context.Context, url string, auth manifest.Auth, opts ClientOptions) (*ClientSet, error) {
 	var (
 		classicClient, client *corerest.Client
```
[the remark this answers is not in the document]
arthurpitman: No, it is OK for the current design because either they are set to non-nil values or this function errors out:

https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/bad912d17a3c594ee78b5196194aab774b4e79ce/pkg/client/clientset.go#L325-L337

We could certainly improve the clientset creation though

tomazpu (approved):

PR #1619, opened 2024-11-04
fix: Skip caching for configs that are set to skip

tomazpu:
…skip parameters set to true

Skip caching for configs of type classic and setting that have the skip parameter set to true. As the configs won't be deployed there is no need to chache them. With this being skipped, we also prevent performance degradation compared to the previous version.


jskelin (approved):

jskelin:
Is it possible to add a test (e.g. in `Test_gatherPreloadConfigTypeEntries_OneEntryPerConfigType`)?

PR #1621, opened 2024-11-06
chore: Update core library with changes in account management API

arthurpitman:


pkg/account/deployer/client_test.go
```diff
@@ -20,32 +20,29 @@ package deployer
...
-   "items": [
-	 {
 	   "uid": "3288032b-9bdc-4480-bb11-2ec0ad2610b2",
 	   "email": "abcd@ef.com",
 	   "emergencyContact": false,
 	   "userStatus": "PENDING",
-	   "type": "DEFAULT"
-	 }
-   ]
+	   "groups": []
```
tomazpu, line 45: The `groups` payload can't influence any test outcome?

pkg/account/downloader/internal/http/client.go
```diff
@@ -53,7 +54,7 @@ func (c *Client) GetGroupsForUser(ctx context.Context, userEmail string, account
 }
 
 func (c *Client) GetPolicies(ctx context.Context, account string) ([]accountmanagement.PolicyOverview, error) {
-	r, resp, err := c.PolicyManagementAPI.GetPolicyOverviewList(ctx, "account", account).Execute()
+	r, resp, err := c.PolicyManagementAPI.GetPolicyOverviewList(ctx, account, "account").Execute()
```
tomazpu, line 57: Can we move "account" into a constant?

pkg/account/deployer/client_test.go
arthurpitman: `groups`, even if empty is required as part of the response payload.

pkg/account/downloader/internal/http/client.go
arthurpitman: We could, along also e.g. with `"environment"`, but we must decide where these should go. I am also not convinced of the benefit of maintaining separate clients for download and deployment. Ideally we should follow this PR with one consisting of improvements, I think @jskelin had something in the works.

jskelin (approved):

jskelin (dismissed):

tomazpu (dismissed):

tomazpu (approved):

PR #1622, opened 2024-11-07
Consider Skip parameter during preload caching

tomazpu:
Skip caching for configs of type classic and setting that have the skip parameter set to true. As the configs won't be deployed there is no need to chache them. With this being skipped, we also prevent performance degradation compared to the previous version.

### DIFF
--- a/pkg/deploy/preload_test.go
+++ b/pkg/deploy/preload_test.go
@@ -359,3 +359,62 @@ func Test_ScopedConfigsAreNotCached(t *testing.T) {
 		})
 	}
 }
+
+func Test_gatherPreloadConfigTypeEntries_WithSkipParam(t *testing.T) {
+	entries := gatherPreloadConfigTypeEntries(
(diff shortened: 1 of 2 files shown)


pkg/deploy/preload_test.go
```diff
@@ -359,3 +359,50 @@ func Test_ScopedConfigsAreNotCached(t *testing.T) {
 		})
 	}
 }
+
+func Test_gatherPreloadConfigTypeEntries_WithSkipParam(t *testing.T) {
```
jskelin, line 363: What is your opinion if we divide this test into 3 subtests:
 * skip = true
 * skip = false
 * skip is not defined (aka default value)
 
Is it going to be more readable?

jskelin (approved):

Laubi (dismissed):

Laubi (approved):

Laubi (approved):

PR #1623, opened 2024-11-07
chore: Split DTClient into Classic and Settings client

arthurpitman:
This PR takes the first step and splits DTClient into a Classic client and a settings client. This is needed to ensure that a settings client can be created for both classic and platform environments, while classic client can only be created for classic environments.

This PR makes ~~no~~ _a small_ attempt to improve naming and code organization, but more work is necessary,  for example, a follow up PR should move both clients into separate packages.


jskelin (dismissed):

tomazpu (approved):
I agree with what we discussed in the daily, the naming is not optimal so i did not write any comments on renaming the dtClient throw the code. 

In my opinion it's a good step, but follow-ups of maybe moving the clients to a builder pattern so the creation looks better on the client code(clientset) and into separate package.

pkg/deploy/deploy_test.go
```diff
@@ -95,8 +95,8 @@ func TestDeployConfigGraph_SingleConfig(t *testing.T) {
 		},
 	}
 
-	dummyClient := dtclient.DummyClient{}
-	clientSet := &client.ClientSet{ClassicClient: &dummyClient, SettingsClient: &dummyClient}
+	dummyClient := dtclient.DummyConfigClient{}
+	clientSet := &client.ClientSet{ConfigClient: &dummyClient, SettingsClient: &dtclient.DummySettingsClient{}}
```
tomazpu, line 99: Does it make sense to pass both parameters in the same way, or is the `dummyClient` needed in a variable as it's used later on?

arthurpitman: dummyClient is being used later, as the classic dummy client records entries when requests are made. But this is something that should be re-examined and made uniform for all dummy clients.

jskelin (approved):

PR #1633, opened 2024-11-21
Update settings api to use `ResourceContext` over `ModificationInfo`

tomazpu:
Update settings API to use `ResourceContext` over `ModificationInfo`, `ModificationInfo` is deprecated by the API and only used as a fallback


pkg/client/dtclient/settings_client.go
```diff
@@ -60,6 +117,13 @@ type SettingsModificationInfo struct {
 	NonModifiablePaths []interface{} `json:"nonModifiablePaths"`
 }
 
+type SettingsResourceContext struct {
+	Operations         []string      `json:"operations"`
+	Movable            *bool         `json:"modifications:movable"`
```
tomazpu, line 122: To the reviewer. I chose a pointer over its own `struct` type, the problem here is that the param `Movable` is optional, and logic code in method `isMovable()` depends on it. 

If preferred, I could change it to a `NullBool` type and implement the unmarshaling myself. Since is not done anywhere else in the code i chose `*bool` over `NullBool`.

pkg/client/dtclient/settings_client.go
```diff
@@ -41,6 +42,13 @@ import (
...
 
+// SettingsResourceContext.Operations possibilities
+const (
+	DeleteOperation = "delete"
+	WriteOperation  = "write"
+	//ReadOperation   = "read"
+)
+
 // DownloadSettingsObject is the response type for the ListSettings operation
 type DownloadSettingsObject struct {
```
jskelin, line 53: Why not move to a separate file

pkg/client/dtclient/settings_client.go
jskelin: By now, I wasn't aware of the NullBool type in GO. But as I see it, it is part of the SQL library and for that reason wouldn't it even consider. For me pointer is just enough

jskelin (approved):

pkg/client/dtclient/settings_client.go
tomazpu: I would suggest doing it in a separate ticket when we rename `dtclinet` folder to `settings` as the call in code would also be `settings.DownloadSettingsObject`. If you have a strong opinion on this I can do it now, but it will be a much bigger PR then.

pkg/client/dtclient/settings_client.go
```diff
@@ -50,6 +58,55 @@ type DownloadSettingsObject struct {
 	Scope            string                    `json:"scope"`
 	Value            json.RawMessage           `json:"value"`
 	ModificationInfo *SettingsModificationInfo `json:"modificationInfo"`
```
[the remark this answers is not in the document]
tomazpu: I need to say my personal opinion is as little comments in code as possible as they a distraction so i would like to agree on amount of comments maybe in a future TechTime :)

pkg/client/dtclient/settings_client.go
```diff
@@ -60,6 +117,13 @@ type SettingsModificationInfo struct {
 	NonModifiablePaths []interface{} `json:"nonModifiablePaths"`
 }
 
+type SettingsResourceContext struct {
+	Operations         []string      `json:"operations"`
+	Movable            *bool         `json:"modifications:movable"`
+	ModifiablePaths    []interface{} `json:"modifications:modifiablePaths"`
```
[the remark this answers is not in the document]
tomazpu: Good point, did it how it was in `ModificationInfo` :)

arthurpitman (approved):

jskelin (approved):

Laubi (approved):

PR #1634, opened 2024-11-21
fix: Support deeper unique settings properties

arthurpitman:
With the current matching algorithm, we can only support non-nested constraints. With this PR, recursive matching is introduced, and so we can apply even nested constraints.

### DIFF
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -632,37 +632,33 @@ func isSameValueForKey(targetPath string, c1 []byte, c2 []byte) (same bool, matc
 	}
 
 	keys := explodePath(targetPath)
-	value1, err := recursiveSearch(u, keys)
-	if err != nil {
-		return false, nil, err
-	}
+	value1 := recursiveSearch(u, keys)
+
 	u = nil
 	if err := json.Unmarshal(c2, &u); err != nil {
 		return false, nil, fmt.Errorf("failed to unmarshal data for key %q: %w", targetPath, err)
 	}
 
-	value2, err := recursiveSearch(u, keys)
-	if err != nil {
-		return false, nil, err
-	}
+	value2 := recursiveSearch(u, keys)
 
-	if cmp.Equal(value1, value2) {
+	if value1 != nil && value2 != nil && cmp.Equal(value1, value2) {
 		return true, value1, nil
 	}
 	return false, nil, nil
 }
 
-func recursiveSearch(nestedMap map[string]any, keys []string) (any, error) {
+// Recursive search allows for nil values in case a field is not in the payload
+func recursiveSearch(nestedMap map[string]any, keys []string) any {
 	currentMap := nestedMap
 	value, found := currentMap[keys[0]]
 	if found {
 		if nestedMap, ok := value.(map[string]interface{}); ok && len(keys) > 1 {
 			return recursiveSearch(nestedMap, keys[1:])
 		}
-		return value, nil
+		return value
 	}
 
-	return nil, fmt.Errorf("key: %q not found", keys[0])
+	return nil
 }
 
 // explodePath splits targetPath by "/", this is the format of settings api.
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -625,22 +625,50 @@ func doObjectsMatchBasedOnUniqueKeys(uniqueKeys []string, source SettingsObject,
 	return true, matches, nil
 }
 
-func isSameValueForKey(key string, c1 []byte, c2 []byte) (same bool, matchingVal any, err error) {
+func isSameValueForKey(targetPath string, c1 []byte, c2 []byte) (same bool, matchingVal any, err error) {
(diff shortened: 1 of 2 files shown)


pkg/client/dtclient/settings_client_test.go
```diff
@@ -517,6 +556,22 @@ func Test_findObjectWithSameConstraints(t *testing.T) {
 					},
 				},
 			},
+			{
```
jskelin, line 559: I would suggest adding the next error cases to document that for matching the full path needs to be same:

```
					schema: Schema{
						UniqueProperties: [][]string{
							{"B"},
						},
					},
					source: SettingsObject{
						SchemaId: "schemaID", Content: []byte(`{"A" : {"B" : "x" } }`),
					},
					objects: []DownloadSettingsObject{
						{Value: []byte(`{"A" : {"B" : "x" } }`)},
					},
``` 

```
					schema: Schema{
						UniqueProperties: [][]string{
							{"B"},
						},
					},
					source: SettingsObject{
						SchemaId: "schemaID", Content: []byte(`{"B" : "x" } `),
					},
					objects: []DownloadSettingsObject{
						{Value: []byte(`{"A" : {"B" : "x" } }`)},
					},
```

tomazpu: thanks, good suggestion :)

pkg/client/dtclient/settings_client_test.go
```diff
@@ -477,6 +477,87 @@ func Test_findObjectWithSameConstraints(t *testing.T) {
...
+				expected: nil,
+			},
+			{
+				name: "objects with no constraint fields in payload - no match",
+				given: given{
+					schema: Schema{
+						UniqueProperties: [][]string{
+							{"A"},
+							{"B"},
+							{"A/B"},
```
jskelin, line 526: What will happen if under UniqueProperties `{"R"}` is added?

pkg/client/dtclient/settings_client.go
```diff
@@ -563,24 +563,48 @@ func doObjectsMatchBasedOnUniqueKeys(uniqueKeys []string, source SettingsObject,
...
 	if err := json.Unmarshal(c1, &u); err != nil {
-		return false, nil, fmt.Errorf("failed to unmarshal data for key %q: %w", key, err)
+		return false, nil, fmt.Errorf("failed to unmarshal data for key %q: %w", targetPath, err)
 	}
-	v1 := u[key]
 
+	keys := explodePath(targetPath)
+	value1 := recursiveSearch(u, keys)
+
+	u = nil
```
jskelin, line 575: I would use a different variable... but that is just a minor thing

jskelin (approved):

pkg/client/dtclient/settings_client.go
```diff
@@ -563,24 +563,48 @@ func doObjectsMatchBasedOnUniqueKeys(uniqueKeys []string, source SettingsObject,
 	return true, matches, nil
 }
 
-func isSameValueForKey(key string, c1 []byte, c2 []byte) (same bool, matchingVal any, err error) {
+func isSameValueForKey(targetPath string, c1 []byte, c2 []byte) (same bool, matchingVal any, err error) {
 	u := make(map[string]any)
```
Laubi, line 567: Could you please call this.... something else? :)
`UnmarshaledConfig`? Better names are welcome

pkg/client/dtclient/settings_client.go
Laubi, line 575: Instead of reusing `u`, I'd go with a new var. Setting it to `null` is pointless, as an error is returned anyway if it can't be unmarshaled.

pkg/client/dtclient/settings_client.go
```diff
@@ -563,24 +563,48 @@ func doObjectsMatchBasedOnUniqueKeys(uniqueKeys []string, source SettingsObject,
...
-		return false, nil, fmt.Errorf("failed to unmarshal data for key %q: %w", key, err)
+		return false, nil, fmt.Errorf("failed to unmarshal data for key %q: %w", targetPath, err)
 	}
-	v2 := u[key]
 
-	if cmp.Equal(v1, v2) {
-		return true, v1, nil
+	value2 := recursiveSearch(u, keys)
+
+	if value1 != nil && value2 != nil && cmp.Equal(value1, value2) {
```
Laubi, line 582: Shouldn't it just be `cmp.Equal(v1, v2)` here?
If both keys are `null`, it should return true as well, shouldn't it?

tomazpu: This is the case where we want to prevent a nil matching a nil. The empty value of a constrain should not match.

Laubi: True. Could you pls add a comment 🥺, because I won't know this detail in 2 weeks? :D

pkg/client/dtclient/settings_client_test.go
jskelin: I would appreciate it if you put it in a test. Recursion is always tricky.

Laubi (approved):

PR #1636, opened 2024-11-22
Allow Settings API with OAuth credentials

tomazpu:
This PR removes the validation check for Settings API in download and deploy to fail if Token is not set. This allows to use OAuth credentials with Settings API.

### DIFF
--- a/cmd/monaco/deploy/deploy.go
+++ b/cmd/monaco/deploy/deploy.go
@@ -292,6 +292,10 @@ func validateAuthenticationWithProjectConfigs(projects []project.Project, enviro
 						if environments[envName].Auth.Token == nil {
 							return fmt.Errorf("API of type '%s' requires a token for environment '%s'", conf.Type, envName)
 						}
+					case config.SettingsType:
+						if environments[envName].Auth.Token == nil && environments[envName].Auth.OAuth == nil {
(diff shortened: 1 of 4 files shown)


cmd/monaco/deploy/deploy.go
```diff
@@ -288,8 +288,7 @@ func validateAuthenticationWithProjectConfigs(projects []project.Project, enviro
 					}
 
 					switch conf.Type.(type) {
-					case config.ClassicApiType,
-						config.SettingsType:
```
arthurpitman, line 292: The problem is the default case also doesn't allow settings with a token.

arthurpitman: This is causing problems with `cmd/monaco/integrationtest/v2 TestSkip/with_env_override_and_skip_via_env_var`:  https://github.com/Dynatrace/dynatrace-configuration-as-code/actions/runs/11974548692/job/33385877965?pr=1636#step:4:2002

arthurpitman: Settings just need to be accepted for both possibilities, as long as there is OAuth or a token.

tomazpu: Just saw it :), I created a validation case for `SettingsApiType`

arthurpitman (approved):

PR #1638, opened 2024-11-26
feat: Limit max concurrent deployments

arthurpitman:
This PR adds a opt-in limit on concurrent deployments: To enable it, set `MONACO_CONCURRENT_DEPLOYMENTS` to a non-zero value.

The limiter can help prevent excessive resource usage in scenarios where Monaco would otherwise start many deployments.

### DIFF
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -22,11 +22,13 @@ import (
 	"time"
 
 	coreapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
 
 	gonum "gonum.org/v1/gonum/graph"
 	"gonum.org/v1/gonum/graph/simple"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/dynatrace"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/environment"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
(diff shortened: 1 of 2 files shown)


pkg/deploy/deploy.go
```diff
@@ -59,12 +61,19 @@ type DeployConfigsOptions struct {
...
 var (
-	lock sync.Mutex
+	lock                         sync.Mutex
+	concurrentDeploymentsLimiter *rest.ConcurrentRequestLimiter
 
 	skipError = errors.New("skip error")
 )
 
 func Deploy(ctx context.Context, projects []project.Project, environmentClients dynatrace.EnvironmentClients, opts DeployConfigsOptions) error {
+	maxConcurrentDeployments := environment.GetEnvValueIntLog(environment.ConcurrentDeploymentsEnvKey)
```
Laubi, line 71: Should we move this to a `func init` block? It is a global thing, and I don't think the initialization here is the best.

I'd love to have it passed down in the functions to have it per function call, but that would mean more changes :/

arthurpitman: I did try it in an init block, but then the logs were first, which I don't like. (I would like the logs, just to make it easy for us to identify if it is being used / causing problems)

We could pass it around via context, what do you think? Then it wouldnt be global, so could be initialized here in `Deploy(...)`

pkg/deploy/deploy.go
```diff
@@ -244,6 +253,11 @@ func removeChildren(ctx context.Context, parent, root graph.ConfigNode, configGr
 }
 
 func deployConfig(ctx context.Context, c *config.Config, clientset *client.ClientSet, resolvedEntities config.EntityLookup) (entities.ResolvedEntity, error) {
+	if concurrentDeploymentsLimiter != nil {
```
jskelin, line 256: The caller should be responsible for handling that.
Since you limit this on a clientSet level, I think it would be better to put it  higher in the code.

Laubi (approved):

PR #1639, opened 2024-11-26
Chore/simplify project by removing redundancy

jskelin:


pkg/project/v2/project.go
```diff
@@ -48,25 +51,29 @@ type Project struct {
...
-	dependencies, found := p.Dependencies[environment]
-
-	if !found {
-		return false
-	}
-
-	for _, dep := range dependencies {
-		if dep == project.Id {
-			return true
+	for c := range slices.Values(p.ConfigList()) {
```
Laubi, line 68: This method just went from a `O(n)` to an `O(n^4)` (`O(n^2)` from ConfigList) + `O(n^2)` from this method). 
Furthermore, in ConfigList we do collect the configs now *on every* call, just to find the dependencies. This is an intended redundency so that other parts of Monaco work in a timely manner. We took much time to optimize this method (and the whole topology sort) to what it is today to optimize it. 

This refactoring here is not acceptable, and must not be done.

Laubi (changes requested):
This PR is a hard reject from me. Apart from the no-go refactoring of `HasDependencyOn`, the rest of the changes are, in my opinion, not better, but make it more complicated and less performent. 

I do agree, that there are some things we can do better, but we need to talk about them and plan them properly. There are some reasons why we do have some redundency.

PR #1641, opened 2024-12-02
fix(Settings): App settings can now be created

Laubi:
App-settings work a bit different from builtin-settings. When creating app-settings, the originObjectId must not be set, while the 'normal' settings do allow that.

To fix that, the approach has been changed. We assume that we can't find the correct object that requires updating, and then search for it. First, by the unique-constrains of the schema, then by the legacy-external-id, and finally by the external-id and origin-id.

Note that currently, the later methods potentially overwrite the previous methods. This can be fixed in a later refactoring.

Furthermore, this commit introduces multiple new tests that test the behavior and fixes existing ones (plus correct assert calls).

### DIFF
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -361,10 +361,21 @@ func (d *SettingsClient) handleUpsertUnsupportedVersion(ctx context.Context, obj
 
 }
 
+// Upsert creates or updates remote settings objects.
+// The logic to find the correct object to update is as follows:
+//  1. We try to match the unique-constrains of the object
+//  2. We try to find the correct object by checking the legacy-external-id, the external-id, as well as the given originObjectId
+//
+// If we find an object, we update it. If we don't, a new one will be created.
+//
+// Note: If the Dynatrace version of the remote system is <262, nothing will be performed and an error is returned.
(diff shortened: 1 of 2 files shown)


arthurpitman:
> First, by the unique-constrains of the schema, then by the legacy-external-id, and finally by the external-id and origin-id.
> 
> Note that currently, the later methods potentially overwrite the previous methods.

Just to get this straight, that means effectively we first look for an object with origin-id, then external-id, then legacy-external-id, and finally we look for a match using schema constraints.

pkg/client/dtclient/settings_client.go
```diff
@@ -436,7 +449,7 @@ func (d *SettingsClient) Upsert(ctx context.Context, obj SettingsObject, upsertO
 		}
 	}
 
-	payload, err := buildPostRequestPayload(ctx, obj, externalID, upsertOptions.InsertAfter)
+	payload, err := buildPostRequestPayload(ctx, remoteObjectId, obj, externalID, upsertOptions.InsertAfter)
```
arthurpitman, line 452: If it is not possible to have both a remoteObjectId and an externalID here, should we have two separate functions for building post requests? Alternatively we can examine this is the larger refactoring.

arthurpitman (approved):

PR #1646, opened 2024-12-09
chore: remove getters for ClientSet

jskelin:


tomazpu (changes requested):
When searching over the whole project i see that the code still uses `.Settings()`. Please also check for all other getters.

jskelin:
> When searching over the whole project i see that the code still uses `.Settings()`. Please also check for all other getters.
> 
> Edit: checked in a call was a false positive with not so optimal naming of mocks in tests.

There is no other appearance or method with the same name as the deleted getters

tomazpu (approved):

PR #1651, opened 2024-12-13
feat(segments): delete functionality for `segments`

jskelin:
With these changes, delete functionality for segments is supported. To declare `segment` config for delete, in `delete.yaml` file add an entry like this for indirect reference:
```
- type: segment
  project: my-project
  id: monaco-config-id
```
or for direct reference like next:
```
- type: segment
  objectId: origin-object-ID
```


pkg/delete/delete_test.go
```diff
@@ -1149,3 +1150,94 @@ func TestDelete_Documents(t *testing.T) {
 		assert.Error(t, err)
 	})
 }
+
+type stubClient struct {
```
arthurpitman, line 1154:
```suggestion
type segmentsStubClient struct {
```

jskelin: done

arthurpitman (approved):

PR #1652, opened 2024-12-13
refactor: Rename `filersegments` -> `segments`

jskelin:
The name of the feature is `segment`. With this PR, it is aligned with it.

### DIFF
--- a/pkg/persistence/config/writer/config_writer_test.go
+++ b/pkg/persistence/config/writer/config_writer_test.go
@@ -1027,8 +1027,8 @@ func TestWriteConfigs(t *testing.T) {
 			},
 		},
 		{
-			name:    "Grail filter-segment",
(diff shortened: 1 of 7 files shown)


pkg/persistence/config/writer/config_writer_test.go
```diff
@@ -1068,24 +1068,24 @@ func TestWriteConfigs(t *testing.T) {
 			},
 		},
 		{
-			name:    "Grail filter-segment should fail if FF MONACO_FILTER_SEGMENTS is not set",
-			envVars: map[string]string{featureflags.Temporary[featureflags.GrailFilterSegment].EnvName(): "false"},
+			name:    "Segment should fail if FF MONACO_SEGMENTS is not set",
```
Laubi, line 1071:
```suggestion
			name:    "Segment should fail if FF MONACO_FEAT_SEGMENTSis not set",
```

jskelin: done

PR #1655, opened 2024-12-18
fix: `insertAfter` can reference other Monaco projects

Laubi:
We did not take other projects into account when resolving the reference.

To ensure this is not an issue in the future, an e2e test is also added.

### DIFF
--- a/pkg/deploy/internal/setting/validation_test.go
+++ b/pkg/deploy/internal/setting/validation_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"golang.org/x/exp/maps"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/coordinate"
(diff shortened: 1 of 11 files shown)


pkg/deploy/internal/setting/validation_test.go
```diff
@@ -246,15 +247,50 @@ func TestInsertAfterSameScopeValidator(t *testing.T) {
...
+			sourceConfig: config.Config{
+				Coordinate:  coordinate.Coordinate{Project: "project-id", Type: "type-a", ConfigId: "config-x"},
+				Environment: "env",
+				Type:        &config.SettingsType{SchemaId: "type-a"},
+				Parameters: map[string]parameter.Parameter{
+					config.InsertAfterParameter: refParam.New("other-project", "type-a", "config-y", "id"),
+					config.ScopeParameter:       valueParam.New("environment"),
+				},
+			},
+			otherProjectConfigs: []config.Config{
```
jskelin, line 261: Can you just add another config with the same ID and type, but from the 3rd project?

```
				{
					Coordinate:  coordinate.Coordinate{Project: "wrong-project", Type: "type-a", ConfigId: "config-y"},
					Environment: "env",
					Type:        &config.SettingsType{SchemaId: "type-a"},
					Parameters: map[string]parameter.Parameter{
						config.ScopeParameter: valueParam.New("environment"),
					},
				},
			},
```

jskelin (changes requested):

pkg/deploy/internal/setting/validation_test.go
Laubi: Did add the config

jskelin (approved):

tomazpu (approved):

PR #1657, opened 2024-12-20
refactor(FeatureFlag): simplify the usage of FF

jskelin:
Instead of 
```go
	if !featureflags.Permanent[featureflags.SkipVersionCheck].Enabled() {
		go setVersionNotificationStr(&versionNotification)
	}
```
with this refactoring the call looks like:
```go
	if !featureflags.SkipVersionCheck.Enabled() {
		go setVersionNotificationStr(&versionNotification)
	}
```

Additionally, with this reshaping and using methods instead of the map, it is harder to cause panic by targeting non-existing FF.
```go
if featureflags.Permanent["DOESNT_EXISTS"].Enabled() { ... }  // panic

if "DOESNT_EXISTS".Enabled() { ... }  // compile error
```


internal/featureflags/featureflag.go
```diff
@@ -17,23 +17,33 @@
...
 	"os"
 	"strconv"
 	"strings"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+)
+
+type (
+	// FeatureFlagID is the environment variable name
+	FeatureFlagID string
```
jskelin, line 29: I think the best name is just `ID`.
Then on the caller side, it would be `featureFlags.ID`.
```go
	tests := []struct {
		name         string
		given        downloadConfigsOptions
		want         wantDownload
		featureFlags map[featureflags.ID]bool
	}{
```

jskelin:
This [TODO](https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/768d5ac63d0a86204bb18f0ae1a7b6bc193d58fd/pkg/graph/graph.go#L268) already existed in code.

jskelin:
I would also suggest renaming the package from `featureflags` to simply `feature`. With that, calling in the code would look like this:
```
	if feature.SkipVersionCheck.Enabled() {
		go setVersionNotificationStr(&versionNotification)
	}

File: internal/featureflags/temporary.go [hunk 2]
```diff
@@ -16,62 +16,39 @@
...
-	},
-	IgnoreSkippedConfigs: {
-		envName:        string(IgnoreSkippedConfigs),
-		defaultEnabled: false,
-	},
-	Segments: {
-		envName:        string(Segments),
-		defaultEnabled: false,
-	},
+var temporary = map[FeatureFlag]defaultValue{
```
arthurpitman, line 46: Maybe rename like this:

```suggestion
var temporaryDefaultValues = map[FeatureFlag]defaultValue{
```

internal/featureflags/permanent.go
```diff
@@ -16,103 +16,61 @@
...
-	LogToFile: {
-		envName:        "MONACO_LOG_FILE_ENABLED",
-		defaultEnabled: true,
-	},
-	UpdateNonUniqueByNameIfSingleOneExists: {
-		envName:        "MONACO_FEAT_UPDATE_SINGLE_NON_UNIQUE_BY_NAME",
-		defaultEnabled: true,
-	},
+// permanent featureFlagOld - features we want to be able to toggle long-term, instead of removing them after a stabilization period.
+var permanent = map[FeatureFlag]defaultValue{
```
arthurpitman, line 62:
```suggestion
var permanentDefaultValues = map[FeatureFlag]defaultValue{
```

internal/featureflags/temporary.go
jskelin: done

internal/featureflags/permanent.go
jskelin: done

internal/featureflags/featureflag.go
```diff
@@ -17,49 +17,63 @@
...
-
-// EnvName gives back the environment variable name for
-// the feature flag
-func (ff FeatureFlag) EnvName() string {
-	return ff.envName
+	v, f = permanent[ff]
+	if f {
+		return enabled(ff, v)
+	}
+	panic(fmt.Sprintf("unknown feature flag %s", ff))
```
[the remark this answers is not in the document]
jskelin: Changed. Now, it panics in cases when an FF is unknown or defined as temporary as well as permanent (tested manually).

PR #1660, opened 2025-01-02
feat(segments): deploy support for `segments` API

tomazpu:
This PR adds support for the `segments` API in monaco.

Follow-up is planned to move the logic of adding `uid` and `owner` to the lib.

### DIFF
--- a/pkg/client/dummy_clientset.go
+++ b/pkg/client/dummy_clientset.go
@@ -160,22 +160,22 @@ func (c *DummyOpenPipelineClient) Update(_ context.Context, _ string, _ []byte) 
 
 type DummySegmentClient struct{}
 
-func (c *DummySegmentClient) List(ctx context.Context) (segments.Response, error) {
-	return segments.Response{}, fmt.Errorf("unimplemented")
+func (c *DummySegmentClient) List(_ context.Context) (segments.Response, error) {
+	return segments.Response{}, nil
 }
 
-func (c *DummySegmentClient) GetAll(ctx context.Context) ([]segments.Response, error) {
-	return []segments.Response{}, fmt.Errorf("unimplemented")
+func (c *DummySegmentClient) GetAll(_ context.Context) ([]segments.Response, error) {
+	return []segments.Response{}, nil
 }
 
-func (c *DummySegmentClient) Delete(ctx context.Context, id string) (segments.Response, error) {
-	return segments.Response{}, fmt.Errorf("unimplemented")
+func (c *DummySegmentClient) Delete(_ context.Context, _ string) (segments.Response, error) {
+	return segments.Response{}, nil
 }
 
-func (c *DummySegmentClient) Upsert(ctx context.Context, id string, data []byte) (segments.Response, error) {
-	return segments.Response{}, fmt.Errorf("unimplemented")
+func (c *DummySegmentClient) Upsert(_ context.Context, _ string, _ []byte) (segments.Response, error) {
+	return segments.Response{}, nil
 }
 
-func (c *DummySegmentClient) Get(ctx context.Context, id string) (segments.Response, error) {
-	return segments.Response{}, fmt.Errorf("unimplemented")
+func (c *DummySegmentClient) Get(_ context.Context, _ string) (segments.Response, error) {
+	return segments.Response{}, nil
 }
--- a/pkg/client/dummy_clientset.go
+++ b/pkg/client/dummy_clientset.go
@@ -26,19 +26,19 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients/automation"
 	buckets "github.com/dynatrace/dynatrace-configuration-as-code-core/clients/buckets"
 	documents "github.com/dynatrace/dynatrace-configuration-as-code-core/clients/documents"
-	grailfiltersegment "github.com/dynatrace/dynatrace-configuration-as-code-core/clients/grailfiltersegments"
 	openpipeline "github.com/dynatrace/dynatrace-configuration-as-code-core/clients/openpipeline"
+	segments "github.com/dynatrace/dynatrace-configuration-as-code-core/clients/segments"
 	dtclient "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/dtclient"
 )
 
 var DummyClientSet = ClientSet{
-	ConfigClient:             &dtclient.DummyConfigClient{},
-	SettingsClient:           &dtclient.DummySettingsClient{},
-	AutClient:                &DummyAutomationClient{},
-	BucketClient:             &DummyBucketClient{},
-	DocumentClient:           &DummyDocumentClient{},
-	OpenPipelineClient:       &DummyOpenPipelineClient{},
-	GrailFilterSegmentClient: &DummySegmentsClient{},
+	ConfigClient:       &dtclient.DummyConfigClient{},
+	SettingsClient:     &dtclient.DummySettingsClient{},
+	AutClient:          &DummyAutomationClient{},
+	BucketClient:       &DummyBucketClient{},
+	DocumentClient:     &DummyDocumentClient{},
+	OpenPipelineClient: &DummyOpenPipelineClient{},
+	SegmentClient:      &DummySegmentsClient{},
 }
 
 var _ AutomationClient = (*DummyAutomationClient)(nil)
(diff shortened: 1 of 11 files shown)


pkg/client/dummy_clientset.go
```diff
@@ -155,3 +157,25 @@ func (c *DummyOpenPipelineClient) GetAll(ctx context.Context) ([]coreapi.Respons
 func (c *DummyOpenPipelineClient) Update(_ context.Context, _ string, _ []byte) (openpipeline.Response, error) {
 	return openpipeline.Response{}, nil
 }
+
+type DummySegmentClient struct{}
+
+func (c *DummySegmentClient) List(ctx context.Context) (segments.Response, error) {
+	return segments.Response{}, fmt.Errorf("unimplemented")
```
arthurpitman, line 164: I think we need an implementation that just does nothing here, similar to OpenPipeline, as this is also used by dry run.

Currently segment dry run reports: `Deployment failed - Monaco Error: failed to deploy segment with externalId: ..., with error: failed to fetch segment object: unimplemented`

arthurpitman: This probably applies to all the methods in the dummy client.

Laubi: Maybe we should rename it at some point to dry-run-client

pkg/deploy/internal/segment/segment.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+type jsonResponse struct {
+	UID        string `json:"uid"`
+	Owner      string `json:"owner"`
+	ExternalId string `json:"externalId"`
+}
+
+func Deploy(ctx context.Context, client DeploySegmentClient, properties parameter.Properties, renderedConfig string, c *config.Config) (entities.ResolvedEntity, error) {
+	var request map[string]any
```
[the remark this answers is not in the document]
Laubi: If yes, please change it also in the parameters of the called functions

pkg/deploy/internal/segment/segment.go
```diff
@@ -0,0 +1,192 @@
...
+		return api.Response{}, deployErrors.NewConfigDeployErr(c, fmt.Sprintf("failed to upsert segment with id %q", id)).WithError(err)
+	}
+
+	return responseUpsert, nil
+}
+
+func createResolveEntity(id string, externalId string, properties parameter.Properties, c *config.Config) entities.ResolvedEntity {
+	properties[config.IdParameter] = id
+	return entities.ResolvedEntity{
+		EntityName: externalId,
```
Laubi, line 178: Note to self: I think we can get rid of this `EntityName` in the future. It has only one read access, and that's in a test. Nothing to do here for now

pkg/deploy/internal/segment/segment_test.go
```diff
@@ -0,0 +1,348 @@
...
+						"name":        "test-segment-post-match",
+						"description": "post - update from monaco - change - 2",
+						"isPublic":    false,
+						"owner":       "79a4c92e-379b-4cd7-96a3-78a601b6a69b",
+						"externalId":  "project_segment:segment:some-id",
+					}),
+				}, nil
+			},
+			getAllStub: func() ([]segments.Response, error) {
+				panic("should not be called")
```
Laubi, line 98: Any chance we can inject `testing.T` here and call `t.Fatal()`? I see that it's not trivial, but I'd prefer it to panicing 🤔

pkg/deploy/deploy_test.go
```diff
@@ -1298,3 +1299,56 @@ func TestDeployConfigGraph_CollectsAllErrors(t *testing.T) {
...
+			},
+			featureFlag: featureflags.Segments.EnvName(),
+			configType:  config.SegmentID,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name+" | FF Enabled", func(t *testing.T) {
+			t.Setenv(tt.featureFlag, "true")
+			err := deploy.Deploy(context.TODO(), tt.projects, c, deploy.DeployConfigsOptions{})
```
Laubi, line 1344: minor nit, because I don't find too much: `context.Background()` would be the correct one to use ;)
https://pkg.go.dev/context#Background

However, starting with go1.24, it will be anyway `testing.Context`, so no need to worry here :)

Laubi (changes requested):
Please double check your go-import-sort settings :)

pkg/client/dummy_clientset.go
tomazpu: @arthurpitman thanks i will split it up between dry-run-client and test client.

pkg/deploy/internal/segment/segment.go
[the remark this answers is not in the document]
tomazpu: I was struggling with the name :), no idea how I didn't come to `requestPayload`

pkg/deploy/internal/segment/segment_test.go
tomazpu: I'll have a look for a more elegant solution 👍

Laubi (dismissed):

pkg/delete/delete_test.go
```diff
@@ -1149,93 +1148,48 @@ func TestDelete_Documents(t *testing.T) {
...
-				{
-					Type:           "segment",
-					OriginObjectId: "originObjectID",
-				},
-			},
-		}
 		err := delete.Configs(context.TODO(), client.ClientSet{SegmentClient: &c}, given)
-		assert.NoError(t, err)
-		assert.True(t, c.called, "delete should have been called")
+		//DummyClient returns unimplemented error on every execution of any method
```
arthurpitman, line 1166: I think we need to update this too:
```suggestion
		// fakeClient returns unimplemented error on every execution of any method
```

pkg/delete/delete_test.go
```diff
@@ -1149,93 +1148,48 @@ func TestDelete_Documents(t *testing.T) {
...
-				return segments.Response{StatusCode: http.StatusOK}, nil
-			},
-		}
+	t.Run("With Enabled Segment FF", func(t *testing.T) {
+		t.Setenv(featureflags.Segments.EnvName(), "true")
 
 		err := delete.All(context.TODO(), client.ClientSet{SegmentClient: &c}, api.APIs{})
-		assert.NoError(t, err)
-		assert.True(t, c.called, "delete should have been called")
+		//Dummy client returns unimplemented error on every execution of any method
```
arthurpitman, line 1185:
```suggestion
		// fakeClient returns unimplemented error on every execution of any method
```

Laubi (approved):

pkg/deploy/internal/segment/segment.go
```diff
@@ -58,12 +59,21 @@ func Deploy(ctx context.Context, client deploySegmentClient, properties paramete
...
-			return entities.ResolvedEntity{}, fmt.Errorf("failed to deploy segment with externalId: %s : %w", externalId, err)
+			return entities.ResolvedEntity{}, err
 		}
 
-		return createResolveEntity(id, properties, c), nil
+		if getResponse.StatusCode != http.StatusNotFound {
+
+			id, err := deployWithOriginObjectId(ctx, client, c, requestPayload)
+			if err != nil {
+				return entities.ResolvedEntity{}, fmt.Errorf("failed to deploy segment with externalId: %s : %w", externalId, err)
```
arthurpitman, line 72: Is this error message correct? Or should it reference originObjectId?

Laubi (approved):

pkg/deploy/internal/segment/segment_test.go
jskelin, line 1: I see all the test cases I could think of. 
I would suggest improving the description of the tests.
For example, instead of "deploy with objectOriginId - success POST" something more descriptive. "if create with objectOrigingID fails with 404, create using externalID.

pkg/deploy/internal/segment/segment.go
```diff
@@ -0,0 +1,157 @@
...
+	UID        string `json:"uid"`
+	ExternalId string `json:"externalId"`
+}
+
+func Deploy(ctx context.Context, client deploySegmentClient, properties parameter.Properties, renderedConfig string, c *config.Config) (entities.ResolvedEntity, error) {
+	ctx = logr.NewContext(ctx, log.WithCtxFields(ctx).GetLogr())
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
+	defer cancel()
+
+	externalId, err := idutils.GenerateExternalIDForDocument(c.Coordinate)
```
jskelin, line 53: nit: by go recommendation, id should be all capital --> externalID

jskelin (approved):
only small suggestions, otherwise it's OK

pkg/deploy/internal/segment/segment_test.go
tomazpu: I'll improve on the naming. Thanks!

PR #1671, opened 2025-01-16
feat(documents): Support `Launchpad` documents

Laubi:
This change allows users to download and deploy Dynatrace `Launchpad` documents.

Example:
```yaml
configs:
- id: my-launchpad
  type:
    document:
      kind: launchpad
  config:
    name: My super awesome launchpad!
    template: './my-launchpad.json'
```

Yes: Users can now define `Launchpad` documents


pkg/deploy/internal/document/document.go
```diff
@@ -148,17 +148,22 @@ func createResolvedEntity(documentName string, id string, coordinate coordinate.
 	}
 }
 
-func getDocumentAttributesFromConfigType(t config.Type) (documents.DocumentType, bool, error) {
+var documentMapping = map[config.DocumentKind]documents.DocumentType{
```
tomazpu, line 151: Does it make sense to centralize this in a helper and use it also in the `download.go`? Just with reverse mapping :)

pkg/download/document/download.go
```diff
@@ -32,17 +33,34 @@ import (
...
+var documentMapping = map[string]config.DocumentKind{
+	documents.Dashboard: config.DashboardKind,
+	documents.Notebook:  config.NotebookKind,
+	documents.Launchpad: config.LaunchpadKind,
+}
+
 func Download(client client.DocumentClient, projectName string) (v2.ConfigsPerType, error) {
-	result := make(v2.ConfigsPerType)
+	// due to the current test setup, the types must be downloaded in order. This should be changed eventually
+	var typesToDownload = []documents.DocumentType{
```
tomazpu, line 44: could this just be replaced with `slices.Collect(maps.Values(documentMapping))`?

Laubi: Tried it, but the issue with maps is that the order is by design not specified. And as the comment states, the test setup is at fault that it has to be in a certain order :(

pkg/deploy/internal/document/document.go
Laubi: I thought about that as well, but could not find a good position. I don't know why the deploy/download/delete code is sooo split up :(

Any good idea where to put it? /internal/document package maybe somewhere?

tomazpu: I understand I think `/internal/document/` sounds good.

tomazpu: Lets move this to a refactoring as discussed in the call :)

tomazpu (approved):

PR #1673, opened 2025-01-17
fix: Skip reference parameter creation if parameter is unused in template

arthurpitman:
This PR ensures that no reference parameter is added to a config if the parameter is unused in the template.

Due to the way the code currently works, the fix is made in both dependency resolvers. This could maybe be improved in a follow up PR, however changes to this code have performance implications and thus should be measured.

A test is also added demonstrating the correct behavior.

In practice the simplest way to demonstrate it is to create a classic config with a name that occurs often in configs out of context. A good example is a network zone with the name `_`. 

This string occurs often in templates however as a false positive, which is detected via a regex:

https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/329021241d62ed9f90bf6421982611aeca7c51da/pkg/download/dependency_resolution/resolver/shared.go#L69-L82

The simple fix here just checks if the `replaceAll` function actually changed the content of the template and if not, skips parameter creation.

### DIFF
--- a/pkg/download/dependency_resolution/dependency_resolution_test.go
+++ b/pkg/download/dependency_resolution/dependency_resolution_test.go
@@ -138,6 +138,41 @@ func TestDependencyResolution(t *testing.T) {
 			},
 		},
 		{
+			"no parameter created for false positive",
+			project.ConfigsPerType{
+				"api": []config.Config{
+					{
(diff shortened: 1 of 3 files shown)


jskelin (approved):
LGTM

PR #1676, opened 2025-01-22
feat: Download service users

arthurpitman:
This PR adds functionality to download service-users.

### DIFF
--- a/pkg/account/downloader/internal/http/client.go
+++ b/pkg/account/downloader/internal/http/client.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/accounts"
@@ -44,7 +45,7 @@ func (c *Client) GetUsers(ctx context.Context, accountUUID string) ([]accountman
 func (c *Client) GetServiceUsers(ctx context.Context, accountUUID string) ([]accountmanagement.ExternalServiceUserDto, error) {
 	serviceUsers := []accountmanagement.ExternalServiceUserDto{}
 	const pageSize = 10
-	for page := 1; ; page++ {
+	for page := (int32)(1); page < math.MaxInt32; page++ {
 		r, err := c.getServiceUsersPage(ctx, accountUUID, page, pageSize)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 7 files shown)


pkg/account/downloader/internal/http/client.go
```diff
@@ -41,6 +41,29 @@ func (c *Client) GetUsers(ctx context.Context, accountUUID string) ([]accountman
 	return r.Items, nil
 }
 
+func (c *Client) GetServiceUsers(ctx context.Context, accountUUID string) ([]accountmanagement.ExternalServiceUserDto, error) {
+	serviceUsers := []accountmanagement.ExternalServiceUserDto{}
+	const pageSize = 10
+	for page := 1; ; page++ {
+		r, resp, err := c.ServiceUserManagementAPI.GetServiceUsersFromAccount(ctx, accountUUID).Page(float32(page)).PageSize(pageSize).Execute()
+		defer closeResponseBody(resp)
```
tomazpu, line 49: Using defer in a loop will release the resources at the end of all iterations of the loop. I suggest moving this code into a separate method, calling it in a loop, and deferring in the method. In this way, the release of resources can happen earlier.

pkg/account/downloader/internal/http/client.go
```diff
@@ -41,6 +41,29 @@ func (c *Client) GetUsers(ctx context.Context, accountUUID string) ([]accountman
 	return r.Items, nil
 }
 
+func (c *Client) GetServiceUsers(ctx context.Context, accountUUID string) ([]accountmanagement.ExternalServiceUserDto, error) {
+	serviceUsers := []accountmanagement.ExternalServiceUserDto{}
```
tomazpu, line 45: nit: I think the convention on an empty slice would be declaring it like `var serviceUsers []accountmanagement.ExternalServiceUserDto`

pkg/account/downloader/internal/http/client.go
```diff
@@ -41,6 +41,29 @@ func (c *Client) GetUsers(ctx context.Context, accountUUID string) ([]accountman
...
+	const pageSize = 10
```
tomazpu, line 46: There are max 10 pages?

arthurpitman: We need to pick something here, yes ;)

pkg/account/downloader/internal/http/client.go
arthurpitman: Here I guess it depends what we want: In the case that there are no service users, do we want an empty slice or nil?  I prefer an empty slice.

pkg/account/downloader/internal/http/client.go
arthurpitman: But this is actually the page size, that is, the number of service users retrieved per request. There is no maximum number of pages retrieved.

pkg/account/downloader/internal/http/client.go
tomazpu: Would a named param be then an option? Just thinking out loud, maybe stupid idea :).

jskelin (changes requested):

tomazpu (approved):

jskelin (approved):

PR #1687, opened 2025-02-06
feat(SLO): create slo, config type and feature flag

tomazpu:
This PR introduces the config type `slo` into monaco with the feature flag to toggle it on or off(by default).

### DIFF
--- a/internal/featureflags/temporary.go
+++ b/internal/featureflags/temporary.go
@@ -40,6 +40,9 @@ const (
 	// or when enabled only in string values within the JSON.
 	// Introduced: v2.19.0
 	OnlyCreateReferencesInStringValues FeatureFlag = "MONACO_FEAT_ONLY_CREATE_REFERENCES_IN_STRINGS"
+	// ServiceLevelObjective toggles whether slo configurations are downloaded and / or deployed.
+	// Introduced: v2.19.0
+	ServiceLevelObjective FeatureFlag = "MONACO_FEAT_SLO"
 )
 
 // temporaryDefaultValues defines temporary feature flags and their default values.
@@ -53,4 +56,5 @@ var temporaryDefaultValues = map[FeatureFlag]defaultValue{
 	Segments:                           true,
 	ServiceUsers:                       false,
 	OnlyCreateReferencesInStringValues: false,
+	ServiceLevelObjective:              false,
 }
--- a/pkg/persistence/config/internal/persistence/type_definition.go
+++ b/pkg/persistence/config/internal/persistence/type_definition.go
@@ -31,7 +31,7 @@ import (
 const (
 	BucketType                = "bucket"
 	SegmentType               = "segment"
-	ServiceLevelObjectiveType = "slo"
+	ServiceLevelObjectiveType = "slo-v2"
 )
 
 type TypeDefinition struct {
--- a/pkg/persistence/config/internal/persistence/type_definition.go
+++ b/pkg/persistence/config/internal/persistence/type_definition.go
@@ -29,8 +29,9 @@ import (
 )
 
 const (
-	BucketType  = "bucket"
-	SegmentType = "segment"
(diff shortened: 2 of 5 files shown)


pkg/persistence/config/internal/persistence/type_definition.go
```diff
@@ -29,8 +29,9 @@ import (
 )
 
 const (
-	BucketType  = "bucket"
-	SegmentType = "segment"
+	BucketType                = "bucket"
+	SegmentType               = "segment"
+	ServiceLevelObjectiveType = "slo"
```
jskelin, line 34: Identifier should be `slo-v2`

pkg/persistence/config/loader/config_loader_test.go
```diff
@@ -1225,6 +1225,62 @@ configs:
...
+			name:             "SLO config with FF on",
+			envVars:          map[string]string{featureflags.ServiceLevelObjective.EnvName(): "true"},
+			filePathArgument: "test-file.yaml",
+			filePathOnDisk:   "test-file.yaml",
+			fileContentOnDisk: `
+configs:
+- id: slo-config-id
+  config:
+    template: 'profile.json'
+  type: slo
```
jskelin, line 1238: the name of the type is: `slo-v2`

jskelin (changes requested):
The identifier is `slo-v2` (instead of `slo`, which is already reserved)

internal/featureflags/temporary.go
```diff
@@ -40,6 +40,9 @@ const (
 	// or when enabled only in string values within the JSON.
 	// Introduced: v2.19.0
 	OnlyCreateReferencesInStringValues FeatureFlag = "MONACO_FEAT_ONLY_CREATE_REFERENCES_IN_STRINGS"
+	// ServiceLevelObjective toggles whether slo configurations are downloaded and / or deployed.
+	// Introduced: v2.19.0
+	ServiceLevelObjective FeatureFlag = "MONACO_FEAT_SLO"
```
arthurpitman, line 45:
```suggestion
	ServiceLevelObjective FeatureFlag = "MONACO_FEAT_SLO_V2"
```

pkg/persistence/config/loader/config_loader_test.go
```diff
@@ -1225,6 +1225,62 @@ configs:
 `,
 			wantErrorsContain: []string{"unknown API: segment"},
 		},
+		{
```
jskelin, line 1228: Is it possible to put this in a separate test? Maybe even in a separate file? 
This table taste is huge.

pkg/persistence/config/writer/config_writer_test.go
```diff
@@ -1087,6 +1087,67 @@ func TestWriteConfigs(t *testing.T) {
 			},
 			expectedErrs: []string{"config.Segment"},
 		},
+		{
```
jskelin, line 1090: Is it possible to put this in a separate test? Maybe even in a separate file?
This table taste is huge.

arthurpitman (dismissed):

pkg/persistence/config/loader/config_loader_test.go
tomazpu: The test file is 1k5 lines of code. If we want to do it, I can but in a separate PR.

pkg/persistence/config/writer/config_writer_test.go
jskelin: We discussed and the proper refactoring is going to be as a separate task

pkg/persistence/config/loader/config_loader_test.go
jskelin: We discussed and the proper refactoring is going to be a separate task

jskelin (approved):

PR #1688, opened 2025-02-06
ci: Allow to overwrite the ko base image path

Laubi:
This change allows us to use a different proxy instead of querying docker directly, potentially breaking the pipeline because of "You have reached your pull rate limit".

By default, it still pulls from the docker-hub

Documentation about ko: https://ko.build/configuration/

no

### DIFF
--- a/.ci/jenkins/tools/ko.groovy
+++ b/.ci/jenkins/tools/ko.groovy
@@ -28,10 +28,13 @@ private def registrySecret(String registry) {
 
     return [[path        : "${path}",
              secretValues: [
-                 [envVar: 'registry', vaultKey: 'registry', isRequired: true],
-                 [envVar: 'repo', vaultKey: 'repo', isRequired: true],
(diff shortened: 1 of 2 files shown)


tomazpu (dismissed):

arthurpitman (commented):
Looks good, but I'm curious: how does the `3.20` in `alpine:3.20` get updated? Will this still happen with this change?

Laubi:
> Looks good, but I'm curious: how does the `3.20` in `alpine:3.20` get updated? Will this still happen with this change?

In the Makefile.

You specify `KO_DEFAULTBASEIMAGE=$(KO_BASE_IMAGE_PATH)/alpine:3.20`, and `alpine:3.20` will be used.
The `base_image_path` is basically the registry from which the image should be pulled, which is `docker.io` by default, and we can overwrite it.

tomazpu (approved):

PR #1692, opened 2025-02-07
refactor: eliminate support archive global variable `SupportArchive`

jskelin:
Since recently, `context.Context` has not been introduced into Monaco CLI. For keeping information that support-archive is needed, the global variable `SupportArchive` has been used.
This PR makes information available via the `context.Context`.

Instead of the solution where printing log files is done via the defer function, `cobra.OnFinalize` option is used.

### DIFF
--- a/cmd/monaco/delete/delete.go
+++ b/cmd/monaco/delete/delete.go
@@ -43,7 +43,7 @@ func Delete(ctx context.Context, environments manifest.Environments, entriesToDe
 			log.WithCtxFields(ctx).Warn("Delete file contains Dynatrace Platform specific types, but no oAuth credentials are defined for environment %q - Dynatrace Platform configurations won't be deleted.", env.Name)
 		}
 
-		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{})
+		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth)
 		if err != nil {
 			return fmt.Errorf("failed to create API client for environment %q due to the following error: %w", env.Name, err)
 		}
--- a/cmd/monaco/delete/delete.go
+++ b/cmd/monaco/delete/delete.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/support"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
@@ -44,7 +43,7 @@ func Delete(ctx context.Context, environments manifest.Environments, entriesToDe
 			log.WithCtxFields(ctx).Warn("Delete file contains Dynatrace Platform specific types, but no oAuth credentials are defined for environment %q - Dynatrace Platform configurations won't be deleted.", env.Name)
 		}
 
-		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{SupportArchive: support.SupportArchive})
+		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{})
 		if err != nil {
 			return fmt.Errorf("failed to create API client for environment %q due to the following error: %w", env.Name, err)
 		}
--- a/cmd/monaco/dynatrace/dynatrace.go
+++ b/cmd/monaco/dynatrace/dynatrace.go
@@ -189,7 +189,7 @@ func CreateEnvironmentClients(ctx context.Context, environments manifest.Environ
 			continue
 		}
 
-		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{})
+		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth)
 		if err != nil {
 			return EnvironmentClients{}, err
 		}
--- a/cmd/monaco/dynatrace/dynatrace.go
+++ b/cmd/monaco/dynatrace/dynatrace.go
@@ -114,7 +114,7 @@ func isPlatformEnvironment(ctx context.Context, env manifest.EnvironmentDefiniti
 }
 
 // CreateAccountClients gives back clients to use for specific accounts
-func CreateAccountClients(manifestAccounts map[string]manifest.Account) (map[account.AccountInfo]*accounts.Client, error) {
+func CreateAccountClients(ctx context.Context, manifestAccounts map[string]manifest.Account) (map[account.AccountInfo]*accounts.Client, error) {
 	concurrentRequestLimit := environment.GetEnvValueIntLog(environment.ConcurrentRequestsEnvKey)
 	accClients := make(map[account.AccountInfo]*accounts.Client, len(manifestAccounts))
 	for _, acc := range manifestAccounts {
@@ -132,7 +132,7 @@ func CreateAccountClients(manifestAccounts map[string]manifest.Account) (map[acc
 			WithRetryOptions(&client.DefaultRetryOptions).
 			WithAccountURL(accountApiUrlOrDefault(acc.ApiUrl))
 
-		if support.SupportArchive {
+		if v := ctx.Value(support.SupportArchive{}); v != nil && v.(*support.SupportArchive).Value {
 			factory = factory.WithHTTPListener(&corerest.HTTPListener{Callback: trafficlogs.GetInstance().LogToFiles})
 		}
 
@@ -189,7 +189,7 @@ func CreateEnvironmentClients(ctx context.Context, environments manifest.Environ
 			continue
 		}
 
-		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{SupportArchive: support.SupportArchive})
+		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{})
 		if err != nil {
 			return EnvironmentClients{}, err
 		}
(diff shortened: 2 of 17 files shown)


pkg/client/clientset.go
```diff
@@ -248,7 +248,7 @@ func CreateClientSet(ctx context.Context, url string, auth manifest.Auth, opts C
 		WithRetryOptions(&DefaultRetryOptions).
 		WithRateLimiter(true)
 
-	if opts.SupportArchive {
+	if v := ctx.Value(support.SupportArchive{}); v != nil && v.(*support.SupportArchive).Value {
```
arthurpitman, line 251: This line is different to the one in `dynatrace.go`; in any case I'd like to see the check put in a function in the `support` package

cmd/monaco/runner/runner.go
```diff
@@ -76,7 +80,9 @@ Examples:
     monaco deploy service.yaml -e dev`,
 
 		PersistentPreRun: func(cmd *cobra.Command, args []string) {
-			fileBasedLogging := featureflags.LogToFile.Enabled() || support.SupportArchive
+			cmd.Context().Value(support.SupportArchive{}).(*support.SupportArchive).Value = supportArchive
```
arthurpitman, line 83: let's also make this a detail in the `support` package

cmd/monaco/delete/delete.go
```diff
@@ -36,15 +35,15 @@ import (
...
 	var envsWithDeleteErrs []string
 	for _, env := range environments {
-		ctx := context.WithValue(context.TODO(), log.CtxKeyEnv{}, log.CtxValEnv{Name: env.Name, Group: env.Group})
+		ctx := context.WithValue(ctx, log.CtxKeyEnv{}, log.CtxValEnv{Name: env.Name, Group: env.Group})
 		if containsPlatformTypes(entriesToDelete) && env.Auth.OAuth == nil {
 			log.WithCtxFields(ctx).Warn("Delete file contains Dynatrace Platform specific types, but no oAuth credentials are defined for environment %q - Dynatrace Platform configurations won't be deleted.", env.Name)
 		}
 
-		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{SupportArchive: support.SupportArchive})
+		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{})
```
tomazpu, line 46: `client.CreateClientSet` method has now changed and always passes empty `client.ClientOptions{}`. Why then not remove it?

cmd/monaco/purge/purge.go
```diff
@@ -73,10 +72,10 @@ func purgeConfigs(environments []manifest.EnvironmentDefinition, apis api.APIs)
 	return nil
 }
 
-func purgeForEnvironment(env manifest.EnvironmentDefinition, apis api.APIs) error {
-	ctx := context.WithValue(context.TODO(), log.CtxKeyEnv{}, log.CtxValEnv{Name: env.Name, Group: env.Group})
+func purgeForEnvironment(ctx context.Context, env manifest.EnvironmentDefinition, apis api.APIs) error {
+	ctx = context.WithValue(ctx, log.CtxKeyEnv{}, log.CtxValEnv{Name: env.Name, Group: env.Group})
 
-	clients, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{SupportArchive: support.SupportArchive})
+	clients, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth, client.ClientOptions{})
```
tomazpu, line 78: Same here

tomazpu (changes requested):
pkg/download/bucket/download.go
pkg/download/document/download.go

Still have context.TODO()

cmd/monaco/runner/runner.go
```diff
@@ -49,7 +50,9 @@ func RunCmd(fs afero.Fs, command *cobra.Command) error {
 			}
 		}
 	}()
-	err := command.Execute()
+
+	ctx = context.WithValue(ctx, support.SupportArchive{}, supportArchivePtr)
```
Laubi, line 54: The key should not be set like this, please follow the best practices for defining context keys and setting/receiving context values.
After quick Google search: https://mainfunction.org/articles/golang-context-with-value-best-practices/

This article also gives information on why: https://medium.com/@sanhdoan/go-context-best-practices-a-better-way-to-store-values-3c1fb3bf4e84

Laubi (changes requested):
Please follow the best practices for contexts, I linked two articles. 
Please also update your PR title to match all other open pull requests

pkg/client/clientset.go
jskelin: done

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -132,7 +132,7 @@ func CreateAccountClients(manifestAccounts map[string]manifest.Account) (map[acc
 			WithRetryOptions(&client.DefaultRetryOptions).
 			WithAccountURL(accountApiUrlOrDefault(acc.ApiUrl))
 
-		if support.SupportArchive {
+		if v := ctx.Value(support.SupportArchive{}); v != nil && v.(support.SupportArchive).Value {
```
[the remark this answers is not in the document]
jskelin: done

cmd/monaco/runner/runner.go
jskelin: done

cmd/monaco/delete/delete.go
jskelin: That structure has more fields. In the production part of the code, we always pass empty struct, but in one or two places in the test part of the code, it is used.
I would refactor it as part of a separate issue.

cmd/monaco/runner/runner.go
jskelin: done

jskelin:
> pkg/download/bucket/download.go pkg/download/document/download.go pkg/client/dtclient/settings_client.go
> 
> Still have context.TODO()

The scope for this was only `cmd` package, not the entire code base

cmd/monaco/supportarchive/supportarchive.go
```diff
@@ -29,11 +30,39 @@ import (
...
-func Archive(fs afero.Fs) error {
+// NewContextWithSupportArchive creates a new child context that has a signal that the support archive is required.
+// If the command has already been called, it returns the given context.
+func NewContextWithSupportArchive(ctx context.Context) context.Context {
+	if ctx.Value(ctxKey{}) != nil {
+		return ctx
+	}
+
+	key := &ctxKey{}
+	return context.WithValue(ctx, ctxKey{}, key)
```
Laubi, line 45: Please have a look at the articles I sent again.
This solution is better, but still not where it should be. 

Why even use pass struct?

cmd/monaco/supportarchive/supportarchive.go
```diff
@@ -29,11 +30,39 @@ import (
...
+	return v != nil && v.(*ctxKey).value
+}
+
+// Enable sets the support archive as enabled in the given context.
+// Prior to this, NewContextWithSupportArchive needs to be called
+func Enable(ctx context.Context) {
+	v := ctx.Value(ctxKey{})
+
+	if v == nil {
+		panic("support archive not created")
```
Laubi, line 59: Why? Why not just set the value?

Laubi (dismissed):

cmd/monaco/supportarchive/supportarchive.go
arthurpitman: We can't just set the value because it would set it in _this_ `context.Context` if it doesnt exist rather than the parent  (or great-great-uhr-alt-old-context), meaning when we go to check if the support archive has been enabled later on, we wont have that information anymore.

cmd/monaco/delete/delete.go
jskelin: The `client.ClientOptions{}` is removed from the signature of the function. For testing purpiose, a new function `client.CreateClientSetWithOptions` is created.

cmd/monaco/supportarchive/supportarchive.go
jskelin: Now two different struct are used, one as key, second as value holder

cmd/monaco/runner/runner.go
jskelin: Now two different struct are used, one as key, second as value holder

cmd/monaco/supportarchive/supportarchive.go
jskelin: Context Value is used here in a different way.
At the very beginning of "Monaco" execution, there is a defer function that prints a support archive even if panic occurs. On the other hand, information about whether the archive should be printed is collected later in the program. For that reason, it is a 2-step process (ContextWithSupportArchive and Enable). First needs to happen at the very beginning before the defer function is specified, and second at the point when cobra framework gives us values of flags.
To avoid using global variables (even non-exported), the value in the context is a pointer. With it, we have the backdoor to alter the value later during the execution and get information in the defer function.

jskelin: We discovered the cobra.OnFinalize command, which works well for this use case.

cmd/monaco/runner/runner.go
```diff
@@ -61,8 +51,20 @@ func BuildCmd(fs afero.Fs) *cobra.Command {
 	return BuildCmdWithLogSpy(fs, nil)
 }
 
+func writeSupportArchive(fs afero.Fs) func() {
+	return func() { // writing the support archive is a deferred function call in order to guarantee that a support archive is also written in case of a panic
```
tomazpu, line 55: Old comment from copy/paste still here. Please remove it.

jskelin: removed

pkg/client/clientset.go
```diff
@@ -226,7 +226,11 @@ func validateURL(dtURL string) error {
 	return nil
 }
 
-func CreateClientSet(ctx context.Context, url string, auth manifest.Auth, opts ClientOptions) (*ClientSet, error) {
+func CreateClientSet(ctx context.Context, url string, auth manifest.Auth) (*ClientSet, error) {
+	return CreateClientSetWothOtipons(ctx, url, auth, ClientOptions{})
+}
+
+func CreateClientSetWothOtipons(ctx context.Context, url string, auth manifest.Auth, opts ClientOptions) (*ClientSet, error) {
```
arthurpitman, line 233: Typo:

```suggestion
func CreateClientSetWothOptions(ctx context.Context, url string, auth manifest.Auth, opts ClientOptions) (*ClientSet, error) {
```

jskelin: done

tomazpu (dismissed):

tomazpu (approved):

cmd/monaco/supportarchive/supportarchive.go
jskelin: This is outdated

arthurpitman (approved):
🚀

PR #1693, opened 2025-02-07
fix: Skip duplicate downloaded values

arthurpitman:
In some scenarios, we have observed an environment may list multiple configurations for download with the same ID. As Monaco cannot distinguish between these, this PR adds a warning and removes the duplicate.

### DIFF
--- a/pkg/download/classic/download.go
+++ b/pkg/download/classic/download.go
@@ -61,6 +61,8 @@ func Download(client client.ConfigClient, projectName string, apisToDownload api
 				return
 			}
 
+			foundValues = checkAndRemoveValuesWithDuplicateIDs(currentApi, foundValues)
+
 			foundValues = filterConfigsToSkip(currentApi, foundValues, filters)
 			if len(foundValues) == 0 {
 				log.WithFields(field.Type(currentApi.ID)).Debug("No configs of type '%s' to download", currentApi.ID)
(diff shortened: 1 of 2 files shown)


pkg/download/classic/download.go
```diff
@@ -82,6 +84,22 @@ func Download(client client.ConfigClient, projectName string, apisToDownload api
 	return results, nil
 }
 
+func checkAndRemoveValuesWithDuplicateIDs(api api.API, originalValues values) values {
+	seenIDs := make(map[string]struct{}, len(originalValues))
+	filteredValues := make(values, 0, len(originalValues))
+
+	for _, v := range originalValues {
+		if _, alreadySeen := seenIDs[v.value.Id]; alreadySeen {
+			log.Warn("Received multiple '%s' configs with the same ID '%s'; skipping duplicate named '%s'", api.ID, v.value.Id, v.value.Name)
```
jskelin, line 93: Is it a good idea to include only one configuration and reject all others? 
In this way, the complete Monaco process will end successfully, and a user can interpret that as meaning that all configurations have been downloaded.

Laubi: As discussed, it's ok as is here

Laubi (approved):

tomazpu (approved):

PR #1699, opened 2025-02-11
fix(settings): Download ordered settings needs to take scopes into account

Laubi:
When downloading settings objects, they are returned ordered.
We utilize that to set the `insertAfter` parameter always to the previous setting if the schema supports ordering.
The issue is that we did not take the scope into account. the `insertAfter` only works within the same scope, so referencing an object from another scope will cause an error.
This changeset changes this by 'remembering' the last settings object for the given scope.

Yes, currently, the download breaks for settings.


arthurpitman (approved):
LGTM

PR #1701, opened 2025-02-12
feat(SLO): download support for SLO API

Kirdock:
Introduces the download command for SLOs

### DIFF
--- a/cmd/monaco/download/download_command.go
+++ b/cmd/monaco/download/download_command.go
@@ -101,6 +101,10 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
 		cmd.Flags().BoolVar(&f.onlySegments, "only-segments", false, "Only download segment configurations, skip all other configuration types")
 	}
 
+	if featureflags.ServiceLevelObjective.Enabled() {
+		cmd.Flags().BoolVar(&f.onlySLOsV2, "only-slo-v2", false, "Only download SLO-V2, skip all other configuration types")
+	}
+
 	err := errors.Join(
 		cmd.RegisterFlagCompletionFunc("token", completion.EnvVarName),
 		cmd.RegisterFlagCompletionFunc("oauth-client-id", completion.EnvVarName),
--- a/cmd/monaco/download/download_command.go
+++ b/cmd/monaco/download/download_command.go
@@ -17,6 +17,7 @@ package download
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 
@@ -32,6 +33,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client"
 	clientAuth "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/auth"
 	versionClient "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/version"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest"
 )
 
@@ -102,7 +104,7 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
 	}
 
 	if featureflags.ServiceLevelObjective.Enabled() {
-		cmd.Flags().BoolVar(&f.onlySLOsV2, "only-slo-v2", false, "Only download SLO-V2, skip all other configuration types")
+		cmd.Flags().BoolVar(&f.onlySLOsV2, "only-slo-v2", false, fmt.Sprintf("Only download %s, skip all other configuration types", config.ServiceLevelObjectiveID))
 	}
 
 	err := errors.Join(
--- a/cmd/monaco/download/downloadopts.go
+++ b/cmd/monaco/download/downloadopts.go
@@ -34,6 +34,7 @@ type downloadConfigsOptions struct {
 	onlyDocuments    bool
 	onlyOpenPipeline bool
 	onlySegment      bool
+	onlySLOV2        bool
 }
 
 func (opts downloadConfigsOptions) valid() []error {
--- a/pkg/download/slo/download.go
+++ b/pkg/download/slo/download.go
@@ -38,7 +38,7 @@ func Download(client DownloadSloClient, projectName string) (project.ConfigsPerT
 	result := project.ConfigsPerType{}
 	downloadedConfigs, err := client.List(context.TODO())
 	if err != nil {
-		log.WithFields(field.Type(config.ServiceLevelObjectiveID), field.Error(err)).Error("Failed to fetch the list of existing SLO-v2 configs: %v", err)
+		log.WithFields(field.Type(config.ServiceLevelObjectiveID), field.Error(err)).Error("Failed to fetch the list of existing %s configs: %v", config.ServiceLevelObjectiveID, err)
 		// error is ignored
 		return nil, nil
 	}
@@ -47,7 +47,7 @@ func Download(client DownloadSloClient, projectName string) (project.ConfigsPerT
 	for _, downloadedConfig := range downloadedConfigs.All() {
 		c, err := createConfig(projectName, downloadedConfig)
 		if err != nil {
-			log.WithFields(field.Type(config.ServiceLevelObjectiveID), field.Error(err)).Error("Failed to convert SLO-v2: %v", err)
+			log.WithFields(field.Type(config.ServiceLevelObjectiveID), field.Error(err)).Error("Failed to convert %s: %v", config.ServiceLevelObjectiveID, err)
 			continue
 		}
 		configs = append(configs, c)
--- a/pkg/download/slo/download.go
+++ b/pkg/download/slo/download.go
@@ -1,0 +1,92 @@
+/*
+ * @license
+ * Copyright 2025 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package slo
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/coordinate"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/template"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/download/internal/templatetools"
+	project "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project/v2"
+)
+
+type DownloadSloClient interface {
+	List(ctx context.Context) (api.PagedListResponse, error)
+}
+
+func Download(client DownloadSloClient, projectName string) (project.ConfigsPerType, error) {
+	result := project.ConfigsPerType{}
+	downloadedConfigs, err := client.List(context.TODO())
+	if err != nil {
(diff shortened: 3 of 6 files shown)


pkg/download/config_creation/config_creation.go
jskelin, line 1: By GO convention, snake_case and camelCase are not allowed
https://go.dev/blog/package-names

go.sum
```diff
@@ -11,6 +11,8 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/dynatrace/dynatrace-configuration-as-code-core v0.7.1-0.20250207115859-d33243b1fdf6 h1:F3cTkffc2xH6IK/MS3gPzNgpjcxG71ik72dTrfnv9ig=
 github.com/dynatrace/dynatrace-configuration-as-code-core v0.7.1-0.20250207115859-d33243b1fdf6/go.mod h1:QUFQtBjb6BkrzqBzCisqg4c1olv2FWC3rhYnxAyQoRo=
+github.com/dynatrace/dynatrace-configuration-as-code-core v0.7.1-0.20250211150749-1391cd8fbd8b h1:uUspeVi3gCtcttfWEOnwnkN9PAoDfDbnNwjOMrv2E0M=
```
jskelin, line 14: After updating the project with a newer version, run `go mod tidy` to remove reference to the previous version (ln. 12-13)

pkg/download/config_creation/config_creation.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/coordinate"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/template"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/download/internal/templatetools"
+)
+
+// CreateConfig creates a config based on a given JSON payload and removes properties like the id (idPropertyKey), version and externalId
+func CreateConfig(projectName string, data []byte, configType config.Type, idPropertyKey string) (config.Config, error) {
```
jskelin, line 29: I would reduce the scope of this function to just removing unwanted entries. Than we could reuse it in every client. For example, the bucket client needs to delete:
```Go
	r.Delete(bucketName)
	r.Delete(status)
	r.Delete(version)
	r.Delete(updatable)
```
The signature should look something like: 
```Go
func Remove(data []byte, keys ... string) (json.JsonRaw, error)
```

What's your opinion @arthurpitman ?

pkg/download/slo/download.go
```diff
@@ -0,0 +1,51 @@
...
+		log.WithFields(field.Type(config.ServiceLevelObjectiveID), field.Error(err)).Error("Failed to fetch the list of existing SLOs: %v", err)
+		// error is ignored
+		return nil, nil
+	}
+
+	var configs []config.Config
+	for _, downloadedConfig := range downloadedConfigs.All() {
+		c, err := config_creation.CreateConfig(projectName, downloadedConfig, config.ServiceLevelObjective{}, "id")
+		if err != nil {
+			log.WithFields(field.Type(config.ServiceLevelObjectiveID), field.Error(err)).Error("Failed to convert SLO: %v", err)
```
jskelin, line 43: Is it possible to add ID of the config?

cmd/monaco/download/downloadopts.go
```diff
@@ -34,6 +34,7 @@ type downloadConfigsOptions struct {
 	onlyDocuments    bool
 	onlyOpenPipeline bool
 	onlySegment      bool
+	onlySLOs         bool
```
jskelin, line 37: nit: I would use `onlySLOv2` as we have just `SLO` in classic api

cmd/monaco/download/download_command.go
```diff
@@ -101,6 +101,10 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
 		cmd.Flags().BoolVar(&f.onlySegments, "only-segments", false, "Only download segment configurations, skip all other configuration types")
 	}
 
+	if featureflags.ServiceLevelObjective.Enabled() {
+		cmd.Flags().BoolVar(&f.onlySLOs, "only-slos", false, "Only download SLOs, skip all other configuration types")
```
jskelin, line 105: For a user, it should be always `slo-v2`. Same as the type name.
As we also have `slo` in the classic endpoint, it could be confusing.

pkg/download/slo/download.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/download/config_creation"
+	project "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project/v2"
+)
+
+func Download(client client.ServiceLevelObjectiveClient, projectName string) (project.ConfigsPerType, error) {
```
jskelin, line 30: It's a good practice to use interfaces.
In GO, they are on the consumer side, as opposed to, e.g., Java, where they are on the implementation side.

Kirdock: Done ✅

pkg/download/slo/download.go
Kirdock: Actually, no. We only have the ID after successfully parsing the JSON payload

pkg/download/config_creation/config_creation.go
Kirdock: I adapted it to only return the id and the json payload. The config creation is then again up to the segments/slos

pkg/download/config_creation/config_creation.go
Kirdock: We already have it quite a lot in the code, also for tests (`.*_test`). As we don't provide monaco as a lib, it's up to us if we align with the guideline or not.
Based on the decision we should then also adapt the rest that is not aligned

go.sum
Kirdock: Noted, thanks.
Rebased to main, therefore the change is not needed anymore

cmd/monaco/download/downloadopts.go
Kirdock: done ✅

cmd/monaco/download/download_command.go
Kirdock: done ✅ 
changed to `only-slo-v2`

pkg/download/config_creation/config_creation.go
Kirdock: I adapted it again. Now it's also possible to use it for download buckets. I haven't touched it yet, because this would change the error logs

regarding the delete suggestion: I would not vote for that as it would only transform it basically from vertical to horizontal with no real value

pkg/download/config_creation/config_creation.go
tomazpu: I agree with @jskelin. The naming convention for a package name is flatcase `config_creation` would become something like `configcreation`(talking about line 17, other examples in code: dependencygraph, cmdutils). Same for filename.

pkg/download/config_creation/config_creation_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+type uidStruct struct {
+	Uid string `json:"uid"`
+}
+
+type idStruct struct {
+	Id string `json:"id"`
+}
+
+func TestCreateConfig(t *testing.T) {
```
tomazpu, line 35: Here, I would create a general test for (invalid/valid input) and a test method pre-config type. The specifics can then be tested in separation. 

I can imagine else this test will grow quite quickly and become hard to read.

tomazpu, line 35: I would also add some additional error test cases. Specially the field that I have in mind is `structPointer` as declared as any. The specific test for `replaceParam`.

pkg/download/segment/download.go
```diff
@@ -20,17 +20,20 @@ import (
...
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/coordinate"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/template"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/download/internal/templatetools"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/download/config_creation"
 	project "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project/v2"
 )
 
+type requiredSegmentProps struct {
+	UId string `json:"uid"`
```
tomazpu, line 34: nit: a time ago we agreed on capitalizing Id -> ID. In your case, `UID`.

pkg/download/slo/download.go
```diff
@@ -0,0 +1,79 @@
...
+type DownloadSloClient interface {
+	List(ctx context.Context) (api.PagedListResponse, error)
+}
+
+func Download(client DownloadSloClient, projectName string) (project.ConfigsPerType, error) {
+	result := project.ConfigsPerType{}
+	downloadedConfigs, err := client.List(context.TODO())
+	if err != nil {
+		log.WithFields(field.Type(config.ServiceLevelObjectiveID), field.Error(err)).Error("Failed to fetch the list of existing SLOs: %v", err)
+		// error is ignored
```
tomazpu, line 46: nit: i don't see the need for the comment

pkg/download/slo/download.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, nil
+	}
+
+	configs := make([]config.Config, 0)
```
tomazpu, line 50: If at this time you know the number of configs would be great to declare it like `configs := make([]config.Config, len(downloadedConfigs))`

As you are new to go maybe this article can give some insights:
https://sampath04.hashnode.dev/make-your-go-code-efficient-using-make-when-creating-slices

Kirdock: I also thought about that but decided against it due to possible errors that could occur. Then the slice may not be filled.
I'll add it then 👍

pkg/download/config_creation/config_creation.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/download/internal/templatetools"
+)
+
+type PreparedConfig struct {
+	JSONString string
+	Parameters map[string]parameter.Parameter
+}
+
+// PrepareConfig returns the given id (idPropertyKey) and the JSON string without the given properties
+func PrepareConfig(data []byte, structPointer any, deletedProperties []string, replaceParam string) (PreparedConfig, error) {
```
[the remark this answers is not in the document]
Kirdock: Done.
Tell me if you want to split it differently or if we should completely change the implementation

pkg/download/config_creation/config_creation_test.go
Kirdock: Done ✅

tomazpu (dismissed):

pkg/download/config_creation/config_creation.go
Kirdock: Decision: Leave the duplicated code in for now and do some refactoring later on

pkg/download/slo/download.go
```diff
@@ -0,0 +1,92 @@
...
+
+type DownloadSloClient interface {
+	List(ctx context.Context) (api.PagedListResponse, error)
+}
+
+func Download(client DownloadSloClient, projectName string) (project.ConfigsPerType, error) {
+	result := project.ConfigsPerType{}
+	downloadedConfigs, err := client.List(context.TODO())
+	if err != nil {
+		log.WithFields(field.Type(config.ServiceLevelObjectiveID), field.Error(err)).Error("Failed to fetch the list of existing SLOs: %v", err)
```
jskelin, line 41: The error message should be:
"Failed to fetch the list of existing SLO-v2 configs:..."
Or something similar

pkg/download/slo/download.go
```diff
@@ -0,0 +1,92 @@
...
+		// error is ignored
+		return nil, nil
+	}
+
+	var configs []config.Config
+	for _, downloadedConfig := range downloadedConfigs.All() {
+		c, err := createConfig(projectName, downloadedConfig)
+		if err != nil {
+			log.WithFields(field.Type(config.ServiceLevelObjectiveID), field.Error(err)).Error("Failed to convert SLO: %v", err)
```
jskelin, line 50: Use the term `SLOv2` in the error message

cmd/monaco/download/download_command.go
jskelin: Personally, I would use string formatting (fmt.Sprintf) and `config.ServiceLevelObjectiveID` as an argument. Then, the same term would be used everywhere. 

What do you think @arthurpitman ?

jskelin (approved):
nice work

jskelin (approved):

PR #1702, opened 2025-02-12
chore: Update Go to version 1.24

Laubi:
Update go.mod version to go `1.24.0`.
Furthermore, update all workflows and Jenkins files to use this new version. Last, a test was failing with 'non-constrant string provided' to Errorf, which this commit also resolves.


Kirdock:
Because I saw it in another go version bump, do we also need to increase the version for semgrep?
https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1391/files#diff-943c91ac8e5085136f071a01d6ad70ffe7beefb141dd4d45c3fc75c52c3c9dbbR22

.github/workflows/dependencies-and-licenses.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 #v5.3.0
         with:
-          go-version: '~1.23'
+          go-version: '~1.24'
```
jskelin, line 20: Use go.mod to get go version version from go.mod. 
Take a look: https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

pkg/persistence/config/writer/config_writer.go
```diff
@@ -769,11 +772,12 @@ func newDetailedConfigWriterError(context *serializerContext, err error) configE
 	}
 }
 
-func fmtDetailedConfigWriterError(context *serializerContext, format string, args ...interface{}) configError.DetailedConfigWriterError {
+func fmtDetailedConfigWriterError(context *serializerContext, err error) configError.DetailedConfigWriterError {
```
arthurpitman, line 775: Can we remove this and just use the function above, `newDetailedConfigWriterError(...)`? It now looks just the same I think.

Laubi: Awesome catch, did not see that one

jskelin (changes requested):

.github/workflows/dependencies-and-licenses.yml
arthurpitman: Like we could have `go-version-file: 'path/to/go.mod'` and then we wouldn't need to update it separately here?

jskelin: > Like we could have `go-version-file: 'path/to/go.mod'` and then we wouldn't need to update it separately here?

exactley

Laubi: The main goal of this PR is updating the Go version to a new version, and everything beyond is already stretching that.
Since this `go-version-file` is a different approach to what we have now, I don't see this as a blocker for this PR.
However, I will create a followup-PR introducing this.

Laubi:
> Because I saw it in another go version bump, do we also need to increase the version for semgrep? I would assume no https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1391/files#diff-943c91ac8e5085136f071a01d6ad70ffe7beefb141dd4d45c3fc75c52c3c9dbbR22

Since the check works fine, I'd say no.
https://github.com/Dynatrace/dynatrace-configuration-as-code/actions/runs/13372996992/job/37345673812?pr=1702

Kirdock (approved):

arthurpitman (approved):
LGTM

jskelin (approved):

PR #1706, opened 2025-02-17
feat(support-archive): add deployment report to archive

Kirdock:
It adds the generated deployment report if enabled to the archive

### DIFF
--- a/cmd/monaco/supportarchive/supportarchive.go
+++ b/cmd/monaco/supportarchive/supportarchive.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/spf13/afero"
 
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/environment"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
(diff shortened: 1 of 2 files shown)


cmd/monaco/supportarchive/supportarchive.go
```diff
@@ -56,6 +57,9 @@ func Write(fs afero.Fs) error {
 		ffState,
 		log.MemStatFilePath(),
 	}
+	if reportFilename, ok := os.LookupEnv(environment.DeploymentReportFilename); ok && len(reportFilename) > 0 {
```
tomazpu, line 60: The method `os.LookupEnv` returns true if more than one environment variable is found with this name. So, I don't see a need in `len(reportFilename)`. Correct me if I'm wrong.

Kirdock: The environment variable can be set but still be empty string
>  If the variable is present in the environment the value (which may be empty) is returned and the boolean is true

Kirdock: I would vote for using `Getenv` as we don't have any error-case in thise case

arthurpitman (approved):

PR #1707, opened 2025-02-17
chore: pass right context to clients and functions

Kirdock:
Forwards the right context to clients and therefore they are now correctly cancelled if related parent contexts are cancelled.
One `context.Background` is still in there but this is the one on top of `main.go`

Depending on which PR is merged first, the SLO feature needs to be updated either here or in the related PR

Assumption:
For the account deployer `\pkg\account\deployer\deployer.go` I had two solutions
1. add a context to the properties of the deployer and add it in the constructors (`NewAccountDeployer` and `WithMaxConcurrentDeploys`). This would result in less forwarding of the context
2. add the context to the `Deploy` function and forward it to the related calls

I went for option 2. Tell me if I should go for option 1

### DIFF
--- a/cmd/monaco/account/delete.go
+++ b/cmd/monaco/account/delete.go
@@ -16,6 +16,14 @@ package account
 
 import (
 	"fmt"
+	"path/filepath"
+
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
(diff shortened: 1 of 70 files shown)


Kirdock:
Will split this one up

PR #1708, opened 2025-02-17
feat(settings): Allow inserting in `front` and `back`

Laubi:

### DIFF
--- a/cmd/monaco/integrationtest/v2/test-resources/settings-ordered/insert-position/both-back-and-front-are-set/config.yaml
+++ b/cmd/monaco/integrationtest/v2/test-resources/settings-ordered/insert-position/both-back-and-front-are-set/config.yaml
@@ -1,0 +1,27 @@
+# Test if setting `insertAfter` to `BACK` works as expected.
+#
+# First config, `first`, is deployed and will be added to the front
+# Second config, `last`, will be added to the back. The phantom reference enforces the deployment order
+
+configs:
+- id: first
+  config:
+    template: url-sampling.json
+  type:
+    settings:
+      schema: builtin:url-based-sampling
+      scope: environment
+- id: last
+  config:
+    template: url-sampling.json
+    parameters:
+      phantom:
+        type: reference
+        configId: first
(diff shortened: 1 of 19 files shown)


cmd/monaco/integrationtest/v2/test-resources/settings-ordered/insert-position/both-back-and-front-are-set/config.yaml
```diff
@@ -0,0 +1,27 @@
...
+# Second config, `last`, will be added to the back. The phantom reference enforces the deployment order
+
+configs:
+- id: first
+  config:
+    template: url-sampling.json
+  type:
+    settings:
+      schema: builtin:url-based-sampling
+      scope: environment
```
Laubi, line 13: add insertAfter: FRONT

pkg/deploy/internal/setting/setting.go
```diff
@@ -95,6 +100,24 @@ func Deploy(ctx context.Context, settingsClient client.SettingsClient, propertie
 
 }
 
+// parseInsertAfter finds and parses the `insertAfter parameter.
+//
+// If the parameter is not set, [dtclient.InsertPositionFront] is returned to indicate that the settings object should be added to the front of all configs.
+// This behavior is to stay compatible with earlier version
```
Laubi, line 106: Add the behavior to the docs about nil values, ordered schemas, etc

pkg/deploy/internal/setting/setting.go
```diff
@@ -95,6 +100,24 @@ func Deploy(ctx context.Context, settingsClient client.SettingsClient, propertie
...
+func parseInsertAfter(properties parameter.Properties) (*string, error) {
```
Laubi, line 107: `getAndParseInsertAfter`

Laubi (commented):
Default behavior should not be to push it to the front, but keep it in the back

Laubi:
Closed as I will split it up in two separate PRs

PR #1710, opened 2025-02-18
refactor: Use t.Context() in integration tests

Laubi:
Some of our e2e tests failed in the recent builds, and we are not able to pinpoint the issue. Locally, the issues do not appear, only on the GitHub pipelines.

However, we were able to deduce some information:
1) It has been happening since the Go1.24 upgrade (note: It is not proven that that is really the issue, or if it just did not appear before that)
2) We recently fixed `createMemStatFile` to be able to cancel on context-cancel. 

This PR introduces the following changes:
1) All e2e runs now use the same method (`monaco.Run`)
2) `monaco.Run` takes `*testing.T` as a parameter
3) Update the calling code to use the context.

This hopefully resolves the issue in future runs. At worst, it doesn't do that, but we have a more streamlined interface and use the context properly, which is still a win.


cmd/monaco/integrationtest/utils/monaco/cmd.go
```diff
@@ -17,49 +19,39 @@
...
-	c = strings.TrimPrefix(c, "monaco ")
+	t.Logf("Running command: %s", command)
+	c = strings.TrimPrefix(c, prefix)
 
 	args := strings.Split(c, " ")
 
 	cmd := runner.BuildCmd(fs)
 	cmd.SetArgs(args)
-	return runner.RunCmd(context.TODO(), cmd)
+	return runner.RunCmd(t.Context(), cmd)
```
Laubi, line 56: Let's manually cancel this context as well

tomazpu (approved):

cmd/monaco/integrationtest/utils/monaco/cmd.go
```diff
@@ -21,45 +23,40 @@ import (
...
+var spacesRegex = regexp.MustCompile(`\s+`)
 
-func RunWithFSf(fs afero.Fs, command string, args ...any) error {
-	return RunWithFs(fs, fmt.Sprintf(command, args...))
-}
-
-func RunWithFs(fs afero.Fs, command string) error {
+// Run is the entrypoint to run monaco for all integration tests.
+// It requires to specify the full command (`monaco [deploy]....`) and sets up the runner.
+func Run(t *testing.T, fs afero.Fs, command string) error {
```
arthurpitman, line 40: Question: do we need to tie this function to `t *testing.T` or would `context.Context` be a better choice? The latter would require a different log line inside but nothing more (I think). Another option might be to go for `*testing.TB`, although this would be a simple change to do in a follow-up if we need it.

Laubi: Context would be enough for sure, having it in also does not hurt IMO, marking the function explicitly as 'testing' function.

arthurpitman (approved):

cmd/monaco/integrationtest/utils/monaco/cmd.go
arthurpitman: Let's leave it is as is then 👍

PR #1712, opened 2025-02-19
refactor(clients): rename dummy clients to dry-run clients

tomazpu:
To prevent future confusion over the dummy client. I am renaming it to dry run client as agreed.


arthurpitman (changes requested):
I think we should not use dry run clients in tests that don't test the dry run. Could we separate these?

Laubi (dismissed):

PR #1713, opened 2025-02-19
refactor(idutils): rename GenerateExternalIDForDocument to GenerateExternalID

tomazpu:
As the `GenerateExternalIDForDocument` is being used also in other API than Documents, it was renamed to `GenerateExternalID`

### DIFF
--- a/internal/idutils/external_id.go
+++ b/internal/idutils/external_id.go
@@ -19,6 +19,7 @@ package idutils
 import (
 	"encoding/base64"
 	"fmt"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/coordinate"
 )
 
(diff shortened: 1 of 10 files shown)


internal/idutils/external_id.go
```diff
@@ -51,8 +52,8 @@ func GenerateExternalIDForSettingsObject(c coordinate.Coordinate) (string, error
 
 type ExternalIDGenerator func(coordinate.Coordinate) (string, error)
 
-// GenerateExternalIDForDocument generates an external ID for a document configuration. It is under 50 characters long and uses at most only "a-z", "A-Z", "0-9" and "-".
-func GenerateExternalIDForDocument(c coordinate.Coordinate) (string, error) {
+// GenerateExternalID generates an external ID for a document configuration. It is under 50 characters long and uses at most only "a-z", "A-Z", "0-9" and "-".
```
arthurpitman, line 55:
```suggestion
// GenerateExternalID generates an external ID for a configuration. It is under 50 characters long and uses at most only "a-z", "A-Z", "0-9" and "-".
```

arthurpitman, line 55: Just an idea: As the function takes the coordinate, could it absorb the function above for settings and then provide the correct externalID after examining the config type?

tomazpu: I like the idea, lets align at the daily :)

Laubi: Kind of like that as well, though we do increase cohesion 🤔

Laubi (approved):

arthurpitman (approved):

PR #1714, opened 2025-02-19
feat(settings): Allow insertAfter to contain more than just references.

Laubi:

Allows defining value, environment, and other values.\
Additionally moves the `insertAfter` fetching into Deploy.\
Adds minor change to add parameters to a interface method, as they were not clear.

This is the first PR of two. 
1) This one allows users to define more than just references while 
2) the next one will allow users to define magic values to insert at the begin and end of a list.

Yes, users will be able to define more than just References in their `insertAfter` settings.

### DIFF
--- a/pkg/deploy/deploy.go
+++ b/pkg/deploy/deploy.go
@@ -21,14 +21,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/internal/segment"
-
-	coreapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
-	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
-
 	gonum "gonum.org/v1/gonum/graph"
 	"gonum.org/v1/gonum/graph/simple"
 
+	coreapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/dynatrace"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/environment"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
@@ -45,6 +42,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/internal/classic"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/internal/document"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/internal/openpipeline"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/internal/segment"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/internal/setting"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/internal/slo"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/internal/validate"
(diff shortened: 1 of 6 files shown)


pkg/deploy/internal/setting/setting_test.go
```diff
@@ -151,7 +151,7 @@ func TestDeploySetting_ManagementZone_FailToDecodeMZoneID(t *testing.T) {
 		Parameters: testutils.ToParameterMap(parameters),
 	}
 	props := map[string]interface{}{"scope": "environment"}
-	resolvedEntity, err := Deploy(context.TODO(), c, props, "", conf, "")
+	resolvedEntity, err := Deploy(t.Context(), c, props, "", conf)
```
tomazpu, line 154: nit: I see you need to update the signature of the Deploy and are using the new `t.Context()` here but not in the changed methods on line 73, 85, 104, 1029. I suspect the others will be updated in a separate PR?

pkg/persistence/config/loader/config_entry_loader.go
```diff
@@ -229,20 +228,12 @@ func getConfigFromDefinition(
...
-		}
-
-		r, isRef := insertAfterParam.(*reference.ReferenceParameter)
-		if !isRef {
-			return config.Config{}, []error{fmt.Errorf("failed to parse insertAfter: cannot use parameter-type %q. Allowed types: %v", insertAfterParam.GetType(), reference.ReferenceParameterType)}
-		}
-
-		if r.Property != "id" {
-			return config.Config{}, []error{fmt.Errorf("failed to parse insertAfter: property field of reference parameter %q must be %q", insertAfterParam, "id")}
+			return config.Config{}, []error{fmt.Errorf("failed to parse insertAfter parameter: %w", err)}
```
tomazpu, line 236: There is missing coverage on this error, would be great if you could add test case to cover this error.

pkg/persistence/config/loader/config_loader_test.go
```diff
@@ -549,11 +591,32 @@ configs:
...
+						Type:     "builtin:profile.test",
+						ConfigId: "profile-id",
+					},
+					Type: config.SettingsType{
+						SchemaId:      "builtin:profile.test",
+						SchemaVersion: "1.0",
+					},
+					Template: template.NewInMemoryTemplate("profile.json", "{}"),
+					Parameters: config.Parameters{
+						config.NameParameter:        &value.ValueParameter{Value: "Star Trek > Star Wars"},
```
tomazpu, line 610: Star Trek < Star Wars

Laubi: This is factually very wrong. Star Trek is Star Wars light years ahead.

pkg/deploy/internal/setting/setting.go
```diff
@@ -34,12 +34,17 @@ import (
...
 	t, ok := c.Type.(config.SettingsType)
 	if !ok {
 		return entities.ResolvedEntity{}, errors.NewConfigDeployErr(c, fmt.Sprintf("config was not of expected type %q, but %q", config.SettingsTypeID, c.Type.ID()))
 	}
 
+	var insertAfter string
+	if ia, ok := properties[config.InsertAfterParameter]; ok {
+		insertAfter = ia.(string)
+	}
+
```
arthurpitman, line 47: What about extracting this into a function in `pkg/deploy/internal/extract/configname.go`, _renaming that file_, and checking the type, like for `extract.Scope(...)` below?

pkg/deploy/deploy.go
```diff
@@ -23,6 +23,7 @@ import (
 
 	coreapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/internal/segment"
```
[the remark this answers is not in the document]
Laubi: Error, fixed, but changed it in a way that the ordering is as intended.

pkg/deploy/internal/setting/setting_test.go
Laubi: Since we have multiple PRs open changing the context in tests, I would do minimal changes right now. I will even revert this change, and we swoop in later to properly do it.

pkg/persistence/config/loader/config_entry_loader.go
Laubi: Thanks for pointing out, I  added a test

pkg/deploy/internal/setting/setting.go
Laubi: I am happy to create a function, but since it's not shared and only used in settings, I would not move it to a shared package. 
Both `extract.Scope` and `extract.ConfigName` are used in more than one config deploy.

Laubi: Additionally, the functionality will change again in the followup PR :)

arthurpitman: As we now have it in a function, I would be in favor of adding a type check for the cast to string.

tomazpu (approved):

pkg/deploy/internal/setting/setting.go
Laubi: We discussed it offline, and we will add the type checking in the followup PR.

PR #1722, opened 2025-02-19
chore: pass correct context

Kirdock:
Passes the correct context to account related functions


jskelin (approved):
lgtm

PR #1723, opened 2025-02-19
refactor(segment): fix some typos and formatting

jskelin:
This PR corrects a few typos and aligned names with go standard


Kirdock (approved):

pkg/delete/internal/segment/delete.go
```diff
@@ -89,20 +89,20 @@ func findEntryWithExternalID(ctx context.Context, c client, dp pointer.DeletePoi
 		return "", fmt.Errorf("unable to generate externalID: %w", err)
 	}
 
-	var foundUid []string
+	var found []string
```
Laubi, line 92: Since `found` is an array, I would specify what we found. 
I agree, `foundUuid` is wrong, but I still think it gives more context -> `foundUUIDs` would be better

Laubi (approved):

PR #1725, opened 2025-02-19
chore(integration): pass right context

Kirdock:
Passes the correct context in integration test related functions

### DIFF
--- a/cmd/monaco/integrationtest/v2/delete_integration_test.go
+++ b/cmd/monaco/integrationtest/v2/delete_integration_test.go
@@ -20,17 +20,16 @@ package v2
 
 import (
 	"fmt"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"path/filepath"
 	"testing"
 
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"golang.org/x/net/context"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/integrationtest"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/integrationtest/utils/monaco"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/testutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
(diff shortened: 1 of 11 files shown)


cmd/monaco/integrationtest/v2/delete_integration_test.go
```diff
@@ -258,7 +258,7 @@ environmentGroups:
 	clientSet := integrationtest.CreateDynatraceClients(t, env)
 
 	// check the setting was deleted
-	integrationtest.AssertSetting(t, context.TODO(), clientSet.SettingsClient, config.SettingsType{SchemaId: "builtin:tags.auto-tagging"}, envName, false, config.Config{
+	integrationtest.AssertSetting(t, t.Context(), clientSet.SettingsClient, config.SettingsType{SchemaId: "builtin:tags.auto-tagging"}, envName, false, config.Config{
```
jskelin, line 261: Is it possible to get context from the `t` when it's needed inside the `AssertSetting` function?

NOTE: the unofficial convention is to have context as the first argument.

jskelin (changes requested):
1. Is it possible in `integrationtest.AssertSetting` to get the context from the `t` (which is already and argument), rather than passing it as the second argument?
2. `cmd.Run` (and all variations) are meant to use in integration test. For that reason, context should be same as the one used in a test.

cmd/monaco/integrationtest/v2/delete_integration_test.go
Kirdock: I decided not to use the t.Context inside because we have another place (`assert.go`) where we are adding some values to the context. Should I use a pointer context instead and forward nil if it's not needed?

cmd/monaco/integrationtest/utils/monaco/cmd.go
```diff
@@ -61,5 +60,5 @@ func RunWithFs(fs afero.Fs, command string) error {
 
 	cmd := runner.BuildCmd(fs)
 	cmd.SetArgs(args)
-	return runner.RunCmd(context.TODO(), cmd)
+	return runner.RunCmd(cmd.Context(), cmd)
```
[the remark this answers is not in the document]
Kirdock: Because there is some refactoring ongoing in this file, I decided to keep the changes here minimal, else we would do the changes twice
https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1710/files#diff-653abb875d042d23f442eb5e19a9f33c2e2cd1d50f17bd3d5c8386f1ba6d9f3b

cmd/monaco/integrationtest/v2/delete_integration_test.go
Laubi: It should be possible to pass `testing.T` here. If not, we need to adapt the testing code.

Laubi: I looked into the code, and the ctx is used for logging - IMO, the client (settings client) should not be aware of logging, but I see why it is currently used that way. 

Let's keep it that way atm

cmd/monaco/integrationtest/utils/monaco/cmd.go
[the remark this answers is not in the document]
Laubi: It's ok, I think having the correct context here makes sense. Who knows when I will merge the followup :D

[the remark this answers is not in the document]
Laubi: Klaus and I had a look at the PRs and code underneath. 
We discussed the best way forward is the following approach: 
1) We revert this change, and still use context.TODO in the tests - this PR changes the other contexts. 
2) My PR #1710 then will do the proper refactoring.

@jskelin are you ok with that?

cmd/monaco/integrationtest/v2/delete_integration_test.go
jskelin: In the call, we agreed to investigate the possibility of moving context creation inside the AssertSettings and AssertConfig.

Kirdock: Done 👍

PR #1727, opened 2025-02-20
chore(jenkins): update security context of container

Kirdock:
Updates the security context, so there aren't any admin/root rights


.ci/jenkins_agents/ca-jenkins-agent.yaml
```diff
@@ -2,16 +2,29 @@ apiVersion: v1
 kind: Pod
 spec:
   containers:
-    - name: jnlp # nosemgrep: yaml.kubernetes.security.privileged-container.privileged-container
-      image: registry.lab.dynatrace.org/keptn/ca-jenkins-agent
+    - name: ca-jenkins-agent
+      image: golang:1.24.0-bookworm
```
Laubi, line 6: @Laubi discuss with @stefaneberl how we can appraoch this.

PR #1728, opened 2025-02-20
feat(account): Delete service users

arthurpitman:
This PR adds the ability to delete service users from accounts.
This PR does the bare minimum to implement the functionality, we should do some follow ups to reduce code duplication etc.

### DIFF
--- a/pkg/account/delete/client.go
+++ b/pkg/account/delete/client.go
@@ -70,8 +70,9 @@ func (c *AccountAPIClient) DeleteUser(ctx context.Context, email string) error {
 	return nil
 }
 
-// DeleteServiceUser removes the service user with the given name from the account.
-// Returns error if any API call fails unless the user is already not present (HTTP 404)
+// DeleteServiceUser retrieves all service users, looks up the service user by name and removes it from the account.
+// Returns error if any API call fails unless the user is already not present, either in the list when looking by name or during the subsequent delete call.
+// In addition, an error is returned if multiple service users are found with the same name.
 func (c *AccountAPIClient) DeleteServiceUser(ctx context.Context, name string) error {
 	uid, err := c.getServiceUserIDByName(ctx, c.accountUUID, name)
 	if err != nil {
--- a/pkg/account/delete/client.go
+++ b/pkg/account/delete/client.go
@@ -52,7 +52,7 @@ func NewAccountAPIClient(accountUUID string, restClient *accounts.Client) Client
 	}
 }
 
-// NotFoundError is an error signifying that the desired resource was not found. Generally this error can be treated as a successful "deletion" of the resource.
+// ResourceNotFoundError is an error signifying that the desired resource was not found. Generally this error can be treated as a successful "deletion" of the resource.
 type ResourceNotFoundError struct {
 	Identifier string
 }
--- a/pkg/account/delete/client.go
+++ b/pkg/account/delete/client.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math"
 	"net/http"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/accounts"
@@ -120,8 +119,9 @@ func (c *AccountAPIClient) getServiceUserIDByName(ctx context.Context, accountUU
 
 func (c *AccountAPIClient) getServiceUsers(ctx context.Context, accountUUID string) ([]accountmanagement.ExternalServiceUserDto, error) {
 	serviceUsers := []accountmanagement.ExternalServiceUserDto{}
-	const pageSize = 100
-	for page := (int32)(1); page < math.MaxInt32; page++ {
+	const pageSize = 1000
+	page := (int32)(1)
+	for {
 		r, err := c.getServiceUsersPage(ctx, accountUUID, page, pageSize)
 		if err != nil {
 			return nil, err
@@ -132,6 +132,7 @@ func (c *AccountAPIClient) getServiceUsers(ctx context.Context, accountUUID stri
 		if r.NextPageKey == nil {
 			break
 		}
+		page++
 	}
 
 	return serviceUsers, nil
--- a/pkg/account/delete/client.go
+++ b/pkg/account/delete/client.go
@@ -21,15 +21,18 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/clients/accounts"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+	accountmanagement "github.com/dynatrace/dynatrace-configuration-as-code-core/gen/account_management"
 )
 
 // Client for deleting resources from the Account Management API
 type Client interface {
 	DeleteUser(ctx context.Context, email string) error
+	DeleteServiceUser(ctx context.Context, name string) error
 	DeleteGroup(ctx context.Context, name string) error
 	DeleteAccountPolicy(ctx context.Context, name string) error
 	DeleteEnvironmentPolicy(ctx context.Context, environment, name string) error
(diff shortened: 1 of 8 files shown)


pkg/account/delete/loader.go
```diff
@@ -126,16 +140,17 @@ func parseDeleteFileDefinition(definition FileDefinition) (Resources, error) {
 			case "environment":
 				environmentPolicies = append(environmentPolicies, EnvironmentPolicy{Name: parsed.Name, Environment: parsed.Level.Environment})
 			default:
-				return Resources{}, newDeleteEntryParserError(fmt.Sprintf("%v", e), i, fmt.Sprintf(`unkown policy level %q - needs to be one of "account","environment"`, parsed.Level))
+				return Resources{}, newDeleteEntryParserError(fmt.Sprintf("%v", e), i, fmt.Sprintf(`unknown policy level %q - needs to be either "account" or "environment"`, parsed.Level))
 			}
 		default:
-			return Resources{}, newDeleteEntryParserError(fmt.Sprintf("%v", e), i, fmt.Sprintf(`unkown type %q - needs to be one of "user","group","policy"`, parsed.Type))
+			return Resources{}, newDeleteEntryParserError(fmt.Sprintf("%v", e), i, fmt.Sprintf(`unknown type %q - needs to be one of "user", "service-user", "group", "policy"`, parsed.Type))
```
Kirdock, line 146: We should not include `service-user` in here. On the one hand it's behind a feature flag and on the other hand we are always printing that service-user may not be set

pkg/account/delete/delete.go
```diff
@@ -60,55 +67,116 @@ func (a Account) String() string {
 	return fmt.Sprintf("%q (%s)", a.Name, a.UUID)
 }
 
-// AccountResources removes all given Resources from the given Account
+// DeleteAccountResources removes all given Resources from the given Account
 // Returns an error if any resource fails to be deleted, but attempts to delete as many resources as possible and only returns an error at the end.
-func AccountResources(ctx context.Context, account Account, resourcesToDelete Resources) error {
+func DeleteAccountResources(ctx context.Context, account Account, resourcesToDelete Resources) error {
```
Kirdock, line 72: Functions should not start with the package name, because it would be redundant `delete.Delete...`

pkg/account/delete/loader.go
arthurpitman: Good catch :)
I will remove `service-user` for the  moment, but maybe we remove the examples from the message - better to encourage people to read the docs.

pkg/account/delete/delete.go
arthurpitman: I agree - but would change the package structure. I've opened a `tech-time` about this, so maybe we can come up with something better.

pkg/account/delete/client.go
```diff
@@ -68,6 +71,79 @@ func (c *AccountAPIClient) DeleteUser(ctx context.Context, email string) error {
...
+		return "", NotFoundErr
+	}
+
+	return uid, nil
+}
+
+func (c *AccountAPIClient) getServiceUsers(ctx context.Context, accountUUID string) ([]accountmanagement.ExternalServiceUserDto, error) {
+	serviceUsers := []accountmanagement.ExternalServiceUserDto{}
+	const pageSize = 100
+	for page := (int32)(1); page < math.MaxInt32; page++ {
```
jskelin, line 119: I don't see the benefit of this extra caution.
In case the Dynatrace is broken, it's more likely that the anti-DoS guardian will activate and the request will fail.
Even when the whole loop succeeds, and for loop bump in the upper limit, this code doesn't return an error. Instead, it pretends everything is OK, giving a false positive appearance.

On the other hand, it brings unwanted complexity to the code. And make it harder to understand.

Kirdock (commented):
Currently, for me it's not working.
If a service user doesn't have a description, an error occurs while fetching the service-users but this error is discarded somewhere in between. The result is then treated as empty, which is wrong

pkg/account/delete/client.go
```diff
@@ -50,23 +52,104 @@ func NewAccountAPIClient(accountUUID string, restClient *accounts.Client) Client
 	}
 }
 
-// NotFoundErr is a sentinel error signifying that the resource desired to be deleted was not found. Generally this error can be treated as a succeful "deletion" of the resource.
-var NotFoundErr = errors.New("nothing with given name found")
+// NotFoundError is an error signifying that the desired resource was not found. Generally this error can be treated as a successful "deletion" of the resource.
```
tomazpu, line 55: Comment not matching struct name after renaming the struct

pkg/account/delete/client_test.go
```diff
@@ -117,8 +118,57 @@ func TestAccountAPIClient_DeleteGroup(t *testing.T) {
...
 
 		err = accountClient.DeleteGroup(t.Context(), "test-group")
-		assert.ErrorIs(t, err, delete.NotFoundErr)
+
+		notFoundErr := &delete.ResourceNotFoundError{}
+		assert.ErrorAs(t, err, &notFoundErr)
 	})
+
+	t.Run("successful delete of service user with no description", func(t *testing.T) {
+		server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
```
Kirdock, line 127: [nit]: maybe it makes sense to have it like in `settings_client_test.go` with mux and handleFun in here and other functionsc. Would be easier to read and we would not need the switch.

pkg/account/delete/client.go
```diff
@@ -50,23 +52,104 @@ func NewAccountAPIClient(accountUUID string, restClient *accounts.Client) Client
...
+		return "", err
+	}
+
+	uid := ""
+	for _, s := range serviceUsers {
+		if s.Name == name {
+			if uid != "" {
+				return "", fmt.Errorf("found multiple service users with name '%s'", name)
+			}
+			uid = s.Uid
```
tomazpu, line 110: Could you return already here? Simplifies the method return at the bottom.

pkg/account/delete/client_test.go
```diff
@@ -117,8 +118,57 @@ func TestAccountAPIClient_DeleteGroup(t *testing.T) {
 		accountClient := delete.NewAccountAPIClient("1234", accounts.NewClient(restClient))
 
 		err = accountClient.DeleteGroup(t.Context(), "test-group")
-		assert.ErrorIs(t, err, delete.NotFoundErr)
+
+		notFoundErr := &delete.ResourceNotFoundError{}
+		assert.ErrorAs(t, err, &notFoundErr)
 	})
+
+	t.Run("successful delete of service user with no description", func(t *testing.T) {
```
Kirdock, line 126: We should move this one below to the service user test. In here we have the delete group tests. We can even replace the `successful delete` test with this one

pkg/account/delete/delete.go
```diff
@@ -60,55 +67,121 @@ func (a Account) String() string {
...
+	totalErrorCount := deleteUsers(ctx, account, resourcesToDelete.Users) +
```
tomazpu, line 73: Could the delete calls be done here in Parralel?

pkg/account/delete/client.go
arthurpitman: An earlier return is not possible because we must check for multiple matches (which should produce an error a couple of lines above)

pkg/account/delete/client.go
arthurpitman: Fixed 👍

pkg/account/delete/delete.go
```diff
@@ -60,55 +67,121 @@ func (a Account) String() string {
...
-			log.Error("Failed to delete group %q from account %s: %v", group.Name, account, err)
-			deleteErrors++
-		} else {
+	return errCount
+}
+
+func deleteServiceUsers(ctx context.Context, account Account, serviceUsers []ServiceUser) int {
+	errCount := 0
+	for _, user := range serviceUsers {
+		err := account.APIClient.DeleteServiceUser(ctx, user.Name)
```
Kirdock, line 109: Should we maybe refactor this one a bit? Currently, it fetches all available service users for each entry in the service users that should be deleted

pkg/account/delete/delete.go
arthurpitman: I'm not sure, let's save that for a follow up

tomazpu (dismissed):

pkg/account/delete/delete.go
arthurpitman: I have added this to our follow ups 👍

pkg/account/delete/client_test.go
arthurpitman: Added to follow-ups

pkg/account/delete/client_test.go
arthurpitman: Good catch, not sure how it ended up there 😅 . I would keep both tests for the moment.

Kirdock (approved):

PR #1729, opened 2025-02-20
refactor(JSONObject): extend JSONObject.Delete() to delete multiple entries

jskelin:
This PR brings small improvement which enables deleting multiple entries at once.
Instead current
```GO
object.Delete("key1")
object.Delete("key2")
```
it is possible to: 
```GO
object.Delete("key1", "key2")
```

### DIFF
--- a/pkg/download/internal/templatetools/template.go
+++ b/pkg/download/internal/templatetools/template.go
@@ -18,6 +18,7 @@ package templatetools
 
 import (
 	"encoding/json"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter/value"
 )
 
@@ -65,10 +66,11 @@ func (o JSONObject) ToJSON(pretty bool) ([]byte, error) {
 		bytes, err = json.Marshal(o)
 	}
 	return bytes, err
-
 }
 
 // Delete removes a key-value pair for the specified key from JSONObject.
-func (o JSONObject) Delete(key string) {
-	delete(o, key)
+func (o JSONObject) Delete(keys ...string) {
+	for _, k := range keys {
+		delete(o, k)
+	}
 }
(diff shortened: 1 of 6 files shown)


pkg/download/internal/templatetools/template.go
```diff
@@ -18,13 +18,14 @@ package templatetools
...
 	"encoding/json"
+
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter/value"
 )
 
 type JSONObject map[string]any
 
 // NewJSONObject is a function that creates a JSONObject from a raw JSON byte slice.
-func NewJSONObject(raw []byte) (JSONObject, error) {
+func NewJSONObject(raw json.RawMessage) (JSONObject, error) {
```
Laubi, line 28: I am not sure why you changed this here to `json.RawMessage`.
Without this change, I would've approved the PR immediately, as it's straightforward enough. 

Since `json.RawMessage` is also not a type alias but a real type, I am really not sure if we should introduce it everywhere 🤔

arthurpitman: I agree - is there a need for this change? I believe the reason for `json.RawMessage` is be able to implement `Marshaler` and `Unmarshaler`.

arthurpitman (commented):
Overall this is stylistic:

- The current version,

  ```
  r.Delete(bucketName)
  r.Delete(status)
  r.Delete(version)
  r.Delete(updatable)
  ```
  is easier to read.

- The proposed version `r.Delete(bucketName, status, version, updatable)` is less verbose.

pkg/download/internal/templatetools/template.go
jskelin: The usage of `json.RawMessage` emphasizes that the input should be JSON.

Could you explain the problems that defining an argument as `json.RawMessage` causes?

Laubi (approved):

arthurpitman (approved):

PR #1731, opened 2025-02-20
chore(core): update core and pass context

Kirdock:
Updates core and sets the context that is now needed


arthurpitman (approved):
I will approve and merge this. I don't think the failing E2E test `Delete sub path API configurations` is related to this, but will investigate it separately.

PR #1732, opened 2025-02-21
feat(slo-v2): add purge functionality

jskelin:
This PR brings purge functionality for SLO.

### DIFF
--- a/pkg/delete/internal/slo/delete.go
+++ b/pkg/delete/internal/slo/delete.go
@@ -122,6 +122,27 @@ func isAPIErrorStatusNotFound(err error) bool {
 	return apiErr.StatusCode == http.StatusNotFound
 }
 
+func DeleteAll(ctx context.Context, c client) error {
+	items, err := c.List(ctx)
+	if err != nil {
+		return err
+	}
+
+	var retErr error
+	for _, i := range items.All() {
+		var e entry
+		if err := json.Unmarshal(i, &e); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/delete/internal/slo/delete.go
```diff
@@ -122,6 +122,27 @@ func isAPIErrorStatusNotFound(err error) bool {
...
+		if err := json.Unmarshal(i, &e); err != nil {
+			retErr = errors.Join(retErr, err)
+			continue
+		}
+		err := deleteSingle(ctx, c, pointer.DeletePointer{Type: string(config.ServiceLevelObjectiveID), OriginObjectId: e.ID})
+		if err != nil {
+			retErr = errors.Join(retErr, err)
+		}
+	}
+	return retErr
```
arthurpitman, line 143: I think we can build a better error with no extra lines of code:
```suggestion
	var errs []error
	for _, i := range items.All() {
		var e entry
		if err := json.Unmarshal(i, &e); err != nil {
			errs = append(errs, err)
			continue
		}
		err := deleteSingle(ctx, c, pointer.DeletePointer{Type: string(config.ServiceLevelObjectiveID), OriginObjectId: e.ID})
		if err != nil {
			errs = append(errs, err)
		}
	}
	return errors.Join(errs...)
```

jskelin: Could you please explain it more?

arthurpitman (approved):

PR #1734, opened 2025-02-21
feat: Optional escaping file parameter

arthurpitman:
This PR adds `escaped`, an optional boolean parameter, to the `file` parameter type. By default, `escaped` is set to `true` to be backwards compatible, but can be set to `false` to disable the escaping of file parameter content.


tomazpu (approved):

cmd/monaco/integrationtest/v2/test-resources/special-character-in-config/project/synthetic-monitor/config.yaml
```diff
@@ -3,6 +3,10 @@ configs:
   config:
     name: Dynatrace Homepage Check
     parameters:
+      request:
+        type: file
+        path: request.json
+        escaped: false
       ua_string: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML,
```
Kirdock, line 10: Currently, this one is still working because nested templating is supported by default. Should we still change it and move it to "request"?

pkg/config/parameter/file/file.go
```diff
@@ -59,8 +62,11 @@ func (f *FileParameter) ResolveValue(context parameter.ResolveContext) (interfac
 	if err != nil {
 		return nil, parameter.NewParameterResolveValueError(context, err.Error())
 	}
+	if f.Escaped {
```
Kirdock, line 65: just to understand it correctly `Escaped` means that it needs to be escaped? If yes, we should probably rename it because from the naming I would assume that it is already escaped

cmd/monaco/integrationtest/v2/test-resources/special-character-in-config/project/synthetic-monitor/config.yaml
arthurpitman: Going to check this out, I think actually this should not work as is, so something is wrong here 🤔

pkg/config/parameter/file/file.go
arthurpitman: This is a good point: @Laubi , what do you think is the clearest name?

pkg/config/parameter/file/file.go
```diff
@@ -73,9 +79,18 @@ func parseFileValueParameter(context parameter.ParameterParserContext) (paramete
...
+		if !ok {
+			return nil, parameter.NewParameterParserError(context, "property `escaped` must be a boolean")
+		}
+		escaped = escapedBool
+	}
+
 	references, ok := context.Value["references"]
 	if !ok {
-		return &FileParameter{Fs: context.Fs, Path: strings.ToString((path))}, nil
+		return &FileParameter{Fs: context.Fs, Path: strings.ToString((path)), Escaped: escaped}, nil
```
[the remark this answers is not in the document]
arthurpitman: Should now be fixed 👍

pkg/config/parameter/file/file.go
```diff
@@ -88,7 +103,7 @@ func parseFileValueParameter(context parameter.ParameterParserContext) (paramete
 		return nil, parameter.NewParameterParserError(context, fmt.Sprintf("invalid parameter references: %v", err))
 	}
 
-	return &FileParameter{Fs: context.Fs, Path: strings.ToString((path)), referencedParameters: referencedParameters}, nil
+	return &FileParameter{Fs: context.Fs, Path: strings.ToString((path)), Escaped: escaped, referencedParameters: referencedParameters}, nil
```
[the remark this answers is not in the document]
arthurpitman: Should now be fixed 👍

pkg/config/parameter/file/file.go
arthurpitman: Consensus:  let's go with `escape`

cmd/monaco/integrationtest/v2/test-resources/special-character-in-config/project/synthetic-monitor/config.yaml
arthurpitman: I have updated `request` to include the missing `references` section. However, currently the file parameter doesn't enforce this. I will tackle that in a follow up PR.

arthurpitman: Update: this is now fixed: the file parameter enforces it.

tomazpu (approved):

PR #1735, opened 2025-02-21
test(slo-v2): add e2e tests

jskelin:
This PR adds e2e tests for slo-v2 configs

### DIFF
--- a/cmd/monaco/integrationtest/v2/all_configs_integration_test.go
+++ b/cmd/monaco/integrationtest/v2/all_configs_integration_test.go
@@ -38,6 +38,7 @@ func TestIntegrationAllConfigsClassic(t *testing.T) {
 	manifest := configFolder + "manifest.yaml"
 
 	t.Setenv(featureflags.OpenPipeline.EnvName(), "true")
+	t.Setenv(featureflags.ServiceLevelObjective.EnvName(), "true")
 
 	targetEnvironment := "classic_env"
 
@@ -55,6 +56,7 @@ func TestIntegrationAllConfigsPlatform(t *testing.T) {
 	manifest := configFolder + "manifest.yaml"
 
 	t.Setenv(featureflags.OpenPipeline.EnvName(), "true")
+	t.Setenv(featureflags.ServiceLevelObjective.EnvName(), "true")
 
 	targetEnvironment := "platform_env"
 
@@ -89,6 +91,7 @@ func runDeployCommand(t *testing.T, fs afero.Fs, manifest, specificEnvironment s
 func TestIntegrationValidationAllConfigs(t *testing.T) {
 	t.Setenv("UNIQUE_TEST_SUFFIX", "can-be-nonunique-for-validation")
 	t.Setenv(featureflags.OpenPipeline.EnvName(), "true")
+	t.Setenv(featureflags.ServiceLevelObjective.EnvName(), "true")
 
 	fs := afero.NewCopyOnWriteFs(afero.NewOsFs(), afero.NewMemMapFs())
 
--- a/cmd/monaco/integrationtest/v2/all_configs_integration_test.go
+++ b/cmd/monaco/integrationtest/v2/all_configs_integration_test.go
@@ -37,9 +37,6 @@ func TestIntegrationAllConfigsClassic(t *testing.T) {
 	configFolder := "test-resources/integration-all-configs/"
 	manifest := configFolder + "manifest.yaml"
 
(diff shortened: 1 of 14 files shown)


Laubi (changes requested):
Please add

- dedicated tests for SLOs
- Add test cases for the `config_restore_e2e_test.go` (they are deployed, downloaded, deployed again)

cmd/monaco/integrationtest/v2/slo-v2_integration_test.go
```diff
@@ -0,0 +1,135 @@
...
+			assert.NotContains(t, externalIDs, externalID)
+		})
+	})
+}
+
+func parseExternalIDs(t *testing.T, response api.PagedListResponse) []string {
+	allObjects := response.All()
+	externalIds := make([]string, 0, len(allObjects))
+	for _, obj := range allObjects {
+		externalIds = append(externalIds, parseSloV2Payload(t, obj).ExternalId)
```
Laubi, line 108: Since you only use the `ExternalId` at all, maybe `parseSlo..` should return it directly. E.g. `extractSloV2ExternalId(t, obj)`.

cmd/monaco/integrationtest/v2/all_configs_integration_test.go
```diff
@@ -38,6 +38,7 @@ func TestIntegrationAllConfigsClassic(t *testing.T) {
 	manifest := configFolder + "manifest.yaml"
 
 	t.Setenv(featureflags.OpenPipeline.EnvName(), "true")
+	t.Setenv(featureflags.ServiceLevelObjective.EnvName(), "true")
```
Laubi, line 41: I don't think this will do anything - same with the OpenPipeline a line above. 

Without looking deeper, I would think that `TestIntegrationAllConfigs**Classic**` does not use OAuth credentials, and thus the API is not available with API Tokens. 

Thus, I don't think those two lines are actually necessary

cmd/monaco/integrationtest/v2/test-resources/slo-v2/project/config.yaml
```diff
@@ -0,0 +1,16 @@
...
+- id: sli-reference
+  type: slo-v2
+  config:
+    name: sli-reference
+    template: './sli-reference.json'
+    parameters:
+      extractedIDs:
+        type: value
+        value:
+          id_HOST_DED360997F66F0C2: HOST-DED360997F66F0C2
```
Laubi, line 16: Please don't use real ME-ids. And since it's not important for the test, I would even just let it stay in the json template

cmd/monaco/integrationtest/v2/test-resources/integration-download-configs-platform/project/slo-v2/config.yaml
```diff
@@ -0,0 +1,16 @@
...
+- id: sli-reference
+  type: slo-v2
+  config:
+    name: sli-reference
+    template: './sli-reference.json'
+    parameters:
+      extractedIDs:
+        type: value
+        value:
+          id_HOST_DED360997F66F0C2: HOST-DED360997F66F0C2
```
Laubi, line 16: Same as the other comment

cmd/monaco/integrationtest/v2/slo-v2_integration_test.go
```diff
@@ -0,0 +1,135 @@
...
+			err := monaco.RunWithFSf(fs, "monaco deploy %s --project=%s --verbose", manifestPath, project)
+			assert.NoError(t, err)
+
+			err = monaco.RunWithFSf(fs, "monaco deploy %s --project=%s --verbose", manifestPath, project)
+			assert.NoError(t, err)
+
+			sloV2Client := createSloV2Client(t, fs, manifestPath, environment)
+			result, err := sloV2Client.List(t.Context())
+			assert.NoError(t, err)
+			externalIDs := parseExternalIDs(t, result)
```
Laubi, line 59: Maybe `extractExternalIDs`? Because you don't parse the externalIDs, you parse the json response to get the externalIds

Laubi (commented):
Looks solid, some minor comments

cmd/monaco/integrationtest/v2/all_configs_integration_test.go
Kirdock: Actually it's still needed. Here all conifgs are read, including that ones that are not really needed and an invalid type would result in an error

arthurpitman (approved):
LGTM

Laubi (approved):

PR #1736, opened 2025-02-23
test: use FeatureFlag instead of string to define wanted FF

jskelin:
This PR introduces small changes in the definition of e2e tests, which, in theory, should result in less error-prone coding.


jskelin:
> [!NOTE]
> Failing quality (function with eight arguments) in inherited.

tomazpu (approved):

tomazpu (approved):

PR #1739, opened 2025-02-24
refactor(reporter): Flush every 3 seconds

Kirdock:
The report was only flushed at the end but should be every 3 seconds instead.
In addition: The file did not close properly in case of an error during marshal or write

Should we add some sort of tests for this behavior or is the default case (that it is written) sufficient? Not sure how easily this can be tested/mocked

### DIFF
--- a/pkg/project/v2/project_loader.go
+++ b/pkg/project/v2/project_loader.go
@@ -115,7 +115,7 @@ func LoadProjects(ctx context.Context, fs afero.Fs, loaderContext ProjectLoaderC
 			errs = append(errs, loadProjectErrs...)
 			continue
 		}
-		reporter.ReportInfo(fmt.Sprintf("project %q loaded", project.String()))
+		reporter.ReportInfo(fmt.Sprintf("Project %q loaded", project.String()))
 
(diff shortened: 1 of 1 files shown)


pkg/report/reporter.go
```diff
@@ -110,30 +110,36 @@ func (d *defaultReporter) runRecorder(fs afero.Fs, reportFilePath string) error
 		return fmt.Errorf("error open record file: %w", err)
 	}
 
+	defer file.Close()
```
Kirdock, line 113: Is it ok to discard the error or should we handle it about the same like [here](https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/8ae60db1a59345d632b9c12f11b6192bbbeedb59/cmd/monaco/integrationtest/config_rewrite.go#L71-L88)?

Laubi: Let's just log the error

Laubi: As discussed offline: Let's move the file opening to the calling function and pass in a writer & flusher

Laubi (approved):

PR #1740, opened 2025-02-24
feat(settings): Insert at the very front and back

Laubi:
Followup to #1714

#### What this PR does / Why we need it

This change allows users to put settings objects to the very front/back of the list. 

To do that, they must specify either `FRONT` or `BACK` in their `insertAfter` configuration.

### DIFF
--- a/cmd/monaco/integrationtest/v2/settings_integration_test.go
+++ b/cmd/monaco/integrationtest/v2/settings_integration_test.go
@@ -42,6 +42,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/coordinate"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest"
+	manifestloader "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest/loader"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project/v2/sort"
 )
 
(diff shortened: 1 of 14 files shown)


cmd/monaco/integrationtest/v2/settings_integration_test.go
```diff
@@ -287,3 +294,161 @@ func createSettingsClient(t *testing.T, env manifest.EnvironmentDefinition, opts
...
+
+		first := settingsExternalIdForTest(t, coordinate.Coordinate{Project: project, Type: schema, ConfigId: "first"}, tc)
+		second := settingsExternalIdForTest(t, coordinate.Coordinate{Project: project, Type: schema, ConfigId: "second"}, tc)
+
+		assert.Equal(t, len(list)-2, findPositionWithExternalId(t, list, first))
+		assert.Equal(t, len(list)-1, findPositionWithExternalId(t, list, second))
+	})
+}
+
+//func TestOrdered_InsertAtFrontWorks(t *testing.T) {
```
Laubi, line 328: Remove/update test

pkg/client/dtclient/settings_client_test.go
```diff
@@ -44,7 +46,7 @@ func TestNewClassicSettingsClient(t *testing.T) {
 
 		client, err := NewClassicSettingsClient(corerest.NewClient(server.URL(), server.Client()))
 		assert.NoError(t, err)
-		assert.Equal(t, settingsSchemaAPIPathClassic, client.settingsSchemaAPIPath)
+		assert.Equal(t, "/api/v2/settings/schemas", client.settingsSchemaAPIPath)
```
Kirdock, line 49: What was the reason for replacing it with the actual value?
On the one side yes, tests should be rather static and not depend on such values but on the other side, why just this one and not the others like `settingsObjectAPIPathClassic`

cmd/monaco/integrationtest/v2/settings_integration_test.go
```diff
@@ -287,3 +294,161 @@ func createSettingsClient(t *testing.T, env manifest.EnvironmentDefinition, opts
...
+		list, err := sClient.List(t.Context(), schema, dtclient.ListSettingsOptions{
+			DiscardValue: true,
+		})
+
+		assert.GreaterOrEqual(t, len(list), 2, "At least the two configs in the test should be deployed")
+
+		first := settingsExternalIdForTest(t, coordinate.Coordinate{Project: project, Type: schema, ConfigId: "first"}, tc)
+		second := settingsExternalIdForTest(t, coordinate.Coordinate{Project: project, Type: schema, ConfigId: "second"}, tc)
+
+		assert.Equal(t, len(list)-2, findPositionWithExternalId(t, list, first))
```
Kirdock, line 323: Should we maybe add `Greater` instead (also for other tests)? This would then less interfere with other E2E in parallel.
```go
assert.Greater(t, findPositionWithExternalId(t, list, second), findPositionWithExternalId(t, list, first))
```

cmd/monaco/integrationtest/v2/settings_integration_test.go
```diff
@@ -287,3 +294,161 @@ func createSettingsClient(t *testing.T, env manifest.EnvironmentDefinition, opts
...
+
+		assert.GreaterOrEqual(t, len(list), 2, "At least the two configs in the test should be deployed")
+
+		// Verify that last is actually the first object
+		first := settingsExternalIdForTest(t, coordinate.Coordinate{Project: project, Type: schema, ConfigId: "first"}, tc)
+		assert.Equal(t, 0, findPositionWithExternalId(t, list, first))
+
+		// Verify that last is actually the last object
+		last := settingsExternalIdForTest(t, coordinate.Coordinate{Project: project, Type: schema, ConfigId: "last"}, tc)
+		assert.Equal(t, len(list)-1, findPositionWithExternalId(t, list, last))
```
Kirdock, line 417: Could interfere with other E2E in parallel but I don't know another solution other than checking it's not the first one

Kirdock (approved):

cmd/monaco/integrationtest/v2/settings_integration_test.go
Laubi: Yes, this can interfere, but most of the time only one e2e test is running.

If this is a problem, we could do the following: Use (or introduce) a schema that is ordered on multiple entities, and write in each test-run to another entity. This would separate the test runs

cmd/monaco/integrationtest/v2/settings_integration_test.go
Laubi: I am not a fan of that, as these asserts here are the core of the test. I expect them to be precisely where they are.

Kirdock (dismissed):

pkg/client/dtclient/settings_client.go
```diff
@@ -707,6 +722,31 @@ func buildPostRequestPayload(ctx context.Context, remoteObjectId string, obj Set
 	return dest.Bytes(), nil
 }
 
+// insertAfterToPayloadValue converts the insertAfter value to the propler
```
arthurpitman, line 725:
```suggestion
// insertAfterToPayloadValue converts the insertAfter value to the proper
```

pkg/client/dtclient/settings_client.go
```diff
@@ -707,6 +722,31 @@ func buildPostRequestPayload(ctx context.Context, remoteObjectId string, obj Set
...
+// value required in the payload.
+//
+// For POST (only method that we use), it must be as follows:
+//
+//   - insert to the front: `insertAfter` to ”
```
arthurpitman, line 730:
```suggestion
//   - insert to the front: `insertAfter` to ""
```

pkg/client/dtclient/settings_client.go
```diff
@@ -707,6 +722,31 @@ func buildPostRequestPayload(ctx context.Context, remoteObjectId string, obj Set
...
+		return nil
+	case InsertPositionFront:
+		var empty = ""
+		return &empty
+	default:
+		return insertAfter
+	}
+}
+
 // parsePostResponse unmarshalls and parses the settings response for the post request
```
arthurpitman, line 750: While I see it, considering the spelling of the function:
```suggestion
// parsePostResponse unmarshals and parses the settings response for the post request
```

pkg/client/dtclient/settings_client_test.go
```diff
@@ -960,7 +1069,7 @@ func TestUpsertSettingsFromCache(t *testing.T) {
 	numAPIGetCalls := 0
 	numAPIPostCalls := 0
 	server := httptest.NewTLSServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
-		if req.URL.Path == settingsSchemaAPIPathClassic+"/some:schema" {
+		if req.URL.Path == (settingsSchemaAPIPathClassic)+"/some:schema" {
```
arthurpitman, line 1072: I'm curious about the `(`  and the `)`

cmd/monaco/integrationtest/v2/test-resources/settings-ordered/insert-position/both-back-and-front-are-set/config.yaml
```diff
@@ -0,0 +1,24 @@
+# Test if setting `insertAfter` to `BACK` works as expected.
```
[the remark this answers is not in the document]
arthurpitman: Although would this test be better with three configs? One with neither front nor back?

cmd/monaco/integrationtest/v2/test-resources/settings-ordered/insert-position/insert-after-set-to-back/config.yaml
```diff
@@ -0,0 +1,27 @@
+# Test if setting `insertAfter` to `BACK` works as expected.
+#
+# First config, `first`, is deployed and will be added to the front
+# Second config, `back`, will be added to the back. The phantom reference enforces the deployment order
```
arthurpitman, line 4: 🤔 
```suggestion
# Second config, `second`, will be added to the back. The phantom reference enforces the deployment order
```

arthurpitman (approved):
LGTM

PR #1750, opened 2025-02-26
test: Both name and id rewrite functions should skip lines ending with `#monaco-test:no-replace`

arthurpitman:
This PR updates name and id rewrite functions to skip lines ending with `#monaco-test:no-replace`

### DIFF
--- a/cmd/monaco/integrationtest/rewrite_functions.go
+++ b/cmd/monaco/integrationtest/rewrite_functions.go
@@ -35,42 +35,46 @@ func GetAddSuffixFunction(suffix string) func(line string) string {
 }
 
 func ReplaceName(line string, idChange func(string) string) string {
+	if strings.HasSuffix(line, "#monaco-test:no-replace") {
+		return line
+	}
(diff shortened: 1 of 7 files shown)


cmd/monaco/integrationtest/rewrite_functions.go
```diff
@@ -35,42 +35,46 @@ func GetAddSuffixFunction(suffix string) func(line string) string {
 }
 
 func ReplaceName(line string, idChange func(string) string) string {
+	if strings.HasSuffix(line, "#monaco-test:no-replace") {
```
Kirdock, line 38: So if there is a space after the comment, this won't work, but both approaches are fine for me.

Kirdock (approved):

cmd/monaco/integrationtest/rewrite_functions.go
arthurpitman: That's by design - it must be exactly this line ending,

PR #1751, opened 2025-02-26
test: Improve TestOrderedSettings test

arthurpitman:
This PR improves `TestOrderedSettings` by testing using the scope of a random host entity ID, which prevents tests from interfering with each other.

This test works by creating two scoped settings objects in one deployment and then re-deploying with the order reversed and checking that the order has been updated.


cmd/monaco/integrationtest/v2/settings_integration_test.go
```diff
@@ -182,26 +187,39 @@ func TestDeploySettingsWithUniqueProperties_ConsidersScopes(t *testing.T) {
...
+	expectedExternalIdForAAA, err := idutils.GenerateExternalIDForSettingsObject(coordinate.Coordinate{Project: "project", ConfigId: "aaa", Type: "builtin:processavailability"})
+	assert.NoError(t, err)
+	expectedExternalIdForBBB, err := idutils.GenerateExternalIDForSettingsObject(coordinate.Coordinate{Project: "project", ConfigId: "bbb", Type: "builtin:processavailability"})
+	assert.NoError(t, err)
+
 	configFolder := "test-resources/settings-ordered/order1"
 	manifestPath := configFolder + "/manifest.yaml"
 
-	RunIntegrationWithCleanup(t, configFolder, manifestPath, "", "SettingsOrdered", func(fs afero.Fs, _ TestContext) {
+	t.Setenv("MONACO_TARGET_ENTITY_SCOPE", host)
```
tomazpu, line 207: I would probably move this to be between lines 193 and 194.

tomazpu (dismissed):

Kirdock (dismissed):
LGTM
Just one thing to understand it better: Why does the order change? In the first step the settings are created and the second one should update the objects because they already exist right? I would not expect an order-change on update

tomazpu (approved):

PR #1752, opened 2025-02-27
fix: Support archive should only include memstat log if feature flag enabled

arthurpitman:
This PR only includes the memstat log in the support archive if the feature flag is enabled because otherwise it doesnt exist.

This fixes the error:

```
2025-02-27T10:22:48+01:00	error	Encountered error creating support archive. Archive may be missing or incomplete: encountered multiple errors: [ unable to add .logs/20250227-102246-memstat.log file to archive support-archive-20250227-102246.zip: open .logs/20250227-102246-memstat.log: no such file or directory ]
```

### DIFF
--- a/cmd/monaco/integrationtest/v2/supportarchive_report_test.go
+++ b/cmd/monaco/integrationtest/v2/supportarchive_report_test.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


Kirdock (approved):

arthurpitman:
I see that our E2E tests don't cover these options, I will extend them.

Laubi (dismissed):

tomazpu (approved):

Laubi (approved):

PR #1756, opened 2025-02-28
fix: Allow relative paths in file parameters

arthurpitman:
This PR allows a file parameter to specify a template file using a relative path. This means the following is now accepted:

```
parameters:
  javascript:
    type: file
    path: ../common/script.js
```

where for example, the `config.yaml` is `./workflow/config.yaml` and the referenced `/script.js` is `../common/script.js`

### DIFF
--- a/pkg/config/parameter/file/file.go
+++ b/pkg/config/parameter/file/file.go
@@ -18,6 +18,7 @@ package file
 
 import (
 	"fmt"
+	"path/filepath"
 
 	"github.com/spf13/afero"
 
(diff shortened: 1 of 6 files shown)


Kirdock (dismissed):

pkg/config/parameter/file/file.go
```diff
@@ -93,14 +94,18 @@ func parseFileValueParameter(context parameter.ParameterParserContext) (paramete
 		return nil, parameter.NewParameterParserError(context, "missing filesystem handle to load parameter")
 	}
 
-	path, ok := context.Value["path"]
+	p, ok := context.Value["path"]
 	if !ok {
 		return nil, parameter.NewParameterParserError(context, "missing property `path`")
 	}
 
+	path := strings.ToString(p)
```
Laubi, line 102: I think we should perform some proper validation here and type-cast it. Imo it should error if it's not a string.

Laubi: Line #130 also no longer needs the `strings.ToString(path)`, right?

pkg/config/parameter/parameters.go
```diff
@@ -81,6 +83,7 @@ func (p ParameterReference) String() string {
 }
 
 type ParameterParserContext struct {
+	Folder        string
```
Laubi, line 86: I would go with `workingDirectory` or `workingDir` instead of `Folder`

Kirdock (approved):

PR #1757, opened 2025-03-04
feat: extend `SettingsClient` with `AccessControl`

tomazpu:
Introduction of `AccessControl` interface and implementation to the `SettingsClient`.
```
GetPermission(context.Context, string) (dtclient.PermissionObject, error)
UpsertPermission(context.Context, string, dtclient.PermissionObject) error
DeletePermission(context.Context, string) error
```

I decided for an Upsert over Create and Update, since the handling of both is 99% same.

### DIFF
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -198,6 +198,8 @@ const (
 
 type TypeAccessor string
 
+const AllUsers TypeAccessor = "all-users"
+
 type (
 	// SettingsObject contains all the information necessary to create/update a settings object
 	SettingsObject struct {
@@ -233,11 +235,11 @@ type (
 	}
 
 	Accessor struct {
-		Type string `json:"type"`
+		Type TypeAccessor `json:"type"`
 	}
 	PermissionObject struct {
 		Permissions []TypePermissions `json:"permissions"`
-		Accessor    Accessor          `json:"accessor"`
+		Accessor    *Accessor         `json:"accessor,omitempty"`
 	}
 
 	postResponse struct {
@@ -268,11 +270,12 @@ type (
 )
 
 const (
-	settingsSchemaAPIPathClassic  = "/api/v2/settings/schemas"
-	settingsSchemaAPIPathPlatform = "/platform/classic/environment-api/v2/settings/schemas"
-	settingsObjectAPIPathClassic  = "/api/v2/settings/objects"
-	settingsObjectAPIPathPlatform = "/platform/classic/environment-api/v2/settings/objects"
-	settingsPermissionAPIPath     = "/settings/objects/{objectId}/permissions/all-users"
+	settingsSchemaAPIPathClassic      = "/api/v2/settings/schemas"
+	settingsSchemaAPIPathPlatform     = "/platform/classic/environment-api/v2/settings/schemas"
+	settingsObjectAPIPathClassic      = "/api/v2/settings/objects"
+	settingsObjectAPIPathPlatform     = "/platform/classic/environment-api/v2/settings/objects"
+	settingsPermissionAllUsersAPIPath = "/settings/objects/{objectId}/permissions/all-users"
+	settingsPermissionAPIPath         = "/settings/objects/{objectId}/permissions"
 )
 
 func WithExternalIDGenerator(g idutils.ExternalIDGenerator) func(client *SettingsClient) {
@@ -860,8 +863,8 @@ func (d *SettingsClient) Delete(ctx context.Context, objectID string) error {
 	return nil
 }
 
-func (d *SettingsClient) GetPermission(ctx context.Context, id string) (PermissionObject, error) {
-	path, err := getPermissionPathWithID(id)
+func (d *SettingsClient) GetPermission(ctx context.Context, objectID string) (PermissionObject, error) {
+	path, err := getPermissionPathWithID(objectID, "GET")
 	if err != nil {
 		return PermissionObject{}, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/client/dtclient/settings_client.go
```diff
@@ -841,3 +862,106 @@ func (d *SettingsClient) Delete(ctx context.Context, objectID string) error {
...
+	}
+
+	resp, err := coreapi.AsResponseOrError(d.client.GET(
+		ctx,
+		path,
+		corerest.RequestOptions{CustomShouldRetryFunc: corerest.RetryIfTooManyRequests},
+	))
+
+	apiError := coreapi.APIError{}
+	// when the API returns a 404 it means that you don't have permission (no-access)
```
Laubi, line 879: or it could be that the object does not exist

pkg/client/dtclient/settings_client.go
```diff
@@ -841,3 +862,106 @@ func (d *SettingsClient) Delete(ctx context.Context, objectID string) error {
...
 	return nil
 }
+
+func (d *SettingsClient) GetPermission(ctx context.Context, objectID string) (PermissionObject, error) {
+	path, err := getPermissionPathWithID(objectID)
+	if err != nil {
+		return PermissionObject{}, err
+	}
+
+	resp, err := coreapi.AsResponseOrError(d.client.GET(
```
Laubi, line 872: Any reason you preferred the deprecated over the non-deprecated methods?

pkg/client/dtclient/settings_client.go
```diff
@@ -841,3 +862,106 @@ func (d *SettingsClient) Delete(ctx context.Context, objectID string) error {
...
+		return err
+	}
+
+	payload, err := json.Marshal(permission)
+	if err != nil {
+		return fmt.Errorf("failed to marshal permission: %w", err)
+	}
+
+	// In the first sep we try to update the remote object
+	_, err = coreapi.AsResponseOrError(d.client.PUT(
```
Laubi, line 908: Should we do it 'correctly'? Aka fetch the permissions object, and if it does not exist, then create it, otherwise update it? I know that we have this update-first-and-if-fail-create code multiple times, but imo it's just hacky.

pkg/client/dtclient/settings_client.go
```diff
@@ -841,3 +862,106 @@ func (d *SettingsClient) Delete(ctx context.Context, objectID string) error {
...
+
+	if err != nil {
+		return fmt.Errorf("failed to delete permission object: %w", err)
+	}
+
+	return nil
+}
+
+// When creating a permission we need to point to settingsPermissionAPIPath else we point to settingsPermissionAllUsersAPIPath
+func getPermissionPathWithID(id string) (string, error) {
```
Laubi, line 961: You say in the doc that this handles both cases (all users and 'normal' one), but in the end you handle only the all-users case. If you keep it to all-users only, please update the function name

pkg/client/dtclient/settings_client.go
```diff
@@ -841,3 +862,106 @@ func (d *SettingsClient) Delete(ctx context.Context, objectID string) error {
...
+	if err != nil {
+		return err
+	}
+	_, err = coreapi.AsResponseOrError(d.client.DELETE(
+		ctx,
+		path,
+		corerest.RequestOptions{CustomShouldRetryFunc: corerest.RetryIfTooManyRequests},
+	))
+
+	if err != nil {
```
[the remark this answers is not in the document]
Laubi: Good question. I agree that it should be like in the other cases, but it feels also funny that we just ignore if we did not find the object. Why did we execute a delete at all if it does not exist?

Laubi (changes requested):
Some comments, but overall, a easy to read, good to understand code. Like it!

pkg/client/dtclient/settings_client.go
tomazpu: To consistently use the same method over the client. In the deprecated call, we're using the new method. Maybe we should change it for all methods in the client in a follow-up?

pkg/client/dtclient/settings_client.go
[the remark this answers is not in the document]
tomazpu: As discussed in daily we handle a 404 in the del call of monaco.

pkg/client/dtclient/settings_client.go
```diff
@@ -251,10 +274,12 @@ type (
...
 const (
-	settingsSchemaAPIPathClassic  = "/api/v2/settings/schemas"
-	settingsSchemaAPIPathPlatform = "/platform/classic/environment-api/v2/settings/schemas"
-	settingsObjectAPIPathClassic  = "/api/v2/settings/objects"
-	settingsObjectAPIPathPlatform = "/platform/classic/environment-api/v2/settings/objects"
+	settingsSchemaAPIPathClassic      = "/api/v2/settings/schemas"
+	settingsSchemaAPIPathPlatform     = "/platform/classic/environment-api/v2/settings/schemas"
+	settingsObjectAPIPathClassic      = "/api/v2/settings/objects"
+	settingsObjectAPIPathPlatform     = "/platform/classic/environment-api/v2/settings/objects"
+	settingsPermissionAllUsersAPIPath = "/settings/objects/{objectId}/permissions/all-users"
```
Kirdock, line 281:
```suggestion
	settingsPermissionAllUsersAPIPath = "/platform/classic/environment-api/v2/settings/objects/{objectId}/permissions/all-users"
```

pkg/client/dtclient/settings_client.go
```diff
@@ -251,10 +274,12 @@ type (
...
+	settingsPermissionAPIPath         = "/settings/objects/{objectId}/permissions"
```
Kirdock, line 282:
```suggestion
	settingsPermissionAPIPath         = "/platform/classic/environment-api/v2/settings/objects/{objectId}/permissions"
```

Kirdock (changes requested):

pkg/client/dtclient/settings_client.go
```diff
@@ -223,6 +238,14 @@ type (
 		TotalCount int        `json:"totalCount"`
 	}
 
+	Accessor struct {
+		Type TypeAccessor `json:"type"`
+	}
+	PermissionObject struct {
+		Permissions []TypePermissions `json:"permissions"`
+		Accessor    *Accessor         `json:"accessor,omitempty"`
```
Kirdock, line 246: Can the accessor actually be undefined? I would expect that it is always set to "all-users", "group" or "user".
When looking at the POST API, a missing accessor leads to a contraint violation

tomazpu: In a PUT it can be undefined. Since you would send the request at for example the all-users endpoint.

Kirdock: Maybe a follow up if we should add different structs for this case.

tomazpu: I agree, but also keep in mind monaco should not reimplement domain constrains of the payload.

pkg/client/dtclient/settings_client.go
Laubi: I think we should have consistent usage, and not using deprecated functions. A unification would be appreciated :)

Laubi (approved):

PR #1764, opened 2025-03-05
fix: Add logging and timeout to context for `GetDynatraceClassicURL()`

arthurpitman:
This PR adds logging and timeout to the `context.Context` used in `metadata.GetDynatraceClassicURL(...)`

### DIFF
--- a/pkg/client/metadata/metadata.go
+++ b/pkg/client/metadata/metadata.go
@@ -21,9 +21,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
 
 	coreapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	corerest "github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 )
 
 const ClassicEnvironmentDomainPath = "/platform/metadata/v1/classic-environment-domain"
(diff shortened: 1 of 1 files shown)


Laubi (approved):

pkg/client/metadata/metadata.go
```diff
@@ -43,9 +47,13 @@ func (u classicEnvURL) GetURL() string {
...
 }
 
-// GetDynatraceClassicURL tries to fetch the URL of the classic environment using the API of a platform enabled
-// environment
+// GetDynatraceClassicURL tries to fetch the URL of the classic environment using the API of a platform enabled environment.
+// The call has a timeout of 15 seconds.
 func GetDynatraceClassicURL(ctx context.Context, platformClient corerest.Client) (string, error) {
+	ctx = logr.NewContext(ctx, log.WithCtxFields(ctx).GetLogr())
+	ctx, cancel := context.WithTimeout(ctx, 15*time.Second)
+	defer cancel()
```
Kirdock, line 55: Do we even need cancel here? The context is only used for the API call and after that it's not used anymore, so `cancel` would noch make any difference

arthurpitman: I think so: the pattern from the docs for `context.WithTimeout(...)`:
```
func slowOperationWithTimeout(ctx context.Context) (Result, error) {
  ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
  defer cancel()  // releases resources if slowOperation completes before timeout elapses
  return slowOperation(ctx)
}
```

PR #1766, opened 2025-03-05
feat(report): report cyclic loading errors

Kirdock:
Adds errors that occur when resolving and sorting environment configs


cmd/monaco/integrationtest/v2/loading-report_test.go
```diff
@@ -93,7 +93,7 @@ func TestLoadingReport(t *testing.T) {
 					Type:  report.TypeLoad,
 					Time:  report.JSONTime(time.Now()),
 					State: report.StateError,
-					Error: "cannot parse definition in `invalid-classic-with-settings\\config.yaml`: config api type (notification) configuration can only reference IDs of other config api types - parameter \"alertingProfileId\" references \"builtin:alerting.profile\" type",
+					Error: "config api type (notification) configuration can only reference IDs of other config api types - parameter \"alertingProfileId\" references \"builtin:alerting.profile\" type",
```
Kirdock, line 96: Removed the start of it
Reason:
- It's anyway a contains match
- This would need some adjustments (`\` vs `/` on different operating systems)

pkg/deploy/deploy.go
```diff
@@ -122,6 +129,18 @@ func Deploy(ctx context.Context, projects []project.Project, environmentClients
 	return nil
 }
 
+func getSortedEnvConfigs(g graph.ConfigGraphPerEnvironment, environmentClients dynatrace.EnvironmentClients) (map[string][]graph.SortedComponent, error) {
```
arthurpitman, line 132: This function could also just take the environment names, obtained from `environmentClients.Names()`, similar to `graph.New(...)`.

arthurpitman: Or it could even move to the `graph` package or `ConfigGraphPerEnvironment`type?

pkg/deploy/deploy.go
```diff
@@ -84,27 +84,35 @@ func Deploy(ctx context.Context, projects []project.Project, environmentClients
 		}
 		errors.As(validationErrs, &deploymentErrs)
 	}
+
+	reporter := report.GetReporterFromContextOrDiscard(ctx)
+
+	preloadCaches(ctx, projects, environmentClients)
```
arthurpitman, line 90: From a performance point of view, does it make sense to preload the caches (which can take a long time) after building the graph and sorting the configs?

Kirdock: Yes, you are right. It would be better to first validate the cyclic dependency and only after that do the preload of the cache. Adjusted 👍

arthurpitman (approved):

tomazpu (approved):

PR #1770, opened 2025-03-06
feat(slo): fail when slo-v1 payload is sent to v2 and vice versa

Kirdock:
Fails now if an SLO-V1 payload is sent to SLO-V2 and vise versa


arthurpitman:
What happens if you try to deploy an `slo-v2` payload as an `builtin:slo` config?

Kirdock:
> What happens if you try to deploy an `slo-v2` payload as an `builtin:slo` config?

So you mean if an SLO-V2 config gets deployed to SLO-V1 (type: api: slo)?
It fails with `tried to deploy an slo-v2 configuration to slo-v1`. So basically the second test within `cmd/monaco/integrationtest/v2/slo_request_handling_test.go`
Basically if `evaluationType` is present, then it's assumed to ve slo-v1 and if not it's assumed to be slo-v2

Or are you referring to something else?

pkg/deploy/internal/slo/slo.go
```diff
@@ -83,6 +85,11 @@ func Deploy(ctx context.Context, client deployServiceLevelObjectiveClient, prope
 	}
 
 	//Strategy 3 is to create a new slo
+	if isSloV1, parseErr := api.IsSloV1Payload(requestPayload); parseErr != nil {
```
tomazpu, line 88: We already have unmarshaled the payload at the beginning line 54. We're adding an `externalID`, with a little change we can save some time here. Also, we should validate not only for create but also update.

pkg/client/dtclient/config_client.go
```diff
@@ -135,6 +135,14 @@ func (d *ConfigClient) UpsertByName(ctx context.Context, a api.API, name string,
 	if a.ID == api.Extension {
```
tomazpu, line 135: Also i would do the change in same style you did with slo in `pkg/deploy/internal/setting/setting.go`

tomazpu (changes requested):

arthurpitman:
> Or are you referring to something else?

I was wondering about an `slo-v2` payload going to a
```
type:
  settings:
    schema: builtin:slo
```

config. Would is that possible

Kirdock:
> > Or are you referring to something else?
> 
> I was wondering about an `slo-v2` payload going to a
> 
> ```
> type:
>   settings:
>     schema: builtin:slo
> ```
> 
> config. Would that be possible?

This would lead to the question: Which schemas do we want to differentiate and do we even want it in the first place? What does the rest say?
@Laubi @tomazpu

pkg/client/dtclient/config_client.go
```diff
@@ -132,9 +132,17 @@ func (d *ConfigClient) ExistsWithName(ctx context.Context, api api.API, name str
 }
 
 func (d *ConfigClient) UpsertByName(ctx context.Context, a api.API, name string, payload []byte) (entity DynatraceEntity, err error) {
-	if a.ID == api.Extension {
+	if isExtension(a) {
 		return d.uploadExtension(ctx, a, name, payload)
 	}
+
+	if isSlo(a) {
+		if isSloV1, parseErr := isSloV1Payload(payload); parseErr != nil {
```
tomazpu, line 140: At first, glance hard to read since the method is `isSloV1Payload`, and I would expect that I would go into the if it is, but then I see that an error is returned. Then, there is another if for the payload is V1. 

I would probably refactor it to something that the `isSloV1Payload` would only return an error. The detail about the error being a parse error or a validation error the code here doesn't care it would return in any case...

```
err := isSloV1Payload(payload)
if isSlo(a) && err != nil {
   return DynatraceEntity{}, err
}
```

Kirdock: I'll keep the validate check still inside the isSlo block, because we don't need to validate each and every payload.
Will also rename it to validateSloV1Payload because "is..." would implicitly tell that we have a boolean value

tomazpu (approved):

arthurpitman (approved):

PR #1771, opened 2025-03-06
chore: Use `go-version-file` instead of `go-version` in workflows

Laubi:
So we don't have do adapt the version in the future 🥳

### DIFF
--- a/.github/workflows/pr-static-code-analysis.yml
+++ b/.github/workflows/pr-static-code-analysis.yml
@@ -18,13 +18,13 @@ jobs:
       contents: read
       checks: write
     steps:
+      - name: ⬇️ Check out code into the Go module directory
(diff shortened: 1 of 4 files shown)


.github/workflows/pr-static-code-analysis.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: 🛠️ Set up Go 1.x
         uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 #v5.3.0
         with:
-          go-version: '~1.24'
+          go-version-file: go.mod
```
Kirdock, line 24: Needs to be after the checkout :)

Kirdock (approved):

PR #1773, opened 2025-03-10
fix: Update Buckets client

arthurpitman:


arthurpitman:
No longer needed - we have already updated!

PR #1776, opened 2025-03-11
feat: Deploy service users

arthurpitman:


pkg/account/deployer/client.go
```diff
@@ -247,6 +249,168 @@ func (d *accountManagementClient) upsertUser(ctx context.Context, userId string)
...
+// ResourceNotFoundError is an error signifying that the desired resource was not found.
+type ResourceNotFoundError struct {
+	Identifier string
+}
+
+func (e ResourceNotFoundError) Error() string {
+	return fmt.Sprintf("resource '%s' not found", e.Identifier)
+}
+
+func (c *accountManagementClient) getServiceUserIDByName(ctx context.Context, name string) (string, error) {
```
tomazpu, line 308: I see that once you're using it with d and once with c, if you could rename that to all be the same.

pkg/account/deployer/client.go
```diff
@@ -247,6 +249,168 @@ func (d *accountManagementClient) upsertUser(ctx context.Context, userId string)
...
+
+func (d *accountManagementClient) createServiceUser(ctx context.Context, dto accountmanagement.ServiceUserDto) (string, error) {
+	uuidDto, resp, err := d.client.ServiceUserManagementAPI.CreateServiceUserForAccount(ctx, d.accountInfo.AccountUUID).ServiceUserDto(dto).Execute()
+	defer closeResponseBody(resp)
+	if err = handleClientResponseError(resp, err, "failed to create service user"); err != nil {
+		return "", err
+	}
+
+	if uuidDto == nil {
+		return "", errors.New("the received data are empty")
```
tomazpu, line 277: Should the error message point more toward that an ID was not returned from the API?

pkg/account/deployer/client.go
```diff
@@ -247,6 +249,168 @@ func (d *accountManagementClient) upsertUser(ctx context.Context, userId string)
...
+}
+
+func (c *accountManagementClient) getServiceUsersPage(ctx context.Context, page int32, pageSize int32) (*accountmanagement.ExternalServiceUsersPageDto, error) {
+	r, resp, err := c.client.ServiceUserManagementAPI.GetServiceUsersFromAccount(ctx, c.accountInfo.AccountUUID).Page(page).PageSize(pageSize).Execute()
+	defer closeResponseBody(resp)
+	if err = handleClientResponseError(resp, err, "failed to get service users"); err != nil {
+		return nil, err
+	}
+	if r == nil {
+		return nil, errors.New("the received data are empty")
```
tomazpu, line 409: Same here maybe a better error message makes sense?

pkg/account/deployer/client.go
```diff
@@ -247,6 +249,168 @@ func (d *accountManagementClient) upsertUser(ctx context.Context, userId string)
...
+func (c *accountManagementClient) getServiceUserEmailByName(ctx context.Context, name string) (string, error) {
+	serviceUser, err := c.getServiceUserByName(ctx, name)
+	if err != nil {
+		return "", err
+	}
+
+	return serviceUser.Email, nil
+}
+
+//nolint:dupl
```
tomazpu, line 326: I wonder if its better to have the duplication shown to have this on the SonarCloud dashboard.

pkg/account/deployer/client.go
arthurpitman: I'll go with `c` for the moment, one step close towards having a shared client 👍

pkg/account/deployer/deployer.go
```diff
@@ -286,6 +306,23 @@ func (d *AccountDeployer) deployUserBindings(ctx context.Context, users map[acco
...
+//nolint:dupl
+func (d *AccountDeployer) deployServiceUserBindings(ctx context.Context, serviceUsers []account.ServiceUser, dispatcher *Dispatcher) {
+	for _, serviceUser := range serviceUsers {
+		serviceUser := serviceUser
+		deployUserBindingsJob :=
+			func(wg *sync.WaitGroup, errCh chan error) {
+				defer wg.Done()
+				d.logger.Info("Updating group bindings for service user %s", serviceUser.Name)
+				if err := d.updateServiceUserGroupBindings(d.logCtx(ctx), serviceUser); err != nil {
+					errCh <- fmt.Errorf("unable to deploy user binding for account %s: %w", d.accClient.getAccountInfo().AccountUUID, err)
```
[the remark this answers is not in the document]
arthurpitman: Good point, I just checked and many of the errors returned from the  code paths will include such information, but not all. Added to follow ups 👍

pkg/account/deployer/client.go
```diff
@@ -247,6 +249,168 @@ func (d *accountManagementClient) upsertUser(ctx context.Context, userId string)
...
+	serviceUser, err := c.getServiceUserByUid(ctx, uid)
+	if err != nil {
+		return "", err
+	}
+
+	return serviceUser.Email, nil
+}
+
+//nolint:dupl
+func (c *accountManagementClient) getServiceUserByUid(ctx context.Context, uid string) (*accountmanagement.ExternalServiceUserDto, error) {
```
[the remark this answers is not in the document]
arthurpitman: I will try this out

pkg/account/deployer/client.go
Kirdock: Or how about `this`? 😅

pkg/account/deployer/client.go
arthurpitman: I believe these type of error cannot occur and these checks can be removed. Added to follow ups though

Kirdock (approved):

PR #1778, opened 2025-03-12
feat(settings): add deploy for ACL all-users

Kirdock:
Adds support for ACL "all-users" in settings on deploy

I added some testing utils, which reduce the overhead for the testing server a lot.
I also added an exception for 404 of the permission deploy. We can discuss this one. Currently a delete is always one call. If we would add a GET check before it would be 1 (not exists) to 2  (exists, GET + DELETE) API calls


pkg/api/utils.go
Laubi, line 1: Please don't call anything `utils`, think of a telling name. Here it could be `errorchecks` or something similar.
Furthremore, since you make a public function, please find all usages of the same code. There are multiple occurances.

Laubi (changes requested):

pkg/client/dtclient/settings_client.go
```diff
@@ -616,10 +619,59 @@ func (d *SettingsClient) Upsert(ctx context.Context, obj SettingsObject, upsertO
...
+		return nil
+	}
+	permissions := getPermissionsFromConfig(allUserPermission)
+
+	// if we don't have any permissions to update and the schema does not support ACL, we return (success)
+	if schema, schemaExists := d.schemaCache.Get(schemaId); schemaExists && (schema.OwnerBasedAccessControl == nil || !*schema.OwnerBasedAccessControl) && permissions == nil {
+		return nil
+	}
+
+	if permissions == nil {
```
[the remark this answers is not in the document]
Kirdock: You are right. I'll adjust the read-logic (in the deploy I don't know if `permissions` is set or not)

pkg/api/utils.go
Kirdock: I'll move it to core and create another PR to introduce it

pkg/client/dtclient/settings_client.go
```diff
@@ -616,10 +619,59 @@ func (d *SettingsClient) Upsert(ctx context.Context, obj SettingsObject, upsertO
 		return DynatraceEntity{}, err
 	}
 
+	permErr := d.modifyPermission(ctx, entity.Id, obj.SchemaId, upsertOptions.AllUserPermission)
+	if permErr != nil {
+		return DynatraceEntity{}, permErr
```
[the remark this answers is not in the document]
Kirdock: Context added 👍

pkg/client/dtclient/settings_client.go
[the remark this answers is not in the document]
Kirdock: Regarding `If the 'permissions' is set, but allusers is not, fallback to 'None'`, at the moment we can't differentiate if it's defined and empty vs not defined at all.
Should I modify and parse the config differently/again in order to make this work (setting to none)? 
```yaml
configs:
- id: profile-id
  config:
    name: 'my config'
    template: 'profile.json'
    originObjectId: origin-object-id
  type:
    settings:
      schema: 'builtin:profile.test'
      schemaVersion: '1.0'
      scope: 'tenant'
      permissions:
```

If something inside `permissions` is defined it will work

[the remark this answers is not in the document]
Kirdock: added the above mentioned code/functionality ✔️

pkg/client/dtclient/settings_client.go
```diff
@@ -616,10 +618,56 @@ func (d *SettingsClient) Upsert(ctx context.Context, obj SettingsObject, upsertO
 		return DynatraceEntity{}, err
 	}
 
+	permErr := d.modifyPermission(ctx, entity.Id, obj.SchemaId, upsertOptions.AllUserPermission)
+	if permErr != nil {
+		return DynatraceEntity{}, fmt.Errorf("failed to modify permissions of Settings object with externalId %s: %w", externalID, permErr)
```
Laubi, line 623:
```suggestion
		return DynatraceEntity{}, fmt.Errorf("failed to modify permissions of settings object with externalId %s: %w", externalID, permErr)
```

pkg/client/dtclient/settings_client.go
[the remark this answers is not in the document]
Kirdock: After some discussion, we decided to remove this functionality again

pkg/client/dtclient/settings_client.go
```diff
@@ -608,18 +610,59 @@ func (d *SettingsClient) Upsert(ctx context.Context, obj SettingsObject, upsertO
...
+		err := d.UpsertPermission(ctx, objectID, PermissionObject{
+			Permissions: permissions,
+			Accessor:    &Accessor{AllUsers},
+		})
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
tomazpu, line 652:
```suggestion
func (d *SettingsClient) modifyPermission(ctx context.Context, objectID string, allUserPermission config.AllUserPermissionKind) error {
	permissions := getPermissionsFromConfig(allUserPermission)

	if permissions != nil {
		return d.UpsertPermission(ctx, objectID, PermissionObject{
			Permissions: permissions,
			Accessor:    &Accessor{AllUsers},
		})
	}

	return d.DeletePermission(ctx, objectID)
}
```
I would simplify this as the method returns error and you not wrapping the error anyway. Also removed the else as I don't like it :). I ran the unit test and all passed, but please re-check.

pkg/deploy/internal/setting/setting.go
```diff
@@ -61,8 +61,9 @@ func Deploy(ctx context.Context, settingsClient client.SettingsClient, propertie
 	}
 
 	insertOptions := dtclient.UpsertSettingsOptions{
```
tomazpu, line 63: NIT: Should this be called `upsertOptions`?

pkg/persistence/config/internal/persistence/type_definition.go
```diff
@@ -379,3 +378,16 @@ func (c TypeDefinition) MarshalYAML() (interface{}, error) {
 	}
 	return nil, fmt.Errorf("unknown type: %T", c.Type)
 }
+
+// getAllUserPermission returns the allUsers permission and falls back to "none" if permission are set but allUsers is missing
+func getAllUserPermission(p *PermissionDefinition) *config.AllUserPermissionKind {
```
tomazpu, line 383: I really like it with a separate method and an explanation of what is going on here 👍

tomazpu (commented):
Just some suggestion, looking good.

tomazpu (approved):

Laubi (approved):

PR #1782, opened 2025-03-13
feat(settings): Require OAuth for Settings config with defined `permission`

tomazpu:
Validation that OAuth is defined for the usage of Permissions on the `SettingsType`.


cmd/monaco/deploy/deploy_test.go
```diff
@@ -525,6 +530,32 @@ func Test_ValidateAuthenticationWithProjectConfigs(t *testing.T) {
 			},
 			"",
 		},
+		{
+			"token manifest with settings api and permissions",
```
Kirdock, line 534:
```suggestion
			"OAuth manifest with settings api and permissions",
```

cmd/monaco/deploy/deploy.go
```diff
@@ -287,6 +287,10 @@ func validateAuthenticationWithProjectConfigs(projects []project.Project, enviro
 							return fmt.Errorf("API of type '%s' requires a token for environment '%s'", conf.Type, envName)
 						}
 					case config.SettingsType:
+						t, ok := conf.Type.(config.SettingsType)
+						if ok && t.AllUserPermission != nil && environments[envName].Auth.OAuth == nil {
+							return fmt.Errorf("using permission property on settings API requires a OAuth, schema '%s' enviroment '%s'", t.SchemaId, envName)
```
Kirdock, line 292: maybe like this?
```suggestion
							return fmt.Errorf("using permission property on settings API requires OAuth, schema '%s' enviroment '%s'", t.SchemaId, envName)
```

cmd/monaco/deploy/deploy.go
```diff
@@ -287,6 +287,10 @@ func validateAuthenticationWithProjectConfigs(projects []project.Project, enviro
 							return fmt.Errorf("API of type '%s' requires a token for environment '%s'", conf.Type, envName)
 						}
 					case config.SettingsType:
+						t, ok := conf.Type.(config.SettingsType)
+						if ok && t.AllUserPermission != nil && environments[envName].Auth.OAuth == nil {
+							return fmt.Errorf("Using permission property on settings API requires OAuth, schema '%s' enviroment '%s'", t.SchemaId, envName)
```
Kirdock, line 292: Missed that one. It should start lower case right? ON the other hand we have the `API` uppercase in the other errors

tomazpu: I got confused, about why others start uppercase. I see now its the API acronym :D. Thanks for pointing it out 👍

Kirdock (dismissed):

Kirdock (dismissed):

cmd/monaco/deploy/deploy.go
```diff
@@ -287,6 +287,10 @@ func validateAuthenticationWithProjectConfigs(projects []project.Project, enviro
 							return fmt.Errorf("API of type '%s' requires a token for environment '%s'", conf.Type, envName)
 						}
 					case config.SettingsType:
+						t, ok := conf.Type.(config.SettingsType)
```
Kirdock, line 290: Technically the `ok` is not needed becasue the switch-case already checks it, but also fine if we leave this one

Kirdock (approved):

PR #1788, opened 2025-03-18
fix(account): Allow multiple service users with the same name iff originObjectId is specified

arthurpitman:
As account service users are not uniquely identified by name, but rather an ID, this PR refactors to allow this:
- stores service users in a slice rather than a map with name as key
- adds checks for service user ambiguity:
  - same name is allowed as long as `originObjectId` is specified and
  - `originObjectId` must be unique
- adds some test cases demonstrating that service users with the same name but different originObjectIds can be loaded and written, while ambiguous ones cannot


Kirdock (commented):
I already started with the review before the PR was set to draft again, just in case you are wondering why I comment on a draft

arthurpitman:
> I already started with the review before the PR was set to draft again, just in case you are wondering why I comment on a draft

No problem, I've just decided to redo this a bit, first changing some other things.

pkg/persistence/account/loader/load.go
```diff
@@ -157,12 +155,35 @@ func addResourcesFromFile(res persistence.Resources, file persistence.File) erro
 
 	if featureflags.ServiceUsers.Enabled() {
 		for _, su := range file.ServiceUsers {
-			if _, exists := res.ServiceUsers[su.Name]; exists {
-				return fmt.Errorf("found duplicate service user with name %q", su.Name)
+			for _, existingServiceUser := range res.ServiceUsers {
+				if areServiceUsersAmbiguous(su, existingServiceUser) {
+					return fmt.Errorf("found ambiguous service user with name %q", su.Name)
```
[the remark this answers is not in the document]
arthurpitman: fixed

pkg/account/downloader/service_users.go
```diff
@@ -71,10 +71,10 @@ func (a *Downloader) serviceUsers(ctx context.Context, groups Groups) (ServiceUs
 	return retVal, nil
 }
 
-func (sus ServiceUsers) asAccountServiceUsers() map[account.ServiceUserId]account.ServiceUser {
-	retVal := make(map[account.ServiceUserId]account.ServiceUser, len(sus))
+func (sus ServiceUsers) asAccountServiceUsers() []account.ServiceUser {
```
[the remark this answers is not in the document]
arthurpitman: Basically I agree, I'll add it to our list of follow ups.

pkg/account/persistence/loader/load.go
```diff
@@ -46,31 +46,33 @@ func LoadResources(fs afero.Fs, workingDir string, projects manifest.ProjectDefi
...
 			}
 			resources.Users[us.Email.Value()] = us
 		}
 
 		if featureflags.ServiceUsers.Enabled() {
 			for _, su := range res.ServiceUsers {
-				if _, exists := resources.ServiceUsers[su.Name]; exists {
-					return nil, fmt.Errorf("service user with name %q already defined in another project", su.Name)
+				for _, existingServiceUser := range resources.ServiceUsers {
+					if err := verifySerivceUsersAreNotAmbiguous(su, existingServiceUser); err != nil {
```
Kirdock, line 71: Just to get it right, we are using `verifySerivceUsersAreNotAmbiguous` two times, one to check if it's unique for one project and one for checking if it's unique for all projects.
This means we actually don't need to check it within `Load` if we change it to be private (no usages besides tests).

Because the check is done across projects it would make sense to also mention the project in the error message

tomazpu: I would suggest an improvement in error handling here in general. It may make sense to introduce an error type that gets the project. Also splitting the big for loop into smaller methods, so we have less complexity.

pkg/account/persistence/loader/load.go
```diff
@@ -205,12 +205,36 @@ func addResourcesFromFile(res *account.Resources, file persistence.File) error {
...
-			res.ServiceUsers[su.Name] = transformServiceUser(su)
+			res.ServiceUsers = append(res.ServiceUsers, serviceUser)
 		}
 	}
 
 	return nil
 }
+
+// verifySerivceUsersAreNotAmbiguous returns an error iff the two objects could refer to the same underlying service users.
+func verifySerivceUsersAreNotAmbiguous(su1 account.ServiceUser, su2 account.ServiceUser) error {
```
tomazpu, line 222: Typo in `Serivce`

tomazpu (approved):

pkg/account/persistence/loader/load.go
arthurpitman: Totally agree - it is already in the (big) list of follow ups 👍

arthurpitman: I have improved the error messages through a small refactoring, but would like to clarify the role of projects in account management before doing too much more.

tomazpu (dismissed):

Kirdock (dismissed):

pkg/account/persistence/loader/load.go
```diff
@@ -38,44 +38,54 @@ var ErrMixingDelete = errors.New("mixing account resources with a delete file de
...
 
-		for _, gr := range res.Groups {
-			if _, exists := resources.Groups[gr.ID]; exists {
-				return nil, fmt.Errorf("group with id %q already defined in another project", gr.ID)
-			}
-			resources.Groups[gr.ID] = gr
+	return allResources, nil
+}
+
+func addProjectResources(resources *account.Resources, res *account.Resources) error {
```
d0weinberger, line 56: From the function signature and the argument naming, I find it hard to tell which is the source and which is the target `Resources`

d0weinberger: Kinda [nit], but since this function only depends on `account.Resources`, you could also move it there. What about something like the following?
`func (r account.Resources) AddProjectResources(projectResources account.Resources) error`

pkg/account/persistence/loader/load.go
```diff
@@ -112,7 +122,7 @@ func findAndLoadResources(fs afero.Fs, rootPath string) (*account.Resources, err
 		Policies:     make(map[string]account.Policy),
 		Groups:       make(map[string]account.Group),
 		Users:        make(map[string]account.User),
-		ServiceUsers: make(map[string]account.ServiceUser),
+		ServiceUsers: make([]account.ServiceUser, 0),
```
d0weinberger, line 125: Just a suggestion that came to mind when reading the purpose of the PR, but please disregard if it doesn't make any sense: Instead of having a slice, what about `map[serviceUserKey]account.ServiceUser`, where `serviceUserKey` is a struct containing both the name and an optional originObjectId?

pkg/account/persistence/loader/load.go
arthurpitman: > Kinda [nit], but since this function only depends on `account.Resources`, you could also move it there. What about something like the following? `func (r account.Resources) AddProjectResources(projectResources account.Resources) error`

This function also uses `verifyServiceUsersAreNotAmbiguous(...)` which is also used by `addResourcesFromFile(...)`, I would like to address this too, but not in this PR.

pkg/account/persistence/loader/load.go
arthurpitman: With the exception of checking for duplicates, we don't actually need to look up service users, so I'm not sure this would make sense. But I take the point that the data structure isn't ideal, and will add it to the follow up list.

Kirdock (approved):

PR #1795, opened 2025-03-19
test: Fix first/last e2e test

Laubi:
This Pr fixes a flaky test by replacing it with two other ones.
The flakiness was as the configs were deployed in parallel, and the ordering was up to chance.

The new tests test:
* If a config already exists, first/last should work as expected
* If deployed twice, first/last should work as expected.

### DIFF
--- a/cmd/monaco/integrationtest/utils/monaco/cmd.go
+++ b/cmd/monaco/integrationtest/utils/monaco/cmd.go
@@ -47,7 +47,6 @@ func Run(t *testing.T, fs afero.Fs, command string) error {
 	if !strings.HasPrefix(c, prefix) {
 		return fmt.Errorf("command must start with '%s'", prefix)
 	}
-	t.Logf("Running command: %s", command)
 	c = strings.TrimPrefix(c, prefix)
 
 	args := strings.Split(c, " ")
@@ -58,5 +57,10 @@ func Run(t *testing.T, fs afero.Fs, command string) error {
 	// explicit cancel for each monaco run invocation
 	ctx, cancel := context.WithCancel(t.Context())
 	defer cancel()
-	return runner.RunCmd(ctx, cmd)
+
+	t.Logf("Running command: %s", command)
+	err := runner.RunCmd(ctx, cmd)
+	t.Logf("Finished command: %s", command)
+
+	return err
 }
--- a/cmd/monaco/integrationtest/v2/settings_integration_test.go
+++ b/cmd/monaco/integrationtest/v2/settings_integration_test.go
@@ -398,7 +398,7 @@ func TestOrdered_InsertAtFrontAndBackWorks(t *testing.T) {
 	const manifestFile = configFolder + "/manifest.yaml"
 
 	const specificEnvironment = "platform"
-	const project = "both-back-and-front-are-set"
+	const project = "both-back-and-front-are-set-with-initial"
 	const schema = "builtin:url-based-sampling"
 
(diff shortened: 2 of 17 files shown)


pkg/client/dtclient/settings_client.go
```diff
@@ -616,7 +616,11 @@ func (d *SettingsClient) Upsert(ctx context.Context, obj SettingsObject, upsertO
 		return DynatraceEntity{}, err
 	}
 
-	log.WithCtxFields(ctx).Debug("Created/Updated object %s (%s) with externalId %s", obj.Coordinate.ConfigId, obj.SchemaId, externalID)
+	insertAfterForLogging := "<nil>"
+	if upsertOptions.InsertAfter != nil {
+		insertAfterForLogging = *upsertOptions.InsertAfter
```
arthurpitman, line 621: I think these lines would be better in a function

cmd/monaco/integrationtest/v2/settings_integration_test.go
```diff
@@ -429,6 +429,48 @@ func TestOrdered_InsertAtFrontAndBackWorks(t *testing.T) {
...
+		t.Log("Monitored entity ID for testing ('MONACO_TARGET_ENTITY_SCOPE') =", pgiMeId)
+
+		// first
+		err := monaco.Run(t, fs, fmt.Sprintf("monaco deploy %s --project %s --verbose", manifestFile, project))
+		require.NoError(t, err)
+		// second
+		err = monaco.Run(t, fs, fmt.Sprintf("monaco deploy %s --project %s --verbose", manifestFile, project))
+		require.NoError(t, err)
+
+		integrationtest.AssertAllConfigsAvailability(t, fs, manifestFile, []string{project}, specificEnvironment, true)
```
arthurpitman, line 453: Why do we need this call to `integrationtest.AssertAllConfigsAvailability(...)` (and if so, why only once)?

arthurpitman: Reflecting, I guess it might be something to check in our tests in general: if we are explicitly checking deployment manually we might not need this.

cmd/monaco/integrationtest/utils/monaco/cmd.go
```diff
@@ -58,5 +57,10 @@ func Run(t *testing.T, fs afero.Fs, command string) error {
 	// explicit cancel for each monaco run invocation
 	ctx, cancel := context.WithCancel(t.Context())
 	defer cancel()
-	return runner.RunCmd(ctx, cmd)
+
+	t.Logf("Running command: %s", command)
```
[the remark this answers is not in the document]
Laubi: I'd not say, as with this it states the 'full' invocation: `monaco deploy ...`

cmd/monaco/integrationtest/v2/settings_integration_test.go
Laubi: Is there a reason you are concerned about having it in?

pkg/client/dtclient/settings_client.go
tomazpu: 9/10 times, I would also be on the side of a separate function, but here my gut feeling is that the if is so simple that i would make it more complicated. Since I would need to go into the function and then back up to the caller code. 
But if in the future this if needs to do more things then override the default value i would move it into a separate function.

arthurpitman (approved):

tomazpu (approved):

PR #1798, opened 2025-03-20
fix: correctly save environment urls in downloaded manifest file

d0weinberger:
If a manifest file is used in the monaco download command and the environment url is given via environment variable, the same variable name is present in the manifest file of the download folder, whereas previously only the value was stored.

Currently, if you use the `monaco download` command with a manifest file, and an environment url in this manifest file is defined via an environment variable, the manifest file generated in the download folder will still contain the constant url string instead of an environment variable name.

So, current behavior is:

Manifest file with environment url defined like so:
```url:
      type: environment
      value: ENVIRONMENT_URL
```
used in `monaco download`, where `ENVIRONMENT_URL=env.url.com`, will lead to a generated manifest file with the following content:

```url:
      value: env.url.com
```
**This pull request** changes that. The output will then be:

```url:
      type: environment
      value: ENVIRONMENT_URL
```

Tests have been adapted/improved.

Yes, see the description of "What this PR does" above.

### DIFF
--- a/cmd/monaco/download/download_configs_test.go
+++ b/cmd/monaco/download/download_configs_test.go
@@ -44,6 +44,22 @@ import (
 )
 
 func TestDownloadConfigsBehaviour(t *testing.T) {
+	var downloadOptions = downloadOptionsShared{
+		environmentURL: manifest.URLDefinition{
+			Type:  manifest.ValueURLType,
+			Value: "testurl.com",
+		},
+		auth: manifest.Auth{
+			Token: &manifest.AuthSecret{
+				Name:  "TEST_TOKEN_VAR",
+				Value: "test.token",
+			},
+		},
+		outputFolder:           "folder",
+		projectName:            "project",
+		forceOverwriteManifest: false,
+	}
+
 	tests := []struct {
 		name                      string
 		givenOpts                 downloadConfigsOptions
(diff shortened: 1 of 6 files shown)


pkg/download/download_writer_test.go
```diff
@@ -200,6 +255,10 @@ func TestWriteToDisk(t *testing.T) {
 				t.Errorf("WriteToDisk(): expected manifest %v was not created", expectedManifest)
 			}
 
+			actualManifestContent, err := afero.ReadFile(tt.args.fs, expectedManifest)
+			require.NoError(t, err)
+			assert.Equal(t, tt.wantManifestContent, string(actualManifestContent),
+				"WriteToDisk(): manifest content was expected to be:\n%v\nbut actually was:\n%v", tt.wantManifestContent, string(actualManifestContent))
```
arthurpitman, line 261: Nit: I don't think `WriteToDisk():` serves any purpose here (or anywhere in this test)
```suggestion
				"manifest content was expected to be:\n%v\nbut actually was:\n%v", tt.wantManifestContent, string(actualManifestContent))
```

cmd/monaco/download/download_configs_test.go
```diff
@@ -159,7 +159,10 @@ func TestDownloadConfigsBehaviour(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 
 			tt.givenOpts.downloadOptionsShared = downloadOptionsShared{
```
arthurpitman, line 161: While not introduced here in this PR, would it be possible to move the `downloadOptionsShared` constant out of here and assigned to each test case in the table above? I don't think we should modify the test setup within the test itself.

Kirdock (dismissed):
LGTM!
Please clarify the points @arthurpitman mentioned before merging :)

cmd/monaco/download/download_configs_test.go
d0weinberger: I moved it to the top of the func and assigned it to each test case.

arthurpitman (approved):

PR #1800, opened 2025-03-21
fix: Correctly handle `APIError` in `AssertBucket(...)`

arthurpitman:
This PR  updates `AssertBucket(...)`  and `getBucketWithRetry(...)` to  expect and react to API errors that are returned via `err`, and specifically 404s, rather than assuming these are still part of the `bucket.Response`. I also updated `getBucketWithRetry(...)` to be non-recursive, as I find that easier to understand. While it is hard to verify, I suspect this may be responsible for the instability of the bucket integration tests.

I am also increasing the number of bucket get retries to `120`, that is 2 minutes, the same value as used by default in the core library. In my experience Buckets can be slow to appear, so this is a good idea.

I will follow up this PR with one for the core removing `func (r Response) AsAPIError() (APIError, bool)` and some associated functions as we no-longer use them.

### DIFF
--- a/cmd/monaco/integrationtest/assert.go
+++ b/cmd/monaco/integrationtest/assert.go
@@ -23,9 +23,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/http"
 	"testing"
 	"time"
+
+	coreapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 
 	"github.com/stretchr/testify/assert"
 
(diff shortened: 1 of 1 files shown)


tomazpu (approved):

cmd/monaco/integrationtest/assert.go
```diff
@@ -318,14 +319,21 @@ func AssertBucket(t *testing.T, client client.BucketClient, env manifest.Environ
 	return expectedId
 }
 
-func getBucketWithRetry(ctx context.Context, client client.BucketClient, bucketName string, try, maxTries int) (buckets.Response, error) {
-	resp, err := client.Get(ctx, bucketName)
-	if try < maxTries && resp.StatusCode == http.StatusNotFound {
-		try++
+func getBucketWithRetry(ctx context.Context, client client.BucketClient, bucketName string, maxTries int) (buckets.Response, error) {
```
Kirdock, line 322: Techlnically we could get rid of the response, because we don't need it anymore

Kirdock, line 322: So it changes it from 5 seconds max retry to 120 right?

arthurpitman: yes, correct - i suspect 5 is also not enough

arthurpitman: True 👍

PR #1807, opened 2025-03-25
fix: validator checking unique names doesn't return early anymore on …

d0weinberger:

This fixes a bug where we erroneously return early from unique name validation for classic apis if the scope of the config-to-check differs from a scope of the configs we have already seen before.

**YAML to validate**
```
configs:
  - id: settings-1
    config:
      name: Duplicated
      template: dashboard-share-settings.json
      skip: false
    type:
      api:
        name: dashboard-share-settings
        scope:
          type: value
          value: scope-0
  - id: settings-2
    config:
      name: Duplicated
      template: dashboard-share-settings.json
      skip: false
    type:
      api:
        name: dashboard-share-settings
        scope:
          type: value
          value: scope-1
  - id: settings-3
    config:
      name: Duplicated
      template: dashboard-share-settings.json
      skip: false
    type:
      api:
        name: dashboard-share-settings
        scope:
          type: value
          value: scope-1
```

**Current (wrong) log output:**

```
2025-03-25T11:17:51+01:00	info	Deploying configurations to environment "platform_env"...
2025-03-25T11:17:51+01:00	info	Deploying 3 independent configuration sets in parallel...
2025-03-25T11:17:51+01:00	info	[coord=test:dashboard-share-settings:settings-3][gid=2]	Deploying config
2025-03-25T11:17:51+01:00	info	[coord=test:dashboard-share-settings:settings-1][gid=0]	Deploying config
2025-03-25T11:17:51+01:00	info	[coord=test:dashboard-share-settings:settings-2][gid=1]	Deploying config
```

**Expected log output, introduced with this PR:**

```
2025-03-25T11:30:40+01:00	error	Error: Deployment failed - check logs for details: Errors encountered for 1 environment(s):
	"platform_env": duplicated config name found: configurations test:dashboard-share-settings:settings-3 and test:dashboard-share-settings:settings-2 define the same 'name' "Duplicated"
```

A test has been added to check the behavior. If you execute the test without the fix, it will fail.

Yes, validation will fail now for non-unique names in the same scope, whereas before there were cases where validation would **succeed**, although it shouldn't.

### DIFF
--- a/pkg/deploy/internal/classic/validation.go
+++ b/pkg/deploy/internal/classic/validation.go
@@ -74,41 +74,57 @@ func (v *validator) Validate(_ []project.Project, c config.Config) error {
 	}
 
 	for _, c2 := range v.uniqueNames[c.Environment][a.Api] {
-		// if the configs have a scope and they are different then the configs are unique
-		scope1 := c.Parameters[config.ScopeParameter]
-		scope2 := c2.Parameters[config.ScopeParameter]
-		if !cmp.Equal(scope1, scope2) {
+		if !scopesAreEqual(&c, &c2) {
 			continue
 		}
 
-		// if we are able to resolve the parameters, compare their value
-		cResolvedParams, errc := c.ResolveParameterValues(entities.New())
-		c2ResolvedParams, errc2 := c2.ResolveParameterValues(entities.New())
-		if len(errc) == 0 && len(errc2) == 0 {
-			if cmp.Equal(cResolvedParams[config.NameParameter], c2ResolvedParams[config.NameParameter]) {
-				return fmt.Errorf("duplicated config name found: configurations %s and %s define the same 'name' %q", c.Coordinate, c2.Coordinate, cResolvedParams[config.NameParameter])
-			}
-		}
-
-		// check if (unresolvable) reference parameters are equal
-		if r, ok := c.Parameters[config.NameParameter].(*reference.ReferenceParameter); ok {
-			if r2, ok := c2.Parameters[config.NameParameter].(*reference.ReferenceParameter); ok {
-				if r.Equal(r2) {
-					return fmt.Errorf("duplicated config name found: configurations %s and %s define the same 'name'", c.Coordinate, c2.Coordinate)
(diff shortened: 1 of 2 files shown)


pkg/deploy/internal/classic/validation_test.go
```diff
@@ -139,6 +139,26 @@ func TestValidate_ErrorForSameNameAndScope(t *testing.T) {
...
+	err2 := validator.Validate(newTestClassicConfigForValidation(t, "config2", api.KeyUserActionsMobile, map[string]parameter.Parameter{
+		config.NameParameter:  value.New("name"),
+		config.ScopeParameter: value.New("scope1")}))
+	assert.NoError(t, err2)
+
+	err3 := validator.Validate(newTestClassicConfigForValidation(t, "config3", api.KeyUserActionsMobile, map[string]parameter.Parameter{
+		config.NameParameter:  value.New("name"),
+		config.ScopeParameter: value.New("scope1")}))
+
+	assert.Error(t, err3)
```
Laubi, line 159: Nit: Empty space above where the other ones don't have it.

pkg/deploy/internal/classic/validation.go
```diff
@@ -74,41 +74,57 @@ func (v *validator) Validate(_ []project.Project, c config.Config) error {
...
+
+	return cmp.Equal(scope1, scope2)
+}
+
+// ensureNonDuplicateNames returns an error if a name is used more than once for the same type.
+// If parameter values can be resolved for both configs, we compare name parameters directly.
+// If they can't be resolved, we compare the unresolved NameParameters of the two configs,
+// first assuming both being of type ReferenceParameter, and if that fails, assuming both being of type CompoundParameter.
+// If the fields are equal, a name clash is guaranteed.
+// This check is not perfect. Name clashes are not caught if, e.g., one config is resolvable and the other one isn't.
```
Laubi, line 103: 10/10 comment

d0weinberger: Thx!

PR #1810, opened 2025-03-26
fix: `\n`, `\t`, `\r` are no longer included in ME ids

Laubi:
Fixes a small issue where ME IDs accidently included the `n` from `\n` if there were no extra characters in between.

#### Note for reviewers:

The searching is done after extracting ME IDs, as it can't happen with UUIDs.

### DIFF
--- a/pkg/download/id_extraction/id_extraction.go
+++ b/pkg/download/id_extraction/id_extraction.go
@@ -95,9 +95,23 @@ func ExtractIDsIntoYAML(configsPerType project.ConfigsPerType) (project.ConfigsP
 	return configsPerType, nil
 }
 
+// invalidMeId finds ME IDs in the form of `nABC`, `rABC`, and `tABC`. They are most commonly mistakenly created by `\nABC`.
+var invalidMeId = regexp.MustCompile("[nrt][A-Z]+")
+
 func findAllIds(content string) []string {
 	ids := meIDRegexPattern.FindAllString(content, -1)
+
+	// Super simple, hacky, sanity check to not accidentally include matches for ME IDs in the form of `nABC-123` which result from `\nABC-123`.
+	// Alternatives to the check is validating the content of matches for `\`, but a `\` can also be escaped and would require recursive checking.
+	// To not do that, we simply check if the id starts with `n`, `r`, `t` (from `\n`, `\r`, `\t`), and then only uppercase characters. This is a good check for potential issues.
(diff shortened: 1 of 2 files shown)


pkg/download/id_extraction/id_extraction.go
```diff
@@ -95,9 +95,23 @@ func ExtractIDsIntoYAML(configsPerType project.ConfigsPerType) (project.ConfigsP
 	return configsPerType, nil
 }
 
+// invalidMeId finds ME IDs in the form of `nABC`, `rABC`, and `tABC`. They are most commonly mistakenly created by `\nABC`.
+var invalidMeId = regexp.MustCompile("[nrt][A-Z]+")
```
arthurpitman, line 99: Why is this defined like this when the ME ID is a much more general `[a-zA-Z_]+-[A-Fa-f0-9]{16}`?

arthurpitman: e.g. `nmz-1234567890ABCDEF` from `\nmz-1234567890ABCDEF` would match `meIDRegexPattern` but not `invalidMeId`.

Laubi: TBH, I've never seen a lowercase ME ID, and I think settings even does not allow it.
The sanity check is exactly having a lowercase `n` followed by uppercase characters. If there are only lowercase characters, how can we say that it's not part of the ME ID?

Maybe we need to challenge our assumption that ME IDs can be lowercase, and that would fix it as well.

arthurpitman: OK, that makes sense. The extra check handles exactly this case.

arthurpitman (commented):
After some consideration I'm not sure about this: the fix assumes that ME ids essentially match `[A-Z_]...}` but in that case, why not update `meIDRegexPattern` to begin with? Admittedly, this is more cautious, excluding _exactly those cases_, but it does feel pretty hacky. @tomazpu, what do you think?

arthurpitman (approved):
After quite some discussion, I feel that this is sufficient to handle exactly the case of removing a potential prefix of `n`, `r`, or `t` from a detected ME ID that begins with uppercase characters. This is a low risk option, without any significant performance penalty.
Further options, such as modifying the original ME ID detection regex, trying to remove encoded white space or unencoding the input string will be investigated in a follow up.

tomazpu (approved):

PR #1813, opened 2025-03-27
feat: add SonarQube Scan build step and `sonar-project.properties` to…

tomazpu:
Add's github action for sonar, that runs after the uni-test with coverage where executed.
Removes the GitHub action that displayed the Unit-Test result on PR instead all data is now ingested into Sonar.

This will need to be merged into `main` so Sonar gets result from `main` and can properly display code coverage change.

No


.github/workflows/build-test.yml
```diff
@@ -52,6 +52,14 @@ jobs:
           name: Test Results - Unit
           path: test-result-*.xml
 
+      - name: Coverage
```
Laubi, line 55: Eventually we should get rid of this, as the tests are now run twice, aren't they? (Above in Unit test task)

If they are reported correctly, we can just use those results (?)

pkg/deploy/internal/document/document.go
```diff
@@ -69,7 +68,7 @@ func Deploy(ctx context.Context, client Client, properties parameter.Properties,
 			return createResolvedEntity(documentName, md.ID, c.Coordinate, properties), nil
 		}
 
-		if !isAPIErrorStatusNotFound(err) {
+		if !api.IsNotFoundError(err) {
```
Laubi, line 71: On purpose to have both changesets in this PR? :D

tomazpu: SonarCloud coverage only triggers if more than 20 lines are changed. This is only a test to check that the coverage data is ingested.

.github/workflows/build-test.yml
tomazpu: Yes, true this is just a mid-step, next step would be a cleanup of the old stuff.

pkg/deploy/internal/document/document.go
tomazpu: 20 lines of go production non test code.

Laubi (approved):

PR #1814, opened 2025-03-27
refactor: use isNotFoundError of core lib for 404 checking

Kirdock:
We recently added a 404 check in the core.
Now I replaced all existing 404 checks with the one from core to reduce code duplication.

### DIFF
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -496,7 +496,7 @@ func (d *SettingsClient) handleUpsertUnsupportedVersion(ctx context.Context, obj
 	if err != nil {
 		apiErr := coreapi.APIError{}
 		// Settings API returns 400 StatusBadRequest for 404 StatusNotFound
-		if !errors.As(err, &apiErr) || apiErr.StatusCode != http.StatusBadRequest || apiErr.StatusCode == http.StatusNotFound {
+		if !errors.As(err, &apiErr) || apiErr.StatusCode != http.StatusBadRequest || coreapi.IsNotFoundError(apiErr) {
 			return DynatraceEntity{}, fmt.Errorf("unable to fetch settings object with object id %q: %w", obj.OriginObjectId, err)
 		}
 	}
@@ -942,8 +942,7 @@ func (d *SettingsClient) Get(ctx context.Context, objectId string) (res *Downloa
 func (d *SettingsClient) Delete(ctx context.Context, objectID string) error {
 	_, err := coreapi.AsResponseOrError(d.client.DELETE(ctx, d.settingsObjectAPIPath+"/"+objectID, corerest.RequestOptions{CustomShouldRetryFunc: corerest.RetryIfTooManyRequests}))
 	if err != nil {
-		apiError := coreapi.APIError{}
-		if errors.As(err, &apiError) && apiError.StatusCode == http.StatusNotFound {
+		if coreapi.IsNotFoundError(err) {
 			log.Debug("No settings object with id '%s' found to delete (HTTP 404 response)", objectID)
 			return nil
 		}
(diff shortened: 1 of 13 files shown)


pkg/client/dtclient/settings_client.go
```diff
@@ -1052,9 +1049,8 @@ func (d *SettingsClient) DeletePermission(ctx context.Context, objectID string)
 		corerest.RequestOptions{CustomShouldRetryFunc: corerest.RetryIfTooManyRequests},
 	))
 
-	apiError := coreapi.APIError{}
 	// deployments with "none" for all-user will always try to delete. This could be an update (restricted to shared) or it stays the same (delete 404)
-	if err != nil && (!errors.As(err, &apiError) || apiError.StatusCode != http.StatusNotFound) {
+	if err != nil && !coreapi.IsNotFoundError(err) {
```
tomazpu, line 1053: First part of the if is not needed, as if there is a 404 err the error is also always not nil.

pkg/delete/internal/classic/delete.go
```diff
@@ -42,7 +41,7 @@ func Delete(ctx context.Context, client client.ConfigClient, dps []pointer.Delet
 		var e error
 		if theAPI.HasParent() {
 			parentID, e = resolveIdentifier(ctx, client, theAPI.Parent, toIdentifier(dp.Scope, "", ""))
-			if e != nil && !is404(e) {
+			if e != nil && !coreapi.IsNotFoundError(e) {
```
tomazpu, line 44: same here

pkg/delete/internal/classic/delete.go
```diff
@@ -56,7 +55,7 @@ func Delete(ctx context.Context, client client.ConfigClient, dps []pointer.Delet
 		id := dp.OriginObjectId
 		if id == "" {
 			id, e = resolveIdentifier(ctx, client, &a, toIdentifier(dp.Identifier, dp.ActionType, dp.Domain))
-			if e != nil && !is404(e) {
+			if e != nil && !coreapi.IsNotFoundError(e) {
```
tomazpu, line 58: and here

pkg/delete/internal/classic/delete.go
```diff
@@ -66,7 +65,7 @@ func Delete(ctx context.Context, client client.ConfigClient, dps []pointer.Delet
 			}
 		}
 
-		if e := client.Delete(ctx, a, id); e != nil && !is404(e) {
+		if e := client.Delete(ctx, a, id); e != nil && !coreapi.IsNotFoundError(e) {
```
tomazpu, line 68: and here

pkg/delete/internal/document/delete.go
```diff
@@ -73,7 +72,7 @@ func deleteSingle(ctx context.Context, c client, dp pointer.DeletePointer) error
 	}
 
 	_, err := c.Delete(ctx, id)
-	if err != nil && !isAPIErrorStatusNotFound(err) {
+	if err != nil && !api.IsNotFoundError(err) {
```
tomazpu, line 75: here

pkg/delete/internal/segment/delete.go
```diff
@@ -70,7 +69,7 @@ func deleteSingle(ctx context.Context, c client, dp pointer.DeletePointer) error
 	}
 
 	_, err := c.Delete(ctx, id)
-	if err != nil && !isAPIErrorStatusNotFound(err) {
+	if err != nil && !api.IsNotFoundError(err) {
```
tomazpu, line 72: here

pkg/delete/internal/slo/delete.go
```diff
@@ -69,7 +68,7 @@ func deleteSingle(ctx context.Context, c client, dp pointer.DeletePointer) error
 	}
 
 	_, err := c.Delete(ctx, id)
-	if err != nil && !isAPIErrorStatusNotFound(err) {
+	if err != nil && !api.IsNotFoundError(err) {
```
tomazpu, line 71: here

pkg/client/dtclient/settings_client.go
Kirdock: Here it's actually needed because we are checking for "not" 404.
`IsNotFoundError` returns false if the error is nil, meaning that `!coreapi.IsNotFoundError(err)` will resolve to true if the error is nil

Kirdock: `coreapi.IsNotFoundError(err)` is basically `err != nil && coreapi.IsNotFoundError(err)`
whereas
`!coreapi.IsNotFoundError(err)` is basically `err == nill || !coreapi.IsNotFoundError(err)`

tomazpu (approved):

PR #1815, opened 2025-03-27
feat: Sanitize bucket names

arthurpitman:
This PR adds sanitization of generated bucket names:
- Bucket names are still generated by concatenating project name with the config id: `[projectName]_[configId]`
- Bucket names are sanitized by:
  - converting the name to lower case
  - removing characters such that the result must match `([a-z])([a-z0-9])([a-z0-9_-])+`
  - truncating to a maximum length of 100 characters.

By default this is enabled, however this can be disabled by setting the feature flag `MONACO_SANITIZE_BUCKET_NAMES` to `false`.

Furthermore, it makes sure the generated bucket name is visible in logs and errors by updating several error messages as well as updating the version of the core library used.

### DIFF
--- a/internal/featureflags/temporary.go
+++ b/internal/featureflags/temporary.go
@@ -46,6 +46,9 @@ const (
 	// AccessControlSettings toggles whether settings enable the access control
 	// Introduced: v2.21.0
 	AccessControlSettings FeatureFlag = "MONACO_FEAT_ACCESS_CONTROL_SETTINGS"
+	// SanitizeBucketNames toggles whether bucket names created by Monaco are sanitized or not.
+	// Introduced: v2.23.0
+	SanitizeBucketNames FeatureFlag = "MONACO_FEAT_SANITIZE_BUCKET_NAMES"
 )
 
 // temporaryDefaultValues defines temporary feature flags and their default values.
@@ -61,4 +64,5 @@ var temporaryDefaultValues = map[FeatureFlag]defaultValue{
 	OnlyCreateReferencesInStringValues: false,
 	ServiceLevelObjective:              true,
 	AccessControlSettings:              false,
+	SanitizeBucketNames:                true,
 }
--- a/internal/featureflags/temporary.go
+++ b/internal/featureflags/temporary.go
@@ -48,7 +48,7 @@ const (
 	AccessControlSettings FeatureFlag = "MONACO_FEAT_ACCESS_CONTROL_SETTINGS"
 	// SanitizeBucketNames toggles whether bucket names created by Monaco are sanitized or not.
 	// Introduced: v2.23.0
-	SanitizeBucketNames FeatureFlag = "MONACO_FEAT_SANITIZE_BUCKET_NAMES"
+	SanitizeBucketNames FeatureFlag = "MONACO_SANITIZE_BUCKET_NAMES"
 )
 
 // temporaryDefaultValues defines temporary feature flags and their default values.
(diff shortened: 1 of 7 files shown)


internal/idutils/bucketname_test.go
```diff
@@ -0,0 +1,121 @@
...
+			inputName:          "0abc",
+			expectedOutputName: "abc",
+		},
+		{
+			name:               "Multiple invalid first characters are removed",
+			inputName:          "0_abc",
+			expectedOutputName: "abc",
+		},
+		{
+			name:               "Invalid second character are removed",
```
Kirdock, line 91:
```suggestion
			name:               "Invalid second character is removed",
```

internal/idutils/bucketname.go
```diff
@@ -18,12 +18,51 @@ package idutils
...
+	name = regexp.MustCompile(`^[0-9_-]+`).ReplaceAllString(name, "")
+
+	// delete second character while it is not [a-z0-9]
+	name = regexp.MustCompile(`^([a-z])([_-]+)`).ReplaceAllString(name, "$1")
+
+	// truncate if longer that 100 characters. this only works because name only consists of characters [a-z0-9_-]: each is one byte in UTF-8.
+	if len(name) > 100 {
+		name = name[0:100]
+	}
+	return name
```
Kirdock, line 67: Just wondering, should we fail return an error here if the name is smaller than 3 characters?

internal/featureflags/temporary.go
```diff
@@ -46,6 +46,9 @@ const (
 	// AccessControlSettings toggles whether settings enable the access control
 	// Introduced: v2.21.0
 	AccessControlSettings FeatureFlag = "MONACO_FEAT_ACCESS_CONTROL_SETTINGS"
+	// SanitizeBucketNames toggles whether bucket names created by Monaco are sanitized or not.
+	// Introduced: v2.23.0
+	SanitizeBucketNames FeatureFlag = "MONACO_FEAT_SANITIZE_BUCKET_NAMES"
```
Laubi, line 51: Nit, but we don't need to add `FEAT` to everything, as it's not really a feature but a bugfix.

internal/idutils/bucketname.go
```diff
@@ -18,12 +18,51 @@ package idutils
...
+	if sanitiziedName != name {
+		log.Warn("Bucket name was changed to '%s' from '%s'", sanitiziedName, name)
+	}
+
+	return sanitiziedName
+}
+
+// SanitizeBucketName modifies the specified name to meet the requirements of the bucket-definitions API: pattern: `([a-z])([a-z0-9])([a-z0-9_-])+`,  maxLength: 100.
+// It does this by deleting invalid characters and truncating the result if it is more than 100 characters long.
+func SanitizeBucketName(name string) string {
```
Laubi, line 50: Is this method necessary to be public? I would either 1) make the test file use the same package, or 2) only test the GenerateBucketName, but with all test cases that you already provide.

arthurpitman: Exactly that - I wanted to test just this function but the existing tests are in the `idutils_test` package.

arthurpitman: Thanks, I have fixed the tests now to use `GenerateBucketName` directly.

internal/featureflags/temporary.go
arthurpitman: I've changed this to `MONACO_SANITIZE_BUCKET_NAMES` - less typing 👍

PR #1823, opened 2025-04-01
test: Enable caching for clients used by `AssertAllConfigsAvailability(...)`

arthurpitman:
**--- DO NOT MERGE ---** This PR, while a nice idea, currently breaks E2E tests

This PR introduces a new testing helper function `CreateDynatraceClientsWithCaching(...)` which creates a client set with caching enabled, and uses it within `AssertAllConfigsAvailability(...)`.

In tests which create significant numbers of Settings configs, this should improve performance.

Two examples are `TestPaginationClassic(...)` and `TestPaginationPlatform(...)`.

### DIFF
--- a/cmd/monaco/integrationtest/assert.go
+++ b/cmd/monaco/integrationtest/assert.go
@@ -102,7 +102,7 @@ func AssertAllConfigsAvailability(t *testing.T, fs afero.Fs, manifestPath string
 
 		env := loadedManifest.Environments[envName]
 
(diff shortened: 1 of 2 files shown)


arthurpitman:
Closing this for now as `AssertAllConfigsAvailability()` relies of waiting for configs to become available, which is only visible by actually calling the API again. A better approach would be to parallelize the check of all configs check those that are not available again once caches have been cleared, but this is beyond the scope right now.

PR #1825, opened 2025-04-01
refactor(deploy): add test cases to increase coverage in `deploy.go` 

tomazpu:
This PR adds test-cases to increase code coverage in deploy package.

### DIFF
--- a/pkg/deploy/deploy_test.go
+++ b/pkg/deploy/deploy_test.go
@@ -1300,10 +1300,13 @@ func TestDeployConfigGraph_CollectsAllErrors(t *testing.T) {
 
 }
 
+// Test if the FF gets Disabled/Enabled that the correct error is returned
 func TestDeployConfigFF(t *testing.T) {
+	// Clients here will return an error if reached
 	dummyClientSet := client.ClientSet{
 		SegmentClient:               client.TestSegmentsClient{},
 		ServiceLevelObjectiveClient: client.TestServiceLevelObjectiveClient{},
+		OpenPipelineClient:          client.TestOpenPipelineClient{},
 	}
 	c := dynatrace.EnvironmentClients{
 		dynatrace.EnvironmentInfo{Name: "env"}: &dummyClientSet,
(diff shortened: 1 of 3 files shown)


pkg/deploy/deploy_test.go
```diff
@@ -1379,33 +1412,105 @@ func TestDeployConfigFF(t *testing.T) {
...
+		{
+			name: "default error",
+			updateStub: func() (openpipeline.Response, error) {
+				return openpipeline.Response{}, coreapi.APIError{StatusCode: 0}
+			},
+			expectedErrLog: "Deployment failed - Dynatrace API call unsuccessful",
+		},
+	}
+	logSpy := bytes.Buffer{}
+	log.SetDefaultLogger(loggers.LogOptions{JSONLogging: true, LogSpy: &logSpy})
```
Kirdock, line 1476: What was again the reason, why `log.PrepareLogging` was not used? Too many parameters tests don't need to care about?

tomazpu: That is a very good point. To be honest, I just didn't see it :). I'll have a closer look.

pkg/deploy/deploy_test.go
```diff
@@ -1379,33 +1411,105 @@ func TestDeployConfigFF(t *testing.T) {
...
 				"env": project.ConfigsPerType{
 					"p1": {
 						config.Config{
-							Type:        config.Segment{},
+							Type:        config.OpenPipelineType{},
 							Environment: "env",
 							Coordinate: coordinate.Coordinate{
 								Project:  "p1",
-								Type:     "segment",
+								Type:     "type",
```
Kirdock, line 1437:
```suggestion
								Type:     "openpipeline",
```

pkg/deploy/deploy_test.go
```diff
@@ -1416,6 +1520,76 @@ func TestDeployDryRun(t *testing.T) {
 							},
 							Template: testutils.GenerateDummyTemplate(t),
 						},
+						config.Config{
+							Type:        config.OpenPipelineType{},
+							Environment: "env",
+							Coordinate: coordinate.Coordinate{
+								Project:  "p1",
+								Type:     "segment",
```
Kirdock, line 1528:
```suggestion
								Type:     "openpipeline",
```

Kirdock (dismissed):
LGTM
Comments don't need to be addressed, as the coordinate values should not matter

Kirdock (approved):

PR #1830, opened 2025-04-02
fix(settings): fixed unsafe access during permission load

Kirdock:
Accessing the permission map within a go routine is not thread save, therefore re-writing the implementation, to also align with the one in `settings_client.go`

No

### DIFF
--- a/pkg/download/settings/download.go
+++ b/pkg/download/settings/download.go
@@ -192,9 +192,9 @@ func extractApiErrorMessage(err error) string {
 
 func getObjectsPermission(ctx context.Context, client client.SettingsClient, objects []dtclient.DownloadSettingsObject) (map[string]dtclient.PermissionObject, error) {
 	type result struct {
-		Schema   dtclient.PermissionObject
-		ObjectId string
-		Err      error
+		Permission dtclient.PermissionObject
+		ObjectId   string
+		Err        error
 	}
 	errs := make([]error, 0)
 	resChan := make(chan result, len(objects))
@@ -209,17 +209,15 @@ func getObjectsPermission(ctx context.Context, client client.SettingsClient, obj
 	}
 
 	for i := 0; i < len(objects); i++ {
-		if res := <-resChan; res.Err != nil {
+		res := <-resChan
+		if res.Err != nil {
 			errs = append(errs, res.Err)
-		} else {
-			permissions[res.ObjectId] = res.Schema
+			continue
 		}
+		permissions[res.ObjectId] = res.Permission
 	}
 
-	if len(errs) > 0 {
-		return nil, errors.Join(errs...)
-	}
-	return permissions, nil
+	return permissions, errors.Join(errs...)
 }
 
 func asConcurrentErrMsg(err coreapi.APIError) string {
--- a/pkg/download/settings/download.go
+++ b/pkg/download/settings/download.go
@@ -217,7 +217,11 @@ func getObjectsPermission(ctx context.Context, client client.SettingsClient, obj
 		permissions[res.ObjectId] = res.Permission
 	}
 
-	return permissions, errors.Join(errs...)
+	if len(errs) > 0 {
+		return nil, errors.Join(errs...)
+	}
+
+	return permissions, nil
 }
 
 func asConcurrentErrMsg(err coreapi.APIError) string {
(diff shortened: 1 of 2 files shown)


pkg/download/settings/download.go
```diff
@@ -191,26 +191,31 @@ func extractApiErrorMessage(err error) string {
...
-				return
-			}
-			permissions[obj.ObjectId] = permission
-		}(obj)
+			resChan <- result{permission, obj.ObjectId, err}
+		}(ctx, obj)
 	}
-	wg.Wait()
+
+	for i := 0; i < len(objects); i++ {
```
arthurpitman, line 211: Is it possible to range over the channel until it is closed? 

```suggestion
	for res := range resChan {
```

Kirdock: Actually no, because we don't know when it should be closed. With a wait group we could wait for everything to be done and then close it. The implemenation was the outcome of this one https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1777#discussion_r1993449907

pkg/download/settings/download.go
```diff
@@ -191,26 +191,31 @@ func extractApiErrorMessage(err error) string {
 }
 
 func getObjectsPermission(ctx context.Context, client client.SettingsClient, objects []dtclient.DownloadSettingsObject) (map[string]dtclient.PermissionObject, error) {
-	downloadMutex := sync.Mutex{}
-	wg := sync.WaitGroup{}
-	wg.Add(len(objects))
+	type result struct {
+		Schema   dtclient.PermissionObject
```
d0weinberger, line 195: The name doesn't fit, right? I guess it should be `Permission` and not `Schema`

pkg/download/settings/download_test.go
```diff
@@ -852,6 +852,9 @@ func TestDownloadAll(t *testing.T) {
 				},
 				GetPermissionCalls: 0,
 			},
+			envVars: map[string]string{
+				featureflags.AccessControlSettings.EnvName(): "false",
```
d0weinberger, line 856: I don't understand this change and how it is related to the PR. Could you explain it, please?

d0weinberger (changes requested):

pkg/download/settings/download.go
```diff
@@ -191,26 +191,31 @@ func extractApiErrorMessage(err error) string {
...
-		}(obj)
+			resChan <- result{permission, obj.ObjectId, err}
+		}(ctx, obj)
 	}
-	wg.Wait()
+
+	for i := 0; i < len(objects); i++ {
+		if res := <-resChan; res.Err != nil {
+			errs = append(errs, res.Err)
+		} else {
```
tomazpu, line 214: As discussed in our call, it would be great if we get rid of the else

pkg/download/settings/download.go
```diff
@@ -191,26 +191,31 @@ func extractApiErrorMessage(err error) string {
...
+			permissions[res.ObjectId] = res.Schema
+		}
+	}
+
 	if len(errs) > 0 {
```
tomazpu, line 219: And remove this :)

pkg/download/settings/download_test.go
Kirdock: I was using some envs per default, also one where `AccessControlSettings` was enabled by default. I adjusted the test to always run with `false`. If we don't do it, the test fails, once we flip the FF

pkg/download/settings/download.go
Kirdock: You are right, copy & paste mistake

d0weinberger (dismissed):

tomazpu (dismissed):

pkg/download/settings/download.go
```diff
@@ -191,30 +191,33 @@ func extractApiErrorMessage(err error) string {
...
+		res := <-resChan
+		if res.Err != nil {
+			errs = append(errs, res.Err)
+			continue
+		}
+		permissions[res.ObjectId] = res.Permission
 	}
-	return permissions, nil
+
+	return permissions, errors.Join(errs...)
```
arthurpitman, line 220: small point, but I guess this function can return both some permissions and errors, however the calling function only makes use of one or the other

Kirdock: @tomazpu this would be a valid reason to use the `if` again, I would say.
Now: returns an incomplete result + error
With the if: returns no result + error

tomazpu (approved):

PR #1831, opened 2025-04-03
refactor(deploy): cache per environment

Kirdock:
Fills the cache per environment when it's deployed. So instead of filling the cache with all environments we are filling it when we deploy the config for one environment (only one env in cache at the time).

Yes, less memory consumption if several environments are used. Only one environment will be cached at the time


cmd/monaco/deploy/deploy.go
```diff
@@ -83,7 +83,7 @@ func deployConfigs(ctx context.Context, fs afero.Fs, manifestPath string, enviro
 		return formattedErr
 	}
 
-	err = deploy.Deploy(ctx, loadedProjects, clientSets, deploy.DeployConfigsOptions{ContinueOnErr: continueOnErr, DryRun: dryRun})
+	err = deploy.DeployForAllEnvironment(ctx, loadedProjects, clientSets, deploy.DeployConfigsOptions{ContinueOnErr: continueOnErr, DryRun: dryRun})
```
d0weinberger, line 86: Should be `DeployForAllEnvironments`

pkg/deploy/preload.go
```diff
@@ -89,35 +87,29 @@ func preloadValuesForApi(ctx context.Context, client client.ConfigClient, theApi
...
-				// If the config shall be skipped there is no point in caching it
-				if c.Skip {
-					return
-				}
-				if _, ok := seenConfigTypes[c.Coordinate.Type]; ok {
-					return
-				}
-				seenConfigTypes[c.Coordinate.Type] = struct{}{}
+	for _, p := range projects {
+		p.ForEveryConfigInEnvironmentDo(environment, func(c config.Config) {
```
d0weinberger, line 95: [note] Ah, this is the reason why we still need the environment; because we need it to select the relevant configs.

pkg/deploy/preload_test.go
```diff
@@ -35,7 +39,8 @@ import (
 var dummyConfigClient = &dtclient.DummyConfigClient{}
 var dummySettingsClient = &dtclient.DummySettingsClient{}
 var clientsetEnv1 = &client.ClientSet{ConfigClient: dummyConfigClient, SettingsClient: dummySettingsClient}
-var clientsetEnv2 = &client.ClientSet{ConfigClient: dummyConfigClient, SettingsClient: dummySettingsClient}
+
+//var clientsetEnv2 = &client.ClientSet{ConfigClient: dummyConfigClient, SettingsClient: dummySettingsClient}
```
d0weinberger, line 43: I think you can remove this one, right? If yes, please also rename `clientsetEnv1`to just `clientset`

pkg/project/v2/sort/topologysort/topologysort.go
d0weinberger, line 1: How is topologysort.go related to the refactoring in deploy and preload and (why) is it safe to delete?

Kirdock: When looking through some usages of config related properties, I realized that this file is not used at all, therefore I deleted it

pkg/deploy/preload_test.go
Kirdock: Yes, forgot to remove these.
Adjusted 👍

d0weinberger (dismissed):
LGTM

pkg/deploy/deploy.go
```diff
@@ -130,6 +128,14 @@ func Deploy(ctx context.Context, projects []project.Project, environmentClients
 	return nil
 }
 
+func Deploy(ctx context.Context, clientSet *client.ClientSet, projects []project.Project, sortedConfigs []graph.SortedComponent, environment string) error {
```
arthurpitman, line 131: As the `client.ClientSet` is created outside of this function (and even outside the loop in the calling function), is the memory used by caches actually freed after `Deploy(...)` is done? I guess we need to do this explicitly, either by ensuring that the client set goes out of scope and can be freed, or if this is not possible at the moment, by emptying the caches.

Kirdock: Added a clear function 👍

pkg/deploy/preload.go
```diff
@@ -54,6 +54,16 @@ func preloadCaches(ctx context.Context, projects []project.Project, clientSet *c
 	wg.Wait()
 }
 
+func clearCaches(clientSet *client.ClientSet) {
```
arthurpitman, line 57: I was considering if this should be moved to ClientSet, but considering how preload currently works, I think it makes more sense here.

arthurpitman (approved):
LGTM

PR #1835, opened 2025-04-08
refactor: remove deprecated dashboard-v2

Kirdock:
`dashboard-v2` is something that was only used in Monaco V1. In Monaco V2 it doesn't even work if `dashboard-v2` is provided, as this will lead to be an unknown API.
Therefore, I removed everything related to `dashboard-v2`

No


arthurpitman (approved):
LGTM

PR #1836, opened 2025-04-08
feat(dashboard): fail when dashboard-v1 payload is sent to v2 and vice versa

Kirdock:
Before the config is deployed the payload is validated.
- If it tries to deploy a classic dashboard to platform it will return an error
- If it tries to deploy a platform dashboard to classic it will return an error

Yes. An error will be logged with a better message indicating that the wrong payload (classic vs platform) was used


pkg/client/dtclient/config_client.go
```diff
@@ -897,3 +902,33 @@ func validateSloV1Payload(payload []byte) error {
...
+		// map should only be used in V2
+		if _, isMap := parsedPayload.Tiles.(map[string]any); isMap {
+			return errors.New("tried to deploy a classic dashboard configuration to platform dashboard")
+		}
+		return nil
+	}
+
+	// array should only be used in V1
+	if _, isArray := parsedPayload.Tiles.([]any); isArray {
+		return errors.New("tried to deploy a platform dashboard configuration to classic dashboard")
```
arthurpitman, line 930: Just wondering: do we use the term "platform dashboard" anywhere else? Would that mean anything to a user?

arthurpitman (approved):

pkg/deploy/internal/document/document.go
```diff
@@ -57,6 +58,12 @@ func Deploy(ctx context.Context, client Client, properties parameter.Properties,
 		return entities.ResolvedEntity{}, errors.New("missing name parameter")
 	}
 
+	if documentType == documents.Dashboard {
+		if valErr := dtclient.ValidateDashboardPayload([]byte(renderedConfig), false); valErr != nil {
```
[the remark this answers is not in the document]
Kirdock: With this there may be the wrong error message.
I decided against one function that only checks a single version and using the negation of it for the other one because "tiles are not an array" doesn't mean it's an object (V2), it could be a completely wrong payload and we would say "V1 instead of V2 was used" which will be wrong.
I could split it into two functions though

pkg/client/dtclient/config_client.go
Kirdock: I adjusted it a bit, to not mention "platform"

pkg/client/dtclient/config_client.go
```diff
@@ -897,3 +902,25 @@ func validateSloV1Payload(payload []byte) error {
 	}
 	return nil
 }
+
+var errWrongPayloadType = errors.New("can't deploy a Dynatrace platform dashboard using 'api: dashboard'. Either use 'type: document' with 'kind: dashboard' to deploy a Dynatrace platform dashboard or update your payload to a Dynatrace classic dashboard")
```
Kirdock, line 906: We should align the casing. Either `Classic/Platform` or `classic/platform`

pkg/client/dtclient/config_client.go
```diff
@@ -897,3 +902,25 @@ func validateSloV1Payload(payload []byte) error {
...
+	}
+
+	parsedPayload := DashboardKeys{}
+	err := json.Unmarshal(payload, &parsedPayload)
+	if err != nil {
+		return fmt.Errorf("failed to unmarshal dashboard payload: %w", err)
+	}
+
+	// Tiles should only be a map if Dynatrace platform dashboards configs are defined.
+	// For Dynatrace Platform Dashboards, an array is used.
```
Kirdock, line 921:
```suggestion
	// For Dynatrace classic dashboards, an array is used.
```

Laubi (approved):

PR #1837, opened 2025-04-08
refactor(deploy): move `resolveEntities` one lvl lower into `deployGr…

tomazpu:
…aph`


Kirdock (approved):
So the change is basically like this, right?
From:
image

To:
image

PR #1839, opened 2025-04-09
refactor: Rename package `v2` to `project`

arthurpitman:
As `project` now only contains `v2`, this PR moves everything one level up to `project` and ensures that this package name is used throughout the code instead of `v2`, `project`, `projectv2` or some other names.

The PR is done in two commits to make it easy to review; it should change nothing else and leaves tests as they are.


d0weinberger (approved):
LGTM

PR #1841, opened 2025-04-10
refactor: move deployment limiter into context

Kirdock:
Removes the global variable and puts limiter into the context

Coverage check may fail because of these two lines inside the `if`
```go
if options.ConcurrentDeploymentsLimiter != nil {
	options.ConcurrentDeploymentsLimiter.Acquire()
	defer options.ConcurrentDeploymentsLimiter.Release()
}
```
The question is: do we want to test that?
Add a test that just checks if acquire/release was called? how: replace the limiter struct inside the options with an interface and put a stub into the context, so that the call can be checked.
Should we write an integration test for it? how: stubbed clients that don't respond and we check if upserts were called {limitedConcurrentCount} times

No

### DIFF
--- a/cmd/monaco/deploy/command.go
+++ b/cmd/monaco/deploy/command.go
@@ -24,11 +24,13 @@ import (
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/cmdutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/completion"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/environment"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/files"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/report"
 	monacoVersion "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/version"
 )
@@ -36,7 +38,7 @@ import (
 func GetDeployCommand(fs afero.Fs) (deployCmd *cobra.Command) {
 	var dryRun, continueOnError bool
 	var manifestName string
-	var environment, project, groups []string
+	var selectedEnvs, project, groups []string
 
 	deployCmd = &cobra.Command{
 		Use:               "deploy <manifest.yaml>",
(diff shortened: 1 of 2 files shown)


d0weinberger (approved):
I had one remark about a comment, but other than that: LGTM

d0weinberger (dismissed):

cmd/monaco/deploy/command.go
```diff
@@ -56,11 +58,26 @@ func GetDeployCommand(fs afero.Fs) (deployCmd *cobra.Command) {
...
-			return deployConfigs(ctx, fs, manifestName, groups, environment, project, continueOnError, dryRun)
+			var limiter *rest.ConcurrentRequestLimiter
+			maxConcurrentDeployments := environment.GetEnvValueIntLog(environment.ConcurrentDeploymentsEnvKey)
+
+			if maxConcurrentDeployments > 0 {
+				log.Info("%s set, limiting concurrent deployments to %d", environment.ConcurrentDeploymentsEnvKey, maxConcurrentDeployments)
+				limiter = rest.NewConcurrentRequestLimiter(maxConcurrentDeployments)
+			}
+
+			deployOptions := deployoptions.DeployConfigsOptions{
```
Laubi, line 69: The `context` values should not be used for business logic. 
It makes sense to keep values in for logging, tracing, concurrency, etc.
Values like dry-run or continueOnErr fundamentally change how Monaco behaves. Having them in the context makes it really hard to see how code behaves.

Laubi (changes requested):

cmd/monaco/deploy/command.go
Laubi: Since this change is quite a lot of the PR, I will have a look again after this has been changed.

Kirdock: Changed 👍

pkg/deploy/deploy.go
```diff
@@ -62,18 +62,24 @@ type DeployConfigsOptions struct {
...
+	lock    sync.Mutex
 	errSkip = errors.New("skip error")
 )
 
-func DeployForAllEnvironments(ctx context.Context, projects []project.Project, environmentClients dynatrace.EnvironmentClients, opts DeployConfigsOptions) error {
-	maxConcurrentDeployments := environment.GetEnvValueIntLog(environment.ConcurrentDeploymentsEnvKey)
-	if maxConcurrentDeployments > 0 {
-		log.Info("%s set, limiting concurrent deployments to %d", environment.ConcurrentDeploymentsEnvKey, maxConcurrentDeployments)
-		concurrentDeploymentsLimiter = rest.NewConcurrentRequestLimiter(maxConcurrentDeployments)
+type ctxDeploymentLimiter struct{}
```
Laubi, line 69:
```suggestion
type ctxDeploymentLimiterKey struct{} // or deploymentLimiterKey
```

I did not find a general rule, but the convention is to use `...Key`: https://github.com/golang/go/issues/15229

cmd/monaco/deploy/command.go
```diff
@@ -56,11 +58,19 @@ func GetDeployCommand(fs afero.Fs) (deployCmd *cobra.Command) {
 				return err
 			}
 
-			return deployConfigs(ctx, fs, manifestName, groups, environment, project, continueOnError, dryRun)
+			maxConcurrentDeployments := environment.GetEnvValueIntLog(environment.ConcurrentDeploymentsEnvKey)
```
Laubi, line 61: Is this really necessary to have in the deploy here? The command should IMO not be responsible to set the deployment concurrency - it parses the argument and calls the correct code. I think this should move further down

Kirdock: I thought when it comes to env variable handling it should be rather on top. I'll move it to the previous place then :+1:

arthurpitman (approved):

pkg/deploy/deploy.go
```diff
@@ -62,17 +62,29 @@ type DeployConfigsOptions struct {
...
-	lock                         sync.Mutex
-	concurrentDeploymentsLimiter *rest.ConcurrentRequestLimiter
-
+	lock    sync.Mutex
 	errSkip = errors.New("skip error")
 )
 
+type ctxDeploymentLimiterKey struct{}
+
+func NewContextWithDeploymentLimiter(ctx context.Context, limiter *rest.ConcurrentRequestLimiter) context.Context {
```
Laubi, line 71: Can be private, as we don't expose the functionality

Kirdock: done ✅

Laubi (approved):

PR #1842, opened 2025-04-10
feat: bump golangci-lint and report so sonar

Kirdock:
Upgrade to golangci-lint to v2 and report to sonar.
The new `.golangci.yaml` file was generated via `golangci-lint migrate`. More details [here](https://golangci-lint.run/product/migration-guide/)

**In action:**
I commited a lint error, and it was correctly detected
image

At the moment, we have 70 issues compared to about 33 before (new ones are addressed with https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1843) due to the new `staticcheck` lint rule, which is enabled by default.

**Why the upgrade to golangci-lint v2:**
Easier to generate reports due to new arguments like `--output.checkstyle.path`.

**Why are there so many linters now in the lint file?**
The reason why we have more linters in the `enabled` section is, that the `preset` property is removed now. That's why the `bugs` and `unused` presets  we had were split into the individual linters. https://golangci-lint.run/product/migration-guide/#linterspresets

**Why not 3 workflow files? Why are lint, test and sonar report merged?**
Theoretically, each workflow can upload its artifacts BUT the workflow (Sonar report) that should download them would be cumbersome or filled with workarounds.
The Sonar workflow would need the `run-id` of a workflow in order to download its artifacts. This could be obtained via `github.event.workflow_run.id `. The problem here is that either `lint` or `test` triggered the Sonar workflow, so only one artifact could be downloaded during the workflow run and because there is a new environment on each workflow run, there isn't a nice and clean way to download both artifacts, except downloading the artifact of the triggering workflow and uploading it to the Sonar one.

NONE


arthurpitman (approved):

.github/workflows/build-lint-test.yml
```diff
@@ -0,0 +1,137 @@
+name: Build, Lint and Test
```
Laubi, line 1: Absolute nitpick: Please always use the Oxford comma.
```suggestion
name: Build, Lint, and Test
```
Or `Build, Lint, Test`

.github/workflows/build-lint-test.yml
```diff
@@ -0,0 +1,137 @@
...
+        uses: actions/download-artifact@95815c38cf2ff2164869cbab79da8d1f422bc89e #v4.2.1
+        with:
+          name: lint-report
+
+      - name: ⬆️ SonarQube Scan
+        uses: SonarSource/sonarqube-scan-action@aa494459d7c39c106cc77b166de8b4250a32bb97 # nosemgrep false detection of commit v5.1.0
+        env:
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+
+  upload_event:
```
Laubi, line 129: This event file is used for the `build-test--collect-results.yml` `Test Results` GitHub workflow. 
The referenced workflow triggers upon the deleted workflows, so needs to be changed as well.
This event file here is used to populate the response of the referenced workflow.

Laubi (changes requested):
The `Test Results` workflow needs to be fixed as well

.github/workflows/build-lint-test.yml
Kirdock: Is this something that is deprecated? We don't have any `build-test--collect-results.yml` in our repo. So the upload_event can be removed I guess?

Laubi: Yes, the file has been removed before, sorry for that. The event file upload should not be required anymore, please double check. If not, please remove it.

Laubi (approved):

PR #1844, opened 2025-04-16
test(account): Update existing E2E tests to include service users

arthurpitman:
This PR updates the existing account management E2E tests to include service users.

### DIFF
--- a/cmd/monaco/integrationtest/account/deploy-download_test.go
+++ b/cmd/monaco/integrationtest/account/deploy-download_test.go
@@ -77,39 +77,60 @@ func TestIdempotenceOfDeployment(t *testing.T) {
 	deploy1st := deploy(project, baseFs)
 	download1st := download(project, afero.NewCopyOnWriteFs(baseFs, afero.NewMemMapFs()))
 
+	allDeployedItemsDownloaded := true
 	for _, u := range deploy1st.Users {
-		assert.Contains(t, download1st.Users, u.Email.Value())
+		if !assert.Contains(t, download1st.Users, u.Email.Value()) {
+			allDeployedItemsDownloaded = false
+		}
 	}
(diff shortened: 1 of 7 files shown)


cmd/monaco/integrationtest/account/deploy-download_test.go
```diff
@@ -72,6 +74,9 @@ func TestIdempotenceOfDeployment(t *testing.T) {
 	for _, u := range deploy1st.Users {
 		assert.Contains(t, download1st.Users, u.Email.Value())
 	}
+	for _, deployedServiceUser := range deploy1st.ServiceUsers {
```
Kirdock, line 77: We can remove this loop, because we are testing it already in the loop below

arthurpitman: That's a good question. More to the point, should the test stop here if this fails? Before the second deploy and then download? Here we are interested in did we manage to download what we think we deployed, where as the second time we ultimately want to check did we affect something by re-deploying.

Kirdock: I would vote for `require`. Then I could leave the rest as it is 👍

arthurpitman: Ok, now I see - we probably shouldn't use require, otherwise delete won't run, which is cleaning the test up

Kirdock: We could also `defer` the delete instead

arthurpitman: I was looking into that vs `t.Cleanup(...)`. We currently use `defer` though, so I think I'll go with that for now.

Kirdock (approved):

PR #1845, opened 2025-04-16
refactor(download): Introduce Downloadable interface and use with buckets

d0weinberger:

This is the first of a series of PRs with the goal of making the download code - and later also the deploy and delete code - independent of the concrete feature/API. We introduce a `Downloadable` interface and use for already for downloading buckets. The goal is to use it for all the other clients as well. As soon as all clients adhere to `Downloadable`, we can refactor/simplify `download_configs.go`

NONE


cmd/monaco/download/download_configs.go
```diff
@@ -274,18 +274,24 @@ type downloadFn struct {
...
+
 var defaultDownloadFn = downloadFn{
-	classicDownload:      classic.Download,
-	settingsDownload:     settings.Download,
-	automationDownload:   automation.Download,
-	bucketDownload:       bucket.Download,
+	classicDownload:    classic.Download,
+	settingsDownload:   settings.Download,
+	automationDownload: automation.Download,
+	bucketDownload: func(bucketClient client.BucketClient) Downloadable {
```
Laubi, line 292: FYI for other reviewers: I suggested this to keep the changeset minimal for the start. When we moved more stuff, we can improve this file/package more easily.

Kirdock (dismissed):

cmd/monaco/download/download_configs.go
```diff
@@ -273,22 +273,28 @@ func escapeGoTemplating(c *config.Config) error {
...
 	automationDownload   func(context.Context, client.AutomationClient, string, ...config.AutomationType) (project.ConfigsPerType, error)
-	bucketDownload       func(context.Context, client.BucketClient, string) (project.ConfigsPerType, error)
+	bucketDownload       func(client.BucketClient) Downloadable
 	documentDownload     func(context.Context, client.DocumentClient, string) (project.ConfigsPerType, error)
 	openPipelineDownload func(context.Context, client.OpenPipelineClient, string) (project.ConfigsPerType, error)
 	segmentDownload      func(context.Context, segment.DownloadSegmentClient, string) (project.ConfigsPerType, error)
 	sloDownload          func(context.Context, slo.DownloadSloClient, string) (project.ConfigsPerType, error)
 }
 
 var defaultDownloadFn = downloadFn{
```
arthurpitman, line 291: When we are done with all "features" we should rename this as now calling `bucketDownload(...)` returns a function that can download rather than actually downloading. But while this is in progress this is fine 😃

cmd/monaco/download/download_configs.go
```diff
@@ -273,22 +273,28 @@ func escapeGoTemplating(c *config.Config) error {
 	return nil
 }
 
+type Downloadable interface {
+	Download(ctx context.Context, projectName string) (project.ConfigsPerType, error)
```
arthurpitman, line 277: Should we document the contact of what this function should do? For example I notice at the moment this returns an error, but we are already copying across the behavior of just logging the error inside the function and returning `nil`. Ideally we'd start enforcing the contract now for each config type we move over.

I think we should also document what the `projectName` is for, essentially all configs will have project set to this right?

cmd/monaco/download/download_configs.go
d0weinberger: Yes, exactly, that's the plan :)

cmd/monaco/download/download_configs.go
d0weinberger: Good point ... should we really return `nil` from the bucket download, even though there was an error? I really just copied it without caring much about the implementation details.
I will write a comment.

Kirdock (approved):

PR #1847, opened 2025-04-17
fix(buckets): incorrect template escaping

Kirdock:
On download, If the bucket had a displayName, it was incorrectly escaped in the template.
Before
```json
{
  "displayName": "{{`{{`}}.displayName{{`}}`}}",
}
```

Now
```json
{
  "displayName": "{{.displayName}}",
}
```

Yes. Downloading now creates the right payload and parameters


pkg/download/bucket/testdata/buckets.json
```diff
@@ -6,7 +6,7 @@
             "displayName": "Default metrics (15 months)",
             "status": "active",
             "retentionDays": 462,
-            "metricInterval": "PT1M",
+            "metricInterval": "{{.PT1M}}",
```
arthurpitman, line 9: Is this example purely fictitious? To show that something has `{{` ... `}}` will be escaped? Or is this something that actually happens?

arthurpitman (approved):

pkg/download/bucket/testdata/buckets.json
Kirdock: It's fictional, to test if the escaping works.
At the moment, there isn't a way to set something like `{.something}` except fot the displayName. The other (optional) properties don't allow that

PR #1848, opened 2025-04-17
feat: Enable `MONACO_FEAT_SERVICE_USERS` and `MONACO_FEAT_ACCESS_CONTROL_SETTINGS` feature flags by default

arthurpitman:
This PR enables the `MONACO_FEAT_SERVICE_USERS` feature flag by default, so that service users can be downloaded and deployed  with other account resources.

It also enables the `MONACO_FEAT_ACCESS_CONTROL_SETTINGS` feature flag by default, so that read/write permissions for settings objects can be managed via Monaco for settings schemas which support it.


d0weinberger (approved):
LGTM

PR #1849, opened 2025-04-17
fix: Use `serviceUser` or `serviceUsers` in account resource and delete YAML files

arthurpitman:
The PR updates account management to use `serviceUser` or `serviceUsers` in account resource and delete YAML files to be consistent with our usage of camel rather than kebab case in Monaco.


d0weinberger (approved):
LGTM

PR #1854, opened 2025-04-23
refactor(segments): Use `Downloadable` interface for segments

d0weinberger:

This is the second of a series of PRs with the goal of making the download code - and later also the deploy and delete code - independent of the concrete feature/API. The Downloadable interface is used for downloading segments. The goal is to use it for all the other clients as well. As soon as all clients adhere to Downloadable, we can refactor/simplify download_configs.go

NONE


d0weinberger:
Regarding the code duplications: I don't think it's worthwile to generify creation of configs from api responses across all apis (yet).

pkg/feature/segment/segment.go
```diff
@@ -34,10 +34,18 @@ type DownloadSegmentClient interface {
 	GetAll(ctx context.Context) ([]segments.Response, error)
```
tomazpu, line 34: Just thinking about the folder structure, would it make more sense to call it `pkg/resource/segment/download.go` ?

I think that `feature` is not descriptive enough, and after we start moving in other (deploy, delete) stuff, the name segment also becomes misleading. WDYT?

Laubi (approved):

tomazpu (approved):

PR #1859, opened 2025-04-28
test: remove not needed settings/config tests and move one to the config render test

Kirdock:
The settings and config client had some tests in them that haven't had any use. Rendering of templates and validating cyclic dependencies doesn't happen inside the clients but in the config render and sort.
So
- Removed the unused tests
- Added a new test to the config tests to check if invalid JSON returns an error on template render
- No additional test for the cyclic dependency needed, as this is already tested in `sort_test.go`

NONE


arthurpitman (approved):
Looks good!

Optional idea: tighten the error check in `func TestSortParametersShouldFailOnCircularDependency(t *testing.T)` in `pkg/config/sort_test.go`: I think the errors should be of type `config.CircularDependencyParameterSortError`.

pkg/config/sort_test.go
```diff
@@ -102,7 +105,12 @@ func TestSortParametersShouldFailOnCircularDependency(t *testing.T) {
 
 	_, errs := getSortedParameters(c)
 
-	assert.True(t, len(errs) > 0, "should fail")
+	require.Greater(t, len(errs), 0, "should fail")
```
arthurpitman, line 108:
```suggestion
	require.NotEmpty(t, errs, "should fail")
```

PR #1860, opened 2025-04-28
Refactor/minor renamings in segment and bucket download

d0weinberger:

This PR is a follow-up to #1854 
It features renaming `segment.SegmentAPI` to `segment.Api`, `bucket.BucketAPI` to `bucket.API` and introducing a `Source` interface for the bucket download

NONE


arthurpitman (dismissed):
Only nit: Should we go with `API` rather than `Api`? In general golang writes acronyms like this.

d0weinberger:
Renamed `Api` to `API`

arthurpitman (approved):

PR #1862, opened 2025-04-28
refactor: use different endpoint for classic connection check

Kirdock:
For validating a classic client, the cluster version was retrieved in order to check if the connection is actually working.
The downside of this is, that an additional scope (`DataExport`) is needed for this, which is not optimal.
The connection check is now changed to use the token lookup endpoint. This one doesn't require any specific scopes.
`DataExport` is not required anymore but still optional, as this is still needed for checking deprecated cluster versions that are not compatible with settings.

Tested download and upload of the downloaded items without the DataExport token 👍 

- What exactly happens if a settings object is sent to a cluster that is below the suggested cluster version?
- FYI golangci-lint is a false positive

The `DataExport` scope is not required anymore

### DIFF
--- a/pkg/client/apitoken/client.go
+++ b/pkg/client/apitoken/client.go
@@ -1,0 +1,86 @@
+/*
+ * @license
+ * Copyright 2025 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package apitoken
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	coreapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	corerest "github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+)
+
+const apiTokenPath = "/api/v2/apiTokens"
+
+type Response struct {
+	ID                  string            `json:"id"`
+	Name                string            `json:"name"`
+	Enabled             bool              `json:"enabled"`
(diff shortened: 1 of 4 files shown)


pkg/client/apitoken/client.go
```diff
@@ -0,0 +1,86 @@
...
+	"io"
+	"net/http"
+
+	coreapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	corerest "github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+)
+
+const apiTokenPath = "/api/v2/apiTokens"
+
+type Response struct {
```
arthurpitman, line 33: I was unable to find out which properties of the response are required (if any). Did you manage to find anything beyond https://upd51752.dev.dynatracelabs.com/rest/v2/rest-api-docs/v2/spec3.json?

arthurpitman, line 33: The spec `.../rest/v2/rest-api-docs/v2/spec3.json` doesn't mention which fields are required, but it should be fine if the defaults are OK 👍

Kirdock: I just created a token and validated the fields that are given. If something is not required it won't hurt us, as it will then be the empty value of the type (we don't have a strict validation like in account-management)

pkg/client/apitoken/client.go
```diff
@@ -0,0 +1,86 @@
...
+	resp, err := client.POST(ctx, apiTokenPath+"/lookup", bytes.NewReader(body), corerest.RequestOptions{CustomShouldRetryFunc: corerest.RetryIfTooManyRequests})
+
+	if err != nil {
+		return Response{}, fmt.Errorf("failed to query token metadata: %w", err)
+	}
+
+	response, err := coreapi.NewResponseFromHTTPResponse(resp)
+
+	if err != nil {
+		return Response{}, fmt.Errorf("failed to query token metadata: %w", err)
```
d0weinberger, line 75: You use the same error message here as above when the POST fails.
I think, a different error message should be used to differentiate in the logs between the two cases - failing POST and failing response parsing.

Kirdock: We wrap the error in both cases, so it should be fine.
Anyway, one error is an error during the request (e.g., lost connection) and another one is the status code validation.
I can't think of a good error message for now, maybe you have a better idea for a message because technically both errors relate to a failed request

Kirdock: After some discussion, we decided to leave the message as it is and maybe improve the overall structure in the future for all clients

PR #1868, opened 2025-04-30
test: add tests for dependency resolvers to increase coverage

Kirdock:
It adds some tests for the ahocorasick and basic resolver. The tests inside dependency_resolution are actually not detecting the coverage of the resolvers (probably due to different package/folder). Tests are based on the ones of `dependency_resolution_test`. It's basically a subset of `dependency_resolution_test` and the `ahocorasick resolver` uses the same tests as the `basic resolver`.

There is one thing inside `dependency_resolution_test` that I didn't quite understand. Is it intended that the `Dashboards should not be able to reference...` test uses `dashboard-id` as the config ID for `dashboard 1` AND the same id for the `dashboard-setting`?

NO

### DIFF
--- a/pkg/download/dependency_resolution/resolver/ahocorasick_dep_resolver_test.go
+++ b/pkg/download/dependency_resolution/resolver/ahocorasick_dep_resolver_test.go
@@ -1,0 +1,306 @@
+/*
+ * @license
+ * Copyright 2025 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package resolver_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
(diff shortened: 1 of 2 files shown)


arthurpitman:
About `"Dashboards should not be able to reference a dashboard-share-setting, even if it's the dashboard's share setting"` - I feel we need to explain this test further. I wasn't immediately able to trigger it to fail after commenting out this:

https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/d0836b4c217ef5da6009b9a42b3872975db2d8d4/pkg/download/dependency_resolution/resolver/shared.go#L112-L119

I don't think it strictly matters that the same string is used for the config ID for both the dashboard and dashboard share settings, but I think a real download would use `dashboard-iddashboard-id`.

Kirdock:
> About `"Dashboards should not be able to reference a dashboard-share-setting, even if it's the dashboard's share setting"` - I feel we need to explain this test further. I wasn't immediately able to trigger it to fail after commenting out this:
> 
> https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/d0836b4c217ef5da6009b9a42b3872975db2d8d4/pkg/download/dependency_resolution/resolver/shared.go#L112-L119
> 
> I don't think it strictly matters that the same string is used for the config ID for both the dashboard and dashboard share settings, but I think a real download would use `dashboard-iddashboard-id`.

The critical part I was thinking about is this one
https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/d0836b4c217ef5da6009b9a42b3872975db2d8d4/pkg/download/dependency_resolution/dependency_resolution.go#L98

In our current `Dashboards should not be able to reference a dashboard-share-setting...` test, the dashboard has the same ID as the shared-dashboard, which would lead to overriding the first one right?

arthurpitman:
I have created a follow-up for us to describe and improve the tests in `pkg/download/dependency_resolution/dependency_resolution_test.go`

arthurpitman (approved):

PR #1869, opened 2025-04-30
refactor(classic): use `Downloadable` interface for classic download

d0weinberger:

This is one PR in a series of PRs with the goal of making the download code independent of the concrete feature/API.

To do so, the `Downloadable` interface, is used for downloading classic configs. When, in the end, all resources are downloaded using this interface, we can simplify `download_configs.go` to only depend on `Downloadable` instead of concrete resource APIs. 
Also, a `Source` interface is introduced to the classic download to make it independent of the `ConfigClient` interface

Arguably, the download code could still be simplified and refactored quite a bit. I don't think this should be done in this PR - it might make the changeset a bit complex. What do you think?

NONE


arthurpitman (approved):
Looks good!
The only maybe would be to update the message of the first commit:

refactor(download): Move contents of `download/classic` to `resource/classic`

PR #1871, opened 2025-04-30
refactor(settings): use `Downloadable` interface for settings download

d0weinberger:

This is one PR in a series of PRs with the goal of making the download code independent of the concrete feature/API.

To do so, the `Downloadable` interface, is used for downloading settings. When, in the end, all resources are downloaded using this interface, we can simplify `download_configs.go` to only depend on `Downloadable` instead of concrete resource APIs.
Also, a `Source` interface is introduced to the settings download to make it independent of the `SettingsClient` interface

Arguably, the download code could still be simplified and refactored quite a bit. I don't think this should be done in this PR - it might make the changeset a bit complex. What do you think?

NONE


pkg/resource/settings/download.go
```diff
@@ -50,29 +49,45 @@ type schema struct {
...
+}
+
+type API struct {
+	settingsSource Source
+	filters        Filters
+	schemaIDs      []config.SettingsType
+}
+
+func NewAPI(settingsSource Source, filters Filters, schemaIDs ...config.SettingsType) *API {
+	return &API{settingsSource, filters, schemaIDs}
```
Kirdock, line 65: Because we aren't using `config.SettingsType` in any way except mapping it to a different data-structure, would it make sense to put the settingsType => string mapping in here instead of Download?

tomazpu (approved):

PR #1875, opened 2025-05-06
ci: use golangci-lint and remove double lint reporting

Kirdock:
#### **Why** this PR?
Regarding golangci-lint:
The golangci-lint action is from the maintainers of golangci-lint; it's the suggested way to use golangci-lint, and because of some incidents in the past, Reviewdog should not be used.

Regarding reporting:
Because Sonar does its reporting, we don't need our lint action to report (and fail) an additional time. Therefore, it's enough if the lint report is generated, which is everything Sonar needs. This also adds the ability to ignore specific errors that are reported not to occur again, once ignored.

#### **What** has changed?
- Reviewdog is now removed, and the golangci-lint action is used instead.
- The double reporting is gone, and Sonar will only be reporting.

#### **How** does it do it?
- Using the golangci-lint action.
- Adding an argument not to fail the pipeline if there are any errors.
- Removing the GitHub token from it, as we don't need the additional reporting tools.

#### How is it **tested**?
The lint report is generated. An additional commit with linting errors showed that the action did not fail, and Sonar correctly detected it and failed the pipeline.

Doesn't affect users

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -82,14 +82,11 @@ jobs:
         run: make vet
 
       - name: 🔎 golangci-lint
-        uses: reviewdog/action-golangci-lint@f9bba13753278f6a73b27a56a3ffb1bfda90ed71 #v2.8.0
+        uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 #v8.0.0
         with:
-          go_version_file: go.mod
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          reporter: github-pr-check
-          golangci_lint_version: v2.0.2
-          # the name has to be as the same as the one specified in "sonar-project.properties" and in the upload artifact step
-          golangci_lint_flags: "--output.checkstyle.path=lint-report.xml"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-lint-test.yml
```diff
@@ -75,21 +75,17 @@ jobs:
         with:
           go-version-file: go.mod
 
+      - name: Install golangci-lint
+        run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(go env GOPATH)/bin v2.1.6
```
arthurpitman, line 79: Would downloading the executable directly be an option, similar to https://github.com/golangci/golangci-lint-action/blob/main/src/install.ts#L152? I am unsure about running a script with `HEAD`.

arthurpitman: OR could we even use the official GitHub action and pass it the command line args?

Kirdock: I'll go with the second one 👍

arthurpitman (approved):

tomazpu (approved):

PR #1876, opened 2025-05-06
feat(manifest): `loader.Load(...)` includes unselected environments as skipped with `Skip: true`

arthurpitman:
#### **Why** this PR?
In order to be able to check environment and group overrides in `project.LoadProjects(...)`, something that will be done in a follow up PR, the loaded manifest needs to contain _at least_ the names of **all** environment and groups, and not just those selected to be deployed or deleted.

#### **What** has changed?
An additional field, `Skip` is added to `manifest.EnvironmentDefinition`:
- With `Skip` set to `false`, the environment should be used as previously; it is not skipped and therefore is selected for deployment or deletion.
- With `Skip` set to `true`, the environment is loaded within its group, but is not selected for deployment or deletion. Such an environment won't have its URL or credentials resolved.

#### **How** does it do it?
- `loader.Load(...)` now loads all environments defined in the manifest, setting the `Skip` field appropriately.
- Small changes are made to `dynatrace.VerifyEnvironmentGeneration(...)`, `dynatrace.CreateEnvironmentClients(...)`, `project.toEnvironmentSlice(...)` to make them skip non-Enabled environments.

#### How is it **tested**?
- Existing tests are updated, predominantly to add the `Skip: true` for environments that should be skipped.
- Two new tests are added: `TestLoadConfigFile_ConfigsGeneratedForNonSkippedEnvironments` and `TestLoadConfigFile_NoConfigsGeneratedForSkippedEnvironments`

It has no effect on users.

### DIFF
--- a/cmd/monaco/delete/delete.go
+++ b/cmd/monaco/delete/delete.go
@@ -38,7 +38,7 @@ import (
 func Delete(ctx context.Context, environments manifest.Environments, entriesToDelete delete.DeleteEntries) error {
 	var envsWithDeleteErrs []string
 	for _, env := range environments {
-		if !env.Enabled {
+		if env.Skip {
 			continue
 		}
 		ctx := context.WithValue(ctx, log.CtxKeyEnv{}, log.CtxValEnv{Name: env.Name, Group: env.Group})
--- a/cmd/monaco/delete/delete.go
+++ b/cmd/monaco/delete/delete.go
@@ -38,6 +38,9 @@ import (
 func Delete(ctx context.Context, environments manifest.Environments, entriesToDelete delete.DeleteEntries) error {
 	var envsWithDeleteErrs []string
 	for _, env := range environments {
+		if !env.Enabled {
+			continue
+		}
 		ctx := context.WithValue(ctx, log.CtxKeyEnv{}, log.CtxValEnv{Name: env.Name, Group: env.Group})
 		if containsPlatformTypes(entriesToDelete) && env.Auth.OAuth == nil {
 			log.WithCtxFields(ctx).Warn("Delete file contains Dynatrace Platform specific types, but no oAuth credentials are defined for environment %q - Dynatrace Platform configurations won't be deleted.", env.Name)
--- a/pkg/manifest/loader/manifest_loader.go
+++ b/pkg/manifest/loader/manifest_loader.go
@@ -377,9 +377,10 @@ func parseEnvironments(context *Context, groups []persistence.Group) (map[string
 			}
 			envNames[env.Name] = true
 
-			// skip loading if environments is not empty, the environments does not contain the env name, or the group should not be included
+			// skip actually loading if environments is not empty, the environments does not contain the env name, or the group should not be included
 			if shouldSkipEnv(context, group, env) {
 				log.WithFields(field.F("manifestPath", context.ManifestPath)).Debug("skipping loading of environment %q", env.Name)
+				environments[env.Name] = manifest.EnvironmentDefinition{Enabled: false, Name: env.Name, Group: group.Name}
 				continue
 			}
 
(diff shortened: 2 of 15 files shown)


arthurpitman:
I'm pretty satisified that E2E tests are working with this change, although I've been unable to get them all green in one run here.

Kirdock:
At the moment, we have a small behavior change here
https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1876/files#diff-f0178ccba6ea85e4b333a9ba0913163f69960f5361186b218cab7b25d5936d1cR188
If all definitions are skipped, it resulted in an error before.

Here is also a behavior change, right? https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/485b4087cb5803e9f30f90b5f617368909df32df/cmd/monaco/deploy/deploy.go#L159
Previously, it was marked as an undefined environment

Not sure about this though
https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/485b4087cb5803e9f30f90b5f617368909df32df/pkg/project/project_loader.go#L123

Anyway, it would be nice if @Laubi or @tomazpu could also have a look, as this touches a lot of places :)

pkg/manifest/loader/manifest_loader.go
```diff
@@ -377,9 +377,10 @@ func parseEnvironments(context *Context, groups []persistence.Group) (map[string
 			}
 			envNames[env.Name] = true
 
-			// skip loading if environments is not empty, the environments does not contain the env name, or the group should not be included
+			// skip actually loading if environments is not empty, the environments does not contain the env name, or the group should not be included
 			if shouldSkipEnv(context, group, env) {
 				log.WithFields(field.F("manifestPath", context.ManifestPath)).Debug("skipping loading of environment %q", env.Name)
+				environments[env.Name] = manifest.EnvironmentDefinition{Enabled: false, Name: env.Name, Group: group.Name}
```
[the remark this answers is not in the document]
arthurpitman: So `Skip: true` would mean "don't use this environment"? I like that idea 👍

arthurpitman:
@Kirdock, you're correct: we need to clarify what the expected behavior is. The code:

https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/485b4087cb5803e9f30f90b5f617368909df32df/cmd/monaco/deploy/deploy.go#L157-L161

doesn't really work as configs are only created for environments in the loaded manifest, which means no configs will violate this. There is even a test for this:

https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/485b4087cb5803e9f30f90b5f617368909df32df/cmd/monaco/deploy/deploy_test.go#L166-L181

pkg/manifest/loader/manifest_loader.go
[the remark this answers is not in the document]
arthurpitman: Done 👍

arthurpitman:
In response to `sonarqubecloud`, I'll increase test coverage, particularly around `loader.LoadConfigFile(...)`.

cmd/monaco/delete/delete.go
```diff
@@ -38,6 +38,9 @@ import (
 func Delete(ctx context.Context, environments manifest.Environments, entriesToDelete delete.DeleteEntries) error {
 	var envsWithDeleteErrs []string
 	for _, env := range environments {
+		if env.Skip {
```
Laubi, line 41: We now have a lot of these `env.Skip` checks (7), and I am thinking if it would be possible to not even pass them into the consuming functions, aka filter them beforehand.

tomazpu: I was wondering the same. Would it be possible to add to the `Manifest` struct something like a getter method where only active Environments are returned?

arthurpitman: This is a fair point - let's do this!

I just tried with an iterator, but it doesn't really solve the problem: defining it on the `Manifest` would require changing many function signatures (including this one), definition it on `manifest.Environments` doesn't improve the problem, as you could still forget to use it. So I think I'll go for an actual getter on the `Manifest` that does the filtering when called.

PR #1878, opened 2025-05-08
fix(settings): Fix possible bucket not found error due to not updated datasource

Kirdock:
#### **Why** this PR?
Some endpoints/services use a different endpoint to get/validate the buckets, which could result in a "not found" error, even though our data source tells us the bucket is available.
[Related failed pipeline run](https://github.com/Dynatrace/dynatrace-configuration-as-code/actions/runs/14901283390/job/41853568830)

#### **What** has changed?
The number of retries until it stops trying to create/update a setting that has a bucket reference.

#### **How** does it do it?
By increasing the number of retries from 6 to 12

#### How is it **tested**?
No additional logic was added; just one constant is increased.

A not-found error is less likely to occur when creating settings objects referencing buckets.

### DIFF
--- a/pkg/deploy/internal/setting/setting.go
+++ b/pkg/deploy/internal/setting/setting.go
@@ -67,7 +67,9 @@ func Deploy(ctx context.Context, settingsClient client.SettingsClient, propertie
 	}
 
 	if c.HasRefTo(string(config.BucketTypeID)) {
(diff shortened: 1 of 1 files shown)


d0weinberger (approved):
The change is simple. If increasing MaxRetries fixes the problem and doesn't introduce new ones, I'm fine with it.

arthurpitman (approved):

PR #1879, opened 2025-05-08
refactor(download): downloadFn is removed from download_configs.go 

d0weinberger:
#### **Why** this PR?
This PR is another step in refactoring the download code to make it easier to extend it for new resource types. 

#### **What** has changed?

The core `downloadConfigs` method does not depend on concrete resource types anymore - it only uses `Downloadable`. This also allowed us to get rid of `downloadFn`. 

#### **How** does it do it?

A `prepareDownloadables` function is introduced, which constructs all the resource APIs, based on the `ClientSet` and the `downloadConfigsOptions`. In `downloadConfigs`, this factory function is called, we iterate over all `Downloadable` instances, call `Download`, and collect and return the resulting configs. This is in contrast to the previous implementation, where constructing the "client" and downloading the configs all happened in one go. The new implementation would allow for parallel downloading, if we'd wish to do that.

#### How is it **tested**?

For `TestDownload_Options`, instead of checking, which download functions have been called on `downloadConfigs`, we check, which Downloadable instances have been created by `prepareDownloadables`.
No new tests were added. I will run e2e tests.

It doesn't.

### DIFF
--- a/cmd/monaco/download/download_configs.go
+++ b/cmd/monaco/download/download_configs.go
@@ -306,7 +306,6 @@ func downloadConfigs(ctx context.Context, clientSet *client.ClientSet, apisToDow
 	configs := make(project.ConfigsPerType)
 	if opts.onlyOptions.ShouldDownload(OnlyApisFlag) {
 		if opts.auth.Token != nil {
-			log.Info("Downloading configuration objects")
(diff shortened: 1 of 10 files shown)


Kirdock (approved):
Nice work!

arthurpitman (approved):

PR #1880, opened 2025-05-08
feat(config): Warn when undefined environments or groups are used in overrides

arthurpitman:
#### **Why** this PR?

When copying and pasting configs between projects, it is easy to forget to update any overrides. Currently, this is silent: the overrides simply have no effect, which can have undesirable consequences. 

#### **What** has changed?

With this PR, a warning is emitted when an override references an environment or environment group that is not defined in the manifest.

A decision was made not to emit an error, at least for now, as this would be a breaking change.

#### **How** does it do it?

This PR builds on https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1897, which adds the names of all environments and groups to the loaded manifest, regardless of whether they're selected to be deployed or not.

In this PR, when a config entry is loaded, the environment or group name of each override is checked against those in the manifest.

#### How is it **tested**?

- Existing tests are updated
- Two extra tests are added to that explicitly check that warnings are logged:
  - `TestLoadProjects_GroupOverrideWithUndefinedGroupProducesWarning`
  - `TestLoadProjects_EnvironmentOverrideWithUndefinedEnvironmentProducesWarning`

Users will now receive a warning when an override references an environment or environment group that is not defined in the manifest.

### DIFF
--- a/pkg/config/loader/config_entry_loader.go
+++ b/pkg/config/loader/config_entry_loader.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/spf13/afero"
 
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/coordinate"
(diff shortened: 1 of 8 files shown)


pkg/config/loader/config_entry_loader.go
```diff
@@ -55,7 +56,19 @@ func parseConfigEntry(
 		return nil, []error{newDefinitionParserError(configId, singleConfigContext, err.Error())}
 	}
 
+	for _, group := range definition.GroupOverrides {
```
tomazpu, line 59: I would move each check into a separate function to decrease the complexity and have an opportunity to comment on what is checked under which assumption.

My opinion you decide if it makes sense.

arthurpitman: I have also improved the actual log message.

tomazpu (approved):

Kirdock (approved):

d0weinberger (approved):

PR #1882, opened 2025-05-09
chore(deps): update core lib

Kirdock:
#### **Why** this PR?
To make use of the new retry options, we're updating the core lib.

#### **What** has changed?
Updates the core lib to use the retry delay and max retries for the rest client's CRUD operations.

#### **How** does it do it?
By running `go get` with the latest commit of core.

#### How is it **tested**?
Current tests are executed again.

Doesn't affect users.


Kirdock:
There is already another PR that takes care of this

PR #1883, opened 2025-05-09
refactor(download): a set of smaller code improvements

d0weinberger:
#### **Why** this PR?
While working in the cmd/download package, I came across some low-hanging fruits for refactoring. Following the boy scouts rule, I cleaned up a bit.

#### **What** has changed?

There are three distinct commits and improvements:
 - There was a check in `download_command.go`, which checked whether upload to the same environment was possible. This function was only ever called from `download_configs.go` and felt out-of-place in `download_command.go`. Therefore I moved it.
 - The construction of the `auth` struct from `mapToAuth()` is moved to its own file, `auth.go`.
 - `makeSettingTypes` was removed, as it is unnecessary to wrap the schema IDs in `SettingsType` structs when the very next operation unpacks them again to `[]string` 

#### **How** does it do it?

Three separate commits for each improvement

#### How is it **tested**?

Tests still work, no new tests were added.

It doesn't.

### DIFF
--- a/cmd/monaco/download/auth.go
+++ b/cmd/monaco/download/auth.go
@@ -1,0 +1,65 @@
+/*
+ * @license
+ * Copyright 2025 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package download
+
+import (
+	"fmt"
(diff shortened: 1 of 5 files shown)


cmd/monaco/download/auth.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/secret"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest"
+)
+
+type auth struct {
```
arthurpitman, line 27: This commit I'm not too sure about, as I think it should go further: As the `auth` type is only used in `downloadCmdOptions`, why not include the fields in that type directly and place the logic there? Or with similar code:
https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/363b8742e179e42347a2e01e24f9355754f5a7e0/cmd/monaco/download/download_configs.go#L128-L143

arthurpitman (approved):
Looks good, maybe my comment above can be addressed in a follow up.

d0weinberger:
Thanks @arthurpitman , I'll try further suggested simplifications in a follow-up

PR #1887, opened 2025-05-13
refactor(segment): Update core lib where `api/clients/segments` is removed

tomazpu:
#### **Why** this PR?
This PR updates the core lib version and removes the `segment.Response` from the code base to unify and standardize our responses.

#### **What** has changed?
Updated core lib version where `segment.Response` was removed and the `api/clients/segment` was also removed. This change also needed a change in the monaco codebase as `segment.Response` was used throughout the codebase, and is now replaced by `api.Response`.

#### How is it **tested**?
All unit tests passed, a deployment for segments was done on the command line.

This is a purely technical update that has no effect on users.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -9,8 +9,6 @@ github.com/buger/jsonparser v1.1.1/go.mod h1:6RYKKt7H4d4+iWqouImQ9R2FZql3VbhNgx2
 github.com/cpuguy83/go-md2man/v2 v2.0.6/go.mod h1:oOW0eioCTA6cOiMLiUPZOpcVxMig6NIQQ7OS05n1F4g=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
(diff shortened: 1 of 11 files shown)


arthurpitman (dismissed):

go.sum
```diff
@@ -11,6 +11,8 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/dynatrace/dynatrace-configuration-as-code-core v0.8.1-0.20250424104703-ba9c6e8f54a3 h1:Tmx58EW3EMFltVW6kCwYfPO3RaK46FbFf2XqC+eOy1s=
 github.com/dynatrace/dynatrace-configuration-as-code-core v0.8.1-0.20250424104703-ba9c6e8f54a3/go.mod h1:bOIcK+LsuiLcyqA8XifLSVyfD+zCLnhQJEfPCqnEy2I=
+github.com/dynatrace/dynatrace-configuration-as-code-core v0.8.1-0.20250513082749-4a52a35af3c3 h1:457hWDZrF5vcBwvE5ekKhMNNfpjAoKMIksKTDey5Olc=
```
Kirdock, line 14: Should we also run `go mod tidy` here?

Kirdock (approved):

PR #1889, opened 2025-05-13
chore: Update casc core library to prevent concurrent header reads and writes

arthurpitman:
--- currently just testing for any unforeseen consequences ---


arthurpitman:
Closing this as we have https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1873

PR #1891, opened 2025-05-14
feat: Improve some log and error messages

arthurpitman:
#### **Why** this PR?
This PR contains several improvements I found while working on moving to SLOG. As they are not really related to that work, it makes sense for this to be in a separate PR.

#### **What** has changed?
Some log and error messages have been updated,  but not really _what_ is emitted, _when_, or _why_.

#### **How** does it do it?
Updates the text of some log and error messages. Some light refactoring is performed around this.

#### How is it **tested**?
Existing tests still work - no change. Code coverage of tests is not addressed in this PR.

Users get to read nicer messages.

### DIFF
--- a/pkg/delete/internal/automation/delete.go
+++ b/pkg/delete/internal/automation/delete.go
@@ -18,7 +18,6 @@ package automation
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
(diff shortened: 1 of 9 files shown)


pkg/delete/internal/automation/delete.go
```diff
@@ -102,40 +95,32 @@ func DeleteAll(ctx context.Context, c client) error {
...
-				logger.WithFields(field.Error(err)).Error("Failed to collect Automation objects of type %q - rejected by API: %v", resource, err)
-				errs++
-				continue
-			} else {
-				logger.Error("Failed to collect Automation objects of type %q - network error: %v", resource, err)
-				errs++
-				continue
-			}
+			logger.Error("Failed to collect Automation objects of type '%s': %v", resource, err)
+			errCount++
```
Kirdock, line 106: Is it intended that the `continue` was removed here?

Kirdock (approved):

PR #1894, opened 2025-05-15
refactor(settings): use request retry logic of core instead of custom one

Kirdock:
#### **Why** this PR?
The settings client had its own retry implementation, leading to duplicated code and more maintenance if a change is needed on both ends. Now it's using the retry logic of core and not using a custom one anymore

#### **What** has changed?
The settings client now uses the retry mechanism of core instead of having its own.

#### **How** does it do it?
It's now using the retry functionality of core.

#### How is it **tested**?
Current tests cover it

It slightly changes the logs but nothing else

### DIFF
--- a/pkg/client/dtclient/settings_client.go
+++ b/pkg/client/dtclient/settings_client.go
@@ -581,14 +581,18 @@ func (d *SettingsClient) Upsert(ctx context.Context, obj SettingsObject, upsertO
 		retrySetting = *upsertOptions.OverrideRetry
 	}
 
-	resp, err := coreapi.AsResponseOrError(
-		d.client.POST(ctx, d.settingsObjectAPIPath, bytes.NewReader(payload), corerest.RequestOptions{
-			CustomShouldRetryFunc: func(response *http.Response) bool {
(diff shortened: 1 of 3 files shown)


pkg/client/dtclient/settings_client.go
```diff
@@ -580,7 +581,14 @@ func (d *SettingsClient) Upsert(ctx context.Context, obj SettingsObject, upsertO
 		retrySetting = *upsertOptions.OverrideRetry
 	}
 
-	resp, err := SendWithRetryWithInitialTry(ctx, d.client.POST, d.settingsObjectAPIPath, corerest.RequestOptions{CustomShouldRetryFunc: corerest.RetryIfTooManyRequests}, payload, retrySetting)
+	resp, err := coreapi.AsResponseOrError(
```
tomazpu, line 584: Please use `NewResponseFromHTTPResponse` instead of the deprecated `AsResponseOrError`

tomazpu (approved):

arthurpitman (approved):

PR #1897, opened 2025-05-15
feat(manifest): Loaded manifest includes all environment and group names

arthurpitman:
#### **Why** this PR?
This PR is an alternative to https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1876 and this time adds all environment and environment group names seen during loading directly to the manifest data structure.

This PR is more robust because `manifest.Environments.SelectedEnvironments` is identical to the current `manifest.Environments`, meaning it avoids the effort of having to explicitly deal with `Skip`. However, a lot of effort had to be invested in updating the tests.

#### **What** has changed?
The `Environments` field of `manifest.Manifest` is now a struct of type `manifest.Environments` that holds selected environments as well as the names of all environments and environment groups seen during loading. This extra information will be used https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1880 to emit a warning if an undefined environment or group is referenced.

#### **How** does it do it?
The PR does some renaming and then simply adds the new type and updates tests. The actual change to production code is pretty small.

Best reviewed commit by commit.

#### How is it **tested**?
Existing tests are updated to include the new data structure, which holds information about the names of all environments and environment groups.

No effect on users.

### DIFF
--- a/pkg/manifest/loader/manifest_loader.go
+++ b/pkg/manifest/loader/manifest_loader.go
@@ -202,9 +202,9 @@ func Load(context *Context) (manifest.Manifest, []error) {
 	}
 
 	return manifest.Manifest{
-		Projects:     projectDefinitions,
-		Environments: environmentDefinitions,
-		Accounts:     accounts,
+		Projects:             projectDefinitions,
+		SelectedEnvironments: environmentDefinitions,
+		Accounts:             accounts,
 	}, nil
 }
 
--- a/pkg/manifest/loader/manifest_loader.go
+++ b/pkg/manifest/loader/manifest_loader.go
@@ -180,12 +180,12 @@ func Load(context *Context) (manifest.Manifest, []error) {
 	}
 
 	// environments
-	var environmentDefinitions map[string]manifest.EnvironmentDefinition
+	var environments manifest.Environments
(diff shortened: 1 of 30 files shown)


pkg/manifest/loader/manifest_loader.go
```diff
@@ -180,12 +180,12 @@ func Load(context *Context) (manifest.Manifest, []error) {
...
 	// environments
-	var environmentDefinitions map[string]manifest.EnvironmentDefinition
+	var environments manifest.Environments
 	if len(manifestYAML.EnvironmentGroups) > 0 {
 		var manifestErrors []error
-		if environmentDefinitions, manifestErrors = parseEnvironments(context, manifestYAML.EnvironmentGroups); manifestErrors != nil {
+		if environments, manifestErrors = parseEnvironments(context, manifestYAML.EnvironmentGroups); len(manifestErrors) > 0 {
 			errs = append(errs, manifestErrors...)
-		} else if len(environmentDefinitions) == 0 {
+		} else if len(environments.AllEnvironmentNames) == 0 {
```
Kirdock, line 188: So this is the only behavior change, right?

Kirdock (approved):

tomazpu (approved):

PR #1898, opened 2025-05-16
feat: Use slog for logging

arthurpitman:


internal/log/log.go
```diff
@@ -59,96 +56,111 @@ type CtxKeyAccount struct{}
...
 
 func Debug(msg string, a ...interface{}) {
-	std.Debug(msg, a...)
+	slog.Debug(fmt.Sprintf(msg, a...))
 }
 
-func Level() loggers.LogLevel {
-	return std.Level()
+// PrepareLogging sets up the default slog.Logger using the specified options.
+func PrepareLogging(ctx context.Context, fs afero.Fs, verbose bool, loggerSpy io.Writer, fileLogging bool, enableMemstatLogging bool) {
```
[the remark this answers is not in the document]
arthurpitman: Depending on how we proceed with the refactoring, this could be a follow-up. Currently, doing such a thing would involve touching a lot of code in a lot of places.

However, if we embrace slog fully, we will no longer need this function in its present form, or many of the options. For example, loggerSpy would be something for tests, not for production code.

internal/log/log.go
```diff
@@ -59,96 +56,111 @@ type CtxKeyAccount struct{}
...
 
-	setDefaultLogger(loggers.LogOptions{
-		File:        logFile,
-		ErrorFile:   errFile,
-		JSONLogging: logFormat == loggers.LogFormatJSON,
-		LogLevel:    loglevel,
-		LogSpy:      loggerSpy,
-		LogTimeMode: logTime,
-	})
+func getReplaceAttrFunc() func(groups []string, a slog.Attr) slog.Attr {
```
Kirdock, line 142: Should we maybe structure this a bit differently? Instead of deciding inside this function for every attribute if UTC should be used, we set a different replace function
```go
func getHandlerOptions(level slog.Leveler) *slog.HandlerOptions {
  if useUTC {
    return &slog.HandlerOptions{
		Level:       level,
		ReplaceAttr: getReplaceUTC,
      }
  }
return &slog.HandlerOptions{
  Level:       level,
  }
}

func getReplaceUTC() func(groups []string, a slog.Attr) slog.Attr {
		if a.Key == slog.TimeKey {
			t := a.Value.Time()
			return slog.Attr{Key: slog.TimeKey, Value: slog.TimeValue(t.UTC())}
		}

		return a
}
```
or we decide in a generic `getReplaceAttrFunc` which function to return

internal/log/log.go
```diff
@@ -59,96 +56,111 @@ type CtxKeyAccount struct{}
 // CtxValGraphComponentId context value used for correlating logs that belong to deployment of a sub graph
 type CtxValGraphComponentId int
 
-var (
-	_ loggers.Logger = (*zap.Logger)(nil)
-	_ loggers.Logger = (*console.Logger)(nil)
-)
+const envVarLogFormat = "MONACO_LOG_FORMAT"
```
Kirdock, line 59: Should we document these two somewhere?

internal/log/log.go
```diff
@@ -59,96 +56,111 @@ type CtxKeyAccount struct{}
...
+
+		if errorFile != nil {
+			handlers = append(handlers, getHandler(errorFile, getHandlerOptions(slog.LevelError)))
+		}
 	}
 
-	if c, ok := ctx.Value(CtxGraphComponentId{}).(CtxValGraphComponentId); ok {
-		f = append(f, field.F("gid", c))
+	if len(handlers) == 1 {
+		return handlers[0]
```
Kirdock, line 113: I guess there isn't an easy way to have just one handler? For multiple write operations, `io.MultiWriter` could be used. Not sure how to handle the errorFile handler though

internal/log/log.go
```diff
@@ -172,21 +201,21 @@ func MemStatFilePath() string {
 // may be returned in case of errors.
 // If log directory or logFile creation fails, no log files are returned.
 // If errLog creation fails, a valid logFile is still being returned with an error.
-func prepareLogFiles(ctx context.Context, fs afero.Fs, enableMemstatLogging bool) (logFile afero.File, errFile afero.File, err error) {
+func PrepareLogFiles(ctx context.Context, fs afero.Fs, enableMemstatLogging bool) (logFile afero.File, errFile afero.File, err error) {
```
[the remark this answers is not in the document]
Laubi: Not a huge fan, but understandable - we might want to get rid of it anyway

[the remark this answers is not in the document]
arthurpitman: I figured this was better than splitting the tests into two packages. I presume this function will not stay the same in the long term, as the setup for logging should be simplified once we really use slog.

internal/log/log.go
arthurpitman: For the moment, this is the easiest way as we want different things to go to the different sinks in different formats. For example, in a future PR, we'd like color logs on the console, but plain text going to the files, and as you say, only errors and fatals going to the error file, but all logs going to the log file. Something to revisit this in the future, though, and optimization is certainly possible.  

One last thought: in a future PR, our handler might also add attributes that it pulls from the context. The TeeHandler could do this once, and then, with this current architecture, pass this on a handler that supports color, or, encodes it to JSON, saving us from having to reimplement all this functionality.

Kirdock (approved):

PR #1899, opened 2025-05-19
feat(report): Introduce caching record types

Kirdock:
#### **Why** this PR?
It's not that easy to see/parse report records related to caching.

#### **What** has changed?
- A new record type `CACHE` is added.
- This new type is used for caching related records.

#### **How** does it do it?
By adding a new type and function, and making use of it.

#### How is it **tested**?
The existing test is extended to include caching events

The generated report uses CACHE now instead of LOAD for caching related records.
Before
`{"type":"LOAD","time":"2025-05-19T13:42:45+02:00","state":"SUCCESS","message":"Cached settings values for schema builtin:tags.auto-tagging"}`
After
`{"type":"CACHE","time":"2025-05-19T13:42:45+02:00","state":"SUCCESS","message":"Cached settings values for schema builtin:tags.auto-tagging"}`

#### Notes to the reviewer
Because we haven't used the `error` property and state before, I decided to keep it the same (warning and message).

### DIFF
--- a/pkg/report/reporter.go
+++ b/pkg/report/reporter.go
@@ -60,6 +60,8 @@ type Reporter interface {
 	ReportDeployment(config coordinate.Coordinate, state RecordState, details []Detail, err error)
 	// ReportLoading reports the result of a load config
 	ReportLoading(state RecordState, err error, message string, config *coordinate.Coordinate)
+	// ReportCaching reports the result of caching actions, like settings.
+	ReportCaching(state RecordState, message string)
 	// ReportInfo reports info messages like monaco version or that the deployment succeeded
 	ReportInfo(message string)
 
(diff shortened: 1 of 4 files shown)


pkg/report/reporter.go
```diff
@@ -171,7 +173,7 @@ func (d *defaultReporter) updateSummaryFromRecord(r Record) {
 	case StateError:
 		d.deploymentsErrorCount++
 	default:
-		panic(fmt.Sprintf("unexpected state for deployment event: %s", r.State))
+		return
```
Kirdock, line 176: I also did a minor change here. We have some other record types, and using another one, not related to the summary, should not crash the application

tomazpu (approved):

d0weinberger (approved):

PR #1904, opened 2025-05-22
refactor: restructure verifyEnvironmentGeneration

Kirdock:
#### **Why** this PR?
The verifyEnvironment had some hiccups. It only validated one of the provided credentials (OAuth or token) instead of both, and error handling was caught/logged within the function.

#### **What** has changed?
- Renamed and added some comments to some functions to make the functionality clearer.
- Returns now an error instead of a boolean
- Error handling and logging are done outside

#### How is it **tested**?
Extended existing tests and added new ones for the stuff that was not tested.

Some logs have changed. Also, one log was removed because it had no value.


cmd/monaco/deploy/deploy.go
```diff
@@ -51,9 +52,12 @@ func deployConfigs(ctx context.Context, fs afero.Fs, manifestPath string, enviro
 		return err
 	}
 
-	ok := verifyEnvironmentGen(ctx, loadedManifest.Environments.SelectedEnvironments, dryRun)
-	if !ok {
-		return fmt.Errorf("unable to verify Dynatrace environment generation")
```
Kirdock, line 56: FYI, I removed this line, as it had no additional value to the returned error now

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -33,92 +30,95 @@ import (
...
-		log.Error("Please verify that this environment is a Dynatrace Platform environment.")
-		return false
-	}
-	return true
+	return nil
 }
 
-func handleAuthError(ctx context.Context, env manifest.EnvironmentDefinition, err error) {
+// formatAuthError takes the provided error and wraps it into a more useful error message.
+func formatAuthError(env manifest.EnvironmentDefinition, url string, err error, auth authType) error {
```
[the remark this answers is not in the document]
Kirdock: I think we would probably end up adding a custom error for each possible error inside verifyEnvironment. This will then lead to making them generic and not having a struct for every error, and this will then lead to a struct for all, where we have to rely on the text again.

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -33,92 +30,95 @@ import (
...
 	"golang.org/x/oauth2/clientcredentials"
 )
 
-// VerifyEnvironmentGeneration takes a manifestEnvironments map and tries to verify that each environment can be reached
-// using the configured credentials
-func VerifyEnvironmentGeneration(ctx context.Context, envs manifest.EnvironmentDefinitionsByName) bool {
-	if !featureflags.VerifyEnvironmentType.Enabled() {
-		return true
-	}
+type authType = string
```
Laubi, line 44: This is unnecessary. Same with the constants on L47--48. Inline the strings, remove the type

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -33,92 +30,95 @@ import (
...
+
+const (
+	oAuthType authType = "OAuth"
+	tokenType authType = "token"
+)
+
+var ErrorMissingAuth = errors.New("no token or oAuth credentials provided in the manifest")
+
+// VerifyEnvironmentsAuthentication takes a manifestEnvironments map and tries to verify that each environment can be reached
```
Laubi, line 53:
```suggestion
// VerifyEnvironmentsAuthentication verifies that all environments can be reached with the defined credentials.
// It returns the first error encountered 
```

I don't think it has too much value to specify too much detail about the types or passing it, but really briefly describe the behavior.

cmd/monaco/dynatrace/dynatrace.go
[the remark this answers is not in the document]
Laubi: @Kirdock, could you elaborate on what you mean?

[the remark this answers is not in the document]
Kirdock: Suppose we have a custom error here, like `OAuthError` and `TokenError`. In that case, I assume this will then lead to wanting to have a custom or generic error for every error we create in `verifyEnvironment`, which will then result in refactoring the custom errors into a more generic `AuthError`, like for the permission errors in core now. If that's the case, we again have a problem differentiating the error inside of tests without evaluating the message, if each error is of the same type, unless we add a `type` property to the AuthError, which should be different for each error returned.

Kirdock:
https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1904#discussion_r2111361892

It adds more context to the error returned from the request, referring to the authentication that did not work and the failed request.

cmd/monaco/deploy/deploy.go
```diff
@@ -51,9 +52,12 @@ func deployConfigs(ctx context.Context, fs afero.Fs, manifestPath string, enviro
...
+	if !dryRun && featureflags.VerifyEnvironmentType.Enabled() {
+		if err := dynatrace.VerifyEnvironmentsAuthentication(ctx, loadedManifest.Environments.SelectedEnvironments); err != nil {
+			report.GetReporterFromContextOrDiscard(ctx).ReportLoading(report.StateError, err, "", nil)
+			log.Error("%s", err)
```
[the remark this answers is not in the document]
Kirdock: The thing is, for many errors, we're adding some context, and at some point, it gets a bit redundant.
Assuming that the classic call failed due to a timeout. The resulting error message would be
`please verify that this environment is a Dynatrace Classic environment: could not connect to environment xyz (url) using token authorization: context deadline`.
With an additional context for the logger outside, this will again add more information, which won't add any extra value in my opinion.

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -33,92 +30,87 @@ import (
...
-		ClientSecret: env.Auth.OAuth.ClientSecret.Value.Value(),
-		TokenURL:     env.Auth.OAuth.GetTokenEndpointValue(),
-	}
-
-	if _, err := getDynatraceClassicURL(ctx, env.URL.Value, oauthCreds); err != nil {
-		handleAuthError(ctx, env, err)
-		log.Error("Please verify that this environment is a Dynatrace Platform environment.")
-		return false
+		err = formatAuthError(env, classicURL, err, "token")
+		return fmt.Errorf("please verify that this environment is a Dynatrace Classic environment. %w", err)
```
arthurpitman, line 102: I think "please verify that this environment is a Dynatrace Classic environment" belongs in a log message but not an error message. I would move this out of this function.

cmd/monaco/download/download_configs.go
```diff
@@ -90,9 +90,11 @@ func (d DefaultCommand) DownloadConfigsBasedOnManifest(ctx context.Context, fs a
 		return fmt.Errorf("environment %q was not available in manifest %q", cmdOptions.specificEnvironmentName, cmdOptions.manifestFile)
```
arthurpitman, line 90: Not in this PR, but while I see it: 
```suggestion
		return fmt.Errorf("environment '%s' is not defined in manifest '%s'", cmdOptions.specificEnvironmentName, cmdOptions.manifestFile)
```

cmd/monaco/dynatrace/dynatrace.go
arthurpitman: In fact, the call to `formatAuthError(...)` should also move out, at least to line 81 above, or, both calls in _that_ function could move out. 

In a way, I see `formatAuthError(...)` as a code smell that suggests we should probably just wrap any error returned from `VerifyEnvironmentAuthentication(...)` with the information there rather than calling this twice inside.

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -33,92 +30,87 @@ import (
...
+// validateEstablishClassicConnection checks if a classic connection (via token) can be established. Scopes are not validated.
+func validateEstablishClassicConnection(ctx context.Context, env manifest.EnvironmentDefinition, classicURL string) error {
 	token := env.Auth.Token.Value.Value()
 	client, err := clients.Factory().
-		WithClassicURL(env.URL.Value).
+		WithClassicURL(classicURL).
 		WithAccessToken(token).
 		WithRateLimiter(true).
 		WithRetryOptions(&client.DefaultRetryOptions).
 		CreateClassicClient()
```
arthurpitman, line 95: Do we also have to explicitly enable the support archive here too?

cmd/monaco/deploy/deploy.go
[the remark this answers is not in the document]
arthurpitman: I _prefer_ it on the outside: functions that perform tasks returning errors and then, in the main control flow,  if this cannot be handled any other way, logged there.

cmd/monaco/dynatrace/dynatrace.go
Kirdock: Good catch! Forgot about this one.
Added 👍

cmd/monaco/dynatrace/dynatrace.go
Kirdock: I restructured it to get rid of the formatAuthError

cmd/monaco/deploy/deploy.go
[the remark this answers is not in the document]
Kirdock: Now the error logged would look like this
`could not authorize against environment 'xyz' (url) using OAuth authorization: context deadline`.
Which message would you suggest that would not lead to some redundancy?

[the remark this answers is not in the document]
arthurpitman: It's a good error message. The only thing extra we could log is what it means for the process as a whole. I guess deployment is being aborted? So maybe we could add something so that it reads like `deployment aborted: could not authorize against environment...`

cmd/monaco/download/download_configs.go
```diff
@@ -90,9 +90,11 @@ func (d DefaultCommand) DownloadConfigsBasedOnManifest(ctx context.Context, fs a
...
 	}
 
-	ok := dynatrace.VerifyEnvironmentGeneration(ctx, manifest.EnvironmentDefinitionsByName{env.Name: env})
-	if !ok {
-		return fmt.Errorf("unable to verify Dynatrace environment generation")
+	if featureflags.VerifyEnvironmentType.Enabled() {
+		if err := dynatrace.VerifyEnvironmentAuthentication(ctx, env); err != nil {
+			log.Error("%s", err)
```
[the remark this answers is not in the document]
arthurpitman: Maybe it could then read like `Download aborted: could not authorize against...`

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -211,14 +193,19 @@ func CreateEnvironmentClients(ctx context.Context, environments manifest.Environ
...
 	}
 
-	client, err := clients.Factory().WithPlatformURL(platformURL).WithOAuthCredentials(oauthCreds).CreatePlatformClient(ctx)
+	factory := clients.Factory().WithPlatformURL(platformURL).WithOAuthCredentials(oauthCreds)
+	if supportarchive.IsEnabled(ctx) {
+		factory = factory.WithHTTPListener(&corerest.HTTPListener{Callback: trafficlogs.GetInstance().LogToFiles})
+	}
+	client, err := factory.CreatePlatformClient(ctx)
 	if err != nil {
 		return "", err
```
arthurpitman: Question: should we wrap this error like in line 99 above: `return fmt.Errorf("could not create client: %w", err)`?

Kirdock: I would say so, yes.
Committed 👍

cmd/monaco/deploy/deploy.go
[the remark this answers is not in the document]
Kirdock: I adjusted it, but now I'm wondering why we have this log anyway. Is there anything against removing it?
Before we logged it because the error was not returned, and now we're returning the error, which makes the logging unnecessary.

arthurpitman (approved):

PR #1906, opened 2025-05-23
feat(clients): allow custom http headers

d0weinberger:
#### **Why** this PR?
Allowing users to set additional HTTP headers for requests.

#### **What** has changed?
The multiline environment variable `MONACO_ADDITIONAL_HTTP_HEADERS` is introduced. Its content is parsed. Each line represents an HTTP header. The key- and value part of the header are separated via `:`. These HTTP headers are appended to all requests sent via Monaco.

#### **How** does it do it?
The variable is retrieved and parsed at all places where the client factory is used to create REST clients. It is also parsed once at the start of Monaco to log the header keys, if any are present.

#### How is it **tested**?
To be done.
Currently, only parsing the environment variable contents is tested. Passing it to the clients is not tested yet.

Allowing users to set additional HTTP headers for requests via the environment variable `MONACO_ADDITIONAL_HTTP_HEADERS`.

### DIFF
--- a/cmd/monaco/runner/runner.go
+++ b/cmd/monaco/runner/runner.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/maps"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/account"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/delete"
@@ -29,6 +30,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/purge"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/supportarchive"
 	versionCommand "github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/version"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/environment"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log/field"
@@ -96,6 +98,11 @@ Examples:
 			}
 
 			memory.SetDefaultLimit()
+
+			additionalHeaders := environment.GetAdditionalHTTPHeadersFromEnv()
+			if len(additionalHeaders) > 0 {
+				log.Info("Additional headers %q from '%s' environment variable will be added to HTTP requests", maps.Keys(additionalHeaders), environment.AdditionalHTTPHeaders)
+			}
 		},
 		Run: func(cmd *cobra.Command, args []string) {
 			_ = cmd.Help()
--- a/internal/environment/get_env_value.go
+++ b/internal/environment/get_env_value.go
@@ -19,6 +19,7 @@ package environment
 import (
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 )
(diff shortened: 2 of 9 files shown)


internal/environment/get_env_value.go
```diff
@@ -105,3 +107,28 @@ func GetEnvValueIntLog(env string) int {
...
+		key := strings.TrimSpace(headerKeyAndValue[0])
+		if key == "" {
+			continue // header key must not be empty
+		}
+
+		if len(headerKeyAndValue) != 2 {
+			result[key] = ""
+		} else {
+			result[key] = strings.TrimSpace(headerKeyAndValue[1])
+		}
```
arthurpitman: Maybe it is possible to simplify it using `Cut(...)`?
```suggestion
		key, value, _ := strings.Cut(header, ":")
		key = strings.TrimSpace(key)
		if key == "" {
			continue // header key must not be empty
		}

		result[key] = strings.TrimSpace(value)
	
```

d0weinberger: Ah, didn't know that function! Will try it out, thanks

EDIT: Yeah, looks way cleaner, thx!

cmd/monaco/runner/runner.go
```diff
@@ -96,6 +98,11 @@ Examples:
 			}
 
 			memory.SetDefaultLimit()
+
+			additionalHeaders := environment.GetAdditionalHttpHeadersFromEnv()
+			if len(additionalHeaders) > 0 {
+				log.Info("Additional custom HTTP headers: %q, from '%s' environment variable", maps.Keys(additionalHeaders), environment.AdditionalHttpHeaders)
```
Kirdock: We should use `'%v'` or just `%v` instead of `%q`

Kirdock: On the one side, `%q` makes a lot of sense here, as `[]string{"a b", "c"}` would result in `["a b", "c"]` and `%v` or `%s` result in `[a b c]`

d0weinberger: Okay, then I'll leave it as is, if that's fine with you

Kirdock (approved):

internal/environment/get_env_value.go
```diff
@@ -30,6 +31,7 @@ const (
 	KeyUserActionWebWaitSecondsEnvKey = "MONACO_KUA_WEB_WAIT_SECONDS"
 	MaxFilenameLenKey                 = "MONACO_MAX_FILENAME_LEN"
 	DeploymentReportFilename          = "MONACO_DEPLOYMENT_REPORT_FILENAME"
+	AdditionalHttpHeaders             = "MONACO_ADDITIONAL_HTTP_HEADERS"
```
arthurpitman: Nit: should probably be written `AdditionalHTTPHeaders`

internal/environment/get_env_value.go
```diff
@@ -105,3 +107,24 @@ func GetEnvValueIntLog(env string) int {
 
 	return value
 }
+
+// GetAdditionalHttpHeadersFromEnv gets the content of the env var MONACO_ADDITIONAL_HTTP_HEADERS and parses
+// a map of http headers from it, if present. Headers are separated via '\n'
+// Headers consist of a mandatory key part and a value part which can be empty. Key and value part are separated via ':'
+func GetAdditionalHttpHeadersFromEnv() map[string]string {
```
arthurpitman: Nit: should probably be written `GetAdditionalHTTPHeadersFromEnv`

cmd/monaco/runner/runner.go
arthurpitman: We could increase the informational content of this log, maybe something like:
```suggestion
				log.Info("Additional headers %q from '%s' environment variable will be added to HTTP requests", maps.Keys(additionalHeaders), environment.AdditionalHttpHeaders)
```

arthurpitman (approved):

internal/environment/get_env_value.go
d0weinberger: Good point, changed it like that

internal/environment/get_env_value.go
d0weinberger: Good point, changed it.

cmd/monaco/runner/runner.go
d0weinberger: Good point, copied your suggestion

arthurpitman (approved):

PR #1910, opened 2025-05-30
feat: Extract logging attributes directly from context

arthurpitman:
#### **Why** this PR?
Slog extends logging with additional methods that take a context. The context is then made available to the handler for further processing.

This PR utilizes this to extract log attributes directly from the context, reducing the need to construct individual loggers with this information.

This results in simpler code but also lower memory consumption.

#### **What** has changed?
The biggest change is the introduction of `ContextHandler` which is then used in preference to `log.WithCtxFields(ctx)`.

#### **How** does it do it?
- Logging calls are updated to provide context where possible
- `ContextHandler` is introduced to extract attributes from the context
- Calls to `log.WithCtxFields(ctx)` are removed

#### How is it **tested**?
Tests are added for:
- Additional logging helper functions that accept context
- `ContextHandler`, in a fashion similar to those for `TeeHandler`

While not conclusive, manual testing indicates a reduction in memory consumption compared with before the switch to slog, and importantly, an improvement over using slog and log.WithCtxFields(ctx)`:

| Commit  | Name      | Description  | Average heap allocation bytes over deployment runtime | Total heap allocation bytes  over deployment runtime | 
| --------  | ---------| -------------|------------------| ---------------------------|
| [120fe6a](https://github.com/Dynatrace/dynatrace-configuration-as-code/commit/120fe6aa96367026fde07159f59c59ca2ad8aa0f) |  v2.23.1  |  Pre slog       | 38684584 |  1669212432 |
| [ec1bbee](https://github.com/Dynatrace/dynatrace-configuration-as-code/commit/ec1bbee60cfdb4e7d5a6f0dc3dbd768a7eabb771) |  main |  With slog and `log.WithCtxFields(ctx)` |  40314489 | 1822583240  | 
|  This PR | This PR | With `log.WithCtxFields(ctx) removed | 35988891|  1583269944 |

SonarCloud Code Analysis quality gate is failing as this PR touches a lot of code when updating the logging call that has low coverage. 

No real effect on users.

### DIFF
--- a/pkg/delete/internal/bucket/delete.go
+++ b/pkg/delete/internal/bucket/delete.go
@@ -37,7 +37,7 @@ type client interface {
 
 func Delete(ctx context.Context, c client, entries []pointer.DeletePointer) error {
 	logger := log.WithCtxFields(ctx).WithFields(field.Type("bucket"))
-	logger.Info(`Deleting %d config(s) of type "bucket"...`, len(entries))
+	logger.InfoContext(ctx, `Deleting %d config(s) of type "bucket"...`, len(entries))
 
 	deleteErrs := 0
 	for _, e := range entries {
@@ -77,7 +77,7 @@ func Delete(ctx context.Context, c client, entries []pointer.DeletePointer) erro
 //   - error: After all deletions where attempted an error is returned if any attempt failed.
 func DeleteAll(ctx context.Context, c client) error {
 	logger := log.WithCtxFields(ctx).WithFields(field.Type("bucket"))
-	logger.Info("Collecting Grail Bucket configurations...")
+	logger.InfoContext(ctx, "Collecting Grail Bucket configurations...")
 
 	response, err := c.List(ctx)
 	if err != nil {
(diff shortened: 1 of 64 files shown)


pkg/delete/internal/bucket/delete.go
```diff
@@ -36,7 +36,7 @@ type client interface {
 }
 
 func Delete(ctx context.Context, c client, entries []pointer.DeletePointer) error {
-	logger := log.WithCtxFields(ctx).WithFields(field.Type("bucket"))
+	logger := log.WithFields(field.Type("bucket"))
 	logger.Info(`Deleting %d config(s) of type "bucket"...`, len(entries))
```
Kirdock:
```suggestion
	logger.InfoContext(ctx, `Deleting %d config(s) of type "bucket"...`, len(entries))
```

arthurpitman: I added this to [refactor: Add context to info log calls where available](https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1910/commits/bd99c578acdb070ca2d88cc89f818ecf1514b713)

Kirdock (approved):

Kirdock (approved):

tomazpu (approved):

PR #1912, opened 2025-06-04
refactor(deploy): Introduce `Deployable` and generify SLO

Kirdock:
#### **Why** this PR?
The deploy checks many config types, so adding new config types always comes with some overhead and several places where adjustments are needed to support a new resource.

#### **What** has changed?
A new "Deployable" interface is introduced.
For now, the SLO resource fulfills this interface. Other resources will follow.

#### **How** does it do it?
By creating an SLO-deployable and making use of it.
At the moment, this is only used for the deploy. Caching and validation are out of scope.

#### How is it **tested**?
Existing tests cover it, and new ones were added.

NONE

### DIFF
--- a/pkg/resource/slo/deploy.go
+++ b/pkg/resource/slo/deploy.go
@@ -40,12 +40,20 @@ type deployServiceLevelObjectiveClient interface {
 	Create(ctx context.Context, data []byte) (api.Response, error)
 }
 
+type DeployAPI struct {
+	sloSource deployServiceLevelObjectiveClient
+}
+
+func NewDeployable(sloSource deployServiceLevelObjectiveClient) *DeployAPI {
+	return &DeployAPI{sloSource}
+}
+
 type sloResponse struct {
 	ID         string `json:"id"`
 	ExternalID string `json:"externalId"`
 }
 
-func Deploy(ctx context.Context, client deployServiceLevelObjectiveClient, properties parameter.Properties, renderedConfig string, c *config.Config) (entities.ResolvedEntity, error) {
+func (d DeployAPI) Deploy(ctx context.Context, properties parameter.Properties, renderedConfig string, c *config.Config) (entities.ResolvedEntity, error) {
 	ctx = logr.NewContextWithSlogLogger(ctx, log.WithCtxFields(ctx).SLogger())
 	ctx, cancel := context.WithTimeout(ctx, time.Minute)
 	defer cancel()
@@ -58,7 +66,7 @@ func Deploy(ctx context.Context, client deployServiceLevelObjectiveClient, prope
 
 	//Strategy 1 when OriginObjectId is set we update the object
 	if c.OriginObjectId != "" {
-		_, err = client.Update(ctx, c.OriginObjectId, requestPayload)
+		_, err = d.sloSource.Update(ctx, c.OriginObjectId, requestPayload)
 		if err == nil {
 			return createResolveEntity(c.OriginObjectId, properties, c), nil
 		}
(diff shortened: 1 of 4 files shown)


pkg/resource/resource.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/parameter"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest"
+)
+
+type Deployable interface {
+	// Verify verifies given auth types and configs:
+	// 1. if the needed Auth is set (OAuth/token/platform)
+	// 2. if the needed Auth for a certain config is set (e.g., permissions in settings require a platform connection)
+	// 3. if the configs are valid (e.g., open pipeline: only one kind in one environment)
+	Verify(ev manifest.EnvironmentDefinition, cfg []config.Config) (bool, error)
```
d0weinberger: [suggestion] wdyt about the name `IsApplicableTo` or similar?
It would then read like `"Deployable is applicable to environment and configs"`, i.e.
`deployable.IsApplicableTo(environment, configs)`

pkg/resource/resource.go
```diff
@@ -0,0 +1,74 @@
...
+	// Deploy deploys a given resource and returns the resolved entity
+	Deploy(ctx context.Context, properties parameter.Properties, renderedConfig string, c *config.Config) (entities.ResolvedEntity, error)
+
+	// Cache caches any needed data for a given configType and identifier (schema, api)
+	Cache(c config.Type, identifier string)
```
d0weinberger: If there is only one config type per Deployable, the type shouldn't need to be in the interface, should it?

Kirdock: The thing is that the config type includes other information, like schemaId or apiType

pkg/resource/resource.go
Kirdock: I've split it up into several Verify functions (enabled, verifyAuth, verifryConfig) to separate the different checks that should be present for every deployable.
Would you still rename something?

pkg/resource/resource.go
```diff
@@ -0,0 +1,64 @@
...
+
+type Deployable interface {
+	// Enabled returns true if the Resource is enabled (FF is not false)
+	Enabled() bool
+	// VerifyAuth checks if the needed auth for the given configs is set (classic/platform)
+	VerifyAuth(auth manifest.EnvironmentDefinition, configs []config.Config) error
+	// VerifyConfigs checks if the given configs are valid for a resource (e.g., open pipeline: only one kind in one environment)
+	VerifyConfigs(configs []config.Config) (bool, error)
+	// Deploy deploys a given resource and returns the resolved entity
+	Deploy(ctx context.Context, properties parameter.Properties, renderedConfig string, c *config.Config) (entities.ResolvedEntity, error)
```
tomazpu: I would make the deployable interface only with the Deploy method

tomazpu: I would also move the creation of the resolved entity out of the deploy

tomazpu: Then i would try to provide a struct DTO for the deploy with a min set of requirements that are needed

pkg/resource/slo/deploy.go
```diff
@@ -40,12 +40,20 @@ type deployServiceLevelObjectiveClient interface {
 	Create(ctx context.Context, data []byte) (api.Response, error)
 }
 
+type DeployAPI struct {
+	sloSource deployServiceLevelObjectiveClient
```
tomazpu, line 44: I just wonder if this should also be a pointer.

tomazpu (approved):

pkg/resource/resource.go
tomazpu: This where old comments on the previous version.

pkg/resource/slo/deploy.go
Kirdock: As this is just an interface, I would leave it as it is. Also, to be in line with the downloadable implementation

PR #1913, opened 2025-06-04
refactor: Move e2e tests to `/test` directory

Laubi:
#### **Why** this PR?
Currently, our e2e tests are stored in `./cmd/monaco/integrationtest/{account|v2}`, as well as other places (for example, `./cmd/monaco/deletefile`).
Additionally, nearly all tests live in the same directory, thus making it way harder to know what test requires what test data.
Furthermore, utilities are someone in the same folder with the tests, somewhat other packages, all a big mix. 
Additionally, the test execution for account/configs was based on the path, which made it also error-prone.

This PR changes that. 

#### **What** has changed?

**Note**: This PR changes only the test-code. While all `.yaml` and `.json` have moved, they did not change in content.

1) All e2e tests and utils moved to `./test`.
2) All utilities for the e2e tests are in `./test/internal`.
3) Account tests now use the `account_integration` test flag.
4) The mono-test folder has been split up and moved into more logical groups. Are the groups perfect? No. But its way better than before. This also means, that there is not one huge test-resources folder, but only those per test.
5) Instead of the folder `test-resources`, the folder `testdata` is now used, as it should be. Read here why: `go help packages`.
6) Some files were deleted that were not used at all.

#### **How** does it do it?
See commits. Basically, I started moving some files, and didn't stop until all was done. 
It's reviewable commit-by-commit.

#### How is it **tested**?
They are test, but it's obvious the test utils should be tested as well.

It does not, but it should be easier for us to add/manage tests in the future.

#### Future work
1) Split up the huge `integration-test` test. It tests nearly everything, and is hard to debug/fix. This should be its own PR.
2) Better test the test-utils.
3) Revisit the tests to 1) maybe parallize them, 2) improve them.


cmd/monaco/integrationtest/v2/integration_test_utils.go
```diff
@@ -32,8 +32,8 @@ import (
 
 // TestContext contains all information necessary for the test-run
 type TestContext struct {
-	// suffix contains the suffix which is generated for the test-run.
-	suffix string
+	// Suffix contains the Suffix which is generated for the test-run.
```
d0weinberger: [absolute nit]
should be 
`// Suffix contains the suffix which is generated for the test-run.`

cmd/monaco/integrationtest/v2/settings_integration_test.go
```diff
@@ -620,7 +620,7 @@ func createSettingsClientFromManifest(t *testing.T, fs afero.Fs, manifestPath st
 	return clientSet.SettingsClient
 }
 
-// getRandomMonitoredEntitySuffix gets a random 16 uppercase hexadecimal character string for use as a suffix for creating Dynatrace entity IDs, such as `HOST-...`.
+// getRandomMonitoredEntitySuffix gets a random 16 uppercase hexadecimal character string for use as a Suffix for creating Dynatrace entity IDs, such as `HOST-...`.
```
d0weinberger: [absolute nit]
this line should not be changed.

cmd/monaco/integrationtest/v2/integration_test_utils.go
```diff
@@ -94,16 +94,20 @@ func WithEnvironment(env string) Option {
...
-
 		for k, v := range env {
 			options.envVars[k] = v
 		}
 	}
 }
 
+// WithMeId sets the `MONACO_TARGET_ENTITY_SCOPE` environment variable to the defined entity.
+// The `MONACO_TARGET_ENTITY_SCOPE` should be used in tests that target a specific entity.
+func WithMeId(meId string) Option {
```
d0weinberger, line 104: Not sure why this was added here and where it is coming from

test/slov2/slo-v2_test.go
```diff
@@ -47,13 +48,12 @@ func TestSloV2(t *testing.T) {
 	t.Setenv(featureflags.ServiceLevelObjective.EnvName(), "true")
 
 	t.Run("When deploying two configs, two configs exist", func(t *testing.T) {
-		Run(t, configFolder,
-			Options{
-				WithManifestPath(manifestPath),
```
d0weinberger: Is manifest path not needed anymore or why did you omit it?

test/slov2/slo-v2_test.go
```diff
@@ -87,14 +87,12 @@ func TestSloV2(t *testing.T) {
 	t.Run("With a disabled FF the deploy should fail", func(t *testing.T) {
 		t.Setenv(featureflags.ServiceLevelObjective.EnvName(), "false")
 
-		Run(t, configFolder,
-			Options{
-				WithManifestPath(manifestPath),
-				WithSuffix("SLO-V2"),
```
d0weinberger: Same here: Suffix and Manifest path aren't needed anymore? Same question applies to some other touched files as well.

test/logging/testdata/configs-with-duplicate-ids/manifest.yaml
```diff
@@ -0,0 +1,21 @@
+manifestVersion: 1.0
```
d0weinberger, line 1: Is this a new file or was it moved? If it was moved, shouldn't it be marked as such, or shouldn't there at least be another file marked as deleted? Question applies to the other files as well.

test/commands/all_configs_integration_test.go
```diff
@@ -87,19 +84,9 @@ func TestIntegrationAllConfigsPlatform(t *testing.T) {
 func runDeployCommand(t *testing.T, fs afero.Fs, manifest, specificEnvironment string) {
 	t.Helper()
 
-	reportFile := fmt.Sprintf("report%s.jsonl", time.Now().Format(trafficlogs.TrafficLogFilePrefixFormat))
```
d0weinberger, line 89: why is the report file and the call to `assertReport` removed?

test/logging/testdata/configs-with-duplicate-ids/manifest.yaml
Laubi: Super simple answer: They are now duplicated, so that test data is not randomly shared between tests. 
Tests that did not test the same (e.g. logging, parsing references) are now split up. However, they did use the same testdata set. 
Instead of sharing it, I copied it, as each test should only have their own resources to worry about. I don't want to have a random test failing, just because I delete some resources I didn't think have anything to do with each other.

There are some more of those. `references`, `configs-with-duplicate-ids`. 
The authors of the tests should have created their own data, but I don't want to reinvent all tests here.

cmd/monaco/integrationtest/v2/integration_test_utils.go
Laubi: It's in the commit: https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1913/commits/2039aeeec809f2645ba6a3563e88a23c6f858a3d?file-filters%5B%5D=.go&show-viewed-files=true#diff-a4c9fb3feb5ba26ade01ccb1463d291e28f2287be073dcc7bc5ef88706a75b2eR57
You need to load the diff. 

At the start, I wanted to improve the tests already, but this was way to much work for all of them

test/slov2/slo-v2_test.go
Laubi: It defaults correctly, as per the documentation of the `Run` function. I did it for some configs at the start, but it would have been too much work for all of them. I decided to first move the tests, then improve them iteratively.

test/slov2/slo-v2_test.go
Laubi: Same as above

test/commands/all_configs_integration_test.go
Laubi: Because it had no business there. This is a dirty hack, and we should do it properly. This introduced way more issues when refactoring the code

pkg/project/testdata/configs-account-resources-mixed-same-file/config.yaml
d0weinberger, line 1: Where did this file and other files move? Or are these the ones that weren't used?

test/configtypes/settings/settings_insertafter_test.go
```diff
@@ -28,14 +28,14 @@ import (
...
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/integrationtest/utils/monaco"
-	v2 "github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/integrationtest/v2"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/idutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/dtclient"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/coordinate"
 	manifestloader "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest/loader"
+	assert2 "github.com/dynatrace/dynatrace-configuration-as-code/v2/test/internal/assert"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/test/internal/monaco"
+	v2 "github.com/dynatrace/dynatrace-configuration-as-code/v2/test/internal/runner"
```
d0weinberger: Just wanted to point out that the naming of the imported module is inconsistent across the tests.
Here it is `v2`, in other places it's `runner2` and occasionally it's just `runner`.
Similar thing with `"github.com/dynatrace/dynatrace-configuration-as-code/v2/test/internal/assert"`

Not sure if worth aligning or if it doesn't matter.

pkg/project/testdata/configs-account-resources-mixed-same-file/config.yaml
Laubi: Weren't used

test/configtypes/settings/settings_insertafter_test.go
Laubi: Yes, this is the only test that has it. If nothing really blocks this PR, I would rather work on follow-ups than changing the whole PR

d0weinberger (dismissed):

test/commands/all_configs_integration_test.go
Laubi: But we should definitely add some tests for that, if none exist. But they should test exactly that.

cmd/monaco/integrationtest/v2/integration_test_utils.go
Laubi: since i had to rebase, i fixed it

cmd/monaco/integrationtest/v2/settings_integration_test.go
Laubi: since i had to rebase, i fixed it

test/configtypes/settings/settings_insertafter_test.go
Laubi: since i had to rebase, i fixed it

d0weinberger (approved):

PR #1918, opened 2025-06-10
refactor: Rename all `token` mentions in the code and messages to `API token`

Laubi:
#### **Why** this PR?
This ensures that the manifest is still compatible (stays `token`), but the user *might* be less confused.

And also, we might be less confused.

#### **What** has changed?
Searched and replaced most `token` mentions with `API-token`. 
Notable exclusions are:

* `--token` in the CLI
* `token: ...` in the manifest
* `TokenEndpoint` (SSO token endpoint)

Users will have a harder time differentiating between `token` and platformToken`. These updated messages could help, at least, they should not confuse more.

### DIFF
--- a/cmd/monaco/download/auth.go
+++ b/cmd/monaco/download/auth.go
@@ -25,17 +25,17 @@ import (
 )
 
 type auth struct {
-	token, clientID, clientSecret string
+	apiToken, clientID, clientSecret string
 }
 
 func (a auth) mapToAuth() (*manifest.Auth, []error) {
 	errs := make([]error, 0)
 	mAuth := manifest.Auth{}
 
-	if token, err := readAuthSecretFromEnvVariable(a.token); err != nil {
+	if token, err := readAuthSecretFromEnvVariable(a.apiToken); err != nil {
 		errs = append(errs, err)
 	} else {
-		mAuth.Token = &token
+		mAuth.ApiToken = &token
 	}
 
 	if a.clientID != "" && a.clientSecret != "" {
--- a/cmd/monaco/dynatrace/dynatrace.go
+++ b/cmd/monaco/dynatrace/dynatrace.go
@@ -53,15 +53,15 @@ func VerifyEnvironmentsAuthentication(ctx context.Context, envs manifest.Environ
 	return nil
 }
 
-// VerifyEnvironmentAuthentication checks if the provided token and OAuth credentials of the provided environment are valid.
+// VerifyEnvironmentAuthentication checks if the provided API token and OAuth credentials of the provided environment are valid.
 func VerifyEnvironmentAuthentication(ctx context.Context, env manifest.EnvironmentDefinition) error {
-	if env.Auth.Token == nil && env.Auth.OAuth == nil {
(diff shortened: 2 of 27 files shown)


pkg/client/auth/transport.go
```diff
@@ -20,35 +20,34 @@ import (
...
 	if baseTransport == nil {
 		baseTransport = http.DefaultTransport
 	}
-	t := &TokenAuthTransport{
+	t := &ApiTokenAuthTransport{
 		RoundTripper: baseTransport,
 		header:       http.Header{},
 	}
-	t.setHeader("Authorization", "Api-Token "+token)
+	t.setHeader("Authorization", "Api-ApiToken "+apiToken)
```
Kirdock: Not intended change here

cmd/monaco/download/auth.go
Kirdock, line 1: For consistency reasons, it makes sense to also rename `token` of the `auth` struct above to `apitoken`

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -55,7 +55,7 @@ func VerifyEnvironmentsAuthentication(ctx context.Context, envs manifest.Environ
 
 // VerifyEnvironmentAuthentication checks if the provided token and OAuth credentials of the provided environment are valid.
```
Kirdock: The comment should also be changed to point to the API token. Also down below in the function

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -74,19 +74,19 @@ func VerifyEnvironmentAuthentication(ctx context.Context, env manifest.Environme
 		}
 	}
 
-	if env.Auth.Token != nil {
-		if err := checkClassicConnection(ctx, classicUrl, env.Auth.Token.Value.Value()); err != nil {
+	if env.Auth.ApiToken != nil {
+		if err := checkClassicConnection(ctx, classicUrl, env.Auth.ApiToken.Value.Value()); err != nil {
 			return fmt.Errorf("could not authorize against environment '%s' (%s) using token authorization: %w", env.Name, classicUrl, err)
```
Kirdock: Should be API-token

cmd/monaco/dynatrace/dynatrace.go
```diff
@@ -74,19 +74,19 @@ func VerifyEnvironmentAuthentication(ctx context.Context, env manifest.Environme
...
 		}
 	}
 	return nil
 }
 
 // checkClassicConnection checks if a classic connection (via token) can be established. Scopes are not validated.
```
Kirdock: The comment should also be changed to point to the API token

Kirdock (commented):
- We should also rename the token in `manifest_persistence`.
- At the moment, the casing is inconsistent. `API-Token` vs `API-token`
- Should we then also align the `GetTokenMetadata` function (name and parameter)?

Kirdock (dismissed):

Kirdock (approved):

PR #1921, opened 2025-06-12
refactor: Align deployable and downloadable interfaces

Kirdock:
#### **Why** this PR?
With `Deployable` are coming other API interfaces and sources. `Downloadable` referenced to them as `API` and `Source`, but this can't be used for `Deployable` because it needs something else, and the same name can't be used inside the same package. To make them distinguishable, the API, source interfaces, and constructors now include `Download/Deploy`, and the source is changed to private, as this one isn't used outside of tests.

#### **What** has changed?
- Renamed constructors of downloadables. `NewAPI` => `NewDownloadAPI`.
- Renamed source interface of downloadables and made it private: `Source` to `downloadSource`.
- Moved Deployable and Downloadable to `resource.go`.
- Renamed `NewDeployable` to `NewDeployAPI`, to be aligned with the downloadables.

#### How is it **tested**?
Existing tests cover it

NONE

#### Open for discussion
Should we rename the source of each client's API struct to just `source`? We have `sloSource/automationSource/etc.` for each resource API struct, but this doesn't make much sense, as we're already in the resource package of that, and `source` would be sufficient.


arthurpitman (approved):
Looks good!

I could imagine in the future that we align the resources even more, once we have `Deletable` too, such that resources conform to an interface `Resource`:

```
type Resource interface {
  GetDownloadable(...) Downloadable
  GetDeployable(...) Deployable
  GetDeletable(...) Deletable
}

**Review by arthurpitman — APPROVED** [review]
Looks good!

I could imagine in the future that we align the resources even more, once we have `Deletable` too, such that resources conform to an interface `Resource`:

```
type Resource interface {
  GetDownloadable(...) Downloadable
  GetDeployable(...) Deployable
  GetDeletable(...) Deletable
}

PR #1923, opened 2025-06-12
refactor(deploy): generify openpipeline

Kirdock:
#### **Why** this PR?
In order to make the deploy step more generic, the openpipeline deploy step should satisfy the Deployable interface.

#### **What** has changed?
- Removed the FF for deploy, as this was not used because we already check it when loading the configs.

### DIFF
--- a/pkg/resource/openpipeline/deploy_test.go
+++ b/pkg/resource/openpipeline/deploy_test.go
@@ -49,7 +49,7 @@ func TestDeployOpenPipelineConfig(t *testing.T) {
 	}
 
 	t.Run("Update succeeds", func(t *testing.T) {
-		client := NewMockClient(gomock.NewController(t))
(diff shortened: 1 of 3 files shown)


arthurpitman (approved):

pkg/resource/openpipeline/deploy_test.go
```diff
@@ -49,7 +49,7 @@ func TestDeployOpenPipelineConfig(t *testing.T) {
 	}
 
 	t.Run("Update succeeds", func(t *testing.T) {
-		client := NewMockClient(gomock.NewController(t))
+		client := NewMockDeploySource(gomock.NewController(t))
```
d0weinberger, line 52: -Where is this `NewMockDeploySource` coming from?-

Ahh, it's generated, right.

tomazpu (approved):

PR #1926, opened 2025-06-12
refactor(deploy): generify documents

Kirdock:
#### **Why** this PR?
In order to make the deploy step more generic, the documents deploy step should satisfy the Deployable interface.

#### **What** has changed?
- Deploy makes use of the deployable
- Some constants from the downloadable are moved inside the scope where they're needed in order to prevent name shadowing or clashes.

#### How is it **tested**?
Existing tests are moved and adjusted

NONE


pkg/resource/document/deploy.go
```diff
@@ -35,15 +35,23 @@ import (
 	deployErrors "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/errors"
 )
 
-//go:generate mockgen -source=document.go -destination=document_mock.go -package=document documentClient
-type Client interface {
+//go:generate mockgen -source=deploy.go -destination=document_mock.go -package=document DeploySource
+type DeploySource interface {
```
tomazpu, line 39: As I can see we don't need the Get in the interface as its not used.

pkg/resource/document/deploy.go
```diff
@@ -35,15 +35,23 @@ import (
...
-func Deploy(ctx context.Context, client Client, properties parameter.Properties, renderedConfig string, c *config.Config) (entities.ResolvedEntity, error) {
+type DeployAPI struct {
+	source DeploySource
+}
+
+func NewDeployAPI(source DeploySource) *DeployAPI {
+	return &DeployAPI{source}
+}
+
+func (d DeployAPI) Deploy(ctx context.Context, properties parameter.Properties, renderedConfig string, c *config.Config) (entities.ResolvedEntity, error) {
```
tomazpu, line 60: Some error cases are not handled in the unit tests. I know, you just moved the code. Maybe we can add some test cases if you agree?

- missing name parameter
- error reading received data
- multiple documents found with externalId
- failed to unmarshal dashboard payload

tomazpu (dismissed):

tomazpu (approved):

PR #1927, opened 2025-06-12
refactor(deploy): generify automations

Kirdock:
#### **Why** this PR?
In order to make the deploy step more generic, the automations deploy step should satisfy the Deployable interface.

#### **What** has changed?
- Removed the FF for deploy, as this was not used because we already check it when loading the configs.
- Test package changed to `automation_test` from `automation`
- Using api.Response instead of the alias `automation.Response` to prevent package name clashing

#### How is it **tested**?
Existing tests are moved and adjusted

NONE


pkg/resource/automation/deploy.go
```diff
@@ -31,12 +31,20 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/deploy/errors"
 )
 
-//go:generate mockgen -source=automation.go -destination=automation_mock.go -package=automation automationClient
-type Client interface {
+//go:generate mockgen -source=deploy.go -destination=automation_mock.go -package=automation deploySource
+type deploySource interface {
```
Laubi: Why would this be private?

Kirdock: The initial thought was, if it's not needed from the outside, why export it, but anyway, let's make it public

tomazpu (approved):
The question in my mind is should we also move the `automationutils` closer to the resource. But can be a follow-up.

PR #1928, opened 2025-06-12
feat: Make use of platform tokens

d0weinberger:
#### **Why** this PR?

Tests are adapted and/or extended.

#### **What** has changed?
This PR 
1. updates the core library dependency to get the new platform token client feature,
2. adapts the codebase to other changes that come with the core library upgrade, i.e. API changes for bucket and automation clients,
3. makes use of the platform token clients when creating the client set and determining the Dynatrace Classic URL

ad 2.: Specifically, the type `buckets.Response` was removed and cannot be referenced anymore. Instead, `api.Response` is referenced. Also, changes to automation clients have been made, which need to be addressed. Specifically, the type `automation.Response` was removed and cannot be referenced anymore. Instead, `api.Response` is referenced. Similarly, instead of using `automation.ListResponse`, now `api.PagedListResponse` is used. Also, several types were moved from `api/clients/automation` to `clients/automation`. 

ad 3.: In two places where the client factory is used -- in the creation of the client set and in determining the Dynatrace Classic URL -- platform tokens are now passed to the factory, if they are given in `env.Auth`. 

#### **How** does it do it?

ad 2.: Specifically, the type `buckets.Response` was removed and cannot be referenced anymore. Instead, `api.Response` is referenced. Also, changes to automation clients have been made, which need to be addressed. Specifically, the type `automation.Response` was removed and cannot be referenced anymore. Instead, `api.Response` is referenced. Similarly, instead of using `automation.ListResponse`, now `api.PagedListResponse` is used. Also, several types were moved from `api/clients/automation` to `clients/automation`. 

ad 3.: In two places where the client factory is used -- in the creation of the client set and in determining the Dynatrace Classic URL -- platform tokens are now passed to the factory, if they are given in `env.Auth`. 

#### How is it **tested**?
Existing tests are adapted. Tests for client set creation and environment authentication verification have been extended with test cases featuring platform tokens.

Currently, it doesn't affect users, as the feature is still under development. As soon as it is released, users can use platform tokens instead of OAuth clients to access the Dynatrace Platform.

CA-15411


cmd/monaco/deploy/deploy.go
```diff
@@ -282,15 +278,15 @@ func validateAuthenticationWithProjectConfigs(projects []project.Project, enviro
...
-						if environments[envName].Auth.ApiToken == nil && environments[envName].Auth.OAuth == nil {
-							return fmt.Errorf("API of type '%s' requires an API token or OAuth for environment '%s'", conf.Type, envName)
+						if environments[envName].Auth.ApiToken == nil && !environments[envName].HasPlatformAuth() {
+							return fmt.Errorf("API of type '%s' requires an API token or platform authentication for environment '%s'", conf.Type, envName)
 						}
 					default:
-						if environments[envName].Auth.OAuth == nil {
-							return fmt.Errorf("API of type '%s' requires OAuth for environment '%s'", conf.Type, envName)
+						if !environments[envName].HasPlatformAuth() {
+							return fmt.Errorf("API of type '%s' requires platform authentication for environment '%s'", conf.Type, envName)
```
arthurpitman: As we're using the term platform a lot, we should probably tighten the language here:
- "platform authentication" vs "platform credentials"

And also I notice:
- "API of type ..." vs "... resources"

d0weinberger: I changed it all to "platform credentials".
In logs, "resources" is only really used in relation to the automation client, though.

Kirdock (approved):

pkg/client/clientset.go
```diff
@@ -282,13 +281,23 @@ func CreateClientSetWithOptions(ctx context.Context, url string, auth manifest.A
...
-			}).WithPlatformURL(url)
+			})
+		platformCredentialsGiven = true
+	}
+	if auth.PlatformToken != nil {
+		cFactory = cFactory.WithPlatformToken(auth.PlatformToken.Value.Value())
+		platformCredentialsGiven = true
+	}
+
+	if platformCredentialsGiven {
```
Kirdock, line 299: Technically, we could use `auth.HasPlatformCredentials()` instead, but no strong opinion for not using it.
We could then also move the `if` checks for OAuth and platform token inside this if.
At the moment, the factory includes platform-related settings when creating the classic clients, but this should not be a problem, as the classic client only takes the things it needs. Nvm, it was also like this before

d0weinberger: True, would maybe be easier to read. I can do that in a follow-up

PR #1933, opened 2025-06-16
chore(account): Add missing key to function documenation

jaylinski:
#### **Why** this PR?

Improve docs.

#### **What** has changed?

Function documentation.

#### **How** does it do it?

Adding text.

#### How is it **tested**?

Can't be tested.

Doesn't affect users.


d0weinberger (approved):
Thanks for the PR!

PR #1935, opened 2025-06-17
refactor(deploy): generify classic

Kirdock:
#### **Why** this PR?
In order to make the deploy step more generic, the classic deploy step should satisfy the Deployable interface.

#### **What** has changed?
- Added some tests (coverage was pretty low)
- Refactored some functions to be less complex

#### How is it **tested**?
- Existing tests are moved and adjusted
- New tests added

NONE

## Notes for reviewer
As I commented below, we have some dead code. Let me know, if I should remove or refactor it

### DIFF
--- a/pkg/resource/classic/deploy.go
+++ b/pkg/resource/classic/deploy.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
-	"strings"
 
 	"github.com/go-logr/logr"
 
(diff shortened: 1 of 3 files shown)


pkg/resource/classic/deploy.go
```diff
@@ -87,44 +100,49 @@ func Deploy(ctx context.Context, configClient client.ConfigClient, apis api.APIs
...
-	if val, exists := conf.Parameters[config.NonUniqueNameConfigDuplicationParameter]; exists {
-		resolvedVal, err := val.ResolveValue(parameter.ResolveContext{})
-		if err != nil {
-			return dtclient.DynatraceEntity{}, err
-		}
-		resolvedValBool, ok := resolvedVal.(bool)
-		if !ok {
-			return dtclient.DynatraceEntity{}, err
+			entityUUID = strings.ReplaceAll(entityUUID, "-", "")
+			entityUUID = strings.ToLower(entityUUID)
```
Kirdock: This is basically dead code.
The thing is, we only call this one during `upsertNonUniqueNameConfig`.
This function is only called if `NonUniqueName` is set to `true`, but `UserActionAndSessionPropertiesMobile` doesn't set it to `true`

d0weinberger: I'm just wondering why this handling was added then in the first place. I'm a bit hesitant to remove this without unterstanding the full picture

d0weinberger: Okay, I found the PR where `UserActionAndSessionPropertiesMobile` lost the `NonUniqueName: true`

Ref: https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1410/files#diff-ad64eccd8c16d62a3ad62764b7ab4518e5def0ef3309ab308aa070c9df13db77

The handling here was introduced roughly around the same time, so maybe with the (erroneous) thought in mind, that it still needs to be handled here.
So, I think you're right. It's dead code.

d0weinberger (approved):

PR #1937, opened 2025-06-17
fix: Uses corelib version which closes connections for each request

d0weinberger:


d0weinberger:
I'll close this PR and remove the associated branch. It got us nowhere, and we have a solution for now with the `-p 1`option.

PR #1938, opened 2025-06-17
feat(download): Support direct download with platform token

arthurpitman:
#### **Why** this PR?
Monaco can download configs "directly" by specifying the environment URL and credentials on the CLI. This PR adds support for doing this with a platform token, and adds the necessary `--platform-token PLATFORM_TOKEN` flag.

#### **What** has changed?
Direct download is now possible with:
- just an API token
- **new**: just OAuth credentials, something that has worked with a manifest for some time
- **new**: just a platform token
- an API token and OAuth credentials
- **new**: an API token and a platform token

#### **How** does it do it?
- Some refactoring is done to try to reduce complexity
- A small change to how the manifest file name is handled means that an error is always given if the `--url` flag is used as well as a manifest file name
- The `--platform-token PLATFORM_TOKEN` flag is added
- Logic is updated to add a platform token to the auth structure

#### How is it **tested**?
- Existing tests are updated
- New tests are added with explicit use of a platform token.

- If the `MONACO_FEAT_ENABLE_PLATFORM_TOKENS` is enabled, users can specify a platform for direct download via `--platform-token PLATFORM_TOKEN`

Reference: CA-15413

### DIFF
--- a/cmd/monaco/download/download_command.go
+++ b/cmd/monaco/download/download_command.go
@@ -66,7 +66,7 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
   monaco download [--%s manifest.yaml] --%s MY_ENV ...
 
   # download without manifest
-  monaco download --%s url --%s DT_TOKEN [--%s CLIENT_ID --%s CLIENT_SECRET] ...`, ManifestFlag, EnvironmentFlag, UrlFlag, ApiTokenFlag, OAuthIdFlag, OAuthSecretFlag),
+  monaco download --%s url [--%s DT_TOKEN] [--%s CLIENT_ID --%s CLIENT_SECRET] ...`, ManifestFlag, EnvironmentFlag, UrlFlag, ApiTokenFlag, OAuthIdFlag, OAuthSecretFlag),
 
 		PreRunE: func(cmd *cobra.Command, args []string) error {
 			if f.environmentURL != "" {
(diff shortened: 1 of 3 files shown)


cmd/monaco/download/download_command.go
```diff
@@ -147,31 +163,39 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
...
+}
 
-	return nil
+func preRunChecksForManifestDownload(f downloadCmdOptions) error {
+	switch {
+	case f.apiToken != "" || f.clientID != "" || f.clientSecret != "" || f.platformToken != "":
+		if featureflags.PlatformToken.Enabled() {
+			return fmt.Errorf("'%s', '%s', '%s', and '%s' can only be used with '%s', while '%s' must NOT be set ", ApiTokenFlag, OAuthIdFlag, OAuthSecretFlag, PlatformTokenFlag, UrlFlag, ManifestFlag)
+		}
+		return fmt.Errorf("'%s', '%s', and '%s' can only be used with '%s', while '%s' must NOT be set ", ApiTokenFlag, OAuthIdFlag, OAuthSecretFlag, UrlFlag, ManifestFlag)
```
Laubi: I really think we should never have extracted the flags, this is just not readable on what the message should look like in the end 😞 

Also: Space at the end `... set "`

cmd/monaco/download/download_command.go
```diff
@@ -147,31 +163,39 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
...
+	case f.specificEnvironmentName == "":
```
Laubi: Can we please have ifs instead?

```
if 	case f.specificEnvironmentName == "":
 return ...

if ....
 return ...

return nil

File: cmd/monaco/download/download_command.go (hunk 1 above)
Comment (Laubi) [reply cmd/monaco/download/download_command.go]: Should we also include the `--` before the flags? I think it would be clearer for users that the flags are meant, like in the help.

File: cmd/monaco/download/download_command.go (hunk 2 above)
Comment (arthurpitman) [reply cmd/monaco/download/download_command.go]: I _would_ prefer `if`s personally, but learnt that Golang actually encourages this: https://go.dev/tour/flowcontrol/11: 

> This construct can be a clean way to write long if-then-else chains.

I did work to remove the nested switches, but am ok with both possibilities

File: cmd/monaco/download/download_command.go (hunk 1 above)
Comment (arthurpitman) [reply cmd/monaco/download/download_command.go]: I will fix this and try, in a separate commit, to simplify the language 🚀

File: cmd/monaco/download/download_command.go [hunk 3]
```diff
@@ -147,31 +163,39 @@ func GetDownloadCommand(fs afero.Fs, command Command) (cmd *cobra.Command) {
...
-		}
-	case f.manifestFile != "":
-		switch {
-		case f.apiToken != "" || f.clientID != "" || f.clientSecret != "":
-			return fmt.Errorf("'%s', '%s' and '%s' can only be used with '%s', while '%s' must NOT be set ", ApiTokenFlag, OAuthIdFlag, OAuthSecretFlag, UrlFlag, ManifestFlag)
-		case f.specificEnvironmentName == "":
-			return fmt.Errorf("to download with manifest, '%s' needs to be specified", EnvironmentFlag)
+	case (f.apiToken == "") && (f.clientID == "") && (f.clientSecret == "") && (f.platformToken == ""):
+		if featureflags.PlatformToken.Enabled() {
+			return fmt.Errorf("if '%s' is set, '%s', '%s' and '%s', or '%s' must also be set", UrlFlag, ApiTokenFlag, OAuthIdFlag, OAuthSecretFlag, PlatformTokenFlag)
```
d0weinberger: To me, this read as "If url is set, api token, client id and client secret (all 3 of them!) OR platform token needs to be set." but only api token or only client id and secret are also valid, right?
So, I would go for `if '%s' is set, '%s', or '%s' and '%s', or '%s' must also be set`

arthurpitman: Fixed

cmd/monaco/download/download_command.go
arthurpitman: I added a commit to use `if` statements; let's see which we like better

cmd/monaco/download/download_command_test.go
```diff
@@ -115,22 +157,76 @@ func TestGetDownloadCommand(t *testing.T) {
...
+			projectName: "project",
+			onlyOptions: defaultOnlyOptions,
+		}
+		m.EXPECT().DownloadConfigs(gomock.Any(), gomock.Any(), expected).Return(nil)
+
+		err := m.download("--url http://some.url --platform-token PLATFORM_TOKEN")
+		assert.NoError(t, err)
+	})
+
+	t.Run("Direct download - api token and platform token", func(t *testing.T) {
```
d0weinberger: [nit] `API`

PR #1942, opened 2025-06-23
refactor(account): remove service user feature flag

Kirdock:
#### **Why** this PR?
The service user feature flag is removed, and everywhere it was used, it's replaced by the `true` condition of this flag.

**Issue:** CA-15453

### DIFF
--- a/pkg/account/delete/loader_test.go
+++ b/pkg/account/delete/loader_test.go
@@ -25,12 +25,10 @@ import (
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
 
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/account/delete"
 )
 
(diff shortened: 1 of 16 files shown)


pkg/account/delete/loader_test.go
```diff
@@ -81,19 +79,7 @@ func TestLoader_BasicAllTypesSucceeds(t *testing.T) {
...
-  - type: serviceUser
-    name: my-service-user`)
-
-	_, err := delete.LoadResourcesToDelete(fs, deleteFilename)
-	assert.Error(t, err)
-}
-
-func TestLoader_ServiceUserProducesNoErrorWithFeatureFlag(t *testing.T) {
-	t.Setenv(featureflags.ServiceUsers.EnvName(), "true")
+func TestLoader_ServiceUserProducesNoError(t *testing.T) {
```
arthurpitman: Is this test still required considering a service user is included in `TestLoader_BasicAllTypesSucceeds`?

Kirdock: I'm now wondering, why we had this test in the first place if the one before covers it 🤔 
I'll remove it 👍

PR #1943, opened 2025-06-23
refactor(segments): remove feature flag

Kirdock:
#### **Why** this PR?
The segment feature flag is removed, and everywhere it was used, it's replaced by the `true` condition of this flag.

#### **What** has changed?
Segments FF is removed

**Issue:** CA-15447

### DIFF
--- a/test/commands/config_restore_e2e_test.go
+++ b/test/commands/config_restore_e2e_test.go
@@ -148,7 +148,6 @@ func TestRestoreConfigs_FromDownloadWithPlatformOAuthManifestFile_withPlatformCo
 	subsetOfConfigsToDownload := "alerting-profile,management-zone"
 	suffixTest := "_download_automations"
 
-	t.Setenv(featureflags.Segments.EnvName(), "true")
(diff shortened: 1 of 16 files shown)


d0weinberger (dismissed):

d0weinberger (dismissed):

test/commands/config_restore_e2e_test.go
```diff
@@ -109,7 +109,6 @@ func TestRestoreConfigs_FromDownloadWithPlatformManifestFile_withPlatformConfigs
 	subsetOfConfigsToDownload := "alerting-profile,management-zone"
 	suffixTest := "_download_automations"
```
arthurpitman, line 161: Not strictly related to this PR, why do we have the suffix `_download_automations`?

arthurpitman (approved):

test/commands/config_restore_e2e_test.go
Kirdock: I assume this was a copy&paste thing a while ago or the test got bigger over time (only-automation => everything) 🤔

d0weinberger (approved):

PR #1944, opened 2025-06-24
refactor(slo-v2): remove feature flag

Kirdock:
#### **Why** this PR?
The slo-v2 feature flag is removed, and everywhere it was used, it's replaced by the `true` condition of this flag.

#### **What** has changed?
SLO-V2 FF is removed

**Issue:** CA-15448

### DIFF
--- a/pkg/deploy/deploy_test.go
+++ b/pkg/deploy/deploy_test.go
@@ -1306,35 +1306,6 @@ func TestDeployConfigGraph_CollectsAllErrors(t *testing.T) {
 
 }
 
-// Test if the FF gets Disabled/Enabled that the correct error is returned
-func TestDeployConfigFF(t *testing.T) {
-	// Clients here will return an error if reached
-	dummyClientSet := client.ClientSet{}
-	c := dynatrace.EnvironmentClients{
-		dynatrace.EnvironmentInfo{Name: "env"}: &dummyClientSet,
-	}
-	tests := []struct {
-		name              string
-		projects          []project.Project
-		featureFlag       string
-		configType        config.TypeID
-		expectedErrString string
-	}{}
-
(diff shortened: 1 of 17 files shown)


test/commands/all_configs_integration_test.go
```diff
@@ -61,8 +60,6 @@ func TestIntegrationAllConfigsPlatform(t *testing.T) {
 	manifest := configFolder + "manifest.yaml"
 
 	t.Setenv(featureflags.OpenPipeline.EnvName(), "true")
-	t.Setenv(featureflags.ServiceLevelObjective.EnvName(), "true")
-	t.Setenv(featureflags.AccessControlSettings.EnvName(), "true")
```
d0weinberger: This one is removed by a different PR, actually.

Kirdock: Seems like I double tapped the `CTRL+Y` 😅 .
As this is the default behavior and has been removed in a different PR, it should not matter

pkg/deploy/deploy_test.go
```diff
@@ -1322,32 +1320,7 @@ func TestDeployConfigFF(t *testing.T) {
...
-							},
-						},
-					},
-				},
-			},
-			featureFlag: featureflags.ServiceLevelObjective.EnvName(),
-			configType:  config.ServiceLevelObjectiveID,
-		},
-	}
+	}{}
```
Kirdock: Not sure about this one. At the moment, this test is empty. Should I completely remove it? We might need it again, when adding new configs

d0weinberger: I'd say, currently it's dead code and we could still add it again later when/if we need it 🤷‍♂️

d0weinberger (dismissed):

PR #1945, opened 2025-06-24
refactor(openpipeline): remove feature flag

Kirdock:
#### **Why** this PR?
The openpipeline feature flag is removed, and everywhere it was used, it's replaced by the `true` condition of this flag.

#### **What** has changed?
Openpipeline FF is removed

**Issue:** CA-15449

### DIFF
--- a/pkg/client/test_clientset.go
+++ b/pkg/client/test_clientset.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
-	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients/openpipeline"
 )
 
 // TestSegmentsClient is a fake client that returns an unimplemented error on every execution of any method.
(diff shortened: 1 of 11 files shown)


pkg/client/test_clientset.go
```diff
@@ -69,13 +68,3 @@ func (TestServiceLevelObjectiveClient) Create(ctx context.Context, data []byte)
 func (TestServiceLevelObjectiveClient) Delete(ctx context.Context, id string) (api.Response, error) {
 	return api.Response{}, fmt.Errorf("unimplemented")
 }
-
-type TestOpenPipelineClient struct{}
```
d0weinberger, line 73: So, if OpenPipeline was also present in delete_test.go, we would still need this client, analogously to SLO and segments, right? But as we don't have delete for OpenPipeline, we don't have the test and therefore don't need the client here anymore. Correct?

Kirdock: Correct 👍

arthurpitman (approved):

PR #1950, opened 2025-06-24
refactor: Use `slog.Attr` directly rather than custom `Field` type

arthurpitman:
#### **Why** this PR?
This PR is an important step to using slog directly, as it moves from a custom field type to `slog.Attr`.

This will make it easier to convert each individual log line to use slog.

Furthermore, by reducing the number of loggers and intermediate values created when logging, it should reduce memory pressure/consumption.

This PR should be reviewed commit by commit.

Firstly, renaming is done to bring greater synergy with slog:
- Package `field` named to `attribute`
- Type `Field` renamed to `Attr`
- Functions `WithFields(...) `renamed to `With(...)`
- Function `F(...)` renamed to `Any(...)`

Then, the `field.Field` type is removed in favor of `slog.Attr` and the `With(...)` functions are simplified to just forward to the underlying `slog.Logger`.

Finally, the helper functions to create attributes are updated to create `slog.Attr` directly rather than using intermediate structs.

#### How is it **tested**?
Existing test, with a small number of updates

No effect.

**Issue:**  CA-15581


pkg/config/coordinate/coordinate.go
```diff
@@ -46,6 +46,7 @@ func (c Coordinate) Match(coordinate Coordinate) bool {
 // It returns a group containing the fields of the coordinate so that they appear together in log output
 func (c Coordinate) LogValue() slog.Value {
 	return slog.GroupValue(
+		slog.String("reference", c.String()),
```
Kirdock, line 49: Do we really need this one?
Because we already log the project, type, and configId, I would not see the need to also log the reference (concatenation with `:`)

pkg/account/deployer/deployer.go
```diff
@@ -390,7 +390,7 @@ func (d *AccountDeployer) updateGroupPolicyBindings(ctx context.Context, group a
 	}
 
 	for env, uuids := range envPolicyUuids {
-		d.logger.WithFields().DebugContext(ctx, "Updating environment level policy bindings for group with ID %s and environment with name %s --> %v", remoteGroupId, env, uuids)
+		d.logger.With().DebugContext(ctx, "Updating environment level policy bindings for group with ID %s and environment with name %s --> %v", remoteGroupId, env, uuids)
```
Kirdock: Do we even need the `With` here?

pkg/config/coordinate/coordinate.go
arthurpitman: I would agree, but I left this in case anyone is expecting to find the format `{project}:{type}:{configId}`. @Laubi , what do you think?

internal/log/attribute/attribute.go
```diff
@@ -0,0 +1,70 @@
+/*
```
Laubi: Why an extra package for this? Let's just keep have it in `log`?
It's one less unnecessary import everywhere, and packages should not be so small, but belong together. Since this is about logging, i'd just keep it there

arthurpitman:
> Not sure about the `slog.Any`. We use it for basically everything, even for basic types. While the output doesn't change at all, I'm not sure if this has other impacts. Maybe performance because there are some type checks inside of it, but probably not big enough to consider

I would say it doesn't matter for now, and now that we're using it, we can easily change to something better when we update each log line.

pkg/config/coordinate/coordinate.go
Laubi: Agree. I think the coordinate should not do it at all, but the `Coordinate` function that returns the `Attr`

internal/log/attribute/attribute.go
arthurpitman: I can do this, but will rename the functions in that package from `X(...)` to `XAttr(...)`, otherwise they look awkward and might clash (for example, `Error(...)` already exists).  This is also a temporary measure as I'm not sure how many of these we will ultimately use, as environment and coordinate are normally extracted from the context automatically.

pkg/config/coordinate/coordinate.go
arthurpitman: I was hoping to remove the `CoordinateAttr(...)` in some time and instead rely on the implementation of `slog.LogValuer`.

Laubi: I see, that makes also quite some sense 👍

Laubi: Since the project report no longer depends on the format, we can choose whatever we want. I am happy with `coordinate=a:b:c`

Log line could look like this:
```json
{
  "msg": "Deploying",
  "coordinate" : "a:b:c"
}
```

Laubi (approved):
All my comments can be follow ups

internal/log/attribute/attribute.go
Laubi: What is more - we could move some of the functions to the correct places they are used. E.g. deploy statues should probably be in the deploy package

PR #1952, opened 2025-06-24
Feat/add platform token e2e tests

d0weinberger:
#### **Why** this PR?
Adding some end-to-end tests for platform token authentication

#### **What** has changed?
The dry run test has been extended, as well as the all_configs integration test.

#### **How** does it do it?
The test manifest files have been extended with an environment definition using platform token authentication

It doesn't.

**Issue:** CA-15412


Kirdock (dismissed):

test/commands/testdata/integration-download-configs-platform/platform_token_manifest.yaml
```diff
@@ -0,0 +1,15 @@
+manifestVersion: "1.0"
+projects:
+  - name: project
+environmentGroups:
+  - name: default
+    environments:
+      - name: environment1
```
arthurpitman, line 7: I would suggest renaming these to prevent any confusion, would `platform_env` work?

d0weinberger: Hm, there are many occurrences of `environment1` across different tests. If it's okay with you, I'd like to do that in a follow-up PR

Kirdock (approved):

test/commands/testdata/integration-download-configs-platform/platform_token_manifest.yaml
arthurpitman: Good idea 👍

test/commands/testdata/integration-all-configs/manifest.yaml
arthurpitman, line 1: A bit of a nit, but could we rename `platform_env` to `platform_oauth_env`? It is a little confusing here and even more so in the next commit, particularly with the overrides.

test/commands/testdata/integration-all-configs/project/openpipeline/config.yaml
```diff
@@ -2,7 +2,7 @@ configs:
 - id: events
   config:
     template: events.json
-    skip: true # platform only config
+    skip: true # platform (with OAuth) only config
   environmentOverrides:
     - environment: platform_env
```
arthurpitman, line 7: This would look better if it were `platform_oauth_env`

d0weinberger: Okay, will do in a follow up

test/commands/testdata/integration-all-configs/manifest.yaml
d0weinberger: Okay, will do in a follow up

PR #1956, opened 2025-06-26
refactor: Remove unnecessary `TimelineProvider`

Laubi:
#### **Why** this PR?
I wanted to try out the experimental `synctest` package, which actually works really fine (see first commits), but then dug deeper and deeper into this code and removed quite a bunch of unnecessary functions/code, etc. 

This PR now removes the `TimelineProvider` and `throttle` package

#### **What** has changed?
This PR removes the `TimelineProvider` and `throttle` package.

**Issue:** None


pkg/client/dtclient/list.go
```diff
@@ -40,43 +41,47 @@ const emptyResponseRetryMax = 10
 // as filtering might exclude some entries that where received from the API.
 type AddEntriesToResult func(body []byte) (receivedEntries int, err error)
 
-func listPaginated(ctx context.Context, client *corerest.Client, endpoint string, queryParams url.Values, logLabel string,
-	addToResult AddEntriesToResult) error {
+func listPaginated(ctx context.Context, client *corerest.Client, endpoint string, queryParams url.Values, schemaId string, addToResult AddEntriesToResult) error {
+	logger := slog.With("endpoint", endpoint, "schemaId", schemaId)
```
Kirdock, line 45: At the moment, we don't use slog directly. Should we align it with the rest?

```suggestion
	logger := logger.With("endpoint", endpoint, "schemaId", schemaId)
```

pkg/client/dtclient/list.go
```diff
@@ -100,3 +105,21 @@ func runAndProcessResponse(ctx context.Context, client *corerest.Client, endpoin
...
+// to be used between API calls.
+func generateSleepDuration(backoffMultiplier uint) time.Duration {
+	const backoffTime = 1 * time.Second
+
+	waitNanos, err := rand.Int(backoffTime.Nanoseconds())
+	if err != nil {
+		// Since we are not reliant to cryptographically secure numbers, we can just ignore the error and assign some number.
+		// It's sound enough to just wait the backoff time by default
+		waitNanos = backoffTime.Nanoseconds()
+	}
```
Kirdock, line 122: Do we really need the random wait multiplier?

pkg/client/dtclient/list.go
Laubi: No, this has been discussed with @arthurpitman that we should move to `slog`

pkg/client/dtclient/list.go
Laubi: Using Jitter is a best practice for backoffs. It avoids spikes that can impact a service. Is it absolutely necessary here? Probably not.

Kirdock (approved):

pkg/client/dtclient/list.go
```diff
@@ -40,43 +41,47 @@ const emptyResponseRetryMax = 10
...
-				return fmt.Errorf("received too many empty responses (=%d)", emptyResponseRetryCount)
+			if retryCount >= emptyResponseRetryMax {
+				return fmt.Errorf("received too many empty responses (=%d)", retryCount)
 			}
 
-			emptyResponseRetryCount++
-			throttle.ThrottleCallAfterError(emptyResponseRetryCount, "Received empty array response, retrying with same nextPageKey")
+			retryCount++
+
+			sleepDuration := generateSleepDuration(retryCount)
```
d0weinberger, line 73: [nit] if you log it as `waitDuration`, you could also call the variable `waitDuration`

pkg/client/dtclient/list.go
```diff
@@ -40,43 +41,47 @@ const emptyResponseRetryMax = 10
...
-	emptyResponseRetryCount := 0
+	retryCount := uint(0)
 	for nextPageKey != "" {
 
 		body, receivedCount, err := runAndProcessResponse(ctx, client, endpoint, corerest.RequestOptions{QueryParams: makeQueryParamsWithNextPageKey(endpoint, queryParams, nextPageKey), CustomShouldRetryFunc: corerest.RetryIfTooManyRequests}, addToResult)
 		if err != nil {
 			var apiErr coreapi.APIError
 			if errors.As(err, &apiErr) && apiErr.StatusCode == http.StatusBadRequest {
-				log.WarnContext(ctx, "Failed to get additional data from paginated API %s - pages may have been removed during request.\n    Response was: %s", endpoint, string(body))
+				logger.WarnContext(ctx, "Failed to get additional data from paginated API. Pages may have been removed during request.", "response", string(body))
```
d0weinberger, line 60: Is the endpoint not relevant anymore?

EDIT: Oh wait, it's added using `slog.With` before

PR #1960, opened 2025-06-27
fix: revert back to 1c13eb7cfe77d958a1939c08a290d105ab2d54ef

d0weinberger:

**Issue:**


d0weinberger:
I'll close this PR and remove the associated branch. It got us nowhere, and we have a solution for now with the `-p 1`option.

PR #1964, opened 2025-06-30
refactor(bucket): Introduce Deleter interface and use it with buckets

d0weinberger:
#### **Why** this PR?
We want to introduce a `Deleter` interface which every resource client which wants to featured in the `monaco delete` command must adhere to. The goal is that the `delete` command doesn't know about the details about what it deletes -- it just iterates over a collection of `Deleter`s. 
 
#### **What** has changed?
A `Deleter` interface was introduced which provides two methods: `Delete` for `monaco delete`, and `DeleteAll` for `monaco purge`. 

#### **How** does it do it?
This interface is used with the bucket resources first. In subsequent PRs, the other resources will also implement the `Deleter` interface. In the end, `delete.go` should no longer care about the concrete resource type.

#### How is it **tested**?
Existing tests pass through. Bucket deletion tests are moved from `delete/delete_test.go` to `resource/bucket/delete_test.go`

It doesn't

**Issue:** CA-15292


pkg/resource/bucket/delete.go
```diff
@@ -29,12 +29,20 @@ import (
...
 )
 
-type client interface {
+type DeleteSource interface {
 	Delete(ctx context.Context, id string) (api.Response, error)
 	List(ctx context.Context) (buckets.ListResponse, error)
 }
 
-func Delete(ctx context.Context, c client, entries []pointer.DeletePointer) error {
+type DeleteAPI struct {
```
Laubi: Why do we call these things 'API'? Where is the API part?

I know we have it for the others as well, but it does not make sense to me 🤔

Laubi (approved):

pkg/delete/delete.go
```diff
@@ -96,7 +101,7 @@ func deleteConfig(ctx context.Context, clients client.ClientSet, t string, entri
 		log.With(log.TypeAttr(t)).WarnContext(ctx, "Skipped deletion of %d Classic configuration(s) as API client was unavailable.", len(entries))
 	} else if t == "bucket" {
```
Kirdock, line 102: Just curious, how will the solution in the end look? A map with each and every classic API and resource, where settings is the fallback instead of the switch?
Or will just every client get the delete entries and take the ones that it supports?

pkg/resource/bucket/delete.go
d0weinberger: Maybe I just call it `Deleter`. As in `bucket.Deleter`.

pkg/delete/delete.go
Kirdock: Would we then still need to handle automation differently?

arthurpitman (approved):
Looks good as a first step for buckets

pkg/delete/delete.go
Kirdock: Discussed: As a first step, this is good enough for alignment

Kirdock (approved):

PR #1967, opened 2025-07-01
refactor: Move deploy-specific attributes to their package

Laubi:
#### **Why** this PR?
Super simple PR that moves the deploy-attributes to the `deploy` package

#### **What** has changed?
* Moved deploy-specific attributes
* Changed attr types to be more specific

#### How is it **tested**?
Added some really trivial tests to verify that the structure of the attributes stays the same. 

Doesn't

**Issue:** None

### DIFF
--- a/internal/log/attribute_test.go
+++ b/internal/log/attribute_test.go
@@ -1,0 +1,59 @@
+/*
+ * @license
+ * Copyright 2025 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package log
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/coordinate"
+)
+
+func TestEnvironmentAttr(t *testing.T) {
+	attr := EnvironmentAttr("env", "group")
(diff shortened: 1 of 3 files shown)


internal/log/attribute_test.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+type customError struct {
+}
+
+func (customError) Error() string {
+	return "custom error"
+}
+
+func TestGroupAttr(t *testing.T) {
```
arthurpitman:
```suggestion
func TestErrorAttr(t *testing.T) {
```

internal/log/attribute_test.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func TestCoordinateAttr(t *testing.T) {
+	attr := CoordinateAttr(coordinate.Coordinate{
+		Project:  "p",
+		Type:     "t",
+		ConfigId: "c",
+	})
+
+	assert.Equal(t, "coordinate=p:t:c", attr.String())
```
arthurpitman, line 58: I'm not sure if this test has value, considering `LogValue()` would be called rather than `String()` as (I think) `slog` uses `LogValuer` rather than `Stringer` when resolving values.

Laubi: I agree that the logging would look different, but writing a testing log handler was out of scope for me for now :D

pkg/deploy/deploy.go
```diff
@@ -368,19 +368,41 @@ func deployConfig(ctx context.Context, c *config.Config, deployables resource.De
...
 }
 
 func newContextWithEnvironment(ctx context.Context, env dynatrace.EnvironmentInfo) context.Context {
 	return context.WithValue(ctx, log.CtxKeyEnv{}, log.CtxValEnv{Name: env.Name, Group: env.Group})
 }
+
+const deploymentStatus = "deploymentStatus"
+
+// statusDeploying returns an attribute with deploymentStatus set to 'deploying'.
+func statusDeploying() slog.Attr {
```
arthurpitman, line 391: nit: Shall we rename this `statusDeployingAttr` for consistency?

internal/log/attribute_test.go
arthurpitman: Makes sense 👍

tomazpu (approved):

PR #1968, opened 2025-07-02
feat: Enable `MONACO_FEAT_ENABLE_PLATFORM_TOKENS` feature flag by default

arthurpitman:
#### **Why** this PR?
Now that all parts of platform token support are complete, this feature flag should be enabled by default.

#### **What** has changed?
If `MONACO_FEAT_ENABLE_PLATFORM_TOKENS` is not set, the platform token feature flag is assumed to be enabled by default.

#### **How** does it do it?
Changed the default in `internal/featureflags/temporary.go`.

#### How is it **tested**?
Existing tests cover both the feature flag being turned on and off. Tests are updated to reflect the change in default value.

Yes, platform tokens will be supported by default. Users can disable the functionality by setting  `MONACO_FEAT_ENABLE_PLATFORM_TOKENS` to `false` or `0`.

**Issue:** CA-15415

### DIFF
--- a/cmd/monaco/download/download_command_test.go
+++ b/cmd/monaco/download/download_command_test.go
@@ -85,17 +85,17 @@ func TestGetDownloadCommand(t *testing.T) {
 
 	t.Run("Download using manifest - access token cannot be specified", func(t *testing.T) {
 		err := newMonaco(t).download("--token API_TOKEN")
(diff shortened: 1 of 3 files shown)


arthurpitman:
I don't see any failures in the E2E tests that are relevant to this PR

Kirdock (approved):

PR #1969, opened 2025-07-02
refactor(bucket): use core lib without upsert and active check

Kirdock:
#### **Why** this PR?
The logic for updating a bucket only when it can be is now handled by Monaco and not the core lib. The core lib also removes upsert, which Monaco depends on.

#### **What** has changed?
- The logic for updating a bucket only when it can be, is now handled by Monaco and not the core lib
- The upsert was removed, so Monaco now handles the upsert (create/update) logic.

#### **How** does it do it?
By splitting Upsert into create & update, and adding some checks before it can be updated/deleted.

#### How is it **tested**?
New tests are added, existing ones are adapted, and some existing ones are moved closer to the package that is being tested.

NONE

#### Core changes
https://github.com/Dynatrace/dynatrace-configuration-as-code-core/pull/235

### Notes
Only realized later that there were adjustments of other PRs pointing to the same files. Anyway, a bit of refactoring doesn't hurt I guess 😉 

**Issue:** CA-15100

### DIFF
--- a/pkg/client/clientset.go
+++ b/pkg/client/clientset.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"net/url"
 	"runtime"
-	"time"
 
 	"golang.org/x/oauth2/clientcredentials"
 
@@ -162,7 +161,6 @@ type BucketClient interface {
 	List(ctx context.Context) (buckets.ListResponse, error)
 	Create(ctx context.Context, bucketName string, data []byte) (libAPI.Response, error)
 	Update(ctx context.Context, bucketName string, data []byte) (libAPI.Response, error)
-	Upsert(ctx context.Context, bucketName string, data []byte) (libAPI.Response, error)
 	Delete(ctx context.Context, bucketName string) (libAPI.Response, error)
 }
 
@@ -302,7 +300,7 @@ func CreateClientSetWithOptions(ctx context.Context, url string, auth manifest.A
 			return nil, err
 		}
 
-		bucketClient, err = cFactory.BucketClientWithRetrySettings(ctx, time.Second, 5*time.Minute)
+		bucketClient, err = cFactory.BucketClient(ctx)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/resource/bucket/delete.go
+++ b/pkg/resource/bucket/delete.go
@@ -30,6 +30,7 @@ import (
 )
 
 type DeleteSource interface {
+	Get(ctx context.Context, bucketName string) (api.Response, error)
 	Delete(ctx context.Context, id string) (api.Response, error)
 	List(ctx context.Context) (buckets.ListResponse, error)
 }
@@ -57,7 +58,17 @@ func (d Deleter) Delete(ctx context.Context, entries []pointer.DeletePointer) er
 		}
 
 		logger.DebugContext(ctx, "Deleting bucket '%s'", bucketName)
-		_, err := d.bucketSource.Delete(ctx, bucketName)
+		bucketExists, err := buckets.AwaitActiveOrNotFound(ctx, c, bucketName, maxRetryDuration, durationBetweenRetries)
+		if err != nil {
+			logger.With(log.ErrorAttr(err)).ErrorContext(ctx, "Failed to delete Grail Bucket '%s': %v", bucketName, err)
+			deleteErrs++
+			continue
+		}
+		if !bucketExists {
+			// bucket already deleted
+			continue
+		}
+		_, err = d.bucketSource.Delete(ctx, bucketName)
 		if err != nil {
 			if !api.IsNotFoundError(err) {
 				logger.With(log.ErrorAttr(err)).ErrorContext(ctx, "Failed to delete Grail Bucket '%s': %v", bucketName, err)
@@ -108,8 +119,18 @@ func (d Deleter) DeleteAll(ctx context.Context) error {
 		if buckettools.IsDefault(bucketName.BucketName) {
 			continue
 		}
-
-		_, err := d.bucketSource.Delete(ctx, bucketName.BucketName)
+		// before deleting wait until it can be deleted
+		bucketExists, err := buckets.AwaitActiveOrNotFound(ctx, c, bucketName.BucketName, maxRetryDuration, durationBetweenRetries)
+		if err != nil {
+			logger.ErrorContext(ctx, "Failed to delete Grail Bucket '%s': %v", bucketName.BucketName, err)
+			errCount++
+			continue
+		}
+		if !bucketExists {
+			// bucket already deleted
+			continue
+		}
+		_, err = d.bucketSource.Delete(ctx, bucketName.BucketName)
 		if err != nil {
 			if !api.IsNotFoundError(err) {
 				logger.ErrorContext(ctx, "Failed to delete Grail Bucket '%s': %v", bucketName.BucketName, err)
--- a/pkg/resource/bucket/delete.go
+++ b/pkg/resource/bucket/delete.go
@@ -26,6 +26,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/buckettools"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/idutils"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config/coordinate"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/pointer"
 )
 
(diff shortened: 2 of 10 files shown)


pkg/resource/bucket/deploy.go
```diff
@@ -72,3 +76,16 @@ func (d DeployAPI) Deploy(ctx context.Context, properties parameter.Properties,
...
+func (d DeployAPI) upsert(ctx context.Context, bucketName string, data []byte) error {
+	// Check the status of a bucket (updating/creating/deleting) and if it even exists
+	if bucketExists, err := buckets.AwaitBucketStable(ctx, d.source, bucketName, client.BucketMaxRetryDuration, client.BucketDurationBetweenRetries); err != nil {
+		return err
+	} else if bucketExists {
+		_, err := d.source.Update(ctx, bucketName, data)
+		return err
+	}
+
+	_, err := d.source.Create(ctx, bucketName, data)
```
d0weinberger: So, to be consistent with the behavior as it was before, we would need to call `AwaitBucketStable` again before returning here, don't we? Before the corelib change, `Create` waited for the bucket to become active before it returned.

Kirdock: Part 1 of the removal was about finding out if there are any breaking changes if it is removed (https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1965). If there weren't any breaking changes, we said that we would leave it like that without the active check. Part 1 resulted in breaking changes for update and delete buckets, but not for other related resources.
Because bucket tests and tests that depend on buckets (schema that uses the bucket) are not breaking with the current implementation, I would have seen it as part 1, where we simply remove it.

@Laubi @arthurpitman @tomazpu, your thoughts on that? I could also add the check again if wanted

d0weinberger: It definitely changes the behavior. My comment also applies to the update in line 85 and the delete in `delete.go` by the way. What if a resource that depends on a bucket uses it before it's there? Whether that is an issue we need care about, I don't know.

pkg/client/clientset.go
```diff
@@ -157,12 +157,16 @@ type AutomationClient interface {
 	Delete(ctx context.Context, resourceType automation.ResourceType, id string) (libAPI.Response, error)
 }
 
+const (
+	BucketMaxRetryDuration       = time.Minute * 5
```
d0weinberger: I would move that actually to the `resource/bucket` package. It's bucket internals IMHO not of interest to the clientset.

pkg/resource/bucket/deploy.go
d0weinberger: Could it be that the `Create` response is positive but the bucket never becomes active or something fails on the server side when trying to get the bucket active? That would be an issue that we up til now would maybe fail the deployment on, but with the change, we wouldn't anymore.

pkg/client/clientset.go
Kirdock: Now, after moving the delete stuff to buckets, yes, this is definitely a better place. Will do.

Edit: done 👍

pkg/resource/bucket/deploy.go
Kirdock: If a bucket doesn't become active, this should not be our problem. I would say. Monaco deployed it, so its part is done. 
For other configs that depend on the bucket, we already have a retry mechanism in place (if config references a bucket => other retry settings), so this will still be in place, in case the API can't handle a bucket that is currently creating.

Kirdock: Discussed to keep the status check after create 👍

d0weinberger (approved):

d0weinberger:
LGTM but imho it makes sense to wait for at least one additional review

arthurpitman (commented):
The changes look good, just a couple of ideas:
- Seeing `buckets.AwaitBucketStable(...)` being used, I think it might make sense to rename it `buckets.AwaitActiveOrNotFound(...)`

- Would it possible to remove the duplication between `Delete(...)` and `DeleteAll(...)`?

- Nit: Would it be possible to rename `deleteErrs` to `errorCount`(although we might not need to do this if the implementatons of `Delete` and `DeleteAll` are merged).

pkg/resource/bucket/delete.go
```diff
@@ -99,45 +77,91 @@ func (d Deleter) DeleteAll(ctx context.Context) error {
...
+func responsesToBucketDelete(ctx context.Context, bucketResponses [][]byte, logger *log.Slogger) ([]bucketDelete, int) {
+	var bucketName struct {
+		BucketName string `json:"bucketName"`
+	}
+	bucketDeletes := make([]bucketDelete, 0)
+	errCount := 0
 
+	for _, obj := range bucketResponses {
 		if err := json.Unmarshal(obj, &bucketName); err != nil {
 			logger.ErrorContext(ctx, "Failed to parse bucket JSON: %v", err)
```
arthurpitman: Nit: In this log message, we refer to "bucket JSON". Would "Grail Bucket JSON" be more fitting?

pkg/resource/bucket/delete.go
```diff
@@ -99,45 +77,91 @@ func (d Deleter) DeleteAll(ctx context.Context) error {
...
 		}
-		// before deleting wait until it can be deleted
-		bucketExists, err := buckets.AwaitActiveOrNotFound(ctx, d.bucketSource, bucketName.BucketName, maxRetryDuration, durationBetweenRetries)
+
+		logger := baseLogger
+		if bucketDeleteEntry.coordinate != nil {
+			logger = logger.With(log.CoordinateAttr(*bucketDeleteEntry.coordinate))
+		}
+
+		logger.DebugContext(ctx, "Deleting bucket '%s'", bucketName)
```
arthurpitman: Nit: here also: "bucket" vs "Grail Bucket".

Kirdock: I'll make it `Grail bucket`. Lower case bucket to be aligned with [docs](https://docs.dynatrace.com/docs/discover-dynatrace/platform/grail/data-model#custom-grail-buckets).

pkg/resource/bucket/delete.go
Kirdock: Yes, why not. To be alinged with the rest 👍

pkg/resource/bucket/delete.go
```diff
@@ -83,46 +69,99 @@ func (d Deleter) Delete(ctx context.Context, entries []pointer.DeletePointer) er
...
 			continue
 		}
 
-		_, err := d.bucketSource.Delete(ctx, bucketName.BucketName)
+		logger := baseLogger
+		if bucketDeleteEntry.coordinate != nil {
+			logger = logger.With(log.CoordinateAttr(*bucketDeleteEntry.coordinate))
+		}
+
+		logger.DebugContext(ctx, "Deleting Grail buckets '%s'", bucketName)
```
d0weinberger: Should be singular `bucket`

pkg/resource/bucket/delete.go
```diff
@@ -83,46 +69,99 @@ func (d Deleter) Delete(ctx context.Context, entries []pointer.DeletePointer) er
...
+		bucketExists, err := buckets.AwaitActiveOrNotFound(ctx, d.bucketSource, bucketName, maxRetryDuration, durationBetweenRetries)
+
 		if err != nil {
-			if !api.IsNotFoundError(err) {
-				logger.ErrorContext(ctx, "Failed to delete Grail Bucket '%s': %v", bucketName.BucketName, err)
-				errCount++
-				continue
-			}
+			logger.With(log.ErrorAttr(err)).ErrorContext(ctx, "Failed to delete Grail buckets '%s': %v", bucketName, err)
```
d0weinberger: same here

pkg/resource/bucket/delete.go
```diff
@@ -83,46 +69,99 @@ func (d Deleter) Delete(ctx context.Context, entries []pointer.DeletePointer) er
...
+	errorCount := errorCountParse + errorCountDelete
+
+	if errorCount > 0 {
+		return fmt.Errorf("failed to delete %d Grail bucket configuration(s)", errorCount)
+	}
+
+	return nil
+}
+
+func deletePointerToBucketDelete(entries []pointer.DeletePointer) []bucketDelete {
```
d0weinberger: I was a bit confused when I first read the function name. Its purpose became clear to me then, but I'd suggest a different name. What do you think of `bucketDeletionsFromDeletePointers`?

pkg/resource/bucket/delete.go
```diff
@@ -83,46 +69,99 @@ func (d Deleter) Delete(ctx context.Context, entries []pointer.DeletePointer) er
...
+
+		bucketsDeletes[i] = bucketDelete{
+			bucketName: bucketName,
+			coordinate: &cord,
 		}
+	}
+	return bucketsDeletes
+}
 
+func responsesToBucketDelete(ctx context.Context, bucketResponses [][]byte, logger *log.Slogger) ([]bucketDelete, int) {
```
d0weinberger: I was a bit confused when I first read the function name. Its purpose became clear to me then, but I'd suggest a different name. What do you think of `bucketDeletionsFromListResponses`?

pkg/resource/bucket/delete.go
```diff
@@ -42,33 +44,17 @@ func NewDeleter(bucketSource DeleteSource) *Deleter {
 	return &Deleter{bucketSource: bucketSource}
 }
 
+type bucketDelete struct {
```
d0weinberger: I'm not sure `bucketDelete` is a good name. Maybe `bucketDeletion`? It's also not good, but I don't have a better idea currently. Maybe just `bucket`?

pkg/resource/bucket/delete.go
Kirdock: I'm also not sure about function names.

Image

@arthurpitman @tomazpu any ideas? 😅

Kirdock: After some discussion, we decided to go with `deleteItem` and convert/parse as a prefix

pkg/resource/bucket/delete.go
Kirdock: changed to `deleteItem`

PR #1972, opened 2025-07-03
refactor(delete): segments should adhere to `Deleter` interface

d0weinberger:
#### **Why** this PR?
The previously introduced `Deleter` interface (see #1964 ) is now also used with the deletion of segments.

#### How is it **tested**?

It doesn't

**Issue:**  CA-15697


pkg/delete/delete.go
```diff
@@ -28,11 +28,11 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/internal/automation"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/internal/classic"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/internal/document"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/internal/segment"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/internal/setting"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/internal/slo"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/pointer"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/resource/bucket"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/resource/segment"
```
tomazpu, line 35: Should line 3 the copyright year be updated?

d0weinberger: I think that's just for new/moved files 🤔

PR #1973, opened 2025-07-07
refactor(slo): SLO deletion adheres to `Deleter` interface

d0weinberger:
#### **Why** this PR?
The previously introduced Deleter interface (see https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1964 ) is now also used with the deletion of SLOs.

#### How is it **tested**?

It doesn't

**Issue:** CA-15716


pkg/resource/slo/delete.go
```diff
@@ -29,15 +29,23 @@ import (
...
 )
 
-type client interface {
+type source interface {
 	List(ctx context.Context) (api.PagedListResponse, error)
 	Delete(ctx context.Context, id string) (api.Response, error)
 }
 
-func Delete(ctx context.Context, c client, dps []pointer.DeletePointer) error {
+type Deleter struct {
```
arthurpitman, line 37: Question: would there be value in including an assertion that this implements `delete.Deleter`?

```
var _ delete.Deleter = Deleter{}
```

pkg/resource/slo/delete_test.go
```diff
@@ -29,10 +29,25 @@ import (
 
 	libAPI "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/idutils"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/internal/slo"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/pointer"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/resource/slo"
 )
 
+type deleteStubClient struct {
+	called bool
```
Kirdock: nit: I know it was also called lke this before but `called` may be a bit misleading. `deleteCalled` would fit better here.

Kirdock (approved):

pkg/resource/slo/delete.go
d0weinberger: We also didn't do it for download or deploy. I don't think it's necessary. It wouldn't be a big deal, though. I would just first need to get rid of the cyclic import that would follow from it. If you're okay with that, I wouldn't add an assertion for now.

Kirdock (approved):

PR #1974, opened 2025-07-07
refactor(document): deletion should adhere to `Deleter` interface

d0weinberger:
#### **Why** this PR?
The previously introduced Deleter interface (see https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1964 ) is now also used with the deletion of documents.

#### How is it **tested**?

It doesn't.

**Issue:** CA-15723


pkg/resource/document/delete_test.go
```diff
@@ -0,0 +1,351 @@
...
+			delete: func(uid string) (libAPI.Response, error) {
+				if uid == given.OriginObjectId {
+					return libAPI.Response{}, nil
+				}
+				return libAPI.Response{}, errors.New("some unpredictable error")
+			},
+		}
+
+		err := document.NewDeleter(&c).Delete(t.Context(), []pointer.DeletePointer{given, {OriginObjectId: "bla"}, given}) // the pointer in the middle is to cause error behavior
+		assert.ErrorContains(t, err, "failed to delete 1 document objects(s)")
```
d0weinberger, line 256: I will need to check for the value of `called` here as well.

pkg/resource/document/delete_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+	libAPI "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients/documents"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/idutils"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/pointer"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/resource/document"
+)
+
+type deleteStubClient struct {
+	called bool
```
Kirdock: nit: `called` may be a bit misleading. `deleteCalled` would fit better here.

pkg/resource/document/delete_test.go
```diff
@@ -0,0 +1,360 @@
...
+
+		err := document.NewDeleter(&c).Delete(t.Context(), []pointer.DeletePointer{given, {OriginObjectId: "bla"}, given}) // the pointer in the middle is to cause error behavior
+		assert.ErrorContains(t, err, "failed to delete 1 document objects(s)")
+		assert.True(t, c.called, "delete command wasn't invoked")
+	})
+}
+
+func TestDeleteAll(t *testing.T) {
+
+	t.Run("filter string is as expected", func(t *testing.T) {
```
Kirdock, line 263: nit: Isn't this more about "no delete called" if there aren't any documents?

Kirdock (dismissed):

pkg/resource/document/delete_test.go
d0weinberger: Hm, I mean it's that too but I specifically wanted to have a test for the filter string as well. Maybe I duplicate it with different asserts.

EDIT: done

PR #1975, opened 2025-07-07
refactor(settings): deletion should adhere to `Deleter` interface

d0weinberger:
#### **Why** this PR?
The previously introduced Deleter interface (see https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1964 ) is now also used with the deletion of settings.

#### How is it **tested**?

It doesn't.

**Issue:** CA-15724


pkg/resource/settings/delete_test.go
```diff
@@ -0,0 +1,488 @@
...
+	t.Run("deletion continues even if error occurs during delete", func(t *testing.T) {
+		c := deleteStubClient{
+			listSchemas: func() (dtclient.SchemaList, error) {
+				return []dtclient.SchemaItem{
+					{SchemaId: "builtin:some-schema"},
+					{SchemaId: "builtin:some-other-schema"},
+				}, nil
+			},
+			list: func(schema string, options dtclient.ListSettingsOptions) ([]dtclient.DownloadSettingsObject, error) {
+				if schema == "builtin:some-schema" {
```
Kirdock: Just an idea. Doesn't have to be implemented at all. You could actually get rid of all the `if` conditions if you define a map beforehand.

```Go
responses := map[string]struct{res []dtclient.DownloadSettingsObject; err error}{
		"builtin:some-schema": {
				[]dtclient.DownloadSettingsObject{
					{
						SchemaId: "builtin:some-schema",
						ObjectId: "some-object-1",
					},
					{
						SchemaId: "builtin:some-schema",
						ObjectId: "some-object-2",
					},
				}, nil,
		},
	}



//...
return responses[schema].res, responses[schema].err
```

Kirdock (approved):

pkg/resource/settings/delete_test.go
d0weinberger: Cool idea, I like it! Changed it

pkg/resource/settings/delete_test.go
```diff
@@ -0,0 +1,482 @@
...
+	t.Run("deletion continues even if error occurs during delete", func(t *testing.T) {
+		c := deleteStubClient{
+			listSchemas: func() (dtclient.SchemaList, error) {
+				return []dtclient.SchemaItem{
+					{SchemaId: "builtin:some-schema"},
+					{SchemaId: "builtin:some-other-schema"},
+				}, nil
+			},
+			list: func(schema string, options dtclient.ListSettingsOptions) ([]dtclient.DownloadSettingsObject, error) {
+				responses := map[string][]dtclient.DownloadSettingsObject{
```
Kirdock: We could move this outside this test run, because the one above uses the same objects, and therefore it can be reused, and also because the responses don't need to be re-created on every list call

PR #1977, opened 2025-07-09
refactor(delete): classic delete should adhere to `Deleter` interface

d0weinberger:
#### **Why** this PR?
The previously introduced `Deleter` interface (see https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1964 ) is now also used with the deletion of classic configs. Additionally, a `Purger` interface is introduced for the `DeleteAll`.

#### **What** has changed?
For this, I needed to split the `Deleter` interface. The `DeleteAll` comes now from another interface, named `Purger`. For the classic configs, there are two components: A `Deleter` component in `delete.go` and a `Purger` component in newly created `purge.go`. The reason is that the `DeleteAll` function had a different signature and behavior for classic configs compared to the other resource types. Instead of deleting ALL configs, you could pass a set of specific APIs which were to be purged. Configs of APIs not included there were not touched. It made no sense to change the signature of `DeleteAll` in the `Deleter` interface just because of that. Still, a solution had to be found for allowing selective purging of classic APIs without changing the `DeleteAll` function signature.

#### **How** does it do it?
The solution I chose for handling this is the introduction of a new `Purger` interface, providing `DeleteAll`. It would still not make any sense to include the `api.APIs` argument in the signature of the function. I chose to add it to the `classic.Purger` component instead.

#### How is it **tested**?
There were no previous tests for the delete and the purge. I added tests there.

It doesn't.

**Issue:** CA-15758


pkg/resource/classic/delete.go
```diff
@@ -20,18 +20,29 @@ import (
...
 
 	coreapi "github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
-	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client/dtclient"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/pointer"
 )
 
+type DeleteSource interface {
```
arthurpitman, line 31: I notice `purgeSource` is not exported; does this need to be exported?

d0weinberger: I will export `purgeSource`. I was a bit inconsistent across the PRs with naming and with what is exported and what isn't. I will create a followup-PR afterwards to align this across all `Deleter`s

pkg/resource/classic/purge_test.go
```diff
@@ -0,0 +1,172 @@
...
+				assert.Contains(t, []string{"some-id-1", "some-id-2", "some-id-3", "some-id-4"}, id)
+				assert.Contains(t, []string{api.Autotag, api.AppDetectionRule}, theApi.ID)
+				return nil
+			},
+		}
+
+		err := classic.NewPurger(&c, api.NewAPIs()).DeleteAll(t.Context())
+		assert.NoError(t, err)
+		assert.Equal(t, 4, c.countDeleteCalled,
+			fmt.Sprintf("expected delete to be called 4 times but was called %d times", c.countDeleteCalled))
```
Kirdock, line 78: FYI, `fmt.Sprintf` is optional here, as `assert` supports it out of the box
```go
assert.Equal(t, 4, c.countDeleteCalled, "expected delete to be called 4 times but was called %d times", c.countDeleteCalled)
```

Kirdock (approved):

pkg/resource/classic/purge_test.go
d0weinberger: Thanks for the hint! Will handle that in a follow up if that's okay

tomazpu (approved):

PR #1981, opened 2025-07-10
fix: Parsing account UUID skips resolving environment variable based on context

arthurpitman:
#### **Why** this PR?
In general the manifest loader support skipping the resolution of environment variables, this was not implemented for account UUIDs.

#### **What** has changed?
With this change, account UUIDs using environment variables will not be resolved if that has been disabled via the manifest loader context, making this consistent with the handling for URLs and credentials.

#### **How** does it do it?
A slight refactoring is performed to only check if UUIDs with an actual value are valid UUIDs. Then the loader context is checked prior to resolving the value.

#### How is it **tested**?
Test cases are added to `TestEnvVarResolutionCanBeDeactivated`.

Currently very little effect on users, as this commit is about code consistency.

Technically, if generating a delete file using a manifest that contains accounts where the UUID was specified via an environment variable which wasn't set, this would have incorrectly resulted in an error. With this change, no error is produced.

**Issue:** CA-15587

### DIFF
--- a/pkg/manifest/loader/account.go
+++ b/pkg/manifest/loader/account.go
@@ -19,10 +19,13 @@ package loader
 import (
 	"errors"
 	"fmt"
+	"os"
+
+	"github.com/google/uuid"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest/internal/persistence"
-	"github.com/google/uuid"
-	"os"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


Kirdock (approved):

pkg/manifest/loader/account.go
```diff
@@ -74,41 +77,41 @@ func parseSingleAccount(c *Context, a persistence.Account) (manifest.Account, er
...
-}
+	if u.Type != persistence.TypeEnvironment {
+		return uuid.UUID{}, fmt.Errorf("unexpected type: %q (expected one of %q, %q)", u.Type, persistence.TypeValue, persistence.TypeEnvironment)
+	}
 
-func loadAccountUUID(u persistence.TypedValue) (string, error) {
-	if u.Value == "" {
-		return "", errAccUidMissing
+	if c.Opts.DoNotResolveEnvVars {
+		log.Debug("Skipped resolving environment variable %s based on loader options", u.Value)
```
d0weinberger, line 94: Is this an options user can/would set? Should we then log it as info, maybe?

d0weinberger (approved):

pkg/manifest/loader/account.go
arthurpitman: No, this was something set for particular commands, namely generating delete files and dependency graphs. For this reason, this log is really only for the developer.

arthurpitman: So I think using `debug` level is OK.

PR #1982, opened 2025-07-10
fix: only parse account or environments envs depending on the usage

Kirdock:
#### **Why** this PR?
During `monaco account deploy`, the envs of the environments should not be validated, and for `monaco deploy`, the envs of the accounts should not be validated

#### **What** has changed?
- Env vars of environmentGroups are not validated when using account commands
- Env vars of account are not validated when not using account commands

#### **How** does it do it?
By excluding the parsing of env vars during manifest load.

#### How is it **tested**?
Existing tests cover it and new ones were added that check if invalid envs can and can't be set.

They can now use account commands without the need of providing the envs for non-account related commands and vice versa.

**Issue:** CA-15587


cmd/monaco/completion/completion_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/cmd/monaco/completion"
+)
+
+func TestEnvironmentByArg0(t *testing.T) {
+	manifestPath := "testdata/manifest.yaml"
+	envs, _ := completion.EnvironmentByArg0(nil, []string{manifestPath}, "")
+	assert.ElementsMatch(t, envs, []string{"env1", "env2"})
+}
+
+func TestEnvironmentByManifestFlag(t *testing.T) {
```
arthurpitman: The test is called `TestEnvironmentByManifestFlag` but it calls `AccountsByManifestFlag`?

pkg/manifest/loader/manifest_loader_test.go
```diff
@@ -117,6 +117,8 @@ environmentGroups:
               name: client-secret`
 )
 
+var environmentManifestOptions = Options{RequireEnvironmentGroups: true}
```
arthurpitman: Considering this is just a single line, I don't see a strong benefit of extracting this

Kirdock: I just saw this one as a default config, that each test can use. I can revert it though 👍

cmd/monaco/completion/completion_test.go
Kirdock: name fixed 👍

test/commands/testdata/env_requirements/manifest.yaml
Laubi, line 1: Super duper minor: The indentations are off within the file. We don't have a rule what we want to use for lists, but mixing it within a file is probably something we don't wanna do :D

test/commands/testdata/env_requirements/manifest.yaml
```diff
@@ -0,0 +1,47 @@
+manifestVersion: 1.0
+projects:
+- name: project
+environmentGroups:
+- name: default
+  environments:
+  - name: classic_env
+    url:
+      type: environment
+      value: E_SECRET
```
Laubi: I guess that means "environment-secret", so maybe we should just call it that ;)

Laubi (dismissed):

Laubi (approved):

PR #1985, opened 2025-07-11
refactor(automation): delete should adhere to `Deleter` interface

d0weinberger:
#### **Why** this PR?
The previously introduced Deleter interface (see https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1964 ) is now also used with the deletion of classic configs.

#### **What** has changed and how is it done?
Minor change to the `Delete` so it follows the interface. Before, the specific automation type was passed from the outside. This was only done for logging purposes, as the `DeletePointer` objects themselves carried the necessary type information. Since we only ever call `Delete` with `DeletePointers` of one specific type, the type of the first object is used for logging. This is analoguous to how it works for the `settings.Delete`. 

#### How is it **tested**?
There were no tests previously. I added tests.

It doesn't.

**Issue:** CA-15696


pkg/resource/automation/delete_test.go
```diff
@@ -0,0 +1,305 @@
...
+	coreautomation "github.com/dynatrace/dynatrace-configuration-as-code-core/clients/automation"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/idutils"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/delete/pointer"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/resource/automation"
+)
+
+type deleteStubClient struct {
+	called bool
```
Kirdock: Should we also rename it here to `deleteCalled`, to make it clearer to which call it relates?

Kirdock (dismissed):

pkg/resource/automation/delete_test.go
```diff
@@ -0,0 +1,305 @@
...
+	deleteCalled bool
+	delete       func(id string) (libAPI.Response, error)
+	list         func(resourceType coreautomation.ResourceType) (libAPI.PagedListResponse, error)
+}
+
+func (s *deleteStubClient) List(_ context.Context, resourceType coreautomation.ResourceType) (libAPI.PagedListResponse, error) {
+	return s.list(resourceType)
+}
+
+func (s *deleteStubClient) Delete(_ context.Context, _ coreautomation.ResourceType, id string) (libAPI.Response, error) {
```
arthurpitman: Does it conceptually make sense that we discard the `coreautomation.ResourceType`  parameter? I would argue that we should assert on that too.

d0weinberger: Okay, I will change the tests so we assert on that too.

Kirdock (approved):

PR #1987, opened 2025-07-14
fix(account): only parse selected accounts

Kirdock:
#### **Why** this PR?
When several accounts were defined, the non-selected ones were not excluded when validating the envs. Now only the envs of the selected account are validated

#### **What** has changed?
The env var resolution is skipped for non-selected accounts

#### How is it **tested**?
New tests are added that parse multiple accounts with and without setting the needed env names

They can now deploy account resources without setting the envs for other accounts

**Issue:** CA-15654


pkg/manifest/loader/account.go
```diff
@@ -117,9 +117,13 @@ func parseUUID(value string) (uuid.UUID, error) {
 // parseAccounts converts the persistence definition to the in-memory definition
 func parseAccounts(c *Context, accounts []persistence.Account) (map[string]manifest.Account, error) {
 
-	result := make(map[string]manifest.Account, len(accounts))
+	result := make(map[string]manifest.Account)
```
tomazpu: I would probably still reserve space in memory as the use case with skipping accounts is not that common?

pkg/manifest/loader/manifest_loader.go
```diff
@@ -53,6 +53,12 @@ type Context struct {
 	// If Environments contains items that do not match any environment in the specified manifest file, the loading errors.
 	Environments []string
 
+	// Account is a filter to what account should be loaded.
+	// If it's empty, all accounts are loaded.
+	//
+	// If Account contains an account that does not match any account in the specified manifest file, the loading errors.
+	Account string
```
tomazpu, line 60: If I understand this correctly, I can't skip multiple accounts. But what happens if I assign' acc1, acc2`?

Kirdock: At the moment, we can only define one single account, as the flag is just for one. I didn't want to change the implementation here, so I kept it as a single string.

pkg/manifest/loader/account.go
Kirdock: Would something speak against this one?
```go
length := len(accounts)
if c.Account != "" {
  length = 1
}
result := make(map[string]manifest.Account, length)
```

else I will go you with your suggestion.

Committed the length check solution for now.

pkg/manifest/loader/account_test.go
```diff
@@ -284,6 +290,43 @@ func TestInvalidAccounts(t *testing.T) {
...
+
+	t.Run("Returns all accounts", func(t *testing.T) {
+		t.Setenv("SECRET_2", "secret")
+		parsedAccounts, err := parseAccounts(&Context{}, accounts)
+
+		assert.NoError(t, err)
+		require.Len(t, parsedAccounts, 2)
+	})
+
+	t.Run("Returns an error if not account matches", func(t *testing.T) {
```
arthurpitman: Typo:
```suggestion
	t.Run("Returns an error if no account matches", func(t *testing.T) {
```

arthurpitman (approved):

tomazpu (dismissed):

PR #1989, opened 2025-07-17
refactor(delete): All `Purgers` and `Deleters` are initialized before use

d0weinberger:
#### **Why** this PR?
Visibility and naming of the `DeleteSource` should be the same across all resources. Differences only lead to confusion here. 

Also, since the resource `Purgers` now follow the `Purger` interface, we can initialize these `Purgers` first before iterating over the slice of all `Purgers` and calling `DeleteAll`.

The same is done with `Deleters`.

#### **What** has changed?
Mostly, some unexported interfaces, named `client`  have been renamed to `DeleteSource`. Also, `Purger` structs are initialized first before `DeleteAll` is called. For the `delete.go`, we initialize a `Deleters` struct which consists of two important components:
1. a map which gives a `Deleter` for a config type, called `deleterForType`,
2. a "catch-all" `unknownTypeDeleter` which is used if the config type is not known. For us, this is the `settings.Deleter`. 

This `Deleters` struct is built given a `ClientSet`. For example, for each API present in `api.NewAPIs()`, a map entry is added. If the `ClassicClient` is not nil, we can create a `classic.Deleter` to which each of the API map entries map. If it is nil, the map entries are _still_ created, but they map to `nil`.
If `SettingsClient` is not nil, the `unknownTypeDeleter` is a `settings.Deleter`. Otherwise it is nil.

When the `Deleters.Delete` function is called, we check whether there is an entry in the map for the given type. If there is, we retrieve the entry. If it's nil, then we can't delete. If there is no entry, we check the `unknownTypeDeleter`. If it's nil, we can't delete.

#### How is it **tested**?

Existing tests still work

It doesn't

**Issue:**
CA-15292

### DIFF
--- a/pkg/resource/settings/delete.go
+++ b/pkg/resource/settings/delete.go
@@ -48,7 +48,7 @@ func (d Deleter) Delete(ctx context.Context, entries []pointer.DeletePointer) er
 	schema := entries[0].Type
 
 	logger := log.With(log.TypeAttr(schema))
-	logger.InfoContext(ctx, "Deleting %d settings objects(s) of schema %q...", len(entries), schema)
+	logger.InfoContext(ctx, "Deleting %d settings object(s) of schema %q...", len(entries), schema)
 
 	deleteErrs := 0
 	for _, e := range entries {
@@ -93,7 +93,7 @@ func (d Deleter) Delete(ctx context.Context, entries []pointer.DeletePointer) er
 	}
 
 	if deleteErrs > 0 {
-		return fmt.Errorf("failed to delete %d settings objects(s) of schema %q", deleteErrs, schema)
+		return fmt.Errorf("failed to delete %d settings object(s) of schema %q", deleteErrs, schema)
 	}
 
 	return nil
--- a/pkg/resource/slo/delete.go
+++ b/pkg/resource/slo/delete.go
@@ -52,7 +52,7 @@ func (d Deleter) Delete(ctx context.Context, dps []pointer.DeletePointer) error 
 		}
 	}
 	if errCount > 0 {
-		return fmt.Errorf("failed to delete %d %s objects(s)", errCount, config.ServiceLevelObjectiveID)
+		return fmt.Errorf("failed to delete %d %s object(s)", errCount, config.ServiceLevelObjectiveID)
 	}
 	return nil
 }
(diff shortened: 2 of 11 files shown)


Kirdock (dismissed):

pkg/delete/delete.go
```diff
@@ -92,37 +90,81 @@ func deleteAutomationConfigs(ctx context.Context, autClient client.AutomationCli
...
+
+	var automationDeleter Deleter = nil
+	if clients.AutClient != nil {
+		automationDeleter = automation.NewDeleter(clients.AutClient)
+	}
+	for _, configType := range []config.AutomationResource{config.Workflow, config.SchedulingRule, config.BusinessCalendar} {
+		deleterForConfigType[string(configType)] = automationDeleter
+	}
+
+	if clients.BucketClient != nil {
```
tomazpu, line 131: I would suggest: 
```
deleterForConfigType[string(config.BucketTypeID)] = nil
if clients.BucketClient != nil {
   deleterForConfigType[string(config.BucketTypeID)] = bucket.NewDeleter(clients.BucketClient)
}

```

This way, you remove lots of else statements.

d0weinberger: agreed, looks much better

pkg/resource/slo/delete.go
```diff
@@ -134,7 +134,13 @@ func (d Deleter) DeleteAll(ctx context.Context) error {
 			errs = append(errs, err)
 		}
 	}
-	return errors.Join(errs...)
+
+	retErr := errors.Join(errs...)
+	if retErr != nil {
+		log.ErrorContext(ctx, "Failed to delete all %s configurations: %v", config.ServiceLevelObjective{}, retErr)
```
Kirdock, line 140: At the moment, the error log is a bit inconsistent across all resources. Sometimes the resource type starts with an uppercase letter and sometimes it's the string of the config type (lowercase)
Could you align them?

pkg/resource/settings/delete.go
```diff
@@ -162,7 +162,9 @@ func (d Deleter) DeleteAll(ctx context.Context) error {
 	}
 
 	if errCount > 0 {
-		return fmt.Errorf("failed to delete %d setting(s)", errCount)
+		returnedError := fmt.Errorf("failed to delete %d setting(s)", errCount)
+		log.ErrorContext(ctx, "Failed to delete all Settings 2.0 objects: %v", returnedError)
```
Kirdock, line 166: Should we also align this one with the other messages?

d0weinberger: We need another PR anyways for introducing slog to the Deleters. Would it be fine if we consolidate the logging topics in that followup-PR?

Kirdock (dismissed):

Kirdock (approved):

PR #1991, opened 2025-07-21
feat(classic): support aws credential services endpoint

Kirdock:
#### **Why** this PR?
Users aren't able to modify or download AWS credential service configurations; therefore, we're adding support for it.

#### **What** has changed?
The new endpoint `/api/config/v1/aws/credentials/{SCOPE}/services` was added to our collection.

#### How is it **tested**?
The existing test was extended to also update this resource.

Users can now update and download credentials service resources.

**Issue:**


d0weinberger (dismissed):
LGTM, but about the removal of `metadata` from the response, I can't comment though. I don't know what that could/would cause.

pkg/api/endpoints.go
```diff
@@ -109,6 +110,12 @@ func NewAPIs() APIs {
 			PropertyNameOfGetAllResponse: "dashboards",
 			NonUniqueName:                true,
 		}
+		var awsCredentialsAPI = API{
+			ID:                           AwsCredentials,
+			URLPath:                      "/api/config/v1/aws/credentials",
+			PropertyNameOfGetAllResponse: StandardApiPropertyNameOfGetAllResponse,
+			SkipDownload:                 true,
```
d0weinberger, line 117: Why is download skipped here, actually, and if we skip it for the AwsCredentials, should we also skip it for the AwsCredentialsServices?

pkg/api/endpoints.go
```diff
@@ -241,11 +248,18 @@ func NewAPIs() APIs {
...
-				SkipDownload:                 true,
+				ID:                  AwsCredentialsServices,
+				URLPath:             "/api/config/v1/aws/credentials/{SCOPE}/services",
+				Parent:              &awsCredentialsAPI,
+				SingleConfiguration: true,
+				NonDeletable:        true,
+				// Metadata doesn't seem that it can be updated.
+				// Updating the cluster version or configurationVersions doesn't change anything
+				TweakResponseFunc: func(m map[string]any) {
+					delete(m, "metadata")
```
d0weinberger, line 263: Why is it necessary to remove `metadata` from the response?

pkg/api/endpoints.go
Kirdock: We usually ignore these configs because they're simply not re-deployable, due to the missing secrets that are not present when downloading. On the other side, it would help for the reference to the service settings 🤔 
@Laubi should we enable the download for this one, so that the download for the service settings is correctly referencing the credential?

pkg/api/endpoints.go
Kirdock: If we download it, the config should be re-deployable. The `metadata` is discarded and not even supported by the endpoint, that's why I removed it.
Endpoint schema:
image

So, the metadata is not present on root level

d0weinberger: Let's skip the download.

pkg/api/endpoints.go
Kirdock: We decided to skip the download 👍

tomazpu (approved):

pkg/delete/delete.go
```diff
@@ -70,6 +66,21 @@ func Configs(ctx context.Context, clients client.ClientSet, entriesToDelete Dele
 	return nil
 }
 
+func removeNonDeletableClassicAPIs(ctx context.Context, remainingEntriesToDelete DeleteEntries) {
```
tomazpu, line 69: Should we add a unit test for these cases?

Kirdock: We already have a test that checks if the dashboard share setting is skipped. For the `Parent == nil` case, we actually don't have anything like this defined in our `endpoints.go`.

PR #1992, opened 2025-07-22
feat(settings): support admin mode for permissions

Kirdock:
#### **Why** this PR?
The core lib added some functionality for setting the `adminAccess` property for permissions. As this is breaking our current implementation, we need to update the logic to handle the `adminAccess` correctly.

#### **What** has changed?
- Updated the core lib
- Added a retry mechanism if `adminAccess` doesn't work

#### How is it **tested**?
New tests added that check if the property is correctly set and the retry correctly works.

No change for the users, as the settings client doesn't support the `adminAccess` so far.

**Issue:** CA-15757

## Open for discussion
Should I move the `ExecuteWithAdminAccess` utility function to core?


pkg/client/dtclient/settings_client.go
```diff
@@ -907,11 +909,14 @@ func (d *SettingsClient) UpsertPermission(ctx context.Context, objectID string,
...
-	_, err = d.permissionClient.GetAllUsersAccessor(ctx, objectID)
+	// admin access is validated after it's validated that the object exists.
+	_, err, adminAccess := ExecuteWithAdminAccess(func(adminAccess bool) (coreapi.Response, error) {
+		return d.permissionClient.GetAllUsersAccessor(ctx, objectID, adminAccess)
+	})
 
 	// When there is no error the object is found and we update it.
 	if err == nil {
-		_, err = d.permissionClient.UpdateAllUsersAccessor(ctx, objectID, payload)
+		_, err = d.permissionClient.UpdateAllUsersAccessor(ctx, objectID, adminAccess, payload)
```
d0weinberger, line 919: Don't you need to wrap this with `ExecuteWithAdminAccess` as well?

pkg/client/dtclient/admin_retry.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// ExecuteWithAdminAccess calls a given function with adminAccess and retries it without adminAccess in case there is a 403.
+//
+// returns:
+//   - The response of the API call with adminAccess enabled if the permission is given
+//     or the response of the API call without adminAccess if the permission is not given.
+//   - Any occurring error, not related to the permission error of the adminAccess enabled call.
+//   - The used adminAccess for the returned response.
+func ExecuteWithAdminAccess(call func(adminAccess bool) (api.Response, error)) (api.Response, error, bool) {
```
d0weinberger: Hm, I'm not a fan of the name. From the name I would not expect a retry without admin access. I would only assume "Okay, whatever function I pass here will be executed with admin access = true".

I don't have a good suggestion currently, however.

pkg/client/dtclient/settings_client.go
Kirdock: No. With the GET call beforehand we already get the information if the `adminMode` is `true` or `false`

pkg/client/dtclient/admin_retry.go
Kirdock: In core we name it even `makeRequestWithAdminAccess` 🤔

Laubi: The 'go' way is to call something like those parameters `fn`, like `requestFn`

Laubi: I mean it does not request anything with admin access, it provides a flag that is toggled 🤔 

`tryWithAdminFlag`?

Laubi: As soon as there are 3 return values, return value names are expected.

Kirdock: Or something like this?
- `ExecuteWithAdminFallback`
- `ExecuteWithPermissionRetry`
- `ExecuteWithAccessToggle`

Kirdock: Called it now `doWithAdminAccessRetry`

d0weinberger (approved):

PR #1994, opened 2025-07-23
chore: Remove reviewers from dependabot config

arthurpitman:
#### **Why** this PR?
`reviewers` is no longer supported in `.github/dependabot.yml` as `CODEOWNERS` is used instead as described in https://github.blog/changelog/2025-04-29-dependabot-reviewers-configuration-option-being-replaced-by-code-owners/

Removes `reviewers` from `.github/dependabot.yml`

#### How is it **tested**?
Not required

No effect

**Issue:** NOISSUE


Kirdock (commented):
There is another reviewers field for github actions. This one should be removed too

arthurpitman:
> There is another reviewers field for github actions. This one should be removed too

Fixed, thanks 😄

PR #1996, opened 2025-07-25
feat: Use `slog` for logging in delete and related code paths

arthurpitman:

**Issue:**

### DIFF
--- a/cmd/monaco/delete/delete.go
+++ b/cmd/monaco/delete/delete.go
@@ -17,6 +17,7 @@ package delete
 import (
 	"context"
 	"fmt"
+	"log/slog"
 	"strings"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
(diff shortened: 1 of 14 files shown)


cmd/monaco/delete/delete.go
```diff
@@ -40,18 +41,18 @@ func Delete(ctx context.Context, environments manifest.EnvironmentDefinitionsByN
...
+			slog.WarnContext(ctx, "Delete file contains entries referencing platform configurations, but environment is missing platform credentials. These configurations won't be deleted.")
 		}
 
 		clientSet, err := client.CreateClientSet(ctx, env.URL.Value, env.Auth)
 		if err != nil {
 			return fmt.Errorf("failed to create API client for environment %q due to the following error: %w", env.Name, err)
 		}
 
-		log.InfoContext(ctx, "Deleting configs for environment %q...", env.Name)
+		slog.InfoContext(ctx, "Deleting configs...")
```
tomazpu, line 52: I noticed the environment name was deleted. Is there no value in knowing which environment we are deleting? wdyt?

cmd/monaco/purge/purge.go
```diff
@@ -80,10 +81,10 @@ func purgeForEnvironment(ctx context.Context, env manifest.EnvironmentDefinition
 		return fmt.Errorf("failed to create a client for env `%s`: %w", env.Name, err)
 	}
 
-	log.InfoContext(ctx, "Deleting configs for environment `%s`", env.Name)
+	slog.InfoContext(ctx, "Purging configs...")
```
tomazpu, line 84: same here?

cmd/monaco/delete/delete.go
arthurpitman: Actually, it is already added via [line 42](https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1996/files/611666fa8169dd572a585ee0a01a0260ff86c3f3#diff-865abfef2995909e444fbd2c57d7ab6136fee53d35d58f81b4d200c58b832d02R42). Removing it here stops it being added twice

cmd/monaco/purge/purge.go
arthurpitman: Same magic from line 77.

PR #1997, opened 2025-07-28
fix(delete): Warn for non-deletable configs

arthurpitman:
#### **Why** this PR?
Currently, a warning is emitted for non-deletable configs regardless of whether they occur in the delete file or not.

This PR adds a check to see if a configuration type is actually used before emitting the warning.

#### How is it **tested**?
A test is added that nothing is logged for an empty delete file

Yes, users will now only see a warning if the delete file actually contains non-deletable config types.

**Issue:** NO-ISSUE


pkg/delete/delete.go
```diff
@@ -72,10 +73,14 @@ func removeNonDeletableClassicAPIs(ctx context.Context, remainingEntriesToDelete
 			continue
 		}
 
+		if _, found := remainingEntriesToDelete[classicApi.ID]; !found {
```
d0weinberger, line 76: Why don't we iterate over the keys of `remainingEntriesToDelete` instead of iterating over all APIs?

arthurpitman: We'd still have to check if they're a classic API, so I guess, considering there are probably many settings schemas, the current approach is more efficient?

PR #1998, opened 2025-07-31
feat(delete): Use slog for logging in resource deletion

arthurpitman:
#### **Why** this PR?
As a follow-up to #1996, this PR switches to using `slog` directly for logging within the delete functionality of resources.

Now `slog` is used directly for logging within the delete functionality of resources.
In addition, the content of log messages is improved, especially regarding how an object is selected for deletion: by originObjectId or by coordinate.

#### How is it **tested**?
Existing tests provide coverage.

No real effect other than some changes in the log messages. Functionality remains the same.

**Issue:** CA-15897

### DIFF
--- a/pkg/resource/automation/delete.go
+++ b/pkg/resource/automation/delete.go
@@ -19,6 +19,7 @@ package automation
 import (
 	"context"
 	"fmt"
+	"log/slog"
 
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/api"
 	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients/automation"
@@ -48,8 +49,7 @@ func (d Deleter) Delete(ctx context.Context, entries []pointer.DeletePointer) er
 	}
 	automationResource := entries[0].Type
 
-	logger := log.With(log.TypeAttr(automationResource))
-	logger.InfoContext(ctx, "Deleting %d config(s) of type %q...", len(entries), automationResource)
+	slog.InfoContext(ctx, "Deleting configs", log.TypeAttr(automationResource), slog.Int("count", len(entries)))
 
 	deleteErrs := 0
 	for _, e := range entries {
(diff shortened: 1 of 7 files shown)


pkg/resource/automation/delete.go
```diff
@@ -63,28 +63,28 @@ func (d Deleter) Delete(ctx context.Context, entries []pointer.DeletePointer) er
...
+	var logger *slog.Logger
 	id := dp.OriginObjectId
 	if id == "" {
 		id = idutils.GenerateUUIDFromCoordinate(dp.AsCoordinate())
+		logger = slog.With(log.CoordinateAttr(dp.AsCoordinate()), slog.String("id", id))
+	} else {
+		logger = slog.With(log.TypeAttr(dp.Type), slog.String("id", id))
 	}
 
-	logger.DebugContext(ctx, "Deleting %v with id %q.", dp.Type, id)
```
Kirdock, line 73: Was this intentionally removed?

pkg/resource/document/delete.go
```diff
@@ -58,32 +58,34 @@ func (d Deleter) Delete(ctx context.Context, dps []pointer.DeletePointer) error
...
-		return nil
+		if id == "" {
+			logger.DebugContext(ctx, "No document found with external ID", slog.String("externalId", extID))
+			return nil
+		}
 	}
 
 	_, err := d.source.Delete(ctx, id)
 	if err != nil && !api.IsNotFoundError(err) {
+		logger.ErrorContext(ctx, "Failed to delete document", slog.String("id", id), log.ErrorAttr(err))
```
Kirdock: Would this ID attribute simply override the empty ID set before?

arthurpitman: Actually no, but to prevent potential duplicates I will move adding this id.

pkg/resource/segment/delete.go
```diff
@@ -58,39 +58,45 @@ func (d Deleter) Delete(ctx context.Context, dps []pointer.DeletePointer) error
...
 
-	if id == "" {
-		logger.DebugContext(ctx, "no action needed")
-		return nil
+		if id == "" {
+			logger.DebugContext(ctx, "No segment found with external ID", slog.String("externalId", extID))
+			return nil
+		}
+
+		logger = logger.With(slog.String("id", id))
```
Kirdock, line 81: Will this still lead to two `id` attributes becausee of [line 61](https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1998/files#diff-2e94f31076ecd70aaf7c2be48c8dedd307dae30323e42b09a2b855e9254cef47R61)?

pkg/resource/automation/delete.go
arthurpitman: Yes, as we don't do it for other resources. I would argue we don't need it because the return code paths say what happened and that is more useful. Alternatively, I could add it to the others though.

pkg/resource/segment/delete.go
arthurpitman: No, it shouldn't due to [line 66](https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1998/files/68e5662345863c8fcd6a3cc5aaae446357905973#diff-2e94f31076ecd70aaf7c2be48c8dedd307dae30323e42b09a2b855e9254cef47R66).

Ultimately, it would be nice to refactor automation, document, segment and slo resources to make this clearer but that would exceed the scope of this PR.

Kirdock: ah nvm, this is not the initial instance, got it 👍

tomazpu (approved):

PR #2000, opened 2025-08-06
fix: Nested values now escape correctly.

Laubi:
#### **Why** this PR?
Given this code example:
```yaml
...
properties:
  property:
    type: value
    value:
      subproperty: |
        My
        string
```

When reading the above property, we read it as a `map[any]any` and convert it to a `map[string]any`. Later, we apply recursive escaping functions to the `any` value.
 However, when parsing subproperties, we get a `map[string]map[any]any`, and the recursive functions fail with the second `map[any]any`.

#### **What** has changed?

 This fix now parses the maps recursively during reading, and thus the later recursive functions work properly, and the multiline strings are escaped correctly.

#### **How** does it do it?
N/A
#### How is it **tested**?
Added an e2e test case, as well as unit tests 💪 

Users are not able to use multiline strings in nested properties.

**Issue:** CA-16115

### DIFF
--- a/internal/maps/maps_test.go
+++ b/internal/maps/maps_test.go
@@ -20,12 +20,14 @@ package maps
 
 import (
 	"fmt"
-	"github.com/google/go-cmp/cmp/cmpopts"
-	"reflect"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
 
 func TestToStringMap(t *testing.T) {
+	t.Parallel()
+
 	tests := []struct {
 		name  string
 		input map[interface{}]interface{}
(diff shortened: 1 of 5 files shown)


internal/maps/maps_test.go
```diff
@@ -68,22 +70,24 @@ func TestToStringMap(t *testing.T) {
 				}): 52,
 			},
 		},
+		{
+			"Applies the stringify recursively",
+			map[interface{}]interface{}{
```
Laubi, line 75: Note to reviewers: I kept `interface{}` here as all other tests also use it, and renaming it would introduce a lot of noise for nothing.

test/configuration/escaping/testdata/special-character-in-config/project/auto-tag/_config.yaml
Kirdock, line 1: Is there any reason for the `_` underscore in the file name?

Kirdock (approved):

test/configuration/escaping/testdata/special-character-in-config/project/auto-tag/_config.yaml
Laubi: So that it's always the first in the directory

PR #2002, opened 2025-08-07
fix(account): Increase delay after 429 errors for account management

arthurpitman:
#### **Why** this PR?

Dynatrace Account Management APIs currently enforce a rate limit of 200 requests per minute, however Monaco is unable to react to this. As a workaround, as the limit is reset after a minute, delaying further requests for 60 seconds will allow Monaco to continue using the API. 

#### **What** has changed?
First a minor refactoring is made to create account clients in a single function, so that they behave consistently. Secondly, the retry options are changed.

#### **How** does it do it?
Most importantly the retry options are changed to:
- only retry after a 429 error
- delay for 30 seconds before retrying: this gives a good balance between successfully retrying and providing feedback via log messages.

#### How is it **tested**?
Basic tests are added for successful account management client creation.

Account management deploy, download and delete operations shouldn't fail after receiving a 429 error.

**Issue:** CA-16097

### DIFF
--- a/cmd/monaco/account/delete.go
+++ b/cmd/monaco/account/delete.go
@@ -22,17 +22,14 @@ import (
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/maps"
-	"golang.org/x/oauth2/clientcredentials"
 
-	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients"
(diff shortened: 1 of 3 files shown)


d0weinberger (approved):

arthurpitman:
I have manually tested this for both download and deploy and can confirm that it allows the download or deployment to continue and complete successfully. A delay of both 60 seconds (a single retry suffices) or 10 seconds (with typically 6 or 7 retries) works.

Kirdock (approved):

PR #2006, opened 2025-08-19
backport: fix: reduce amount of segment GET calls during deploy

Kirdock:
Dev backport of https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/2004 to 2.23.x

#### **Why** this PR?
At the moment, there is a GET call for each available segment during deploy. This is not needed as we only need the UID and externalID for the deploy to work; the simple list call is sufficient. The problem is that GetAll may exceed the 1 minute deploy timeout.

#### **What** has changed?
Instead of fetching all information for segments, we're only fetching the needed one.
O(n) => O(1)

#### **How** does it do it?
By calling List instead of GetAll.

#### How is it **tested**?
Current tests are updated to test the list functionality.

They will less likely run into context deadline errors

**Issue:** CA-16213


Laubi:
closed as this PR was just to trigger the e2e tests

PR #2011, opened 2025-08-26
test: make account management tests tenant independent

Kirdock:
#### **Why** this PR?
Some environments were hardcoded and because of this it wasn't possible to execute the tests on other tenants

#### **What** has changed?
Environments and account URL are now taken from the env variables

#### **How** does it do it?
By replacing the environments with env variables in yaml files that can be replaced via `os.ExpandEnv`.
Reference parameters can't be used due to implementation limitations.

#### How is it **tested**?
Tests are working on different tenants and on the current ones.

Doesn't affect users

**Issue:** CA-14365


test/account/runner.go
```diff
@@ -64,16 +64,17 @@ func createAccountClientsFromManifest(t *testing.T, fs afero.Fs, manifestFileNam
 	return accClients
 }
 
-// randomizeYAMLResources loops over each *.yaml file, replaces %RAND% with a random string and returns the random string
+// randomizeAndResolveYAMLResources loops over each *.yaml file, replaces %RAND% with a random string, resolves environment variables set via ${MY_ENV} and returns the random string
```
Laubi, line 67: I do not like that we have multiple different methods of defining those testing replacements, please unify them

test/account/all_resources_integration_test.go
```diff
@@ -49,7 +51,10 @@ func TestDeployAndDelete_AllResources(t *testing.T) {
 		myLocalGroup := "My LOCAL Group%RAND%"
 		myPolicy := "My Policy%RAND%"
 		myPolicy2 := "My Policy 2%RAND%"
-		envVkb := "vkb66581"
+		env1, ok := os.LookupEnv("ENVIRONMENT_ID_2")
```
tomazpu: NIT: could we also name the variable env2, its a bit misleading that we read the ENV variable ID 2 into env 1 :)

PR #2015, opened 2025-09-01
feat: support skipping SSL certificate checks

Kirdock:
#### **Why** this PR?
For some environments, the SSL check doesn't work. To tackle this corner case, the environment variable `MONACO_SKIP_CERTIFICATE_VERIFICATION` is introduced. If this is set to `true`, the SSL certificate check is disabled.

#### **What** has changed?
A custom HTTP client without SSL cert check is set if the env `MONACO_SKIP_CERTIFICATE_VERIFICATION=true` is set.

#### **How** does it do it?
By adding a modified context at the root level with a custom HTTP client if the FF is set.

#### How is it **tested**?
New tests added. Manual tests to check if the oauth2 lib gets the right client/options.

Users can now skip SSL certificate checks by setting `MONACO_SKIP_CERTIFICATE_VERIFICATION=true`

**Issue:**

### DIFF
--- a/cmd/monaco/main.go
+++ b/cmd/monaco/main.go
@@ -26,6 +26,7 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/log"
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/version"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client"
 	monacoVersion "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/version"
 )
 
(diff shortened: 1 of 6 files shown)


d0weinberger (dismissed):

cmd/monaco/main.go
```diff
@@ -34,7 +35,7 @@ func main() {
 	// furthermore it should honor the desired format, such as JSON
 	// full logging is set up in PreRunE method of the root command, created with runner.BuildCli
 	// that is the earliest point calls to log will be also written into files and adhere to user controlled verbosity
-	ctx := context.Background()
+	ctx := client.SetCustomHTTPClientInContext(context.Background())
```
arthurpitman, line 38: I find this solution really elegant. The only thing is, would you miss knowing that this is done here when looking at where the clients are actually created? It's probably not a big issue, but it might be worth adding a comment in the clientset explaining how this works.

pkg/client/htp_client_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/oauth2"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/featureflags"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/client"
+)
+
+func TestSetClient(t *testing.T) {
```
arthurpitman: Should we rename this?

PR #2017, opened 2025-09-02
feat(boundaries): adds loading of boundaries

d0weinberger:
#### **Why** this PR?
This PR introduces loading of boundaries as account management resource.

#### **What** has changed?
Boundaries are loaded, parsed, validated, and transformed to `account.Boundary`. They are not used yet and can also not be referred to in the `groups` definition yet.

#### **How** does it do it and how is it **tested**?
The structure of a boundary with all fields set is as follows:

```
boundaries:
  - name: My Boundary
    id: my-boundary
    query: shared:app-id = "my-app"
    originObjectId: 0cd9c365-ed0e-4ef5-8440-46d02fceea3e
```

The code for loading boundaries is added in the same manner as for the other account management resources. The yaml test files and load_test.go` are extended .

It doesn't yet. A feature flag `MONACO_FEAT_ENABLE_BOUNDARIES` guards loading of boundaries.

**Issue:** CA-16338


pkg/account/persistence/loader/load.go
```diff
@@ -105,17 +113,18 @@ func Load(fs afero.Fs, rootPath string) (*account.Resources, error) {
...
 // HasAnyAccountKeyDefined checks whether the map has any AM key defined.
-// The current keys are `users`, `serviceUsers`, `groups`, and `policies`.
+// The current keys are `users`, `serviceUsers`, `groups`, `policies`, and `boundaries`.
 func HasAnyAccountKeyDefined(m map[string]any) bool {
 	if len(m) == 0 {
 		return false
 	}
 
-	return m[persistence.KeyUsers] != nil || m[persistence.KeyServiceUsers] != nil || m[persistence.KeyGroups] != nil || m[persistence.KeyPolicies] != nil
+	return m[persistence.KeyUsers] != nil || m[persistence.KeyServiceUsers] != nil || m[persistence.KeyGroups] != nil || m[persistence.KeyPolicies] != nil || m[persistence.KeyBoundaries] != nil
```
Kirdock: [nit]
```suggestion
	return m[persistence.KeyUsers] != nil || m[persistence.KeyServiceUsers] != nil || m[persistence.KeyGroups] != nil || m[persistence.KeyPolicies] != nil || m[persistence.KeyBoundaries] != nil && featureflags.Boundaries.Enabled() 
```

pkg/account/persistence/loader/load_test.go
```diff
@@ -245,6 +283,12 @@ func TestLoad(t *testing.T) {
 		assert.Error(t, err)
 	})
 
+	t.Run("Duplicate boundary produces error", func(t *testing.T) {
+		t.Setenv(featureflags.Boundaries.EnvName(), "true")
+		_, err := Load(afero.NewOsFs(), "testdata/duplicate-boundary.yaml")
+		assert.Error(t, err)
+	})
```
Kirdock, line 294: Isn't this already tested with `TestLoadResources_Duplicates`?

d0weinberger: Yes, it actually is, the same goes for the other "Duplicate <type> produces error" tests. I followed the pattern here, but can also remove the test again if you want.

pkg/account/persistence/loader/load.go
d0weinberger: True, will add this

pkg/account/persistence/internal/types/types.go
```diff
@@ -160,9 +224,65 @@ func (ReferenceSlice) JSONSchema() *jsonschema.Schema {
 	}
 }
 
+type AccountSchema Account
+
+// JSONSchema defines a custom schema definition for Group. If the Boundaries feature is not enabled,
```
arthurpitman: Should this be for Account rather than Group?

pkg/account/persistence/internal/types/types.go
```diff
@@ -160,9 +224,65 @@ func (ReferenceSlice) JSONSchema() *jsonschema.Schema {
...
+		policiesBase.Version = ""
+		base.Properties.Set("policies", policiesBase)
+	}
+
+	return base
+}
+
+type EnvironmentSchema Environment
+
+// JSONSchema defines a custom schema definition for Group. If the Boundaries feature is not enabled,
```
arthurpitman: Same here: EnvironmentSchema?

pkg/account/persistence/internal/types/types.go
```diff
@@ -160,9 +224,65 @@ func (ReferenceSlice) JSONSchema() *jsonschema.Schema {
...
+// the schema of AccountWithoutBoundaries and EnvironmentWithoutBoundaries will be returned for the group.
+func (Environment) JSONSchema() *jsonschema.Schema {
```
arthurpitman, line 243: `EnvironmentSchema`?

pkg/account/persistence/internal/types/types.go
d0weinberger: Ah yes, thx!  The whole comment is wrong, currently. A leftover from a different approach I tried before.

pkg/account/persistence/internal/types/types.go
d0weinberger: True, thx!

pkg/account/persistence/internal/types/types.go
```diff
@@ -123,6 +140,53 @@ func (r *Reference) UnmarshalYAML(unmarshal func(any) error) error {
...
+
+	// The shorthand syntax unmarshalling did not work, so it must be the PolicyBinding struct.
+	// A temporary struct is defined. We cannot just use "any" as then the UnmarshalYAML function for Reference would not get called.
+	var temp struct {
+		Type       string
+		Id         string
+		Value      string
+		Policy     *Reference
+		Boundaries ReferenceSlice
+	}
```
arthurpitman: To save redefining the whole struct, this should work:
```suggestion
	type policyBinding PolicyBinding
	var temp policyBinding
```

pkg/account/persistence/loader/validate.go
```diff
@@ -76,31 +77,43 @@ func validateServiceUserReferences(res *account.Resources, serviceUser account.S
...
-				return fmt.Errorf("group '%s' account references missing policy '%s'", group.Name, policyReference.ID())
-			}
+	return nil
+}
+
+func validateGroupReferences(res *account.Resources, group account.Group) error {
+	// check policy bindings in environment policies
+	for _, env := range group.Environment {
+		if err := validatePolicyBindingReferences(res, env.Policies); err != nil {
+			return fmt.Errorf("group '%s' has an invalid policy reference for environment '%s': %v", group.Name, env.Name, err)
```
arthurpitman:
```suggestion
			return fmt.Errorf("group '%s' has an invalid policy reference for environment '%s': %w", group.Name, env.Name, err)
```

pkg/account/persistence/loader/validate.go
```diff
@@ -76,31 +77,43 @@ func validateServiceUserReferences(res *account.Resources, serviceUser account.S
...
+		}
+	}
+	if group.Account != nil {
+		// check policy bindings in account policies
+		if err := validatePolicyBindingReferences(res, group.Account.Policies); err != nil {
+			return fmt.Errorf("group '%s' has an invalid account policy reference: %v", group.Name, err)
```
arthurpitman:
```suggestion
			return fmt.Errorf("group '%s' has an invalid account policy reference: %w", group.Name, err)
```

pkg/account/persistence/loader/validate.go
```diff
@@ -231,3 +269,39 @@ func validateReference(reference types.Reference) error {
...
+	}
+
+	for _, boundaryRef := range policyBinding.Boundaries {
+		if err := validateReference(boundaryRef); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
arthurpitman:
```suggestion
func validatePolicyBinding(policyBinding types.PolicyBinding) error {
	// We don't need to check for the Value field. The Value field is only set when no key is used, i.e.
	// policies:
	//   - My policy
	// In such a case it would be an invalid yaml file anyway if a child element "policy" was to be added.

	if policyBinding.Policy == nil {
		return validateLegacyPolicyBinding(policyBinding)
	}

	if policyBinding.Id != "" {
		return fmt.Errorf("policy definition is ambiguous. Use the 'policy' key only and remove the keys 'id' and 'type' for policy with id '%v'", policyBinding.Id)
	}

	if err := validateReference(*policyBinding.Policy); err != nil {
		return err
	}

	for _, boundaryRef := range policyBinding.Boundaries {
		if err := validateReference(boundaryRef); err != nil {
			return err
		}
	}

	return nil
}

func validateLegacyPolicyBinding(policyBinding types.PolicyBinding) error {
	if policyBinding.Type == types.ReferenceType {
		if policyBinding.Id == "" {
			return errors.New("missing required field 'id' for reference")
		}
	} else if policyBinding.Value == "" {
		return errors.New("missing reference value")
	}

	if len(policyBinding.Boundaries) > 0 {
		return fmt.Errorf("error when loading boundary reference with id %v: boundaries are only supported when using the 'policy' key", policyBinding.Id)
	}
	return nil
}
```

PR #2020, opened 2025-09-03
chore: Upgrade corelib to newest version

d0weinberger:
#### **Why** this PR/What has changed?
The account management client in the core library was updated, therefore we want to update the corelib dependency.

#### How is it **tested**?
TODO: Account management tests are still working

It shouldn't

**Issue:**
CA-16337


pkg/account/deployer/client.go
```diff
@@ -201,9 +201,23 @@ func (c *accountManagementClient) getGroupsByName(ctx context.Context, name stri
 	return groupsMatchingName, nil
 }
 
+func groupToInsertGroup(group Group) accountmanagement.InsertGroupDto {
+	return accountmanagement.InsertGroupDto{
+		Uuid:                     group.Uuid,
```
d0weinberger: Since the field is readOnly, I'd guess this would result in an error with the API if the UUID is set, but in this case I would think it is better to show this error transparently.

EDIT: Had a look at the `deployer.go` and UUID is not used and set there at all:

```
data := accountmanagement.PutGroupDto{
		Name:                     group.Name,
		Description:              &group.Description,
		FederatedAttributeValues: group.FederatedAttributeValues,
	}
```

Kirdock (approved):

Kirdock:
Failing tests are related to context deadlines, so still fine to me 👍

arthurpitman (approved):

PR #2025, opened 2025-09-05
feat(boundary) resource download

d0weinberger:
#### **Why** this PR?
For the boundary support, we also need to support the download of boundaries.

#### **What** has changed?
On `monaco account download`, 
- boundaries are now also downloaded
- boundary references are present in the policy bindings of the downloaded groups

**Example**
In `boundaries.yaml`
```
boundaries:
- id: Myspecialboundary
  name: My special boundary
  query: cloudautomation:event = "helloworld";
  originObjectId: fbca1698-dbdd-4a60-aec5-4da8390efa01
```

In `groups.yaml`
```
- id: group-with-boundaries
  name: group-with-boundaries
  account:
    policies:
    - policy:
        type: reference
        id: mypol
      boundaries:
      - type: reference
        id: Myspecialboundary
```

#### **How** does it do it?
In the `client.go`, all boundaries are loaded. This includes taking care of pagination. Since there is no `NextPageKey`, the loop breaks if the size of the returned boundary list is smaller than the requests list size, assuming that this had been the last page.

These downloaded boundaries are ready to be written to a boundaries file and used for resolving boundary references in policy bindings.

Note: 
Download boundaries is guarded with the env var `MONACO_FEAT_ENABLE_BOUNDARIES`

#### How is it **tested**?
`downloader_test` is extended with a test case containing boundaries and boundary references in policy bindings.

Since the feature is not yet released, it doesn't, but as soon as it is, users can download boundaries and boundary references in policy bindings of groups.
 
**Issue:**
CA-16339

### DIFF
--- a/pkg/account/downloader/internal/http/client.go
+++ b/pkg/account/downloader/internal/http/client.go
@@ -86,7 +86,7 @@ func (c *Client) GetGroupsForUser(ctx context.Context, userEmail string, account
 }
 
 func (c *Client) GetBoundaries(ctx context.Context, accountUUID string) ([]accountmanagement.PolicyBoundaryOverview, error) {
-	boundaries := make([]accountmanagement.PolicyBoundaryOverview, 0)
+	boundaries := []accountmanagement.PolicyBoundaryOverview{}
 	const pageSize = 100
 	for page := (int32)(1); page < math.MaxInt32; page++ {
 		r, err := c.getBoundariesPage(ctx, accountUUID, page, pageSize)
--- a/pkg/account/downloader/internal/http/client_test.go
+++ b/pkg/account/downloader/internal/http/client_test.go
@@ -1,0 +1,152 @@
+/*
+ * @license
+ * Copyright 2025 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package http
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/api/rest"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/clients/accounts"
+	accountmanagement "github.com/dynatrace/dynatrace-configuration-as-code-core/gen/account_management"
+	"github.com/dynatrace/dynatrace-configuration-as-code-core/testutils"
+)
+
+func TestClient_GetBoundaries(t *testing.T) {
+
+	responses := []testutils.ResponseDef{
+		{
+			GET: func(t *testing.T, request *http.Request) testutils.Response {
+				return testutils.Response{
+					ResponseCode: http.StatusOK,
+					ResponseBody: `{
+  "pageSize": 1,
+  "pageNumber": 1,
+  "totalCount": 1,
+  "content": [
+	{
+	  "uuid": "some-fake-uuid",
+	  "name": "Monaco Test Boundary",
+	  "boundaryQuery": "cloudautomation:event = \"helloworld\";",
+	  "levelType": "account",
+	  "levelId": "abcde",
+	  "boundaryConditions": [],
+	  "metadata": null
+	}
+  ]
+}`,
+				}
+			},
+			ValidateRequest: func(t *testing.T, request *http.Request) {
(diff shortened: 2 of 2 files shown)


pkg/account/downloader/downloader_test.go
```diff
@@ -25,18 +25,21 @@ import (
...
 var (
-	accountUUID = "32ee817e-0db4-4f28-b1d1-a2b7032cdf29"
-	groupUUID1  = "27dde8b6-2ed3-48f1-90b5-e4c0eae8b9bd"
-	groupUUID2  = "3c345885-ff01-428b-ba49-b3381819f6dd"
-	toID        = stringutils.Sanitize
-	originalErr = errors.New("original error")
+	accountUUID   = "32ee817e-0db4-4f28-b1d1-a2b7032cdf29"
+	groupUUID1    = "27dde8b6-2ed3-48f1-90b5-e4c0eae8b9bd"
+	groupUUID2    = "3c345885-ff01-428b-ba49-b3381819f6dd"
+	policyUUID1   = "35902b72-b3bb-4de8-a3d2-4131d73170b5"
```
Kirdock: [minor] not used variable

pkg/account/downloader/internal/http/client.go
```diff
@@ -85,6 +85,38 @@ func (c *Client) GetGroupsForUser(ctx context.Context, userEmail string, account
...
+	const pageSize = 100
+	for page := (int32)(1); page < math.MaxInt32; page++ {
+		r, err := c.getBoundariesPage(ctx, accountUUID, page, pageSize)
+		if err != nil {
+			return nil, err
+		}
+
+		boundaries = append(boundaries, r.Content...)
+		// If the amount of boundaries returned on the page is less than the requested 100, we can assume it was the last page.
+		if len(r.Content) < 100 {
```
Kirdock: Same as for https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/2026/files#r2394058765

Kirdock (approved):

pkg/account/downloader/groups.go
```diff
@@ -184,17 +189,17 @@ func getManagementZonesFor(scope string, perDTOs *accountmanagement.PermissionsG
 	return retVal
 }
 
-func getPoliciesFor(binding *accountmanagement.LevelPolicyBindingDto, groupUUID string) []string {
-	var retVal []string
+func getPoliciesFor(binding *accountmanagement.LevelPolicyBindingDto, groupUUID string) PolicyToBoundaries {
+	policies := make(PolicyToBoundaries)
```
Laubi: Use `make` only where you need to set len/cap.
Use  `policies := PolicyToBoundaries{}` here.

pkg/account/downloader/downloader_test.go
```diff
@@ -498,6 +500,158 @@ func TestDownloader_GroupsWithPolicies(t *testing.T) {
 	assert.Empty(t, result.ServiceUsers)
 }
 
+// TestDownloader_GroupsWithPoliciesAndBoundaries tests that downloading groups with policies and boundaries succeeds.
+func TestDownloader_GroupsWithPoliciesAndBoundaries(t *testing.T) {
+	t.Setenv(featureflags.Boundaries.EnvName(), "true")
+	client := http.NewMockhttpClient(gomock.NewController(t))
+	downloader := downloader.New4Test(&account.AccountInfo{
```
Laubi: Not your code, but please rename `New4Test`...
Can be in a follow-up, but why use numbers for words...

pkg/account/downloader/downloader_test.go
```diff
@@ -498,6 +500,158 @@ func TestDownloader_GroupsWithPolicies(t *testing.T) {
...
+		Name:        "test",
+		AccountUUID: accountUUID,
+	}, client)
+
+	accountPolicyOverview := accountmanagement.PolicyOverview{
+		Uuid:      "2ff9314d-3c97-4607-bd49-460a53de1390",
```
Laubi, line 513: Just to make sure, you did not copy them, but generated the UUIDs randomly, correct? 
For testing, I suggest using obviously wrong uuids, so that I don't have to ask this question

pkg/account/downloader/downloader_test.go
```diff
@@ -498,6 +500,158 @@ func TestDownloader_GroupsWithPolicies(t *testing.T) {
...
+	}, nil)
+
+	client.EXPECT().GetUsers(gomock.Any(), accountUUID).Return([]accountmanagement.UsersDto{}, nil)
+	client.EXPECT().GetServiceUsers(gomock.Any(), accountUUID).Return([]accountmanagement.ExternalServiceUserDto{}, nil)
+
+	result, err := downloader.DownloadResources(t.Context())
+	assert.NoError(t, err)
+	require.NotNil(t, result)
+
+	assert.Equal(t, map[account.BoundaryId]account.Boundary{
```
Laubi: More nit than anything, but I would define the map here as an extra variable.

expectedSomething := ....

assert.Equal(t, expected, actual)

Here we have a function call over 60 lines, where 59 of them are the definition of what we expect

pkg/account/downloader/downloader_test.go
d0weinberger: This specific uuid shows up in `downloader_test.go` many times in different test cases. See line 79, e.g. 
I copied it from another test case. I didn't copy it from a real policy that I deployed.

Kirdock (approved):

PR #2026, opened 2025-09-05
feat(boundary) resource deploy

d0weinberger:
#### **Why** this PR?
For the boundary support, we also need to support the deployment of boundaries and referencing boundaries in policy bindings.

#### **What** has changed?
On `monaco account deploy`
- boundaries are now also deployed
- bondary references in policy bindings of groups are recognized, supported, and deployed as well.

**Example**
In `boundaries.yaml`
```
boundaries:
- id: Myspecialboundary
  name: My special boundary
  query: cloudautomation:event = "helloworld";
  originObjectId: fbca1698-dbdd-4a60-aec5-4da8390efa01
```

In `groups.yaml`
```
- id: group-with-boundaries
  name: group-with-boundaries
  account:
    policies:
    - policy:
        type: reference
        id: mypol
      boundaries:
      - type: reference
        id: Myspecialboundary
```

#### **How** does it do it?
In the `client.go`, code for getting boundaries is added. This includes taking care of pagination. Since there is no `NextPageKey`, the loop breaks if the size of the returned boundary list is smaller than the requests list size, assuming that this had been the last page.

Also, code for upserting boundaries and, correspondingly, creating and updating boundaries, is added. There is also a function for getting all boundary ids.

For boundary references in policy bindings, the functions `updateAccountPolicyBindings` and `updateEnvironmentPolicyBindings` are adapted. A helper function `updateBoundariesForPolicyBinding` is introduced.

Boundary IDs are also stored in the `idMap` struct. 

The `deployer.go` ties everything together, by fetching boundary IDs in `fetchExistingResources` and using the upsert mechanism from `client.go` for `deployBoundaries`

#### How is it **tested**?

Tests are missing at the moment.

Since the feature is not yet released, it doesn't, but as soon as it is, users can download boundaries and boundary references in policy bindings of groups.

**Issue:**
CA-16340

### DIFF
--- a/pkg/account/deployer/client.go
+++ b/pkg/account/deployer/client.go
@@ -65,7 +65,7 @@ func (c *accountManagementClient) getBoundaryIds(ctx context.Context) (map[strin
 		return nil, err
 	}
 
-	result := make(map[string]remoteId)
+	result := map[string]remoteId{}
 	for _, bnd := range boundaries {
 		result[bnd.Name] = bnd.GetUuid()
 	}
@@ -155,7 +155,7 @@ func (c *accountManagementClient) createBoundary(ctx context.Context, boundary B
 	}
 
 	if createdBoundary == nil {
-		return "", errors.New("the received data are empty")
+		return "", errors.New("the received data is empty")
 	}
 
 	return createdBoundary.Uuid, nil
@@ -184,7 +184,7 @@ func (c *accountManagementClient) getBoundaryByName(ctx context.Context, name st
 }
 
 func (c *accountManagementClient) getBoundaries(ctx context.Context) ([]accountmanagement.PolicyBoundaryOverview, error) {
-	boundaries := make([]accountmanagement.PolicyBoundaryOverview, 0)
+	boundaries := []accountmanagement.PolicyBoundaryOverview{}
 	const pageSize = 100
 	for page := (int32)(1); page < math.MaxInt32; page++ {
 		r, err := c.getBoundariesPage(ctx, c.accountInfo.AccountUUID, page, pageSize)
@@ -209,7 +209,7 @@ func (c *accountManagementClient) getBoundariesPage(ctx context.Context, account
 		return nil, err
 	}
 	if r == nil {
-		return nil, errors.New("the received data are empty")
+		return nil, errors.New("the received data is empty")
 	}
 	return r, nil
 }
@@ -550,10 +550,8 @@ func (c *accountManagementClient) updateAccountPolicyBindings(ctx context.Contex
 		return fmt.Errorf("group id must not be empty")
 	}
 
-	var policyIds []string
-	if boundariesForPolicyIds == nil {
-		policyIds = []string{}
-	} else {
+	policyIds := []string{}
+	if boundariesForPolicyIds != nil {
 		policyIds = maps.Keys(boundariesForPolicyIds)
 	}
 	data := accountmanagement.PolicyUuidsDto{PolicyUuids: policyIds}
@@ -646,7 +644,7 @@ func (c *accountManagementClient) updateBoundariesForPolicyBinding(ctx context.C
 	data := accountmanagement.AppendLevelPolicyBindingForGroupDto{Boundaries: boundaryIds}
 	resp, err := c.client.PolicyManagementAPI.UpdateLevelPolicyBindingForPolicyAndGroup(ctx, groupId, policyId, levelId, levelType).AppendLevelPolicyBindingForGroupDto(data).Execute()
 	defer closeResponseBody(resp)
-	if err = handleClientResponseError(resp, err, "unable to update boundaries for policy binding between group with UUID "+groupId+", policy with UUID "+policyId+" and boundaries with UUIDs "+fmt.Sprintf("%v", boundaryIds)); err != nil {
+	if err = handleClientResponseError(resp, err, fmt.Sprintf("unable to update boundaries for policy binding between group with UUID %s, policy with UUID %s and boundaries with UUIDs %v", groupId, policyId, boundaryIds)); err != nil {
 		return err
 	}
 	return nil
--- a/pkg/account/deployer/deployer.go
+++ b/pkg/account/deployer/deployer.go
@@ -575,7 +575,7 @@ func (d *AccountDeployer) getAccountPolicyRefs(group account.Group) (map[remoteI
 }
 
 func (d *AccountDeployer) getEnvPolicyRefs(group account.Group) (map[envName]map[remoteId][]remoteId, error) {
-	result := make(map[envName]map[remoteId][]remoteId)
+	result := map[envName]map[remoteId][]remoteId{}
 	var err error
 	if group.Environment != nil {
 		for _, e := range group.Environment {
@@ -597,7 +597,7 @@ func (d *AccountDeployer) getServiceUserGroupRefs(serviceUser account.ServiceUse
 }
 
 func (d *AccountDeployer) processPolicyBindings(items []account.PolicyBinding) (map[remoteId][]remoteId, error) {
-	refs := make(map[remoteId][]remoteId)
+	refs := map[remoteId][]remoteId{}
 	for _, item := range items {
 		polRef := d.policyIdLookup(item.Policy.ID())
 		if polRef == "" {


pkg/account/deployer/client.go
```diff
@@ -96,6 +112,108 @@ func (c *accountManagementClient) getManagementZones(ctx context.Context) ([]Man
...
+	const pageSize = 100
+	for page := (int32)(1); page < math.MaxInt32; page++ {
+		r, err := c.getBoundariesPage(ctx, c.accountInfo.AccountUUID, page, pageSize)
+		if err != nil {
+			return nil, err
+		}
+
+		boundaries = append(boundaries, r.Content...)
+		// If the amount of boundaries returned on the page is less than the requested 100, we can assume it was the last page.
+		if len(r.Content) < 100 {
```
Kirdock: Technically `#results % 100 == 0` would mean another request, even if it's not needed.
In won't lead to an error, as the API just returns an empty array if the page is too high, but because we experience 429 errors quite frequently I would suggest adapting the logic to not do any not needed request.

pkg/account/deployer/deployer.go
```diff
@@ -320,6 +358,15 @@ func (d *AccountDeployer) deployServiceUserBindings(ctx context.Context, service
 	}
 }
 
+func (d *AccountDeployer) upsertBoundary(ctx context.Context, boundary account.Boundary) (remoteId, error) {
+	data := accountmanagement.PolicyBoundaryDto{
+		Name:          boundary.Name,
+		BoundaryQuery: boundary.Query,
+	}
+
+	return d.accClient.upsertBoundary(ctx, boundary.OriginObjectID, data)
```
Kirdock, line 367: At this point, I'm wondering if it makes sense to add something like this
```suggestion
oid := boundary.OriginObjectID
foundId := d.idMap.getBoundaryUUID(boundary.ID)
if oid == "" && foundId != "" {
  oid = foundId
}
return d.accClient.upsertBoundary(ctx, oid, data)
```

pkg/account/deployer/client.go
d0weinberger: As discussed, that's an edge case we don't really antipicate happening, and if it does, it's one extra call, so not really that harmful.

pkg/account/deployer/client.go
```diff
@@ -442,25 +564,47 @@ func (c *accountManagementClient) updateAccountPolicyBindings(ctx context.Contex
...
 	data := accountmanagement.PolicyUuidsDto{PolicyUuids: policyIds}
 	resp, err := c.client.PolicyManagementAPI.UpdatePolicyBindingsToGroup(ctx, groupId, envName, "environment").PolicyUuidsDto(data).Execute()
 	defer closeResponseBody(resp)
 	if err = handleClientResponseError(resp, err, "unable to update policy binding between group with UUID "+groupId+" and policies with UUIDs "+fmt.Sprintf("%v", policyIds)); err != nil {
 		return err
 	}
+
+	if !featureflags.Boundaries.Enabled() {
+		return nil
+	}
```
Kirdock, line 599: Correct me if I'm wrong, but currently the policies are added twice if the FF is enabled right? First without boundaries and second with a boundary, leading to a conflict. The same is true for the account-policy-binding update
```suggestion
if !featureflags.Boundaries.Enabled() {
		data := accountmanagement.PolicyUuidsDto{PolicyUuids: policyIds}
		resp, err := c.client.PolicyManagementAPI.UpdatePolicyBindingsToGroup(ctx, groupId, envName, "environment").PolicyUuidsDto(data).Execute()
		defer closeResponseBody(resp)
		if err = handleClientResponseError(resp, err, "unable to update policy binding between group with UUID "+groupId+" and policies with UUIDs "+fmt.Sprintf("%v", policyIds)); err != nil {
			return err
		}
		return nil
	}
```

Kirdock:
I also recognized that deploying a config with policy-boundary bindings twice leads to a conflict. Seems like `updateBoundariesForPolicyBinding` always adds policy bindings to groups => the first deploy already added them, the second one runs into a conflict because they already exist

pkg/account/deployer/deployer.go
d0weinberger: Not only for boundaries though but for all resources

Kirdock: It can be a follow-up, though. I think we can reduce the amount of API calls here

pkg/account/deployer/client.go
Kirdock: Now, when working on the E2E, it looks like we still need it to remove policy-bindings, so no adjustment needed

pkg/account/deployer/client_test.go
```diff
@@ -1009,6 +1037,247 @@ func TestClient_GetGlobalPolicies(t *testing.T) {
...
+	server := testutils.NewHTTPTestServer(t, responses)
+	defer server.Close()
+	instance := NewClient(account.AccountInfo{Name: "my-account", AccountUUID: "abcde"}, accounts.NewClient(rest.NewClient(server.URL(), server.Client())))
+	id, err := instance.upsertBoundary(t.Context(), "", Boundary{
+		Name:          "Monaco Test Boundary",
+		BoundaryQuery: "cloudautomation:event = \"goodbye\";",
+	})
+	assert.Error(t, err)
+	assert.Zero(t, id)
+	assert.Equal(t, "unable to update boundary with name: Monaco Test Boundary (HTTP 500): {\"error\" : \"some-error\"}", err.Error())
```
Kirdock, line 1157: [minor]: To make it less error-prone to error-message changes, we could assert the error response. (same for the createFails and GetPoliciesFails test)
e.g., 
```go
expectedErrMsg := `{"error" : "some-error"}`
// ...
PUT: func(t *testing.T, request *http.Request) testutils.Response {
				return testutils.Response{
					ResponseCode: http.StatusInternalServerError,
					ResponseBody: expectedErrMsg,
				}
			},
// ...
assert.ErrorContains(t, err, expectedErrMsg)
```

Kirdock (dismissed):

pkg/account/deployer/client_test.go
d0weinberger: In principle, yes, but I don't think these messages will change a lot. For now, I'd argue, asserting on the exact wording creates a stronger test case. If this bothers us too much in the future, we could still change it. Currently, it was easy to add new test cases.

pkg/account/deployer/client.go
```diff
@@ -442,25 +564,47 @@ func (c *accountManagementClient) updateAccountPolicyBindings(ctx context.Contex
...
 	}
 	if groupId == "" {
 		return fmt.Errorf("group id must not be empty")
 	}
-	if policyIds == nil {
+	var policyIds []string
+	if boundariesForPolicyIds == nil {
 		policyIds = []string{}
+	} else {
+		policyIds = maps.Keys(boundariesForPolicyIds)
```
Laubi, line 588: Nit: Idiomatic Go code suggests the following instead:
```
policyIds := []string{}
if boundariesForPolicyIds != nil {
    policyIds = maps.Keys(boundariesForPolicyIds)
}
```

Use `else` branches only where necessary

pkg/account/deployer/client.go
```diff
@@ -488,6 +632,26 @@ func (c *accountManagementClient) deleteAllEnvironmentPolicyBindings(ctx context
...
+		return fmt.Errorf("policy id must not be empty")
+	}
+	if boundaryIds == nil {
+		boundaryIds = []string{}
+	}
+
+	data := accountmanagement.AppendLevelPolicyBindingForGroupDto{Boundaries: boundaryIds}
+	resp, err := c.client.PolicyManagementAPI.UpdateLevelPolicyBindingForPolicyAndGroup(ctx, groupId, policyId, levelId, levelType).AppendLevelPolicyBindingForGroupDto(data).Execute()
+	defer closeResponseBody(resp)
+	if err = handleClientResponseError(resp, err, "unable to update boundaries for policy binding between group with UUID "+groupId+", policy with UUID "+policyId+" and boundaries with UUIDs "+fmt.Sprintf("%v", boundaryIds)); err != nil {
```
Laubi: String cat, Sprintf in the middle, that sure looks funny :D

I'd go with something like 
```
...handleClientResponseError(resp, err, fmt.Sprintf("unable to update boundaries for policy binding between group with UUID %s, policy with UUID %s and boundaries with UUIDs %v", groupId, policyId, boundaryIds))
```

instead, utilizing one sprintf around the whole string, no concats

pkg/account/deployer/client.go
```diff
@@ -427,12 +545,16 @@ func (c *accountManagementClient) updatePermissions(ctx context.Context, groupId
...
 }
 
-func (c *accountManagementClient) updateAccountPolicyBindings(ctx context.Context, groupId string, policyIds []string) error {
+func (c *accountManagementClient) updateAccountPolicyBindings(ctx context.Context, groupId string, boundariesForPolicyIds map[string][]string) error {
 	if groupId == "" {
 		return fmt.Errorf("group id must not be empty")
 	}
-	if policyIds == nil {
+
+	var policyIds []string
```
Laubi: Same here as mentioned before.

the var here should already hold the 'default' value, which is only overwritten if the condition applies.

So `policyIds := []string{}` it, and remove the else statement

pkg/account/deployer/client.go
```diff
@@ -96,6 +112,108 @@ func (c *accountManagementClient) getManagementZones(ctx context.Context) ([]Man
...
+}
+
+func (c *accountManagementClient) getBoundariesPage(ctx context.Context, accountUUID string, page int32, pageSize int32) (*accountmanagement.PolicyBoundaryDtoList, error) {
+	r, resp, err := c.client.PolicyManagementAPI.GetPolicyBoundaries(ctx, accountUUID).Page(page).Size(pageSize).Execute()
+	defer closeResponseBody(resp)
+	if err = handleClientResponseError(resp, err, "failed to get boundaries"); err != nil {
+		return nil, err
+	}
+	if r == nil {
+		return nil, errors.New("the received data are empty")
```
Laubi: data `is` empty

pkg/account/deployer/client.go
```diff
@@ -96,6 +112,108 @@ func (c *accountManagementClient) getManagementZones(ctx context.Context) ([]Man
...
+	}
+	if foundBoundary == nil {
+		return nil, &ResourceNotFoundError{Identifier: name}
+	}
+
+	return foundBoundary, nil
+}
+
+func (c *accountManagementClient) getBoundaries(ctx context.Context) ([]accountmanagement.PolicyBoundaryOverview, error) {
+	boundaries := make([]accountmanagement.PolicyBoundaryOverview, 0)
```
Laubi: Nit: if you initialize without len/cap, don't use make

`boundaries := []accountmanagement.PolicyBoundaryOverview{}`

Fun fact, here you would not need it at all (just use a var declaration without initalization). `append` already initializes it correctly if it's empty. However, to keep the semantics of the return value (not nil, maybe thats required), we can still use the above statement without the `make` call.

pkg/account/deployer/client.go
```diff
@@ -56,6 +59,19 @@ func (c *accountManagementClient) getAccountInfo() account.AccountInfo {
 	return c.accountInfo
 }
 
+func (c *accountManagementClient) getBoundaryIds(ctx context.Context) (map[string]remoteId, error) {
+	boundaries, err := c.getBoundaries(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make(map[string]remoteId)
```
Laubi: Same as below, don't use make where not necessary.

```suggestion
	result := map[string]remoteId{}
```

Kirdock (approved):

PR #2028, opened 2025-09-08
fix:  Only generate references to document configs if ID is prefixed by a forward slash

arthurpitman:
#### **Why** this PR?
With recent changes to the documents API, custom IDs may be used with little restriction. This effectively broke the generation of references to documents, as these IDs often occurred in other situations.

#### **What** has changed?
As a workaround, references are now only generated if the ID is prefixed by a forward slash `/`, with the intent of only creating references in URLs pointing to documents. This is now the default behavior; however, this restriction can be removed by setting the feature flag `MONACO_RESTRICT_DOCUMENT_REFERENCE_CREATION` to `false`.

#### **How** does it do it?
The regex for finding IDs is now dependent on the config type. For document configs, a different prefix -  a forward slash - is required compared to other configs.

#### How is it **tested**?
Existing tests in `pkg/download/dependency_resolution/resolver/shared_test.go` are extended with extra test cases for document configs.

It should make the download of documents with custom IDs feasible, as references won't be created outside of URLs. The previous behavior is still available by disabling the feature flag.

**Issue:** CA-16427

### DIFF
--- a/internal/featureflags/temporary.go
+++ b/internal/featureflags/temporary.go
@@ -1,6 +1,6 @@
 /*
  * @license
- * Copyright 2023 Dynatrace LLC
+ * Copyright 2025 Dynatrace LLC
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
--- a/pkg/download/dependency_resolution/resolver/shared.go
+++ b/pkg/download/dependency_resolution/resolver/shared.go
@@ -1,6 +1,6 @@
 /*
  * @license
- * Copyright 2023 Dynatrace LLC
+ * Copyright 2025 Dynatrace LLC
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
@@ -18,6 +18,7 @@ package resolver
 
 import (
 	"fmt"
+	"log/slog"
 	"regexp"
 	"strings"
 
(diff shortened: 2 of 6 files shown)


internal/featureflags/temporary.go
```diff
@@ -33,6 +33,8 @@ const (
 	SanitizeBucketNames FeatureFlag = "MONACO_SANITIZE_BUCKET_NAMES"
 	// PlatformToken toggles whether the use of platform tokens is enabled or not (deploy, download, read manifest, etc.)
 	PlatformToken FeatureFlag = "MONACO_FEAT_ENABLE_PLATFORM_TOKENS"
+	// RestrictDocumentReferenceCreation toggles whether the creation of references to documents on download is resticted due to the API's support of liberal custom IDs.
+	RestrictDocumentReferenceCreation FeatureFlag = "MONACO_RESTRICT_DOCUMENT_REFERENCE_CREATION"
```
d0weinberger: [nit] This might even become a permanent feature flag, right? So I think `permanent.go` would be a better place for it.

d0weinberger (dismissed):

internal/featureflags/temporary.go
arthurpitman: Good question, @Laubi, what do you think? I would say we can move it then if need be.

tomazpu (dismissed):

internal/featureflags/temporary.go
Laubi: Yes. I think there's no way around it atm. Can be a follow up

pkg/download/dependency_resolution/resolver/shared.go
```diff
@@ -80,24 +81,36 @@ func replaceAll(content string, key string, s string) string {
...
 		return strings.ReplaceAll(content, key, s)
 	}
 
 	return re.ReplaceAllString(content, fmt.Sprintf("$1%s$3", s))
 }
 
+// getMatchingRegEx returns a regex for finding the specified key related to the config type.
+func getMatchingRegEx(key string, configType config.Type) (*regexp.Regexp, error) {
+	if featureflags.RestrictDocumentReferenceCreation.Enabled() {
+		switch configType.(type) {
```
Laubi: Can't that just be `configType.ID() == types.DocumentTypeID`?

Laubi (dismissed):

internal/featureflags/temporary.go
arthurpitman: I have moved it 👍

pkg/download/dependency_resolution/resolver/shared.go
arthurpitman: Done 👍

PR #2033, opened 2025-09-22
fix(document): Skip download of documents with `originExtensionId` set

arthurpitman:
Documents with `originExtensionId` set are ready-made documents and should therefore not be downloaded.

**Issue:**


go.sum
Kirdock, line 1: Please `go mod tidy`

Kirdock (commented):
Regarding testing: To summarize, it is tested because the list was modified to include a new document with `originExtensionId` set and because we are checking the length of the configs nothing changed, correct?

arthurpitman:
> Regarding testing: To summarize, it is tested because the list was modified to include a new document with `originExtensionId` set and because we are checking the length of the configs nothing changed, correct?

Correct. We are adding a document that is discarded and so leaves the documents that are returned unchanged.

Kirdock (approved):

PR #2037, opened 2025-10-01
fix(support-archive): write command errors

Kirdock:
#### **Why** this PR?
The support archive didn't contain all errors that were reported.

#### **What** has changed?
The error returned from the command was not inside the support archive. That was because the errors returned inside the command run weren't logged inside, and the archive is written right after the run via the onFinalize of the command. Therefore, writing the support archive happened before the returned error was handled/logged by the parent that executed the command.

#### **How** does it do it?
By returning a pointer to the supportArchive flag of the command and checking based on this one after the command execution if the support archive should be written or not.

Users now see every logged error in the support archive.

**Issue:** CA-16548

### Regarding Sonar
There is no way to test the missing line in `main.go`, as `os.Exit` leads to a panic in tests

### DIFF
--- a/cmd/monaco/main_test.go
+++ b/cmd/monaco/main_test.go
@@ -1,0 +1,64 @@
+//go:build unit
+
+/*
+ * @license
+ * Copyright 2025 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/testutils"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/timeutils"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/trafficlogs"
+)
+
+func TestExecuteMain(t *testing.T) {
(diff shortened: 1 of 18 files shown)


cmd/monaco/runner/runner.go
```diff
@@ -48,22 +53,28 @@ func RunCmd(ctx context.Context, cmd *cobra.Command) error {
 	return err
 }
 
-func BuildCmd(fs afero.Fs) *cobra.Command {
+// BuildCmd builds the root Monaco command
+// Returns:
+//   - The built command
+//   - A boolean flag indicating if the supportArchive should be created or not, which set during the execution
```
arthurpitman:
```suggestion
//   - A pointer to a boolean flag indicating if the support archive should be created or not, which is set during execution
```

cmd/monaco/runner/runner.go
```diff
@@ -48,22 +53,28 @@ func RunCmd(ctx context.Context, cmd *cobra.Command) error {
...
+	if err := supportarchive.Write(fs); err != nil {
+		log.With(log.ErrorAttr(err)).Error("Encountered error creating support archive. Archive may be missing or incomplete: %s", err)
 	}
 }
 
-func BuildCmdWithLogSpy(fs afero.Fs, logSpy io.Writer) *cobra.Command {
+// BuildCmdWithLogSpy builds the root Monaco command with a given log writer
+// Returns:
+//   - The built command
+//   - A boolean flag indicating if the supportArchive should be created or not, which set during the execution
```
arthurpitman:
```suggestion
//   - A pointer to a boolean flag indicating if the support archive should be created or not, which is set during execution
```

arthurpitman (approved):

cmd/monaco/main_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/testutils"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/timeutils"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/internal/trafficlogs"
+)
+
+func TestExecuteMain(t *testing.T) {
+	fixedTime := timeutils.TimeAnchor().Format(trafficlogs.TrafficLogFilePrefixFormat)
+
+	t.Run("It should gracefully execute the main program", func(t *testing.T) {
```
Laubi, line 36: We don't use "it ..." structure in go. This is not JavaScript Jest :)

Kirdock: On the one hand, why not, and on the other hand, does it really matter for a test name?

PR #2129, opened 2026-04-13
﻿feat: add --log-payloads flag to deploy command

pieterwintmolders-cgk:
#### **Why** this PR?
﻿Debugging Monaco templates and validating rendered payloads before or during deployment is currently not possible.
The final JSON sent to the Dynatrace API is not exposed, making it hard to diagnose template issues or review
changes.
#### **What** has changed?
﻿A new --log-payloads flag has been added to the deploy command. When enabled, the rendered JSON payload for each
configuration is written to an individual file under
.logs/{timestamp}-payloads/{env}/{project}/{type}/{configId}.json.
#### **How** does it do it?
﻿A payloadprinter.Writer is attached to the context when --log-payloads is set. After each config is rendered (just
before deployment), Record() is called to write the payload immediately to its own JSON file. Environment variable
parameter values are automatically redacted to prevent leaking secrets. On completion, a summary log line reports the
count and directory.
#### How is it **tested**?
﻿Unit tests cover:

 - Correct file path generation (including colon-sanitization in type names)
 - Secret redaction for environment variable parameters
 - Graceful handling of unset or empty env vars
 - File write count tracking
 - Context writer injection/retrieval
﻿Opt-in only via --log-payloads — no impact on existing behaviour. Users can now inspect resolved payloads for
debugging:

 monaco deploy manifest.yaml -e acceptance -p my-app --dry-run --log-payloads
 \# outputs to: .logs/20260413-143809-payloads/acceptance/my-app/builtin-alerting.profile/my-profile.json


Laubi:
Hi @pieterwintmolders-cgk, thanks for the contribution. 

We discussed the feature request internally and agree that it would be useful. 
Before we go into a deeper, code-level review, we'd like to request some higher-level changes. Also, you'll need to sign the CLA, otherwise, we can't continue with your contribution.

One general question up front: 
are you aware of the [Monaco support **archive](https://docs.dynatrace.com/docs/deliver/configuration-as-code/monaco/reference/logging#support-archive) `--support-archive`? 
This flag creates multiple logs, inncluding request/response logs. 
For us, your change is still usefull as the yaml configurations are still created with the `--dry-run` command.

Requested changes:
* Use feature flags (environment variables) for the logging/debug behavior. 
We typically change logging/debug behavior behind env vars, for example, `MONACO_LOG_SOURCE` and `MONACO_LOG_MEM_STATS`. 
Users have 
  Users should be able to enable/disable this feature as well using an env var, e.g., `MONACO_WRITE_JSON_CONFIGS`. 
This also gives us room to iterate over the feature and think about how we can best integrate it permanently in Monaco (potentially with a flag in the future). In general, CLI flags are expected to be stable, while feature flags are allowed to evolve.
* Don't mask all env vars
We already have secret-masking code in `internal/secret/masker.go:49`. Could you please check to what extent you can reuse it to mask secrets for this feature?
* Additional feature if you're up to it: if the FF is set and `--support-archive` is passed, it would be great if the created support archive contains all generated files.

Laubi:
It looks like the CLA hasn't been signed yet, and  release workflow was added that weren't part of the original request. 

I close the PR now as it seems it has not been reviewed before the changes were submitted.

PR #2134, opened 2026-05-08
chore: Update to go 1.26.3

arthurpitman:
#### **Why** this PR?
The PR updates to Go version 1.26.3.

- Go version in `go.mod` is updated
- Core library is updated to the latest version
- Snyk configuration is updated to set up go manually based on the version in `go.mod`, rather than using a preconfigured image, which may be unavailable

#### How is it **tested**?
Existing tests

No effect

**Issue:** NA


.github/workflows/snyk.yml
```diff
@@ -15,9 +15,14 @@ jobs:
       contents: read
       pull-requests: write
     steps:
-      - name: Checkout
```
d0weinberger, line 18: Did you remove the step names on purpose?

.github/workflows/snyk.yml
```diff
@@ -15,9 +15,14 @@ jobs:
...
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-      - name: Run Snyk to check for vulnerabilities
-        uses: snyk/actions/golang@9adf32b1121593767fc3c057af55b55db032dc04 # v1.0.0
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: snyk/actions/setup@9adf32b1121593767fc3c057af55b55db032dc04 # v1.0.0
+      - name: 🛠️ Set up Go 1.x
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c #v6.4.0
+        with:
+          go-version-file: go.mod
+      - name: Snyk monitor
```
Laubi: Will this not fail if it isn't available as well? 

`snyk/actions/golang` just execute the `snyk test` command, so it will fail as well, right? If this is the case, I would keep it as is tbh. 
Sure, the setup works, but that was never the question anyway.

https://github.com/snyk/actions/blob/master/golang/action.yml#L10-L12

arthurpitman: Like this we install go, rather than using what ever is in the Snyk image

.github/workflows/snyk.yml
arthurpitman: Good point, will add them back

.github/workflows/snyk.yml
Laubi: It might work, as the docs state that it _uses the commands_. So let's see, I am hopful 🤔 
https://docs.snyk.io/supported-languages/supported-languages-list/go#go-modules-and-the-cli

Laubi (dismissed):

Kirdock (approved):

d0weinberger (approved):

PR #2135, opened 2026-05-12
fix: validate tokenEndpoint domain to prevent OAuth token redirect

Laubi:
#### **Why** this PR?

`tokenEndpoint` in the manifest can be set to an arbitrary URL, allowing an attacker to redirect OAuth token requests to a server they control and steal credentials.

#### **What** has changed?

`tokenEndpoint` is now validated against an allowlist of domains: `*.dynatracelabs.com` and `*.dynatrace.com`. Any other domain is rejected at manifest load time with a clear error.

#### **How** does it do it?

A new `validateTokenEndpointDomain` function parses the URL and checks the hostname. It is called from `parseOAuth` after the token endpoint is resolved. Hardcoded (`type: value`) URLs are always validated; env-var endpoints skip validation only when `DoNotResolveEnvVars` is set (the value is a placeholder in that case).

#### How is it **tested**?

- `TestValidateTokenEndpointDomain` — 8 unit tests covering allowed domains, disallowed domains, and bypass attempts (lookalikes, path-embedded domains, subdomain-of-attacker)
- Two new table-driven cases in `TestLoadManifest` for rejection of arbitrary and lookalike domains
- Three new `Load()`-level tests in `TestEnvVarResolutionCanBeDeactivated` verifying the correct behavior when `DoNotResolveEnvVars` is set (hardcoded valid accepted, hardcoded invalid rejected, env-var skipped)
- Existing test fixtures updated to use valid Dynatrace domains

Users who have set `tokenEndpoint` to a non-Dynatrace domain will see a manifest load error. The fix only affects `tokenEndpoint`; environment URLs and other fields are unchanged.

**Issue:** CA-19421

### DIFF
--- a/pkg/manifest/loader/manifest_loader_test.go
+++ b/pkg/manifest/loader/manifest_loader_test.go
@@ -782,7 +782,7 @@ func TestLoadManifest(t *testing.T) {
 	t.Setenv("empty-env-var", "")
 	t.Setenv("client-id", "resolved-client-id")
 	t.Setenv("client-secret", "resolved-client-secret")
-	t.Setenv("ENV_OAUTH_ENDPOINT", "resolved-oauth-endpoint")
+	t.Setenv("ENV_OAUTH_ENDPOINT", "https://sso.dynatrace.com")
 
 	tests := []struct {
 		name             string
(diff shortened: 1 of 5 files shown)


pkg/manifest/loader/manifest_loader_test.go
```diff
@@ -1810,6 +1810,24 @@ environmentGroups: [{name: b, environments: [{name: c, url: {value: d}, auth: {t
 `,
 			errsContain: []string{"\"nonexistent\" is not a valid URL type"},
 		},
+		{
+			name: "OAuth token endpoint with disallowed domain is rejected",
+			manifestContent: `
+manifestVersion: 1.0
+projects: [{name: a, path: p}]
+environmentGroups: [{name: b, environments: [{name: c, url: {value: d}, auth: {token: {name: e}, oAuth: {clientId: {name: client-id}, clientSecret: {name: client-secret}, tokenEndpoint: {value: https://evil.com/steal}}}}]}]
```
Kirdock, line 1818: [nit] Should we really put some random domains into tests? On the other side, it's not user-facing

Kirdock (approved):

pkg/manifest/loader/manifest_loader_test.go
Laubi: Could remove it with some other dummy domains, but in the end, I don't think it's too bad.

PR #2136, opened 2026-05-18
fix: settings permissions are not downloaded

Kirdock:
#### **Why** this PR?
Settings permissions aren't downloaded when using OAuth/PlatformToken.

#### **What** has changed?
Settings permissions are downloaded

#### **How** does it do it?
By correctly handling the check if permissions should be downloaded or not.

#### How is it **tested**?
Unit tests updated

They can download settings permissions again

**Issue:** CA-19495

### DIFF
--- a/pkg/resource/settings/download.go
+++ b/pkg/resource/settings/download.go
@@ -53,14 +53,14 @@ type DownloadSource interface {
 }
 
 type DownloadAPI struct {
-	settingsSource        DownloadSource
-	filters               Filters
-	specificSchemas       []string
-	classicSettingsSource bool
+	settingsSource       DownloadSource
+	filters              Filters
+	specificSchemas      []string
+	isPlatformConnection bool
 }
 
 func NewDownloadAPI(settingsSource DownloadSource, filters Filters, specificSchemas []string, usePlatform bool) *DownloadAPI {
-	return &DownloadAPI{settingsSource, filters, specificSchemas, usePlatform}
+	return &DownloadAPI{settingsSource: settingsSource, filters: filters, specificSchemas: specificSchemas, isPlatformConnection: usePlatform}
 }
 
 func (a DownloadAPI) Download(ctx context.Context, projectName string) (project.ConfigsPerType, error) {
@@ -191,7 +191,7 @@ func (a DownloadAPI) getPermissions(ctx context.Context, s schema, objects []dtc
 		return nil, nil
 	}
 
-	if a.classicSettingsSource {
+	if !a.isPlatformConnection {
 		lg.WarnContext(ctx, "Skipped getting permissions as download is using classic credentials")
 		return nil, nil
 	}
--- a/test/configtypes/settings/settings_download_permission_test.go
+++ b/test/configtypes/settings/settings_download_permission_test.go
@@ -1,0 +1,117 @@
+//go:build integration
+
+/*
+ * @license
+ * Copyright 2026 Dynatrace LLC
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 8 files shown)


pkg/resource/settings/download.go
```diff
@@ -53,10 +53,10 @@ type DownloadSource interface {
 }
 
 type DownloadAPI struct {
-	settingsSource        DownloadSource
-	filters               Filters
-	specificSchemas       []string
-	classicSettingsSource bool
```
Laubi, line 59: This change looks so strange, but it seems correct when lookint at the code (it feels strange as no calling code is updated, so download.go code feels like unused code, when it isn't). 
What I am missing is somehow a test that verifies is e2e, would that be possible? If i give platform credentials, it does export it correctly. Would that be possible to add?

Kirdock: Integration test added 👍

test/configtypes/settings/settings_download_permission_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.yaml.in/yaml/v2"
+
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/api"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/config"
+	manifestloader "github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/manifest/loader"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project"
+	"github.com/dynatrace/dynatrace-configuration-as-code/v2/test/internal/monaco"
+	runner2 "github.com/dynatrace/dynatrace-configuration-as-code/v2/test/internal/runner"
```
Laubi: [nit] Why a custum import?

test/configtypes/settings/settings_download_permission_test.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+func TestPermissionDownload(t *testing.T) {
+	configFolder := "testdata/settings-acl/acl-write"
```
Laubi: I am really not happy cross-using testing folders. Each test-case should have it's own set of configs. We had a huge mess in the past that required extensive cleanup. 

`settings-acl` is a test with multiple subtests, so if it's possible, I would either move it to the settings_acl test, or duplicate the configs.

test/configtypes/settings/settings_download_permission_test.go
```diff
@@ -0,0 +1,119 @@
...
+			// load downloaded manifest
+			mani, errs := manifestloader.Load(&manifestloader.Context{
+				Fs:           fs,
+				ManifestPath: "download/manifest.yaml",
+				Opts:         manifestloader.Options{RequireEnvironmentGroups: true},
+			})
+			assert.Empty(t, errs, "unexpected error loading manifest")
+
+			projects, errs := project.LoadProjects(t.Context(), fs, project.ProjectLoaderContext{
+				KnownApis:       api.NewAPIs().GetApiNameLookup(),
```
Laubi: [nit] not necessary afaik, as you don't care about configs, right?

test/configtypes/settings/settings_download_permission_test.go
```diff
@@ -0,0 +1,119 @@
...
+				WorkingDir:      "download",
+				Manifest:        mani,
+				ParametersSerde: config.DefaultParameterParsers,
+			}, nil)
+			assert.Empty(t, errs, "unexpected error loading project")
```
Laubi: [nit] require? Don't need to continue if something went wrong

test/configtypes/settings/settings_download_permission_test.go
Laubi: It's a nit, but I'd appreciate this one

Laubi (dismissed):

test/configtypes/settings/settings_download_permission_test.go
Kirdock: Seems like I had another import at first, which required it 🤔.
Removed 👍

test/configtypes/settings/settings_download_permission_test.go
Kirdock: Created another folder `download-acl` 👍

test/configtypes/settings/settings_download_permission_test.go
Kirdock: Removed 👍

Laubi (dismissed):

Laubi (approved):

arthurpitman (approved):

PR #2139, opened 2026-05-19
fix: restrict account-url to dynatrace-controlled URLs

Laubi:
#### **Why** this PR?

Restrics domains to dynatrace-allowed domains to enhance security. 

#### **What** has changed?

Added domain validation for the `apiUrl` field in account configurations, mirroring the existing `tokenEndpoint` validation:
- Account `apiUrl` is now restricted to `*.dynatrace.com` and `*.dynatracelabs.com` domains
- Hostname matching is case-insensitive (RFC-compliant)
- Validation correctly rejects path-traversal and userinfo-prefix masquerade attempts

#### **How** does it do it?

1. Created `validateDynatraceDomain()` function
2. This function checks that URL hostnames match the dynatrace domain allowlist
3. Integrated validation into `account.go`, similar to the existing `parseOAuth()` check
4. Consolidated all validation tests (tokenEndpoint + apiUrl cases) into `validate_endpoint_test.go`

#### How is it **tested**?

- **12 unit tests** for the shared validator covering
- **2 integration tests** for `parseSingleAccount()`

- **Breaking change:** Manifests with account `apiUrl` values pointing outside dynatrace.com/dynatracelabs.com domains will now fail validation with a clear error message.
- This is intentional and security-critical. Users should only ever point `apiUrl` to official Dynatrace environments.
- No impact on legitimate configurations (apiUrl is rarely used; most deployments use the default environment).

**Issue:** N/a


pkg/manifest/loader/validate_endpoint_test.go
```diff
@@ -0,0 +1,158 @@
...
+				if tt.wantErr {
+					assert.Error(t, err)
+				} else {
+					assert.NoError(t, err)
+				}
+			})
+		}
+	})
+
+	t.Run("apiUrl", func(t *testing.T) {
```
Kirdock, line 59: [nit] regarding consistency, the first test is a table test, whereas here, we do it differently again. We could also add them to the previous one

pkg/manifest/loader/validate_endpoint.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		return fmt.Errorf("not a valid URL: %w", err)
+	}
+
+	host := strings.ToLower(parsed.Hostname())
+	if strings.HasSuffix(host, ".dynatracelabs.com") || strings.HasSuffix(host, ".dynatrace.com") {
+		return nil
+	}
+
+	return fmt.Errorf("host %q is not allowed: must be on a '.dynatrace.com' or '.dynatracelabs.com' domain", host)
```
Kirdock, line 38: [nit] Didn't we say that we don't want to use quotes in error messages, only single ones?

Kirdock (approved):

arthurpitman (approved):
Works with my manual testing

PR #2142, opened 2026-05-20
fix: symlink bypass

Laubi:
#### **Why** this PR?
The symlink check introduced in 5224b9fe only called lstat on the final path component. Because lstat transparently follows intermediate symlinks, a symlinked parent directory (e.g. `project/secrets/ -> ~/.ssh/`) would pass the check undetected and the file read would follow the symlink.

#### **What** has changed?
`RejectSymlinkRecursive` now walks every component of the path independently, from the root down, and calls lstat on each one. The first component found to be a symlink causes an immediate rejection.

#### **How** does it do it?
`parentDirectories` decomposes the path into its components. `rejectSymlink` performs a single-component lstat check and returns a typed `symlinkDetectedError` if a symlink is found. `RejectSymlinkRecursive` iterates from outermost to leaf and forwards the first error.

#### How is it **tested**?
Extended unit tests cover: symlinked leaf, symlinked parent, nested/chained symlinks, dangling symlinks, relative targets that escape the project, .. traversal, and paths with no symlinks. Tests use dangling symlinks so no target files need to exist.

Additionally added e2e tests with symlinks

`type: file` parameters and file-based templates that reference paths containing symlinked directories will now be rejected with an error instead of silently reading through them.

**Issue:**

### DIFF
--- a/internal/files/symlink.go
+++ b/internal/files/symlink.go
@@ -40,9 +40,15 @@ func RejectSymlink(fs afero.Fs, path string) error {
 		return nil
 	}
 
+	components := pathComponents(path)
+	return checkComponentsForSymlinks(lstater, path, components)
+}
+
+// pathComponents returns all parent directories and the path itself, from leaf to root.
+// For example, "/a/b/c" returns ["/a/b/c", "/a/b", "/a"].
+func pathComponents(path string) []string {
 	cleaned := filepath.Clean(path)
 
-	// collect path and all of its parent directories up to (but excluding) the filesystem root
 	var components []string
 	for current := cleaned; ; {
 		components = append(components, current)
@@ -52,8 +58,13 @@ func RejectSymlink(fs afero.Fs, path string) error {
 		}
 		current = parent
 	}
+	return components
+}
 
-	// check from outermost component down to the leaf so we fail on the highest offending segment
+// checkComponentsForSymlinks verifies that no path component is a symlink.
+// Checks from outermost component down to the leaf so errors report the highest offending segment.
+func checkComponentsForSymlinks(lstater afero.Lstater, originalPath string, components []string) error {
+	// iterate from outermost (root-most) to leaf
 	for i := len(components) - 1; i >= 0; i-- {
 		component := components[i]
 
@@ -67,12 +78,17 @@ func RejectSymlink(fs afero.Fs, path string) error {
 		}
 
 		if lstatCalled && fi.Mode()&os.ModeSymlink != 0 {
-			if component == cleaned {
-				return fmt.Errorf("file %q is a symbolic link, which is not allowed for security reasons", path)
-			}
-			return fmt.Errorf("path %q contains a symbolic link at %q, which is not allowed for security reasons", path, component)
+			return symlinksError(originalPath, component)
 		}
 	}
 
 	return nil
+}
+
+// symlinksError formats an error message for a symlink found at a path component.
+func symlinksError(fullPath, symlinkComponent string) error {
+	if symlinkComponent == filepath.Clean(fullPath) {
+		return fmt.Errorf("file %q is a symbolic link, which is not allowed for security reasons", fullPath)
+	}
+	return fmt.Errorf("path %q contains a symbolic link at %q, which is not allowed for security reasons", fullPath, symlinkComponent)
 }
--- a/internal/files/symlink.go
+++ b/internal/files/symlink.go
@@ -20,33 +20,47 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"slices"
 
 	"github.com/spf13/afero"
 )
 
-// RejectSymlink checks whether path or any of its parent directory components is a symbolic link
-// and returns an error if so. Symlinks are rejected to prevent reading files outside the project
-// boundary such as credential files.
-//
-// Each component of the path is checked independently because lstat on the final component
-// transparently resolves intermediate symlinked directories, which would otherwise allow a
-// symlinked parent directory to bypass the check (see CWE-59).
+// symlinkDetectedError is returned when a symlink is found at a path component.
+type symlinkDetectedError struct {
+	path string
+}
+
+func (e *symlinkDetectedError) Error() string {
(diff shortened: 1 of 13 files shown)


internal/files/symlink.go
```diff
@@ -19,34 +19,75 @@ package files
...
-	if err != nil || !exists {
-		return nil
-	}
+// symlinkDetectedError is returned when a symlink is found at a path component.
+type symlinkDetectedError struct {
+	path string
+}
 
+func (e *symlinkDetectedError) Error() string {
+	return fmt.Sprintf("%q is a symbolic link, which is not allowed for security reasons", e.path)
```
Kirdock, line 34: [nit] Didn't we say that we don't want to use quotes in error messages, only single ones?

internal/files/symlink.go
```diff
@@ -19,34 +19,75 @@ package files
...
+
+// rejectSymlink returns an error if the path is a symlink, nil otherwise.
+func rejectSymlink(lstater afero.Lstater, path string) error {
+	lstat, lstatCalled, err := lstater.LstatIfPossible(path)
 	if err != nil {
+		if os.IsNotExist(err) {
+			// component does not exist on disk — nothing to check here
+			return nil
+		}
 		return fmt.Errorf("could not check file %q: %w", path, err)
```
Kirdock, line 86: [nit] Didn't we say that we don't want to use quotes in error messages, only single ones?

internal/files/symlink_test.go
Kirdock, line 1: Some tests are failing on Windows due to incorrect path concatenation.

Image

internal/files/symlink_test.go
```diff
@@ -31,125 +31,394 @@ import (
...
+	require.NoError(t, err)
+	return resolved
+}
+
 // errLstatFs wraps an afero.Fs whose LstatIfPossible always returns an error.
 type errLstatFs struct{ afero.Fs }
 
 func (f errLstatFs) LstatIfPossible(name string) (os.FileInfo, bool, error) {
 	return nil, true, fmt.Errorf("simulated lstat error")
 }
```
Kirdock, line 174: [nit]
```suggestion
func (f errLstatFs) LstatIfPossible(_ string) (os.FileInfo, bool, error) {
	return nil, true, assert.AnError
}
```

and evaluating the error with `assert.ErrorIs(t, err, assert.AnError)` in the related test

internal/files/symlink_test.go
```diff
@@ -31,125 +31,394 @@ import (
...
 		filePath := filepath.Join(dir, "regular.yaml")
 		require.NoError(t, os.WriteFile(filePath, []byte("content"), 0644))
 
 		fs := afero.NewOsFs()
-		err := RejectSymlink(fs, filePath)
+		err := RejectSymlinkRecursive(fs, filePath)
 		assert.NoError(t, err)
 	})
 
 	t.Run("returns no error for directory on OsFs", func(t *testing.T) {
```
arthurpitman, line 50: What about a non-existent directory? Or is that effectively the non-existent file case?

test/account/utils_test.go
Kirdock, line 1: [nit] Not related to the PR or to the commit

internal/files/symlink_test.go
```diff
@@ -31,125 +31,394 @@ import (
...
+
+		// project/a/b/c -> /nonexistent/outside
+		require.NoError(t, os.MkdirAll(filepath.Join(projectDir, "a", "b"), 0755))
+		require.NoError(t, os.Symlink("/nonexistent/outside", filepath.Join(projectDir, "a", "b", "c")))
+
+		fs := afero.NewBasePathFs(afero.NewOsFs(), projectDir)
+		err := RejectSymlinkRecursive(fs, "a/b/c/secret.txt")
+		var symErr *symlinkDetectedError
+		require.ErrorAs(t, err, &symErr)
+		assert.Equal(t, "a/b/c", symErr.path)
```
Kirdock: Windows:
expected: "a/b/c"
actual  : "a\\b\\c"

internal/files/symlink_test.go
```diff
@@ -31,125 +31,394 @@ import (
...
+
+		// project/configs/leak -> /nonexistent/outside
+		require.NoError(t, os.MkdirAll(filepath.Join(projectDir, "configs"), 0755))
+		require.NoError(t, os.Symlink("/nonexistent/outside", filepath.Join(projectDir, "configs", "leak")))
+
+		fs := afero.NewBasePathFs(afero.NewOsFs(), projectDir)
+		err := RejectSymlinkRecursive(fs, "configs/leak/sub/x.json")
+		var symErr *symlinkDetectedError
+		require.ErrorAs(t, err, &symErr)
+		assert.Equal(t, "configs/leak", symErr.path)
```
Kirdock: Windows:
expected: "configs/leak"
actual  : "configs\\leak"

internal/files/symlink_test.go
```diff
@@ -31,125 +31,394 @@ import (
...
+	tests := []struct {
+		input string
+		want  []string
+	}{
+		{"/a/b/c", []string{"/a/b/c", "/a/b", "/a", "/"}},
+		{"a/b/c", []string{"a/b/c", "a/b", "a"}},
+		{"/file.txt", []string{"/file.txt", "/"}},
+		{"file.txt", []string{"file.txt"}},
+		{"/a/b/../c", []string{"/a/c", "/a", "/"}},
+		{"a//b///c", []string{"a/b/c", "a/b", "a"}},
```
Kirdock, line 402: Windows:
For all except `file.txt`
`/` => `\`

internal/files/symlink_test.go
Laubi: yes it's like non-existing file. Added test cases as it costs us nothing to have more tests

internal/files/symlink.go
Laubi: We have it currently 50:50 in the codebase. We can go over it in one go and fix it for all files

Kirdock (approved):

PR #2145, opened 2026-05-26
fix: Clients should retry 429 and 503 by default

arthurpitman:
#### **Why** this PR?
- Per-request retry logic for 429 Too Many Requests was scattered across almost every individual API call site.
- It didn't cover 503 Service Unavailable responses, which are also transient and should be retried.

- `GetDynatraceVersion(...)` in `pkg/client/version/version_check.go` and its tests have been removed as they are not used.
- The core library dependency is updated to a version with `RetryIfTooManyRequestsOrServiceUnavailable` and that expects client-level retry behavior.
- Default retry options now include 429 and 503 handling. `DefaultRetryOptions` in `pkg/client/clientset.go` uses `RetryIfTooManyRequestsOrServiceUnavailable` and adds a 1-second `DelayAfterRetry`. This value may need to be tuned.
- This allows per-call-site `CustomShouldRetryFunc` overrides to be removed as they are now configured at the client level. Special care was taken with the classic and settings client as they retried everything on a client level, but then overwrote this at essentially every call site. The exceptions with special behavior are preserved.

#### How is it **tested**?
- Exiting tests

- No effects other than improved retry logic.

**Issue:** CA-19525


Laubi (dismissed):

pkg/client/clientset.go
```diff
@@ -194,7 +195,7 @@ type ServiceLevelObjectiveClient interface {
 
 var DefaultMonacoUserAgent = "Dynatrace Monitoring as Code/" + version.MonitoringAsCode + " " + (runtime.GOOS + " " + runtime.GOARCH)
 
-var DefaultRetryOptions = rest.RetryOptions{MaxRetries: 10, ShouldRetryFunc: rest.RetryIfNotSuccess}
+var DefaultRetryOptions = rest.RetryOptions{MaxRetries: 10, DelayAfterRetry: time.Second, ShouldRetryFunc: rest.RetryIfTooManyRequestsOrServiceUnavailable}
```
arthurpitman, line 198: What value shall we pick here for `DelayAfterRetry`?

arthurpitman: Note: For terraform we use 10 seconds!

d0weinberger: Should we make this adjustable via env var?

arthurpitman: I'm not sure that is a good idea, remembering that the rate limiter is running on top, and classic APIs customize this value anyway.

d0weinberger (approved):

PR #2150, opened 2026-06-08
chore: remove deprecation validator

Kirdock:
#### **Why** this PR?
This removes the validator that checked if some schemas are deprecated. The schemas aren't available anymore, so the deployment fails and warnings are not sufficient anymore.

#### **What** has changed?
Deprecation messages are removed

#### **How** does it do it?
By removing the deprecation validator and tests

#### How is it **tested**?
Removed tests related to it

Instead of warning + fail to deploy it will only be fail to deploy

**Issue:** NOISSUE


pkg/deploy/internal/setting/validation.go
```diff
@@ -28,29 +28,6 @@ import (
 	"github.com/dynatrace/dynatrace-configuration-as-code/v2/pkg/project"
 )
 
-type DeprecatedSchemaValidator struct{}
-
-var deprecatedSchemas = map[string]string{
-	"builtin:span-attribute":       "this setting was replaced by 'builtin:attribute-allow-list' and 'builtin:attribute-masking'",
```
d0weinberger, line 34: Wouldn't it be enough to just empty this list? If we have deprecated schemas again in the future, we could fill the list up again. We don't really need to remove the whole component

Kirdock: We already have a bunch but we don't maintain it and it's hard to maintain. Therefore, I would vote to remove it if we don't use it. We can always add it back in the future if needed

d0weinberger (approved):

PR #2154, opened 2026-06-15
fix: eventual consistency error for failure detection parameter sets

Kirdock:
#### **Why** this PR?
For schema `builtin:failure-detection.environment.parameters` and API `failure-detection-parametersets` there may be eventual consistency errors during delete, which causes the delete to fail.
`Failure detection rule EnvironmentFailureDetectionRules {your-values-here} refers to this parameter ID: <uuid>`

#### **What** has changed?
Custom retry logic added during delete to mitigate eventual consistency errors

#### **How** does it do it?
Adds custom retry logic for api `failure-detection-parametersets` and setting `builtin:failure-detection.environment.parameters` during delete.

#### How is it **tested**?
New tests added and existing integration tests should not be flaky anymore

Delete for the affected API/schema should not fail anymore

**Issue:** CA-19787

### DIFF
--- a/pkg/client/dtclient/config_client.go
+++ b/pkg/client/dtclient/config_client.go
@@ -118,7 +118,11 @@ func (d *ConfigClient) Delete(ctx context.Context, api api.API, id string) error
 	}
 	parsedURL = parsedURL.JoinPath(id)
 
-	_, err = coreapi.AsResponseOrError(d.client.DELETE(ctx, parsedURL.String(), corerest.RequestOptions{}))
+	_, err = coreapi.AsResponseOrError(d.client.DELETE(ctx, parsedURL.String(), corerest.RequestOptions{
+		CustomShouldRetryFunc: func(resp *http.Response) bool {
+			return corerest.RetryIfTooManyRequestsOrServiceUnavailable(resp) || isFailureDetectionParameterSetsAPIError(api, resp)
+		},
+	}))
 	if err != nil {
 		if coreapi.IsNotFoundError(err) {
 			slog.DebugContext(ctx, "No config found to delete (HTTP 404 response)", slog.String("id", id))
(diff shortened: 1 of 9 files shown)


arthurpitman (approved):

pkg/client/dtclient/config_client.go
```diff
@@ -118,7 +118,11 @@ func (d *ConfigClient) Delete(ctx context.Context, api api.API, id string) error
 	}
 	parsedURL = parsedURL.JoinPath(id)
 
-	_, err = coreapi.AsResponseOrError(d.client.DELETE(ctx, parsedURL.String(), corerest.RequestOptions{}))
+	_, err = coreapi.AsResponseOrError(d.client.DELETE(ctx, parsedURL.String(), corerest.RequestOptions{
+		CustomShouldRetryFunc: func(resp *http.Response) bool {
+			return corerest.RetryIfTooManyRequestsOrServiceUnavailable(resp) || isFailureDetectionParameterSetsAPIError(api, resp)
```
Laubi, line 123: [This has recently been changed](https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/2145/changes#diff-466b4649a643061d8a532aaaf9b36bdd3fd8a31ed93f65859675e8fcb6f79fa7R121)

@Kirdock do you think there might be more issues coming to us?

Kirdock: The flaky test now is new, as the schema was just updated, and even without the mentioned PR, it would have led to issues.

For delete: I don't think that other issues will arise.

Before the mentioned PR, the settings_client had a CustomShouldRetry configured only for 429 errors on every call.
Meaning:
Before the PR: Retry 429
After the PR: Retry 429 and 503

My PR: adding conditional retry for 400

So, the functionality was never reduced; it was just extended, because the default was never considered before

Laubi (approved):

PR #2172, opened 2026-08-05
feat(documents): support document labels

d0weinberger:
#### **Why** this PR?
Adding document labels and description support and adapting to changes in the documents corelib client:
- removal of `ExternalID`, which means that `Metadata.ID` is used directly.
- introduction of `Labels` and `Description`
- change of function signatures (i.e. everything that is not content is passed via a `Metadata` struct)

#### **What** has changed?

- Users can now declare `labels` and `description` on any document config entry (dashboard, notebook, launchpad) in their `config.yaml`:
  ```yaml
  type:
    document:
      kind: notebook
      description: This is my notebook
      labels:
        - team:foo
        - env:prod
  ```
- Labels are optional; existing configs without a `labels` field are unaffected.
- Description is option optional; existing configs without a `description` field are unaffected.
- `monaco download` preserves labels and descriptions, so downloaded configs include whatever labels and description the document already has in the environment.

#### **How** does it do it?

- The core library's new `documents.Metadata` struct bundles `ID`, `Name`, `Type`, `IsPrivate`, `Description`, and `Labels` into a single value; replacing the previous positional-parameter signatures with this struct was the natural place to thread labels through without touching every call-site individually.
- `DeployAPI.Deploy` builds the `documents.Metadata` value before calling `upsertDocument`, which then stamps the resolved ID into `metadata.ID` before each `Update`/`Create` call.
- Also, `DeployAPI.Deploy` now fails on an unknown document kind, instead of defaulting to an empty document type.
- `getDocumentAttributesFromConfigType` was removed; its logic is now inlined into `Deploy` as part of the `Metadata` construction.

#### How is it **tested**?

Unit test: 
- `config_loader_test.go`: verifies `labels` and `description` are parsed from YAML into `config.DocumentType`
- `config_writer_test.go`: verifies it is serialized back out.
- Tests in `deploy_test.go` to check that `labels` and `description` reach the `documents.Metadata`passed to the client.
- `download_test.go` updated to assert that labels and description from the API response are captured in the downloaded `config.DocumentType`.

Integration test: 
- `test/configtypes/documents/documents_test.go` deploys a notebook with labels and description and then reads it back from the environment to verify `Labels` and `Description` match what was configured.

Users can now declare `labels` and `description` on any document config entry (dashboard, notebook, launchpad) in their `config.yaml`

**Issue:** PS-47482

### DIFF
--- a/pkg/config/internal/persistence/type_definition.go
+++ b/pkg/config/internal/persistence/type_definition.go
@@ -61,10 +61,11 @@ type AutomationDefinition struct {
 }
 
 type DocumentDefinition struct {
-	Kind    config.DocumentKind `yaml:"kind" json:"kind" mapstructure:"kind"`
-	ID      string              `yaml:"id,omitempty" json:"id,omitempty"  mapstructure:"id"`
-	Private bool                `yaml:"private,omitempty" json:"private,omitempty"  mapstructure:"private"`
-	Labels  []string            `yaml:"labels,omitempty" json:"labels,omitempty" mapstructure:"labels"`
+	Kind        config.DocumentKind `yaml:"kind" json:"kind" mapstructure:"kind"`
+	ID          string              `yaml:"id,omitempty" json:"id,omitempty"  mapstructure:"id"`
+	Private     bool                `yaml:"private,omitempty" json:"private,omitempty"  mapstructure:"private"`
+	Labels      []string            `yaml:"labels,omitempty" json:"labels,omitempty" mapstructure:"labels"`
+	Description *string             `yaml:"description,omitempty" json:"description,omitempty" mapstructure:"description"`
 }
 
 type OpenPipelineDefinition struct {
@@ -185,10 +186,11 @@ func (c *TypeDefinition) parseDocumentType(a any) error {
 	}
 
 	c.Type = config.DocumentType{
-		Kind:     r.Kind,
-		Private:  r.Private,
-		CustomID: r.ID,
-		Labels:   r.Labels,
+		Kind:        r.Kind,
+		Private:     r.Private,
+		CustomID:    r.ID,
+		Labels:      r.Labels,
+		Description: r.Description,
 	}
 
 	return nil
@@ -331,10 +333,11 @@ func (c TypeDefinition) MarshalYAML() (any, error) {
 	case config.DocumentType:
 		return map[string]any{
 			"document": DocumentDefinition{
-				Kind:    t.Kind,
-				Private: t.Private,
-				ID:      t.CustomID,
-				Labels:  t.Labels,
+				Kind:        t.Kind,
+				Private:     t.Private,
+				ID:          t.CustomID,
+				Labels:      t.Labels,
+				Description: t.Description,
 			},
 		}, nil
 
--- a/pkg/config/internal/persistence/type_definition.go
+++ b/pkg/config/internal/persistence/type_definition.go
@@ -64,6 +64,7 @@ type DocumentDefinition struct {
 	Kind    config.DocumentKind `yaml:"kind" json:"kind" mapstructure:"kind"`
 	ID      string              `yaml:"id,omitempty" json:"id,omitempty"  mapstructure:"id"`
 	Private bool                `yaml:"private,omitempty" json:"private,omitempty"  mapstructure:"private"`
+	Labels  []string            `yaml:"labels,omitempty" json:"labels,omitempty" mapstructure:"labels"`
 }
 
 type OpenPipelineDefinition struct {
@@ -187,6 +188,7 @@ func (c *TypeDefinition) parseDocumentType(a any) error {
 		Kind:     r.Kind,
 		Private:  r.Private,
 		CustomID: r.ID,
+		Labels:   r.Labels,
 	}
 
 	return nil
@@ -332,6 +334,7 @@ func (c TypeDefinition) MarshalYAML() (any, error) {
 				Kind:    t.Kind,
 				Private: t.Private,
 				ID:      t.CustomID,
+				Labels:  t.Labels,
 			},
 		}, nil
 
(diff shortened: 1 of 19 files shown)


pkg/config/internal/persistence/type_definition.go
```diff
@@ -64,6 +64,7 @@ type DocumentDefinition struct {
 	Kind    config.DocumentKind `yaml:"kind" json:"kind" mapstructure:"kind"`
 	ID      string              `yaml:"id,omitempty" json:"id,omitempty"  mapstructure:"id"`
 	Private bool                `yaml:"private,omitempty" json:"private,omitempty"  mapstructure:"private"`
+	Labels  []string            `yaml:"labels,omitempty" json:"labels,omitempty" mapstructure:"labels"`
```
Kirdock: Should we also support the other properties that were introduced?
description and isReshareable

pkg/resource/document/deploy.go
```diff
@@ -58,44 +58,44 @@ func (d DeployAPI) Deploy(ctx context.Context, properties parameter.Properties,
...
 	}
-
-	documentKind, isPrivate := getDocumentAttributesFromConfigType(documentType)
 	documentName, ok := properties[config.NameParameter].(string)
 	if !ok {
 		return entities.ResolvedEntity{}, ErrMissingNameParameter
 	}
-
-	if documentKind == documents.Dashboard {
+	docType, found := documentKindToType[documentType.Kind]
```
Kirdock, line 65: Minor logic change here. isPrivate isn't set to false, if the type can't be found. May be worth questioning, why this was introduced in the first place, also the `docType = ""` change.
Introduced with: https://github.com/Dynatrace/dynatrace-configuration-as-code/pull/1492/changes#diff-5b3534d6ad541c4b0f98e266e044538aa73672d91e38e7e54b909d218e7721feR161

Kirdock:
Regarding the failing test: Haven't looked much into it, but maybe it has the same root cause as in our Terraform provider
https://github.com/dynatrace-oss/terraform-provider-dynatrace/pull/1203#discussion_r3721352296

pkg/config/internal/persistence/type_definition.go
d0weinberger: Yes, you're right. Since they also will be added for the Terraform provider, I will also add them here.

pkg/resource/document/deploy.go
d0weinberger: Good catch!

d0weinberger: But I would say, if the doc type cannot be found, the deploy should actually fail.

image

pkg/config/internal/persistence/type_definition.go
d0weinberger: Added

pkg/resource/document/deploy.go
d0weinberger: I will do some more manual testing, but I do not see any reason why we should allow unknown document types to be used. I doubled down on that and throw an error now on unknown document types.

d0weinberger: Okay, @Kirdock , I found it. If you check in `type_definition.go` (here: https://github.com/Dynatrace/dynatrace-configuration-as-code/blob/main/pkg/config/internal/persistence/type_definition.go#L252), you see that we actually fail already if the yaml contains any unknown document types. I therefore opted to just fail the deployment in the client in case of an unknown document type.

pkg/config/internal/persistence/type_definition.go
```diff
@@ -61,9 +61,11 @@ type AutomationDefinition struct {
...
 
 type DocumentDefinition struct {
-	Kind    config.DocumentKind `yaml:"kind" json:"kind" mapstructure:"kind"`
-	ID      string              `yaml:"id,omitempty" json:"id,omitempty"  mapstructure:"id"`
-	Private bool                `yaml:"private,omitempty" json:"private,omitempty"  mapstructure:"private"`
+	Kind        config.DocumentKind `yaml:"kind" json:"kind" mapstructure:"kind"`
+	ID          string              `yaml:"id,omitempty" json:"id,omitempty"  mapstructure:"id"`
+	Private     bool                `yaml:"private,omitempty" json:"private,omitempty"  mapstructure:"private"`
+	Labels      []string            `yaml:"labels,omitempty" json:"labels,omitempty" mapstructure:"labels"`
+	Description *string             `yaml:"description,omitempty" json:"description,omitempty" mapstructure:"description"`
```
Kirdock, line 68: With this, we should probably also update our static schema `cmd/monaco/generate/schemas/json-schemas/monaco-config.schema.json`

pkg/resource/document/download_test.go
```diff
@@ -44,14 +44,15 @@ func TestDownloader_Download(t *testing.T) {
...
+	dashboardDescription := "A getting started dashboard"
 	expectedDashboardConfig := config.Config{
 		Coordinate: coordinate.Coordinate{
 			Project:  "project",
 			Type:     "document",
 			ConfigId: "12345678-1234-1234-1234-0123456789ab",
 		},
 		OriginObjectId: "12345678-1234-1234-1234-0123456789ab",
-		Type:           config.DocumentType{Kind: config.DashboardKind},
+		Type:           config.DocumentType{Kind: config.DashboardKind, Labels: []string{"team:foo", "env:prod"}, Description: &dashboardDescription},
```
Kirdock: [nit]
```suggestion
		Type:           config.DocumentType{Kind: config.DashboardKind, Labels: []string{"team:foo", "env:prod"}, Description: new("A getting started dashboard"),
```

Kirdock (commented):
[nit] Theoretically we can merge a lot of the tests, as we don't need a separate test for each and every property in the YAML/documentMeta

d0weinberger:
> [nit] Theoretically we can merge a lot of the tests, as we don't need a separate test for each and every property in the YAML/documentMeta

Yes, but it also doesn't hurt, I think ... I'd keep the tests the way they are right now, if that's okay.

Kirdock (approved):

PR #2182, opened 2026-08-25
fix(ci): pin version of libs in `scorecard.yml`

tomazpu:
#### **Why** this PR?
Pinning GitHub Actions to exact commit SHAs prevents supply chain attacks

#### **What** has changed?
All four actions in `.github/workflows/scorecard.yml` are now referenced by their full commit SHA

#### **How** does it do it?
Each `uses:` line in the workflow is updated to reference the commit SHA that the respective tag resolved to at the time of this change.

#### How is it **tested**?
The workflow runs on every push to `main`

No functional change.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -21,24 +21,24 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


Kirdock (commented):
Should we also update the versions?
For e.g., actions/upload-artifact there is v7 already

Kirdock (approved):

PR #2191, opened 2026-09-16
feat: download openpipeline configs of all owners

Kirdock:
#### **Why** this PR?
OpenPipeline resources in Monaco currently operate within strict owner-scoped boundaries, preventing administrators from exporting configurations across different owners.

#### **What** has changed?
OpenPipeline configs of different owners can now be downloaded.

#### **How** does it do it?
By introducing an `--admin-access` flag to the download functionality and limiting the flag to OpenPipeline schemas only.

#### How is it **tested**?
Unit tests added
Manual tests

They can now download OpenPipeline configs of different owners.

**Issue:** PS-45539


pkg/client/dtclient/settings_client.go
```diff
@@ -48,6 +50,8 @@ const (
 	WriteOperation  = "write"
 )
 
+var ownerBasedPipelinesRegex = regexp.MustCompile(`builtin:openpipeline\..*\.(ingest-sources|data-forwarding|pipeline-groups|pipelines)`)
```
tomazpu: Would this be better? 
`regexp.MustCompile( ^builtin:openpipeline\..+\.(ingest-sources|data-forwarding|pipeline-groups|pipelines)$, )`

so that the middle part needs to be not empty, and that it needs to end with one of the keys.

tomazpu: Would also recommend packing this into a test, for things that should and should not match.

Kirdock: The `Test_getAdminAccess` in settings_client_test.go already tests that.
Or do you mean something else?

Kirdock: I added two other tests for the `^` and `$` case in the regex, though

pkg/client/dtclient/settings_client.go
```diff
@@ -826,9 +832,10 @@ func (d *SettingsClient) List(ctx context.Context, schemaId string, opts ListSet
 		listSettingsFields = reducedListSettingsFields
 	}
 	params := url.Values{
-		"schemaIds": []string{schemaId},
-		"pageSize":  []string{defaultPageSize},
-		"fields":    []string{listSettingsFields},
+		"schemaIds":   []string{schemaId},
+		"pageSize":    []string{defaultPageSize},
+		"fields":      []string{listSettingsFields},
+		"adminAccess": []string{strconv.FormatBool(getAdminAccess(opts.AdminAccess, schemaId))},
```
arthurpitman: I'd prefer to see a function of the schemaID that does the check together with an `&&`:
```suggestion
		"adminAccess": []string{strconv.FormatBool(opts.AdminAccess && supportsAdminAccess(schemaId))},
```

arthurpitman (approved):

